PR #18, opened 2021-11-22
Tk/tendencies

tomkimpson:
Basic structure for calculating tendencies.

ToDo():

* ~~Configure `geopotential()` properly and link~~
* ~~Define some constants e.g. `coriol`, `j2` , `nlon`, `nlat`, `mx`, `nx` etc.~~
* ~~Clean up `dynamics_tendencies` into readable subroutine structure~~
* Parameterised physics tendencies. **This is a big one** - might be best as another PR once all other machinery is in place


milankl:
The conflicts are because you renamed `gridded` to `convert_to_grid` and `spectral` to `convert_to_spectral`. Happy for renaming, but currently the tests and `src/SpeedyWeather.jl` depend on this, so we should agree on one naming. `spectral_transform()` and `grid_transform()` could be another?

tomkimpson:
Yeah I’ll clean up all conflicts once I’ve made all changes. ‘convert_to_X’ makes most sense to me, but happy with others

Tom

src/diagnostic_variables.jl
```diff
@@ -4,19 +4,65 @@ struct Tendencies{NF<:AbstractFloat}
...
+    v_grid      ::Array{NF,3}  # Gridpoint field of meridional velocity
+
+
+    Tabs_grid_anomaly   ::Array{NF,3}  # Gridpoint field of absolute temperature anomaly [K]
+
+    
 end
 
+
+struct MiscellaneousVariables{NF<:AbstractFloat}
```
milankl, line 40: I think once we know better what all else is needed we can rename this to be more specific?

tomkimpson:
Yeah for sure - it’s just a placeholder at the minute

Tom

tomkimpson:
@milankl I don't know if you want to bother reviewing this in full before merge since it is quite lengthy? Instead I propose merging to give us a general structure for the tendencies in `main` and we can then subsequently make any changes we want in a separate PR?

src/dynamics_tendencies.jl
```diff
@@ -0,0 +1,216 @@
...
+    #..and calculate values
+    for k in 1:nlev
+        u_mean += u_grid[:,:,k]*dhs[k] 
+        v_mean += v_grid[:,:,k]*dhs[k]
+        d_mean += div_grid[:,:,k]*dhs[k]
+    end
+
+    #Now use the mean fields
+    grad!(pres_surf[:,:,l2], dumc[:,:,2], dumc[:,:,3])
+    px = gridded(dumc[:,:,2]*3600, scale=true)
```
milankl, line 32: This still uses `gridded` ...

src/dynamics_tendencies.jl
```diff
@@ -0,0 +1,216 @@
...
+                                   C::Constants{NF}
+                                   ) where {NF<:AbstractFloat}
+
+
+    @unpack pres_surf= Prog 
+    @unpack pres_surf_tend = Diag.Tendencies
+
+    @unpack u_grid,v_grid,div_grid = Diag.gridvars
+    @unpack u_mean,v_mean,d_mean,dumc,px,py = Diag.miscvars
+    @unpack nlev, dhs = C
```
milankl, line 16: `nlev` is currently not propagated into constants, but we can change that. However, you can also just obtain `nlev` from any prognostic/diagnostic variable `var` via `_,_,nlev = size(var)` which is more independent of any renamings that we still might do

src/dynamics_tendencies.jl
```diff
@@ -0,0 +1,216 @@
...
+                                     Diag::PrognosticVariables{NF}, # Diagnostic variables
+                                     C::Constants{NF}
+                                     ) where {NF<:AbstractFloat}
+
+    @unpack u_grid,v_grid,div_grid = Diag.gridvars
+    @unpack u_mean,v_mean,d_mean,sigma_tend,sigma_m,puv,px,py = Diag.miscvars
+    @unpack nlev,dhs = C 
+
+    #Initialise
+    sigma_tend[:,:,:] = 0.0
```
milankl, line 53: what is `sigma`? `σ=p/p0` (fraction of surface pressure) is used for the vertical levels, which I thought about renaming to `σ_levels` for more clearity (I don't think we use the term _level_ for anything but the vertical)

src/dynamics_tendencies.jl
```diff
@@ -0,0 +1,216 @@
...
+
+    @unpack pres_surf= Prog 
+    @unpack pres_surf_tend = Diag.Tendencies
+
+    @unpack u_grid,v_grid,div_grid = Diag.gridvars
+    @unpack u_mean,v_mean,d_mean,dumc,px,py = Diag.miscvars
+    @unpack nlev, dhs = C 
+
+    #Initialise mean fields. 
+    u_mean[:,:] = 0.0
```
milankl, line 19: This causes an error in Julia as the broadcasting has to be triggered with the dot-notation `.=` furthermore there's no need to write `0.0` as it will be converted to eltype of `u_mean` anyway. You can write `u_mean[:,:] .= 0`, but the Julian way of doing it is more `fill!(u_mean,0)`, probably doesn't matter much.

src/dynamics_tendencies.jl
milankl: Oh and neither is `dhs` (what is that?)

src/dynamics_tendencies.jl
```diff
@@ -0,0 +1,216 @@
...
+
+"""
+Compute the temperature anomaly in grid point space
+"""
+function temperature_grid_anomaly!(Diag::PrognosticVariables{NF}, # Diagnostic variables
+                                   C::Constants{NF}
+                                   ) where {NF<:AbstractFloat}
+
+    @unpack temp_grid,temp_grid_anomaly = Diag.gridvars
+    @unpack nlev,tref = C  #Is tref a constant?
```
milankl, line 78: There are currently two temperature reference values declared in `Params`, `temp_ref` the surface absolute temperature referece (288K) and `temp_top` the reference temperature of the stratosphere (216K), if you need more I'd add it to parameters.jl and convert it to the respective number format NF to be held in `Constants`

src/parametrization_tendencies.jl
```diff
@@ -0,0 +1,13 @@
+struct parametrization_tendencies{T<:AbstractFloat}
+
+#Just a placeholder
+
+end
+
+"""
+Compute physical parametrization tendencies
+"""
+function phypar{T}(utend, vtend, ttend, qtend) where {T<:AbstractFloat}
```
milankl, line 10: Generator functions must have the same name as the generated struct, otherwise the `{T}` notation causes an error. They basically extend the default generating function of the struct (which just takes all elements in the structs as arguments) with an additional method (in the multiple-dispatch sense).

src/dynamics_tendencies.jl
```diff
@@ -0,0 +1,216 @@
...
+
+    @unpack div_grid,humid_grid = Diag.gridvars
+    @unpack arbitrary_array,sigma_tend= Diag.miscvars
+    @unpack dhsr, nlev = C
+
+    for k in 2:nlev
+        arbitrary_array[:,:,k] = sigma_tend[:,:,k].*(humid_grid[:,:,k,itr] - humid_grid[:,:,k-1,itr])
+    end
+    
+    arbitrary_array[:,:,2:3] .= zero
```
milankl, line 207: `zero` is a function that requires as a first argument a type. E.g. `zero(NF)`

src/dynamics_tendencies.jl
tomkimpson: Marking resolved. Agreed in Slack to use `gridded`/ `spectral`

src/dynamics_tendencies.jl
tomkimpson: `nlev` now determined via `size`.

`dhs` is the layer thickness. I guess this will be something that will ultimately calculated and held in the `Constants` struct. For now it is just a placeholder

src/dynamics_tendencies.jl
tomkimpson: Yeah this is a notation hangover from previous Speedy versions. `sigma_tend` denotes the vertical velocity in sigma coords. There is also a quantity `sigma_m` which I am not sure how to interpret. See e.g. L222,  223 of https://github.com/eapax/speedy/blob/stochastic_rounding/source/dyn_grtend.f90

This needs to be renamed. How about just `vertical_velocity`?

src/dynamics_tendencies.jl
tomkimpson: Corrected to use `fill!`

src/dynamics_tendencies.jl
tomkimpson: In previous (fortran) Speedy we have `tref` and permutations defined as:

```fortran

do k=1,kx
        tref(k)=288.0_dp*max(0.2_dp,fsg(k))**rgam 
        tref1(k)=rgas*tref(k)
        tref2(k)=akap*tref(k)
        tref3(k)=fsgr(k)*tref(k)
end do
``` 

Like `dhs` I think these can ultimately be calculated in the Constants struct, right?

src/parametrization_tendencies.jl
tomkimpson: Good catch. Fixed

src/dynamics_tendencies.jl
tomkimpson: Fixed

PR #25, opened 2021-12-16
leapfrog! without mem alloc

milankl:
I haven't checked yet whether the leapfrog is correct otherwise. For `l1=1` for example, the william's line calls `A[i,j,1]` which has been changed in the Robert's line. Is that correct?

### DIFF
--- a/src/SpeedyWeather.jl
+++ b/src/SpeedyWeather.jl
@@ -1,7 +1,8 @@
 module SpeedyWeather
 
-    using NetCDF, FFTW, LinearAlgebra, Parameters, Dates
+    using NetCDF, FFTW, LinearAlgebra, Dates
     
+    import Parameters: @with_kw, @unpack
     import FastGaussQuadrature: gausslegendre
 
     export run_speedy, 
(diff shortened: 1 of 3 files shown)


tomkimpson:
The `l1=1` case should have some logic to make `eps = 0` such that both `w1=w2=0`

What is going on on L32/34 re definition of `eps`?

milankl:
- The coefficients now follow [William (2009)](http://www.atmos.albany.edu/facstaff/rfovell/ATM562/williams-2009.pdf) Eq. 7-9 such that `w1 = nu*alpha/2` and `w2 = nu*(1-alpha)/2`
- The `eps` from Sam's version is removed, which was probably an old notation from [an old IFS documentation, 2.2.5](https://www.ecmwf.int/en/elibrary/13272-ifs-documentation-cy23r4-part-iii-dynamics-and-numerical-procedures)
- the term in square brackets in William (2009) Eq. 8 and 9 is only calculated once and for the sake of a better name called `Aupdate`
- `leapfrog!` now takes only spectral fields by using `::AbstractArray{Complex{NF},3}` if the same code is needed for grid space time stepping, we can quickly change that.

milankl:
Sorry, my bad, I now see that for `l1=1` both filters should be disabled, not just the William's filter. Right? Commit coming.

PR #38, opened 2022-02-24
New spectral transform

milankl:


src/legendre.jl
```diff
@@ -180,3 +180,64 @@ function legendre(  input::Array{Complex{NF},2},
...
+    Λ = Matrix{NF}(undef, size(alms)...)    # legendre polynomials (1 latitude only)
+    gn = zeros(Complex{NF}, nfreq)          # phase factors for northern latitudes
+    gs = zeros(Complex{NF}, nfreq)          # phase factors for southern latitudes
+    map = zeros(NF, nlon, nlat)             # output array
+
+    @inbounds for ilat in 1:nlat_nh         # loop over northern latitudes only due to symmetry
+        ilat_s = nlat - ilat + 1            # southern latitude index
+
+        # Recalculate the (normalized) λ_l^m(cos(colat)) factors of the ass. Legendre polynomials
+        AssociatedLegendrePolynomials.λlm!(Λ, lmax, mmax, cos(colat[ilat]))
```
milankl, line 208: @samhatfield you may find this interesting. In this implementation the legendre polynomials are only allocated at the same size as the spectral coefficients (L199) and recalculated in L208 for every latitude `ilat` on a single hemisphere (making use of odd/even symmetry). The recalculation at T31 is about 50% compute time of the spectral transform, so precalculation would be better. At T682 this increases to 70% (even worse) but at that resolution the storage requirement (64 bits) is 7MB with recalculation and 4GB without. The timings exclude loading something like a 4GB array from memory, I've just commented L208 to quickly check this.

samhatfield: So basically, this new method allows us to massively reduce memory requirements at an increase of wall clock time (as yet unquantified as we don't have the whole model yet)? That's a pretty nifty dial to play with.

milankl: Yes, I'm planning to keep the precalculation/recalculation of the polynomials optional, so that we can investigate this issue further once the model is more complete.

samhatfield: Nice work.

PR #39, opened 2022-02-24
Update tendencies structure 

tomkimpson:
Some changes and tests to make tendencies calculations fully consistent and tested.

Work in progress - draft PR


tomkimpson:
Notes:

* Need to establish clearly which are spectral quantities and which are grid point quantities
* Some additional parameters defined in `geometry.jl` (e.g. tref,rgas). Currently not defined correctly - just in place for structure.

Issue #40, opened 2022-02-25
Prepare input data


samhatfield:
speedy.f90 relies on a number of climatological and boundary fields. These are available currently on the T30/96×48 grid but we would like them for all [available resolutions](https://github.com/milankl/SpeedyWeather.jl/issues/22).

I will begin with the orography field as I already have the data for this on a high-resolution TCO1279 (2560×1280, 9 km) grid, taken from the IFS.

samhatfield:
See branch [sh/input-data](https://github.com/samhatfield/SpeedyWeather.jl/tree/sh/input-data).

samhatfield:
Examples:
![orog_T31](https://user-images.githubusercontent.com/8796885/155708062-1979bd0e-b4e6-4927-9e60-67aac33e2ac5.png)
![orog_T85](https://user-images.githubusercontent.com/8796885/155708065-7a30955f-e8b9-4798-84c4-c28e92d97fee.png)
![orog_T682](https://user-images.githubusercontent.com/8796885/155708069-e3f0d872-a1c9-497e-b11e-1ed014c7f0f9.png)

samhatfield:
The Gaussian latitudes for the resulting T31 data do not match those for the currently used orography data:
```
>>> orig = iris.load("orog_actual_T30.nc")[0]; orig.coords("latitude")
[DimCoord(array([ 87.159,  83.479,  79.777,  76.07 ,  72.362,  68.652,  64.942,
        61.232,  57.521,  53.81 ,  50.099,  46.389,  42.678,  38.967,
        35.256,  31.545,  27.833,  24.122,  20.411,  16.7  ,  12.989,
         9.278,   5.567,   1.856,  -1.856,  -5.567,  -9.278, -12.989,
       -16.7  , -20.411, -24.122, -27.833, -31.545, -35.256, -38.967,
       -42.678, -46.389, -50.099, -53.81 , -57.521, -61.232, -64.942,
       -68.652, -72.362, -76.07 , -79.777, -83.479, -87.159]), standard_name='latitude', units=Unit('degrees'), long_name='latitude', var_name='lat')]
>>> new = iris.load("orog_T31.nc")[0]; new.coords("latitude")
[DimCoord(array([ 88.125,  84.375,  80.625,  76.875,  73.125,  69.375,  65.625,
        61.875,  58.125,  54.375,  50.625,  46.875,  43.125,  39.375,
[...]

PR #41, opened 2022-02-25
Orography from high resolution (2048x1024) file

milankl:


milankl:
Addresses #40 for orography

PR #55, opened 2022-04-12
Initialize model to time stepping flow

milankl:


src/run_speedy.jl
```diff
@@ -1,50 +1,47 @@
...
-                          kwargs...                   # all additional non-default parameters
+"""
+    prog_vars, diag_vars, model_setup = initialize_model(NF,kwargs...)
+    
+Initialize the model by returning
+- `prog_vars`, the initial conditions of the prognostic variables
+- `diag_vars`, the preallocated the diagnotic variables (initialised to zero)
+- `model_setup`, the collected pre-calculated structs that don't change throughout integration:
+parametes, constants, geometry, spectral transform, boundaries, diffusion."""
+function initialize_model(::Type{NF}=Float64;       # number format, use Float64 as default
```
milankl, line 26: @tomkimpson I like the idea to have a separate `initialize_model` function, which is also helpful for testing. Hence, I've integrated that into `run_speedy`

src/run_speedy.jl
```diff
@@ -1,50 +1,47 @@
...
+
+    prognostic_vars = initial_conditions(P,B,G)     # initialize prognostic variables
+    diagnostic_vars = DiagnosticVariables(G)        # preallocate all diagnostic variables with zeros
+    
+    return prognostic_vars,diagnostic_vars,M
 end
 
 """Struct holding all the model structs """
-struct Model{NF<:AbstractFloat}
+struct ModelSetup{NF<:AbstractFloat}
```
milankl, line 44: I think I'd like to call it `ModelSetup` instead to make clear it's only the setup of the model, not the model state (which is defined with the prognostic variabels) nor the diagnostic variables. `ModelSetup` is essentially every "constant" whether it's a scalar or an array that doesn't change through model integration.

PR #67, opened 2022-04-29
Build `readable_secs` on `Dates.canonicalize`

giordano:
I don't know if you'll like it as this is a bit more verbose than the current `readable_secs`, but `Dates.canonicalize` allows you to automatically get the readable strings and (_totally biased opinion coming_) I personally find the proposed implementation easier to read.  Let me know what you think and feel free to reject it :slightly_smiling_face:

Side note, I have a collaborator calling from Julia the Speedy model written in Fortran, maybe we should talk at some point :wink:


milankl:
Happy to accept it, but not sure how much this function will be needed once I've migrated things to ProgressMeter.jl

PR #82, opened 2022-05-20
Parametrization of Large-Scale Condensation

white-alistair:


src/SpeedyWeather.jl
```diff
@@ -2,37 +2,37 @@ module SpeedyWeather
...
     import UnicodePlots
     import ProgressMeter
-    
+
     # EXPORT MAIN INTERFACE TO SPEEDY
     export run_speedy, initialize_speedy
 
     # EXPORT STRUCTS
-    export Parameters, GenLogisticCoefs,
+    export Parameters, GenLogisticCoefs, HumidityCoefs,
```
milankl, line 25: I've exported the `GenLogisticCoefs` in case someone wants to change the vertical resolution. However, with many of the other structs, I don't think they need to be exported as I doubt people would use them frequently enough to justify the `GeneratorFunction` instead of `SpeedyWeather.GeneratorFunction` shortcut.

src/SpeedyWeather.jl
```diff
@@ -2,37 +2,37 @@ module SpeedyWeather
...
         PrognosticVariables, DiagnosticVariables
-    
+
     # EXPORT SPECTRAL FUNCTIONS
     export  spectral, gridded,
         spectral_truncation, spectral_interpolation,
         triangular_truncation
 
     include("utility_functions.jl")
-    include("parameter_structs.jl")
```
milankl, line 35: I've called it `parameter_structs` to have a place where structs can be defined before `default_parameters.jl` is included (otherwise you'll get an undefined type error). In think general there's many places to define "constants" which can serve different purposes
- parameters that can be changed through the `run_speedy` interface
- physical/numerical constants that should be defined as such for clarity and readability
- constants that have to be converted to a given number format NF to avoid runtime conversions

src/coefficient_structs.jl
```diff
@@ -0,0 +1,33 @@
...
+These parameters, and the formula where they are used, do not appear in the original
+Speedy documentation.
+"""
+@with_kw struct HumidityCoefs{NF<:Real} <: Coefficients
+    e₀::NF = 6.108e-3
+    C₁::NF = 17.269
+    C₂::NF = 21.875
+    T₀::NF = 273.16
+    T₁::NF = 35.86
+    T₂::NF = 7.66
```
milankl, line 32: Do any of these have any physical meaning? I guess `T₀` (also maybe we should add that to the style guide, is there an advantage in calling things T₀ over T0?) is the conversion between ˚C and K?

src/humidity.jl
```diff
@@ -0,0 +1,63 @@
+"""
+Compute the saturation specific humidity for a single atmospheric level.
+"""
+function get_saturation_specific_humidity(
+    T::Array{NF,2},  # Absolute temperature for a single atmospheric level
+    p::Array{NF,2},  # Normalised surface pressure
+    σ::NF,           # σ of the current atmospheric level
```
milankl, line 7: Are these half of full levels? See `src/geometry.jl`

src/humidity.jl
```diff
@@ -0,0 +1,63 @@
...
+
+    @inbounds for j = 1:nlat, i = 1:nlon
+        if T[i, j] > T₀
+            Qsat[i, j] = e₀ * exp(C₁ * (T[i, j] - T₀) / (T[i, j] - T₁))
+        else
+            Qsat[i, j] = e₀ * exp(C₂ * (T[i, j] - T₀) / (T[i, j] - T₂))
+        end
+    end
+
+    if σ <= 0
```
milankl, line 23: Full sigma levels are never negative, the half levels start at zero for the top of the atmosphere, see https://github.com/milankl/SpeedyWeather.jl/blob/main/input_data/vertical_coordinate/vertical_levels.png I know this is just taken from Fortran but I don't know when this branch would actually be executed?

src/humidity.jl
```diff
@@ -0,0 +1,63 @@
...
+    p::Array{NF,2},  # Normalised surface pressure
+    σ::NF,           # σ of the current atmospheric level
+    M::ModelSetup{NF},
+) where {NF<:AbstractFloat}
+    @unpack nlon, nlat = M.geospectral.geometry
+    @unpack e₀, C₁, C₂, T₀, T₁, T₂ = M.params.humidity_coefs
+
+    Qsat = zeros(nlon, nlat)  # Saturation specific humidity
+
+    @inbounds for j = 1:nlat, i = 1:nlon
```
milankl, line 15: We also need to have the discussion whether for the parametrizations it makes sense to parallelize in the horizontal or in the vertical. Technically all parametrizations should only act on a given atmospheric column, which would motivate a horizontal domain decomposition, however, the arrays are currently layed out in memory as lon,lat,(leapfrog), and then vert such that the dynamical core can act as independent as possible across the vertical. If we want to parallelize single-column style we may want a function instead that takes in a vector which we could obtain from `view(variable,i,j,:)` instead of baking in the lat lon loop

src/humidity.jl
```diff
@@ -0,0 +1,63 @@
...
+        if T[i, j] > T₀
+            Qsat[i, j] = e₀ * exp(C₁ * (T[i, j] - T₀) / (T[i, j] - T₁))
```
milankl, line 17: Do you think it would make sense to write this function already in ˚C rather than having `T` in Kelvin? Also to be more explicit we called temperature `temp` throughout other functions ...

src/humidity.jl
```diff
@@ -0,0 +1,63 @@
...
+Compute the saturation specific humidity for a single atmospheric level.
+"""
+function get_saturation_specific_humidity(
+    T::Array{NF,2},  # Absolute temperature for a single atmospheric level
+    p::Array{NF,2},  # Normalised surface pressure
+    σ::NF,           # σ of the current atmospheric level
+    M::ModelSetup{NF},
+) where {NF<:AbstractFloat}
+    @unpack nlon, nlat = M.geospectral.geometry
+    @unpack e₀, C₁, C₂, T₀, T₁, T₂ = M.params.humidity_coefs
```
milankl, line 11: `M.parameters.abc...`

src/default_parameters.jl
```diff
@@ -47,14 +47,20 @@ With keywords such that default values can be changed at creation.
...
     # PARAMETRIZATIONS
     seasonal_cycle::Bool=true   # Seasonal cycle?
     n_shortwave::Int=3          # Compute shortwave radiation every n steps
     sppt_on::Bool=false         # Turn on SPPT?
+    humidity_coefs::HumidityCoefs = HumidityCoefs()
+    # Large-scale condensation (from table B10)
+    τ::Real = 4.0           # Relaxation time for humidity (hours)
+    RH¹::Real = 0.9         # Relative humidity threshold at σ = 1
+    ΔRH::Real = 0.1         # Vertical range of relative humidity threshold
+    rhb::Real = 0.95            # Relative humidity threshold for boundary layer - TODO(alistair): rename this
```
milankl, line 63: Can we be a bit more explicit here? E.g. `humid_relax_time`, `humid_thresh` (or better, what's this threshold for? condensation? Then maybe `humid_condensation` or similar?

src/humidity.jl
milankl: also what about `humid_saturation`? and then add the information in a comment that it's actually a water vapour pressure [hPa] or similar?

src/humidity.jl
```diff
@@ -0,0 +1,63 @@
+"""
+Compute the saturation specific humidity for a single atmospheric level.
+"""
+function get_saturation_specific_humidity(
+    T::Array{NF,2},  # Absolute temperature for a single atmospheric level
+    p::Array{NF,2},  # Normalised surface pressure
+    σ::NF,           # σ of the current atmospheric level
+    M::ModelSetup{NF},
```
milankl, line 8: Currently `ModelSetup` is an abstract type (so no `NF`). `models.jl` has now `BarotropicModel`, `ShallowWaterModel` and `PrimitiveEquationModel` defined as subtypes of `ModelSetup`, meaning that for precipitation we should write `M::PrimitiveEquationModel` as humidity isn't defined for the other ModelSetups

src/humidity.jl
milankl: Full! I see from below, sorry.

src/humidity.jl
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    return Qsat
+end
+
+"""
+Compute the saturation specific humidity.
+"""
+function get_saturation_specific_humidity(
+    T::Array{NF,3},  # Absolute temperature
```
milankl, line 36: My idea was to move every vertical loop to `src/distributed_vertical.jl` such that we have one place where in the end the `@distributed` will distribute the vertical layers across processors for parallelisation. However, the question whether we actually want to do that or whether the parametrizations should be distributed in the horizontal. In the latter I'd suggest to create a `src/distributed_horizontal.jl` as a central place for the horizontal domain decomposition. Tricky!

src/humidity.jl
```diff
@@ -0,0 +1,63 @@
...
+    p::Array{NF,2},  # Normalised surface pressure
+    M::ModelSetup{NF},
+) where {NF<:AbstractFloat}
+    @unpack nlon, nlat, nlev, σ_levels_full = M.geospectral.geometry
+
+    Qsat = zeros(nlon, nlat, nlev)  # Saturation specific humidity
```
milankl, line 42: This should be preallocated in `src/diagnostic_variables.jl`. Maybe it makes sense to create therein a new struct like `ParametrizationVariables` or `PhysicsVariables` where all parametrization-related arrays are initialised?

src/large_scale_condensation.jl
```diff
@@ -0,0 +1,47 @@
...
+    M::ModelSetup{NF},
+) where {NF<:AbstractFloat}
+    @unpack pres_ref, cp, alhc, alhs, gravity, τ, RH¹, ΔRH, rhb = M.constants  # Let's give some of these constants better names?
+    @unpack nlon, nlat, nlev, σ_levels_half, σ_levels_full, σ_levels_thick =
+        M.geospectral.geometry
+    @unpack temp_grid, humid_grid, pres_surf_grid, cloud_top, precipitation_ls = Diag.grid_variables
+    @unpack temp_tend, humid_tend = Diag.parametrization_tendencies
+    @unpack cloud_top = Diag.intermediate_variables
+
+    # Use mathematical notation for consistency with the documentation
```
milankl, line 12: I see that motivation, and I agree that we may want that for the beginning to make sure everything works well, otherwise I'd vote to break with the mathematical notation. I already struggle to remember what `cp, alhc, alhs, τ, rhb` etc stand for, hence we started using `gravity` instead of `g` etc. But let's discuss that!

src/large_scale_condensation.jl
```diff
@@ -0,0 +1,47 @@
...
+    Q = humid_grid
+    T = temp_grid
+    ∂Q = humid_tend
+    ∂T = temp_tend
+
+    p = exp.(pres_surf_grid)  # Normalised surface pressure - TODO(alistair): check pressure units throughout
+    Qsat = get_saturation_specific_humidity(T, p, M)
+
+    # 1. Tendencies of humidity and temperature due to large-scale condensation
+    @inbounds for k = 2:nlev
```
milankl, line 22: any idea why this starts at 2? Just trying to decipher whether something is not applied in the stratosphere (k=1,2)

src/large_scale_condensation.jl
```diff
@@ -0,0 +1,47 @@
...
+            if Q[i, j, k] > Q_threshold
+                ∂Q[i, j, k] += -(Q[i, j, k] - Q_threshold) / τ  # Formula 22
+                ∂T[i, j, k] += -alhc / cp * min(∂Q[i, j, k], ∂Q_max * p[i, j])  # Formula 23
+                cloud_top[i, j] = min(k, cloud_top[i, j])  # Page 7 (last sentence)
+            end
+        end
+    end
+
+    # 2. Precipitation due to large-scale condensation#
+    @inbounds for k = 2:nlev
```
milankl, line 43: Also feel free to drop the `@inbounds` for development, I usually only add it if I'm 100% sure that I'm indexing correctly (otherwise you may hit a segfault / behaviour is undefined), and only add it if I include a `@boundscheck` at the top of a function to guarantee that array sizes match.

src/coefficient_structs.jl
white-alistair: So these are the parameters for Magnus' formula for saturation vapour pressure, and as far as I can tell they're determined experimentally. So none of them have an obvious physical meaning, except for `T₀` of course, which in this case is actually used to convert from Celsius to Kelvin.

`C₁` and `T₁` are for the saturation vapour pressure over water, while `C₂` and `T₂` are over ice. Speedy uses the first set of parameters if the temperature is above zero, and the second set if it's below zero. From what I can tell this isn't strictly correct, since apparently water in clouds is often cooled below zero, but such an approximation is probably consistent with the approximate nature of Speedy's parameterisations?

There appears to be a long history of coming up with new, ever so slightly different values for these parameters. A lot of them are listed [here](https://www.osti.gov/servlets/purl/548871). Yet nowhere have I managed to find the exact values used in Speedy!

In any case, it seems like the modern accepted values can be found in Table 8.3, p. 211 of the [Springer Handbook of Atmospheric Measurements](https://link.springer.com/book/10.1007/978-3-030-52171-4) ;) (I would certainly keep the original values from Speedy, for now.)

I'll probably refactor this a bit, to have something like a `MagnusCoefficients` struct, with separate instances for water and ice. I think that's the most logically consistent approach, and also makes it easier to substitute more up-to-date values.

white-alistair: Regarding `T₀` over `T0`, it's just a personal preference to have the code resemble the maths (I find it easier to spot mistakes that way). But happy to get rid of them subscripts :)

src/humidity.jl
white-alistair: Ah I didn't spot that!

Is there any benefit to making the `ModelSetup` abstract type parametric, i.e. `ModelSetup{NF}`? Probably not, unless you want to dispatch on it somewhere?

src/humidity.jl
white-alistair: I find it concerning that this slipped through...

src/humidity.jl
white-alistair: > Do you think it would make sense to write this function already in ˚C rather than having `T` in Kelvin?

Just to be clear, do you mean specifically just write this line in Kelvin so that the function more closely resembles the standard Magnus form? i.e. precede this line with a line like `temp_celsius = temp - T0`?

src/humidity.jl
white-alistair: That's a very good point! Full sigma levels can be zero though, right?

In any case I don't really get what's going on here. And I especially wouldn't be comfortable merging this before understanding the use of `p[1, 1]` here. Do you have any idea what that could be about?

src/large_scale_condensation.jl
white-alistair: The original [docs](http://users.ictp.it/~kucharsk/speedy_description/km_ver41_appendixA.pdf) apply this parameterisation in tropospheric layers, which would seem to explain it. But is k=2 also a stratospheric layer? Naive question but how exactly do we know...?

src/SpeedyWeather.jl
white-alistair: Yeah sorry I was a bit hasty in renaming this file and had actually intended to discuss it with you!

In general I found the constants/parameters thing a bit confusing. As you suggest, I think it would be nice to differentiate somehow between universal physical constants that will never change and parameters which might.

src/humidity.jl
white-alistair: `Qsat` gets calculated for all `k`, but is only actually used (so far at least) for `k >= 2`. So this (weird) branch may get executed once for `σ == 0`, but then the result of that calculation never gets used?

src/SpeedyWeather.jl
milankl: What about using "constants" for actual physical constants that nobody would want to change, "parameters" for everything that `run_speedy` may take as an argument and "constants_runtime" for every value that has to be precalculated, probably from parameters, but is constant at runtime. Meaning we should rename the constants.jl file and constants.jl should then be loaded before default_parameters?

src/coefficient_structs.jl
milankl: I like `HumidityCoefficients` and the idea to access them for example via `M.constants.humidity_coefficients`. Not particularly hard feelings for `T₀` over `T0`, I'd just make sure that the lines that use them get a comment like `# T0 = zero ˚C in Kelvin`

src/humidity.jl
milankl: I don't think there's a need to have `ModelSetup` parametric, making sure we don't mix number formats where we don't intend to happens in lower level functions anyway.

src/humidity.jl
milankl: I mean rewriting the entire model such that temperature uses ˚C instead of Kelvin. I expect everything to work very similar, instead of probably radiation, where somewhere the Stefan-Boltzmann ~T^4 should appear. We could then convert from ˚C to Kelvin therein. But from a precision point-of-view using Kelvin is awful as you'll hit massive rounding errors in the time integration, e.g. T = 300K, tendency dT = 0.1K
```julia
julia> T = Float16(300)
Float16(300.0)

julia> dT = Float16(0.1)
Float16(0.1)

julia> T+dT
Float16(300.0)

julia> T2 = Float16(300-273.15)
Float16(26.84)

julia> T2+dT
Float16(26.94)
```
In Kelvin you can't resolve the increment, but in ˚C you can.

milankl: Having said that, we obviously do the time stepping in spectral space, meaning that we'd only have that problem on the `l=m=0` mode which we could also solve with compensated summation, but I think in general, we should try to use ˚C if possible. If it turns out that ˚C is a bad idea then we can still revert back.

src/humidity.jl
milankl: For default resolution the full sigma levels are currently
```julia
julia> m.geospectral.geometry.σ_levels_full
8-element Vector{Float64}:
 0.036597223182346206
 0.11748566048999509
 0.2169349015956573
 0.3402683261876466
 0.4905020311609761
 0.6631496767486345
 0.8333721536349882
 0.9565026461476916
```
while the first half level is 0, and the last is 1, the full levels are the mid-points, so we'll always have 0<sigma<1.

src/large_scale_condensation.jl
milankl: fortran speedy uses the top two levels as the stratosphere. This is hardcoded (e.g. some loops just start at 3), I've tried to generalize that by introducing `n_stratosphere_levels` in `default_parameters.jl` such that this parameter should be used instead given that we may want to have more levels with higher resolution

src/SpeedyWeather.jl
white-alistair: Yeah, I like that approach. I would write it up as a separate issue, just to avoid this PR becoming too bloated.

src/humidity.jl
white-alistair: Ahhh I had full and half the wrong way around :)

white-alistair: Still though, any idea what this actually does?

```julia
@. Qsat = 622.0 * Qsat / (σ * p - 0.378 * Qsat)
```

src/humidity.jl
white-alistair: Agreed it makes sense to use Celsius (or at the very least to try it out). I would definitely write that up as a separate issue though, and keep this PR in Kelvin for now.

src/large_scale_condensation.jl
white-alistair: Good to know. Hard to explain the loop starting at two then, but I guess they're just saying that some large-scale condensation occurs in the lower troposphere but not upper?

In any case good to be aware of this as we experiment with more vertical levels.

milankl: Yes, I honestly believe there to many of theses lines in the Fortran code where it's absolutely unclear why certain choices have been made. I'd suggest to go with your best guess how this generalizes but add a comment like `# used to be k = 2:nlev` so that if we realize that something is wrong we know where we attempted to remove some hardcoding

white-alistair: I like it :+1:

src/humidity.jl
milankl: Something like
image

I believe `622.0` is the ratio of molecular weight of water (2H+O = 18g/mol) to dry air (28.96g/mol, see [here](https://socratic.org/questions/what-is-the-molecular-weight-of-air)) times 1000 for some conversion?

src/humidity.jl
```diff
@@ -0,0 +1,55 @@
...
+    temp_grid         ::Array{NF,3},
+    pres              ::Array{NF,2},
+    model             ::ModelSetup,
+) where {NF<:AbstractFloat}
+    @unpack nlon, nlat, nlev, σ_levels_full = model.geospectral.geometry
+
+    get_saturation_vapour_pressure!(sat_vap_pressure, temp_grid, model)
+
+    for k = 1:nlev, j = 1:nlat, i = 1:nlon
+        sat_spec_humidity[i, j, k] = 0.622 * sat_vap_pressure[i, j, k] /
```
milankl, line 50: `0.622` or `1-0.622` shouldn't be hardcoded in the loop. You can either define them before the loop, e.g. `mol_ratio = convert(NF,0.622)`, `one_minus_mol_ratio = convert(NF,1 - mol_ratio)` or `@unpack` them from some constants?

src/large_scale_condensation.jl
```diff
@@ -0,0 +1,43 @@
...
+    Diag::DiagnosticVariables{NF},
+    M   ::ModelSetup,
+) where {NF<:AbstractFloat}
+    @unpack gravity, RH_thresh_max, RH_thresh_range, RH_thresh_boundary, humid_relax_time = M.constants
+    @unpack cp, alhc = M.parameters
+    @unpack nlon, nlat, nlev, σ_levels_full, σ_levels_thick = M.geospectral.geometry
+    @unpack temp_grid, humid_grid, pres_surf_grid = Diag.grid_variables
+    @unpack temp_tend, humid_tend, sat_spec_humidity, sat_vap_pressure, cloud_top, precip_large_scale = Diag.parametrization_variables
+
+    pres = exp.(pres_surf_grid)  # Normalised surface pressure - TODO(alistair): check pressure units throughout
```
milankl, line 11: ideally we preallocate `pres` somewhere else and reuse it here. As pressure is always surface pressure maybe we should also call it `pres_log_grid` and `pres_grid` to better distinguish between the logarithm of surface pressure and non-log version?

test/Project.toml
```diff
@@ -0,0 +1,5 @@
+[deps]
+Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+Parameters = "d96e819e-fc66-5662-9728-84c9c7592b0a"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
```
milankl, line 5: It shouldn't be necessary to add these dependencies. But feel free to if it's easier. For `Uniform(200,350)` you can do `200 + 150*rand()`, for `@unpack field = struct` you can do `(;field) = struct` (was introduced in Julia Base in 1.6 or 1.7(?) but I used `@unpack` usually as I find it more explicit.

milankl:
Sorry @white-alistair, with the merging of #87 there's now a conflict. In `src/SpeedyWeather.jl` I had to swap the order of the `include`s, so when you resolve this conflict put your includes in, but don't change the order of the existing ones to avoid an `UndefinedError`

src/large_scale_condensation.jl
white-alistair: I would be in favour of renaming `pres_surf_grid` to `pres_log_grid`, because I found it hard to figure out what to call `exp.(pres_surf_grid)` :sweat_smile: 

Question about units: is pressure always hPa?

white-alistair: Also, if we preallocate `pres_grid` somewhere else, then we have to make sure to update it everytime `pres_log_grid` changes. Where exactly should we do that? So we don't accidentally take the exponential twice, or something.

test/Project.toml
white-alistair: Damn you're right, that's definitely an unnecessary dependency. Will remove the test-specific dependencies.

src/diagnostic_variables.jl
```diff
@@ -66,9 +66,44 @@ function GridVariables(G::GeoSpectral{NF}) where NF
...
+Struct holding quantities calculated from the physical parameterisations. All quantities
+are in grid-point space.
+"""
+struct ParametrizationVariables{NF<:AbstractFloat}
+    sat_vap_pressure   ::Array{NF,3}   # Saturation vapour pressure
+    sat_spec_humidity  ::Array{NF,3}   # Saturation specific humidity
+    cloud_top          ::Array{Int,2}  # Cloud-top
+    precip_large_scale ::Array{NF,2}   # Large-scale precipitation
+    humid_tend         ::Array{NF,3}   # Humidity tendencies due to physics
+    temp_tend          ::Array{NF,3}   # Temperature tendencies due to physics
```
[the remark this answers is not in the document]
white-alistair: So the `humid_tend` and `temp_tend` which are already defined in `Tendencies` are in spectral space. 

My original preference for the parameterisations was to simply add up the tendencies as we went along, rather than allocating a new array for each tendency and parameterisation and summing them at the end. However the problem I ran into here (which may be repeated for other parameterisations) is that the humidity tendency due to LSC gets re-used to calculate precipitation. So you have to allocate it either.

For now, I think I would just pre-allocate a new array for each tendency and parameterisation (`humid_tend_lsc`, `temp_tend_lsc`, ...) and then at the end we can optimise it, once we have a clearer picture of how the parameterisations interact?

white-alistair:
@milankl another thing which I forgot to mention: you can see that I included references to the formulas in the original Speedy documentation, where applicable. I would expect to remove these in the future, probably to be replaced by more up-to-date documentation.

src/diagnostic_variables.jl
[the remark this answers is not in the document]
milankl: >However the problem I ran into here (which may be repeated for other parameterisations) is that the humidity tendency due to LSC gets re-used to calculate precipitation. So you have to allocate it either.

I think this is exactly where we have to start thinking about working in column vectors rather than horizontal fields. That would solve many problems. I reckon it might actually be the best to start very early with this restructuring from vert x lon x lat to lon x lat x vert (outer to inner loop). Because then there's no problem to allocate a bunch of vectors for intermediate calculations that get reused for the next horizontal grid point. That saves memory and gives us more flexibility to reuse stuff.

milankl:
> @milankl another thing which I forgot to mention: you can see that I included references to the formulas in the original Speedy documentation, where applicable. I would expect to remove these in the future, probably to be replaced by more up-to-date documentation.

Yeah just stick to the existing references for now, one we've documented our speedy version we can update those references in the comments. In the end I expect us to do things somewhat differently to Fortran speedy so there should be a small describing section in the documentation of how we did it. Hence we also have the freedom to rename variables as we go along for consistency/clarity etc, as we'll document it anyway again.

src/diagnostic_variables.jl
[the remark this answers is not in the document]
white-alistair: Yeah that's a great point, no problem allocating an 8-element vector.

milankl (approved):

PR #90, opened 2022-06-10
How does Float16 behave during timestepping

tomkimpson:
An experiment to throwaway....


milankl:
I close this given that we talked about it and realised that low precision can be unreasonably good at the oscillation equation test case for leapfrog. I don't think there was actually a pull request to come out of this? Feel free to reopen.

Issue #91, opened 2022-06-13
Pre-allocate grid-point pressure values and use consistent naming


white-alistair:
This is a left over from [this comment](https://github.com/milankl/SpeedyWeather.jl/pull/82#discussion_r885489677) on #82

milankl:
On this note, for the shallow water model, I've decided to use the interface displacement η as prognostic variable, so the instantaneous layer thickness h = η + H, whereby H is the thickness at rest and H = H₀ - orography and H₀ a global constant, something like 8-12km probably. Having said that, I don't want to redefine the variable names in `PrognosticVariables` for the different models, so I'll be using `pres` (or even `pres_log` if we decide to redefine it) for η. This is probably an argument to keep this variable name quite generic (I vote for `pres`, as this has a generic meaning, whether you think about logarithm of surface pressure or interface displacement).

white-alistair:
@milankl picking this up again. How about the following approach:

1. Pre-allocate a new grid variable `pres_exp_grid` in the `GridVariables` struct (the name is potentially misleading, but I think all of the options are suboptimal in their own way, so I'd be happy to go with your preference from the previous comment).
2. In `tendencies_parametrizations.jl`, do `pres_exp_grid = exp.(pres_grid)`, i.e. before any of the parametrizations are calculated, since this quantity is used by several of them. I'm not 100% this is the most logical place to do this, but as far as I know we only use this in the parametrizations.

Let me know what you think!

milankl:
Is `pres_grid` (i.e. the logarithm of surface pressure on the grid) even used for the tendencies? If not, we could just have `pres_grid` which is first transformed from spectral `pres` and then calculate the exponent in-place? Then `pres` would always be the logarithm of surface pressure in spectral space and `pres_grid` the surface pressure (hPa) in grid-point space.

white-alistair:
As far as I can see it's only used for the physics, so yeah I think this works. I'll just comment it clearly so newcomers aren't confused :)

PR #107, opened 2022-07-15
Take exponential of pres_grid for parametrizations

white-alistair:
#91 

@milankl let me know if this is consistent with what you had in mind.

I see that `pres_grid` is also used [here](https://github.com/milankl/SpeedyWeather.jl/blob/a1ec01d3fe909e00eaaae43cb78c3cca28d837cb/src/tendencies_dynamics.jl#L340) as well as for the Bernoulli potential, both of which are, I think. only relevant for the shallow water equations.

I'm lacking the big picture slightly, but do you think it's both safe and not too confusing to do the conversion in this way?

### DIFF
--- a/src/large_scale_condensation.jl
+++ b/src/large_scale_condensation.jl
@@ -8,9 +8,7 @@ function get_large_scale_condensation_tendencies!(
     @unpack temp_grid, humid_grid, pres_grid = Diag.grid_variables
     @unpack temp_tend_lsc, humid_tend_lsc, sat_spec_humidity, sat_vap_pressure, cloud_top, precip_large_scale = Diag.parametrization_variables
 
-    pres = exp.(pres_grid)  # Normalised surface pressure - TODO(alistair): check pressure units throughout
-
-    get_saturation_specific_humidity!(sat_spec_humidity, sat_vap_pressure, temp_grid, pres, M)
+    get_saturation_specific_humidity!(sat_spec_humidity, sat_vap_pressure, temp_grid, pres_grid, M)
 
     # 1. Tendencies of humidity and temperature due to large-scale condensation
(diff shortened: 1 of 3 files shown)


milankl:
I think this is fine. While in the primitive equations, `pres` is ln(p), the logarith of surface pressure, I'm just reusing this prognostic variable in the shallow water equations for the interface displacement. So while in the shallow water equations the bernoulli potential is calculated from the interface displacement, it's done via the geopotential in primitive equations. So there shouldn't be any conflict. If we realise there is later on, then it's good we already had that discussion and will surely easily find a way to resolve it. So happy to merge this now!

src/large_scale_condensation.jl
```diff
@@ -40,7 +38,7 @@ function get_large_scale_condensation_tendencies!(
     # 2. Precipitation due to large-scale condensation
     fill!(precip_large_scale, zero(NF))
     for k = k_lsc:nlev
-        Δpₖ = pres * σ_levels_thick[k]  # Formula 4
+        Δpₖ = pres_grid * σ_levels_thick[k]  # Formula 4
```
milankl, line 41: there are some indices to `pres_grid` missing?

milankl: Ah sorry, I get it, this is matrix = matrix * scalar

white-alistair:
> I think this is fine. While in the primitive equations, `pres` is ln(p), the logarith of surface pressure, I'm just reusing this prognostic variable in the shallow water equations for the interface displacement. So while in the shallow water equations the bernoulli potential is calculated from the interface displacement, it's done via the geopotential in primitive equations. So there shouldn't be any conflict. If we realise there is later on, then it's good we already had that discussion and will surely easily find a way to resolve it. So happy to merge this now!

Perfect, thanks!

PR #117, opened 2022-07-29
LowerTriangularMatrix in PrognVars, DiagnVars +restructured

milankl:


milankl:
@white-alistair This PR is a bit hectic, but I've been restructuring the `PrognosticVariables` and `DiagnosticVariables` such that one can now do something like
```julia
julia> using SpeedyWeather
julia> progn,diagn,model = initialize_speedy();
julia> progn.
layers     lmax        mmax        n_leapfrog  nlev        pres
```
so PrognosticVariables contains another struct with `layers` and `pres` (as it's just surface separated). Hence, access to the variables is straight forward like
```julia
julia> progn.layers[1].leapfrog[1].vor
32×32 LowerTriangularMatrix{ComplexF32}:
...
```
and it should be super easy to loop over all layers like (which can be a single loop very early on at every time step)
```julia
for layer in progn.layers
    func!(layer,...)
end
```

milankl:
With `LowerTriangularMatrix` as the default struct for spectral coefficients looping in spectral domain is either as before
```julia
for m in 1:mmax+1
    for l in m:mmax+1
        vor[l,m]
```
or more convenient (and faster)
```julia
for lm in eachharmonic(vor)
    vor[lm]
```
which is equivalent to
```julia
for lm in 1:length(vor)
    vor[lm]
```

white-alistair:
> @white-alistair This PR is a bit hectic, but I've been restructuring the `PrognosticVariables` and `DiagnosticVariables` such that one can now do something like
> 
> ```julia
> julia> using SpeedyWeather
> julia> progn,diagn,model = initialize_speedy();
> julia> progn.
> layers     lmax        mmax        n_leapfrog  nlev        pres
> ```
> 
> so PrognosticVariables contains another struct with `layers` and `pres` (as it's just surface separated). Hence, access to the variables is straight forward like
> 
> ```julia
> julia> progn.layers[1].leapfrog[1].vor
> 32×32 LowerTriangularMatrix{ComplexF32}:
> ...
> ```
> 
> and it should be super easy to loop over all layers like (which can be a single loop very early on at every time step)
> 
> ```julia
> for layer in progn.layers
>     func!(layer,...)
> end
> ```

@milankl looks good, nice one! 

I see that you moved the parametrization vars into a file `column_variables.jl` :+1: On a related note, say there are intermediate quantities used in the calculation of some parametrization, would you put those arrays in the `IntermediateVariables` struct or the `ParametrizationVariables` struct?

Let me know if you want an actual review of this at some point.

milankl:
@white-alistair I answered that question in #118, I suggest creating a `ColumnVariables` struct.

PR #118, opened 2022-08-03
Basics for KernelAbstractions / GPU

maximilian-gelbrecht:
Basics and utility for use of KernelAbstractions / GPU. 

We could also directly integrate it as a field into `ModelSetup` already. Let me know where you want to have it.


maximilian-gelbrecht:
In the longer run, information about the grid should be incorporated into the `workgroup_size`. Also when all computing functions operator on one layer only, we should also be able to get rid of the `ndrange` argument as they are only two array sizes, one for grid space, one for sph.

src/gpu.jl
```diff
@@ -0,0 +1,104 @@
...
+DeviceSetup() = DeviceSetup(Device(), Device_KernelAbstractions(Device()), workgrop_size(Device()))
+DeviceSetup(device::AbstractDevice) = DeviceSetup(device, Device_KernelAbstractions(device), workgrop_size(device))
+DeviceSetup(device::AbstractDevice, n::Integer) = DeviceSetup(device, Device_KernelAbstractions(device), n)
+
+"""
+    workgrop_size(dev::AbstractDevice)
+
+Returns a workgroup size depending on `dev`. WIP: Will be expended in the future to also include grid information. 
+"""
+function workgrop_size(device::AbstractDevice)
```
white-alistair, line 68: Typo ;)

src/gpu.jl
```diff
@@ -0,0 +1,104 @@
...
+Holds information about the device the model is running on and workgroup size. 
+
+* `device::AbstractDevice`: Device the model is running on 
+* `device_KA::KernelAbstractions.Device`: Device for use with KernelAbstractions
+* `n`: workgroup size  
+"""
+struct DeviceSetup{S<:AbstractDevice}
+    device::S # for internal purposes
+    device_KA # for KernelAbstractions
+    n # workgroup size
```
milankl, line 56: Do they not need a type?

milankl:
> Also when all computing functions operate on one layer only, we should also be able to get rid of the `ndrange` argument as they are only two array sizes, one for grid space, one for sph.

The `PrimitiveEquationModel` should work like this. At every time step do

1) LOOP OVER LAYERS transform spectral prognostic variables to grid variables
2) LOOP OVER GRID POINTS parametrizations
3) VERTICAL TERMS (e.g. the surface pressure tendency is the vertical sum of divergence)
4) LOOP OVER LAYERS dynamics tendencies + leapfrogging + output

So after each of 1-4 there needs to be a checkpoint, but the loops are otherwise independent of each other. 2&3 do the coupling of the layers but for 1 & 4 all functions should be just based on 2D fields (whether spectral or grid).

@white-alistair yeah, my plan is to do 2 conceptually in the following way (why I started in column_variables.jl)

```julia
struct ColumnVariables
     u::Vector{NF}
     v::Vector{NF}
     ...
end

for ij in eachgridpoint(grid_vars::GridVariables)
    column = ColumnVariables(grid_vars,ij)   # extract a single column
    parametrizations!(column,grid_vars)      # calculate terms and add to tendencies
end
```
whereby the latter loop can be parallelized. `column` contains everthing that's needed for the parametrizations, but not more (i.e. only vectors of length `nlev`), which means we never have to allocate all those 2D fields for the parametrizations which are currently done in Fortran speedy

src/gpu.jl
maximilian-gelbrecht: You are right, added types

maximilian-gelbrecht:
@milankl I thought about directly adding `DeviceSetup` to all subtypes of `ModelSetup`, or do you think it should go somewhere else? 

Right now, just the constructor: `DeviceSetup(CPUDevice())` would be called and nothing else is changed as it is not used by any function. Later, the users would be able to hand over a `CPUDevice()`, or `GPUDevice()` as a `kwarg` to `run_speedy` and/or `initialize_speedy`.

milankl:
I don't have any strong feelings of _how_ we attach information of the device to `ModelSetup` but I agree that's where it shuold go. In the end we probably can't oversee all the benefits of having it as a field vs having it as a parameter to the type (what I was suggesting on zulip) but maybe that's just something to keep in mind. So go for either now, and if we realize it should be done differently then we change it. Having said that what about
```julia
struct SomeModel{NF,D} <: ModelSetup{NF<:AbstractFloat,D<:AbstractDevive}
     field1::SomeType{NF}
     field2::SomeOtherType{NF}
    device::D
end
```
then we should be able to store all information about the device in `model.device` but also dispatch to CPU or GPU dynamically by defining 
```julia
func(...,::ModelSetup{NF,D<:GPUDevice}) where {NF,D}
func(...,::ModelSetup{NF,D<:CPUDevice}) where {NF,D}
func(...,::ModelSetup{NF,D<:AbstractDevice}) where {NF,D}
```
for those functions where CPU and GPU implementatios differ (line 1 and 2) or where they are the same (line3)

maximilian-gelbrecht:
Yep, I agree and added it to `ModelSetup`

milankl:
Looks like this can be merged now, but before I'll do that I want to merge #117 first. This will likely create some file conflicts here but better this way around than the other way around...

milankl:
After merging #117 I just resolved the conflicts here and happy to merge this now as all tests pass. We don't have a compat yet for the new dependencies `KernelAbstractions`,`CUDA`,`CUDAKernels` and `Adapt` but I reckon the CompatHelper will do that once it's merged anyway. Yay 🥳

PR #122, opened 2022-08-11
AbstractGrids towards HEALPix or Octahedral

milankl:


milankl:
@hottad @samhatfield I've started implementing `AbstractGrid`, with subtypes like `FullGaussianGrid`, `OctahedralGaussianGrid`, and `HEALPixGrid` as a common type structure for any grids we may want to implement. At the moment I'm thinking to define methods like
```julia
get_weights(::FullGaussianGrid,S::SpectralTransform) = S.gaussian_weights
get_weights(::OctahedralGaussianGrid,S::SpectralTransform) = S.gaussian_weights
get_weights(::OctahedralClenshawGrid,S::SpectralTransform) = S.clenshaw_curtis_weights
get_weights(::HEALPixGrid,S::SpectralTransform) = S.riemann_weights
```
so that eventually we can support any grid with a single spectral transform and the specialities of the different grids are defined elsewhere (in `grids.jl`) using dispatch to compile the code automatically depending on the grid type.

What already works is (because `AbstractGrid <: AbstractVector`) and because I've already propagated the indices to the underling data `Vector`
```julia
for ij in eachindex(G)
    G[ij]
end
```
meaning that whatever grid, loops over all grid points do not have to be changed depending on the grid.

PR #123, opened 2022-08-15
Parametrizations with ColumnVariables

milankl:
@white-alistair This PR contains the following
- a `ColumnVariables` struct that holds all necessary variables of a single atmospheric column to calculate the parametrizations, whatever further (column) variables needed should be added here
- the `tendencies_parametrizations!` function now first loops over the horizontal before calling all the parametrizations, not the opposite structure we had before
- for every grid point, the vertical column is first extracted, then compute all paremetrizations, then copy the tendencies back into the horizontal fields of `DiagnosticVariables.tendencies` (where they later can be spectral transformed from)
- consequently, most parametrizations should have a single loop like `for k in eachlayer(column)` and no loop over the horizontal

all of this saves memory (as the column struct can be reused for the next grid point), and makes it easily parallelisable: the `for ij in eachgridpoint(diagn)` is the single loop with independent iterations that can be distributed across as many processors as we want (and < number of grid points)

- I've started reworking `large_scale_condensation.jl` to that new structure to give you an idea how to work with it. You can absolutely take it from here and change things as you see fit.


src/large_scale_condensation.jl
```diff
@@ -1,46 +1,99 @@
...
-        RH_threshold = RH_thresh_max + RH_thresh_range * (σₖ^2 - 1)  # Relative humidity threshold for condensation (Formula 24)
+    # skip the planetary boundary layer, k_lsc is lowest level for precip 
+    for k in eachlayer(column)[k_lsc:end]
+        
+        # Relative humidity threshold for condensation (Formula 24)
+        σₖ² = σ_levels_full[k]^2
+        RH_threshold = RH_thresh_max + RH_thresh_range * (σₖ² - 1)
         if k == nlev
             RH_threshold = max(RH_threshold, RH_thresh_boundary)
         end
```
milankl, line 24: These seem to be constants, for now fine, but we can think about precomputing `RH_threshold` somewhere

src/large_scale_condensation.jl
```diff
@@ -1,46 +1,99 @@
...
-            end
+        humid_tend_max = 10σₖ² / 3600humid_relax_time
+        humid_threshold = RH_threshold * sat_spec_humidity[k]  # Specific humidity threshold for condensation
+
+        if humid[k] > humid_threshold
+            # accumulate in tendencies (nothing is added if humidity not above threshold)
+            humid_tend[k] += -(humid[k] - humid_threshold) / humid_relax_time       # Formula 22
+            temp_tend[k] += -alhc / cp * min(humid_tend[k], humid_tend_max*pres)    # Formula 23
+            column.cloud_top = min(column.cloud_top, k)                             # Page 7 (last sentence)
         end
```
milankl, line 36: The `else` case is not needed anymore as for every grid point we start with zeros in `humid_tend, temp_tend` and can therefore just accumulate in. In `ColumnVariables` `cloud_top` is now just an `Int=0`, so if I understand line 35 correctly, we should initialize this variable with `nlev+1` and only if the threshold is exceed for a lower layer then `cloud_top` is changed? Note that scalars like `column.cloud_top` are immutable (although `column` is mutable) hence the typical `@unpack cloud_top = column` will not update `column.cloud_top`, hence the explicit `column.cloud_top` here.

src/large_scale_condensation.jl
```diff
@@ -1,46 +1,99 @@
...
-        Δpₖ = pres_grid * σ_levels_thick[k]  # Formula 4
-        for j = 1:nlat, i = 1:nlon
-            precip_large_scale[i, j] += -1 / gravity * Δpₖ[i, j] * humid_tend_lsc[i, j, k]  # Formula 25
-        end
+    column.precip_large_scale = 0
+    for k in eachlayer(column)[k_lsc:end]
+        Δpₖ = pres*σ_levels_thick[k]                                        # Formula 4
+        column.precip_large_scale += -1 / gravity * Δpₖ * humid_tend[k]     # Formula 25
+    end
+end
```
milankl, line 45: Do I understand this correctly that `precip_large_scale` is only computed here for output purposes but not actually used further? E.g. it's not written into any tendencies

src/tendencies_parametrizations.jl
```diff
@@ -1,23 +1,40 @@
...
+        get_column!(column,diagn,ij,G)
 
-    # The prognostic variable pres has units of log(hPa), whereas the grid-point pressure
-    # field which we require for the parametrizations has units of hPa, so we take the
-    # exponential here.
-    @. pres_grid = exp(pres_grid)
+        fill!(column.u_tend,0)      # set tendencies to 0 to accumulate +=
+        fill!(column.v_tend,0)      # parametrizations into
+        fill!(column.temp_tend,0)
+        fill!(column.humid_tend,0)
```
milankl, line 26: we may want to add more zero fills here to guarantee that reusing `column` for the next grid point doesn't cause issues. E.g. at the moment we set `precip_large_scale` to zero in `large_scale_condensation!`, maybe it makes sense to define a single point were the state of `column` is reset in order to be reused for the next grid point

src/large_scale_condensation.jl
milankl: resolved with later commits.

src/large_scale_condensation.jl
milankl: We should add a `precip_large_scale` field to DiagnosticVariables if we want to output that field.

src/large_scale_condensation.jl
```diff
@@ -1,46 +1,100 @@
...
+
+    @unpack humid, pres = column            # prognostic variables: specific humidity, surface pressure
+    @unpack temp_tend, humid_tend = column  # tendencies to write into
 
     # 1. Tendencies of humidity and temperature due to large-scale condensation
-    for k = k_lsc:nlev  # Used to be 2:nlev in original Speedy
-        σₖ = σ_levels_full[k]
-        RH_threshold = RH_thresh_max + RH_thresh_range * (σₖ^2 - 1)  # Relative humidity threshold for condensation (Formula 24)
+    # skip the planetary boundary layer, k_lsc is lowest level for precip 
+    for k in eachlayer(column)[k_lsc:end]
```
milankl, line 17: I think you mentioned that `k_lsc` is 2 in Fortran-SPEEDY, meaning the top layer is skipped (not planetary boundary, my bad) Do you know how this relates to `n_stratosphere_levels=2` which was also hard-coded in Fortran-SPEEDY? Just wondering whether they are supposed to be the same, as in, no condensation in the stratosphere.

milankl:
I don't understand why [f726659](https://github.com/milankl/SpeedyWeather.jl/pull/123/commits/f726659e8ca84a4a6e9f11451493aa0366668b33) is failing. We define
```julia
ColumnVariables(;kwargs...) = ColumnVariables{Float64}(;kwargs...)
```
and then the error says you can't do `ColumnVariables(nlev=8)` because `NF` is not defined...?! Commenting the failing lines, let me know if you can reproduce this error locally (I can't).

src/tendencies_parametrizations.jl
```diff
@@ -1,23 +1,35 @@
...
-    Diag::DiagnosticVariables{NF},
-    M,
-) where {NF<:AbstractFloat}
-    @unpack pres_grid = Diag.grid_variables
+function parametrization_tendencies!(   diagn::DiagnosticVariables{NF},
+                                        M::PrimitiveEquationModel,
+                                        ) where NF
+
+    G = M.geometry
+    column = ColumnVariables{NF}(nlev=diagn.nlev)
```
white-alistair, line 16: Will we need to make this thread-safe when it comes to parallelisation? I'm more familiar with multi-threading rather than multi-processing so not 100% sure how it'd work in the latter case.

src/tendencies_parametrizations.jl
```diff
@@ -1,23 +1,35 @@
...
+        # vertical_diffusion!(column,M)
 
-    #Calculate
-    #utend: u-wind tendency (gp)
-    #vtend: v-wind tendency (gp)
-    #ttend: temp. tendency (gp)
-    #htend: spec. hum. tendency (gp)
+        # write tendencies from parametrizations back into horizontal fields
+        write_column_tendencies!(diagn,column,ij)
+    end
```
white-alistair, line 34: :+1:

test/column_variables.jl
```diff
@@ -0,0 +1,45 @@
+@testset "ColumnVariables initialisation" begin
+
+    # no number format provided
+    # column = ColumnVariables(nlev=8)
+    # @test eltype(column.temp) == Float64
+
+    @testset for NF in (Float16,Float32,Float64)
```
white-alistair, line 7: This test includes `Float16` but the following don't.

test/large_scale_condensation.jl
```diff
@@ -1,8 +1,48 @@
...
+    end
+end
+
+@testset "Parametrization: large scale condensation" begin
+    @testset for NF in (Float32,Float64)
+        _, diagn, model = SpeedyWeather.initialize_speedy(NF,model=:primitive)
+
+        column = ColumnVariables{NF}(nlev=diagn.nlev)
+
+        for ij in SpeedyWeather.eachgridpoint(diagn)
```
white-alistair, line 42: If the purpose of this test is to test the function `large_scale_condensation!` then I don't think we need this loop.

src/large_scale_condensation.jl
white-alistair: > In `ColumnVariables` `cloud_top` is now just an `Int=0`, so if I understand line 35 correctly, we should initialize this variable with `nlev+1` and only if the threshold is exceed for a lower layer then `cloud_top` is changed? 

This is from fortran speedy where, if I remember correctly, `cloud_top == nlev + 1` means no clouds. We could easily change that.

> Note that scalars like `column.cloud_top` are immutable (although `column` is mutable) hence the typical `@unpack cloud_top = column` will not update `column.cloud_top`, hence the explicit `column.cloud_top` here.

Good to know, that could easily lead to some nasty bugs!

white-alistair (approved):
Very nice work, I like the overall structure, only minor comments and questions.

src/large_scale_condensation.jl
white-alistair: They are indeed the same. I discovered this line from the original speedy documentation recently:

>  In practice, the top two and the bottom layer provide a ”bulk” representation of the stratosphere and the
planetary boundary layer (PBL) respectively.

src/large_scale_condensation.jl
milankl: Yes we now have cloud_top = nlev+1 at initialisation of `ColumnVariables`, otherwise the `min(cloud_top,k)` doesn't work.

src/large_scale_condensation.jl
milankl: I'll add two fields to DiagnosticVariables.surface, `precip_large_scale` and `precip_convection` and then we can see whether we want to output them?

src/tendencies_parametrizations.jl
milankl: We have `reset_column!` now that does that.

src/column_variables.jl
```diff
@@ -1,34 +1,112 @@
...
+@with_kw mutable struct ColumnVariables{NF<:AbstractFloat}
+
+    # COORDINATES
+    lat::NF = 0                         # latitude [˚N], needed for radiation?
+    lon::NF = 0                         # longitude [˚E], needed for radiation?
+    nlev::Int = 0                       # number of vertical levels
+
+    # PROGNOSTIC VARIABLES
+    vor::Vector{NF} = zeros(NF,nlev)    # vorticity, actually needed in column?
+    div::Vector{NF} = zeros(NF,nlev)    # divergence, actually needed in column?
```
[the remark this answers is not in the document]
milankl: I'll remove them

src/column_variables.jl
```diff
@@ -1,34 +1,112 @@
...
+    # TENDENCIES to accumulate the parametrizations into
+    u_tend::Vector{NF} = zeros(NF,nlev)
+    v_tend::Vector{NF} = zeros(NF,nlev)
+    temp_tend::Vector{NF} = zeros(NF,nlev)
+    humid_tend::Vector{NF} = zeros(NF,nlev)
+
+    # DIAGNOSTIC VARIABLES
+    ## HUMIDITY AND CLOUDS
+    sat_vap_pres::Vector{NF} = zeros(NF,nlev)   # Saturation vapour pressure
+    sat_spec_humid::Vector{NF} = zeros(NF,nlev) # Saturation specific humidity
```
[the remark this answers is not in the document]
milankl: Yes, I'll rename it now in my next commit.

src/column_variables.jl
```diff
@@ -1,34 +1,112 @@
...
+
+    # surface pressure (logarithm used in dynamics, convert back here)
+    C.log_pres = D.surface.pres_grid[ij]
+    C.pres = exp(C.log_pres)
+
+    @inbounds for (k,layer) =  enumerate(D.layers)
+        C.vor[k] = layer.grid_variables.vor_grid[ij]
+        C.div[k] = layer.grid_variables.div_grid[ij]
+        C.u[k] = layer.grid_variables.U_grid[ij]*coslat⁻¹
+        C.v[k] = layer.grid_variables.V_grid[ij]*coslat⁻¹
```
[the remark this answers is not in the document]
milankl: `U,V` are `u*coslat, v*coslat` and are used in the dynamics. The coslat scaling is only undone in functions like `vorticity_fluxes!`. In order to have `u,v` here (which I assume is needed for the surface fluxes or if we want to parametrize any turbulent drag in the planetary boundary layer) we convert `U,V` to `u,v` here.

src/large_scale_condensation.jl
```diff
@@ -1,46 +1,101 @@
...
-                humid_tend_lsc[i, j, k] = zero(NF)
-                temp_tend_lsc[i, j, k] = zero(NF)
-            end
+        humid_tend_max = 10σₖ² / 3600humid_relax_time
+        humid_threshold = RH_threshold * sat_spec_humid[k]      # Specific humidity threshold for condensation
+
+        if humid[k] > humid_threshold
+            # accumulate in tendencies (nothing is added if humidity not above threshold)
+            humid_tend[k] += -(humid[k] - humid_threshold) / humid_relax_time       # Formula 22
+            temp_tend[k] += -alhc / cp * min(humid_tend[k], humid_tend_max*pres)    # Formula 23
```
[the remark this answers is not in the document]
milankl: Why does the order matter? Happy to introduce tendencies from several parametrizations separately, but didn't see a need to do that at the moment.

src/large_scale_condensation.jl
```diff
@@ -1,46 +1,101 @@
...
+function get_saturation_vapour_pressure!(   column::ColumnVariables{NF},
+                                            model::PrimitiveEquationModel{NF}
+                                            ) where {NF<:AbstractFloat}
+
+    @unpack sat_vap_pres, temp = column
+    @unpack e₀, T₀, C₁, C₂, T₁, T₂ = model.parameters.magnus_coefs
+
+    for k in eachlayer(column)
+        # change coefficients for water (temp > T₀) or ice (else)
+        C₁or₂, T₁or₂ = temp[k] > T₀ ? (C₁,T₁) : (C₂,T₂)
```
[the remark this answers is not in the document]
milankl: Done

src/tendencies_parametrizations.jl
milankl: Yes, but that should be just making sure that every thread/processor has its own copy of column so that they don't interfere

test/column_variables.jl
milankl: Problem is that we still can't `initialize_speedy(Float16)` as I haven't yet included GenericFFT.jl, see #31. So Float16 testing wherever possible, but FFTW only allows for Float32/64.

test/large_scale_condensation.jl
milankl: True, doesn't seem to hurt though?

src/large_scale_condensation.jl
milankl: But "no condensation in the stratosphere" would mean that the loop should be `for k in eachlayer(column)[n_stratosphere_levels+1:end]` otherwise we'd allow condensation in the lowermost stratosphere level (which might be fine?) Probably a question for Franco Molteni...

src/large_scale_condensation.jl
milankl: Done with later commits 👇🏼

PR #124, opened 2022-08-15
Grid abstractions +indexing, SpecTrans performance

milankl:
Creates a grid hierarchy for easier extension to future grids. We now have
- FullGaussianGrid and FullLatLonGrid <: AbstractFullGrid <: AbstractGrid
- OctahedralGaussianGrid <: AbstractOctahedralGrid <: AbstractGrid
- HEALPixGrid <: AbstractHEALPixGrid <: AbstractGrid

Such that it should be (super) easy to define new grids as subtypes of those. E.g. we could define a OctahedralClenshawGrid (Octahedral but with equi-angle latitudes) like so
```julia
struct OctahedralClenshawGrid <: AbstractOctahedralGrid end
get_colat(::Type{<:OctahedralClenshawGrid},nlat_half::Integer) = get_colat(FullLatLonGrid,nlat_half)
get_quadrature_weights(::Type{<:OctahedralClenshawGrid},nlat_half::Integer) = clenshaw_curtis_weights(nlat_half)
```
and basically only the latitudes (here colatitudes) would need to be redefined and which quadrature to use. That's it. The indexing for example would be borrowed from its supertype etc. Neat 😎 

@hottad


milankl:
With the new indexing functions `eachring`,`eachgridpoint` and `each_index_in_ring` loops over grids can be constructed with `grid::AbstractGrid` (i.e. any gridded field)
```julia
for i in eachring(grid)
    for ij in each_index_in_ring(grid,i)
        grid[ij] += 1
    end
end
```
when looping over rings, because, e.g., one wants to access `f_coriolis[i]` in the outer loop that only changes with the latitude ring, but not with the indices within. Both loops get be put together with
```julia
for ij in eachgridpoint(grid)
    grid[ij]
end
```

PR #125, opened 2022-08-19
LowerTriangularMatrix extended

milankl:


milankl:
suggested by @samhatfield

milankl:
@samhatfield While in general I think the idea to throw an error at `[i,j]` with $j>i$ is good, I've been struggeling to get this pass all tests and therefore think it may not be a good idea.

The underlying problem is that we've defined `LowerTriangularMatrix <: AbstractMatrix` which is nice, because that means a lot of methods that are already defined for `AbstractMatrix` don't have to be reimplemented, however, it also means that _many_ methods indeed assume that operations like `setindex!(::LowerTriangularMatrix,x,i,j)` are allowed for any `i,j` within `size(::LowerTriangularMatrix)` which in our case with the `BoundsError` for j>i isn't true.

For example with `L::LowerTriangularMatrix`
```julia
L .*= 1
```
may throw an error if that's implemented as a loop over `eachindex(L::AbstractMatrix)`. But in general there are quite some problems with these seemingly trivial operations, check for example
```julia
julia> alms
3×3 LowerTriangularMatrix{ComplexF64}:
 -0.866373-0.547021im       0.0+0.0im           0.0+0.0im
  -0.23095+0.897737im  0.419995-1.1278im        0.0+0.0im
 -0.128295+0.268862im  0.440973-0.39582im  0.939507+0.483643im

julia> alms*1
3×3 Matrix{ComplexF64}:
 -0.866373-0.547021im       0.0+0.0im           0.0+0.0im
  -0.23095+0.897737im  0.419995-1.1278im        0.0+0.0im
 -0.128295+0.268862im  0.440973-0.39582im  0.939507+0.483643im
```
which for some reasons triggers a conversion from `LowerTriangularMatrix` to `Matrix`. Then I've also observed
```julia
julia> alms = LowerTriangularMatrix(randn(ComplexF64,3,3))
3×3 LowerTriangularMatrix{ComplexF64}:
  -0.274421-1.20264im        0.0+0.0im             0.0+0.0im
 -0.0600222-1.42869im    1.29328+0.181297im        0.0+0.0im
 -0.0453674-0.385649im  -0.33839+1.31662im   -0.321018+0.430279im

julia> alms .*= 2
3×3 LowerTriangularMatrix{ComplexF64}:
 -0.548842-2.40529im  0.0+0.0im        0.0+0.0im
 -0.120044-2.85739im  0.0+0.0im        0.0+0.0im
       0.0+0.0im      0.0+0.0im  -0.642036+0.860557im
```
which suddenly produces zeros where there shouldn't be any (this happens when `i,j` are used with an @inbounds for `j>i`, e.g.
```julia
julia> f(L,x,i,j) = @inbounds L[i,j] = x
julia> alms[1,2] = 1
ERROR: BoundsError: attempt to access 3×3 LowerTriangularMatrix{ComplexF64} at index [1, 2]

julia> f(alms,1,1,2)
julia> alms
3×3 LowerTriangularMatrix{ComplexF64}:
   1.09586-0.462199im       0.0+0.0im              0.0+0.0im
 -0.286196-0.178535im   1.08801-0.0100021im        0.0+0.0im
       1.0+0.0im       0.287399-0.376378im   -0.348464-0.972782im
```
The 1 isn't placed in [1,2] but in [3,1] instead...

Long story short: There's clearly more work that needs to be done for a safe implementation of `LowerTriangularMatrix`. Maybe we should even define it as `<: AbstractVector` and add [i,j] indexing as a bonus (which therefore wouldn't be assumed in Julia's Base or elsewhere).

PR #127, opened 2022-08-23
Grid-flexible spectral transform

### LINKED ISSUE #112: A new grid for SpeedyWeather.jl?
At the moment we use a regular Gaussian grid for grid-point space. This comes with advantages of easy storage in a matrix internally and for output but with the disadvantage that a lot of points are put near the poles that aren't actually needed. Alternatives are
- a reduced Gaussian grid (https://confluence.ecmwf.int/display/FCST/Gaussian+grids)
- an octahedral Gaussian grid (https://github.com/jmert/CMB.jl/issues/71)
- a HEALPix grid (https://github.com/jmert/CMB.jl/pull/68, https://en.wikipedia.org/wiki/HEALPix, https://arxiv.org/pdf/astro-ph/0412607.pdf)

This is an issue to discuss and collect ideas and reasons to implement either of these grids at some point in the future.

milankl:
This PR is intended to fix #112 by allowing for a flexible `grid<:AbstractGrid` to be used in the spectral transform.


milankl:
@hottad at the moment this PR lets one use any `grid<:AbstractGrid` without errors, but the transforms are not accurate yet for Octahedral/HEALPix. I assume I got some normalisation in the grid->spectral transform wrong. For a `FullGaussianGrid` we have
```julia
julia> alms = LowerTriangularMatrix(randn(ComplexF64,33,32));
julia> alms[:,1] .= real.(alms[:,1]);
julia> map = gridded(alms,grid=FullGaussianGrid);
julia> alms2 = spectral(map);
julia> imshow(log10.(abs.(alms-alms2)))
```
an absolute error that is O(1e-14) with Float64. That's good.
image
with `FullClenshawGrid` similar, that's great too, except for the last row. I assume this is because we use lmax+2,mmax+1 LowerTriangularMatrices for the vector quantities, but so that's technically T32 which violates the exactness of the Clenshaw-Curtis quadrature with 64 latitudes. I'm not worried about this as the last row is only included for consistent matrix sizes (and the issue disappears if one starts with the lmax+2 row being zeros).
image

However, for Octahedral/HEALPix the absolute errors are O(1e-2) and larger. I believe this is because I messed up the normalisation with the solid angle $\Delta \Omega = sin(\theta) \Delta \theta \Delta \phi$ that a grid point covers. For the full Gaussian grids I use (96x48)
```julia
julia> nlat_half = 24
julia> w = SpeedyWeather.get_quadrature_weights(FullGaussianGrid,nlat_half);
julia> w *= π/2nlat_half
24-element Vector{Float64}:
 0.0002063860165031381
 0.0004795872813548503
 0.0007511832466177125
 ...
 0.0041838316317469116
 0.00421930059577275
 0.004237072348271895
```
so quadrature weights *pi/nlat and the same for the Clenshaw-Curtis grid but with other weights obviously. I believe this has to be changed for Octahedral/HEALPix. I'll try to figure this out in the next days.

milankl:
@hottad With the latest commit the transform errors on the HEALPix grid are much smaller

image

from
```julia
julia> alms = LowerTriangularMatrix(randn(ComplexF64,33,32))
julia> alms[end,:] .= 0
julia> alms[:,1] .= real.(alms[:,1])
julia> map = gridded(alms,grid=HEALPixGrid)
julia> alms2 = spectral(map)
julia> maximum(abs.(alms)-abs.(alms2))
0.018819099369166703

julia> imshow(abs.(alms-alms2))
```
but only when the lon offset rotation is deactivated

src/grids.jl
```diff
@@ -364,13 +379,29 @@ eachgridpoint(grid::G) where {G<:AbstractGrid} = Base.OneTo(get_npoints(G,get_nr
...
+# get_quadrature_weights(::Type{<:HEALPixGrid},nside::Integer) =
+#                 nside_assert(nside) ? 2/12nside^2*[min(4iring,4nside) for iring in 1:2nside] : nothing
+function get_quadrature_weights(::Type{<:HEALPixGrid},nside::Integer)
+    colat = get_colat(HEALPixGrid,nside)
+    sin_colat_Δθ = zero(colat)
+    sin_colat_Δθ[1] = sin(colat[1])*colat[2]/2
+    for j in 2:length(colat)-1
+        sin_colat_Δθ[j] = sin(colat[j])*(colat[j+1]-colat[j-1])/2
+    end
+    sin_colat_Δθ[end] = sin(colat[end])*(π-colat[end])/2
```
milankl, line 399: These are the best quadrature weights for the HEALPix grid I was sofar able to come up with.

src/spectral_transform.jl
```diff
@@ -434,70 +361,88 @@ function gridded!(  map::AbstractGrid{NF},                      # gridded output
...
+            Legendre.unsafe_legendre!(Λw, Λ, lmax, mmax, cos_colat[ilat_n]) : Λs[ilat_n]
+        quadrature_weight = quadrature_weights[ilat_n]  # weights normalised with π/nlat
 
         lm = 1                                          # single index for spherical harmonics
-        for m in 1:mmax+1                               # Σ_{m=0}^{mmax}, but 1-based index
+        for m in 1:min(nfreq,mmax+1)                    # Σ_{m=0}^{mmax}, but 1-based index
+
+            # QUADRATURE WEIGHTS and LONGITUDE OFFSET
+            w = lon_offsets[m,ilat_n]                   # longitude offset rotation
+            # quadrature_weight *= conj(w)                # complex conjugate for back transform
```
milankl, line 441: However, the longitude offset rotation had to be deactivated. In general we do here
```julia
    _, lons = get_colatlons(grid,nresolution)
    lon1s = [lons[each_index_in_ring(grid,i,nresolution)[1]] for i in 1:nlat]
    lon_offsets = [cispi(m*lon1/π) for m in 0:mmax, lon1 in lon1s]
```
with `lon1s` the longitude in radians of the first grid point per ring.

milankl:
@hottad While I'm still not sure about the accuracy of the HEALPix spectral transform, I'm tempted to merge this once all tests pass as all the other structure has been set up such that you can start contributing/experimenting and do whatever you'd like to test with new grids in SpeedyWeather.jl, I think we probably just ran the first HEALPix atmospheric simulation? `FullGaussianGrid` will remain default for now and we can independently improve the transform and support more grids. For now I'd suggest to keep all grid-related stuff in `src/grids.jl` but eventually we could even go to `src/Grids/grids.jl, healpix.jl, octahedral.jl` or outsurce all of that into its won package. But I think that's an overkill for now.

image

hottad:
@milankl Fantastic! 
Thank you for the great job! So shallow water on FllClenshawGrid, OctahedralGaussianGrid and HEALPIxGrid now runs up to at least 10 days and the results are all visually very similar. Impressive!
I will fork this project and  start experimenting probably next week. Congrats!

milankl:
Yeah, the HEALPixGrid still doesn't have the correct longitude offset rotation implemented, that's something that I need to play around with. As you suggest, it's probably some implementation bug that I don't currently understand as the mathematics are rather clear I believe. However, it's still good enough to run even 1000 days or longer and also at different resolution.

PR #128, opened 2022-08-25
Convection

white-alistair:


src/column_variables.jl
```diff
@@ -106,12 +128,20 @@ function reset_column!(column::ColumnVariables{NF}) where NF
     fill!(column.temp_tend,0)
     fill!(column.humid_tend,0)
 
+    # Convection
     column.cloud_top = column.nlev+1
-    column.precip_large_scale = zero(NF)
+    column.conditional_instability = false
+    column.activate_convection = false
```
white-alistair, line 134: @milankl I decided to use these flags instead of the convention that `cloud_top == nlev+1` means no convection, as I think it's cleaner, safer, and more easier to understand!

src/tendencies_parametrizations.jl
```diff
@@ -8,20 +8,23 @@ Extract for each vertical atmospheric column the prognostic variables
...
 then write the tendencies back into a horizontal field of tendencies.
 """
-function parametrization_tendencies!(   diagn::DiagnosticVariables{NF},
-                                        M::PrimitiveEquationModel,
-                                        ) where NF
-
+function parametrization_tendencies!(
+    diagn::DiagnosticVariables{NF},
+    M::PrimitiveEquationModel,
+) where {NF}
```
milankl, line 14: Hilarious how we have the constant battle how to indent the function arguments! Just to clarify, I don't mind either way 😉

src/thermodynamics.jl
```diff
@@ -0,0 +1,216 @@
...
+            (A_full_level[k+1] - A_full_level[k]) *
+            (log(σ_levels_half[k+1]) - log(σ_levels_full[k])) /
+            (log(σ_levels_full[k+1]) - log(σ_levels_full[k]))
+    end
+
+    # Compute the values at the surface separately
+    A_half_level[nlev] =
+        A_full_level[nlev] +
+        (A_full_level[nlev] - A_full_level[nlev-1]) *
+        (log(0.99) - log(σ_levels_full[nlev])) /
```
milankl, line 85: `log0.99 = convert(NF,log(0.99))` directly after the `@unpack`s instead of `log(0.99)` here?

src/thermodynamics.jl
```diff
@@ -0,0 +1,216 @@
...
+        # change coefficients for water (temp > T₀) or ice (else)
+        C, T = temp[k] > T₀ ? (C₁, T₁) : (C₂, T₂)
+        sat_vap_pres[k] = e₀ * exp(C * (temp[k] - T₀) / (temp[k] - T))
+    end
+
+    return nothing
+end
+
+"""
+saturation_specific_humidity!(  column::ColumnVariables{NF},
```
milankl, line 124: Note they have to be indendent to be correctly displayed in the documentation

milankl (commented):
Fantastic! Sorry there's a few comments I should have just directly commited corrections to this branch. Will do next time.

src/column_variables.jl
milankl: Nice, yeah!

src/tendencies_parametrizations.jl
white-alistair: Haha blame [JuliaFormatter.jl](https://github.com/domluna/JuliaFormatter.jl), not me ;)

milankl: Once you've touched every file we'll have a consistent style! YAY!

src/thermodynamics.jl
white-alistair: How about just doing `log(NF(0.99))` here? It's outside of the loop so only gets executed once.

src/column_variables.jl
```diff
@@ -106,12 +128,20 @@ function reset_column!(column::ColumnVariables{NF}) where NF
...
+    column.excess_humidity = zero(NF)
     column.precip_convection = zero(NF)
+    column.cloud_base_mass_flux = zero(NF)
+    fill!(column.net_flux_humid, 0)
+    fill!(column.net_flux_dry_static_energy, 0)
 
+    # Large-scale condensation
+    column.precip_large_scale = zero(NF)
     return nothing
```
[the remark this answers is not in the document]
white-alistair: I think they all need to be reset. In the case of convection, if convection is not activated in a grid cell, then the column variables from the previous grid cell would simply be carried over, meaning we would get some erroneous tendencies and precipitation due to convection.

That said, I think we can get away without resetting `excess_humidity` and `cloud_base_mass_flux`.

src/constants.jl
```diff
@@ -73,6 +82,9 @@ function Constants(P::Parameters)
                             Δt,Δt_unscaled,Δt_sec,Δt_hrs,
                             robert_filter,williams_filter,n_timesteps,
                             output_every_n_steps, n_outputsteps,
-                            drag_strat, RH_thresh_boundary, RH_thresh_range,
-                            RH_thresh_max, humid_relax_time)
+                            drag_strat, RH_thresh_pbl_lsc, RH_thresh_range_lsc,
+                            RH_thresh_max_lsc, humid_relax_time_lsc, pres_thresh_cnv,
+                            RH_thresh_pbl_cnv, RH_thresh_trop_cnv, humid_relax_time_cnv,
+                            max_entrainment, ratio_secondary_mass_flux,
+                            )
```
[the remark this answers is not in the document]
white-alistair: As discussed, let's spin this off into a separate issue for overhauling the general approach to parameters and constants.

src/convection.jl
```diff
@@ -0,0 +1,204 @@
...
+        humid_threshold_above_pbl = RH_thresh_pbl_cnv * sat_humid[nlev-1]
+
+        # The range of this loop requires clarification, but in its current form it means
+        # that the top-of-convection level may be any tropospheric level, excluding the two
+        # layers directly above the PBL.
+        for k = (nlev-3):-1:3
+            # Condition 1: Conditional instability (MSS in PBL < MSS at this half-level)
+            if sat_moist_static_energy_pbl > sat_moist_static_energy_half[k]
+                column.conditional_instability = true
+                column.cloud_top = k
```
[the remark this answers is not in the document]
white-alistair: Yes, in the current implementation (as in fortran speedy) convection needs to be executed first. The cloud-top can then be adjusted (upwards) by large-scale condensation if the cloud-top due to LSC is higher.

src/convection.jl
```diff
@@ -0,0 +1,204 @@
...
+    # Downward fluxes at upper boundary
+    flux_down_humid = mass_flux * humid_top_of_pbl
+    flux_down_dry_static_energy = mass_flux * dry_static_energy_half[nlev-1]
+
+    # Net flux
+    net_flux_dry_static_energy[nlev] = flux_down_dry_static_energy - flux_up_dry_static_energy
+    net_flux_humid[nlev] = flux_down_humid - flux_up_humid
+
+    # 2. Fluxes for intermediate layers
+    for k = (nlev-1):-1:(cloud_top+1)
```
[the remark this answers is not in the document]
white-alistair: Good to know. In this case the order is important. As it happens my first implementation, which tracked more closely my understanding of the scheme as it's described in the documentation, had the order in the opposite direction :)

src/convection.jl
```diff
@@ -0,0 +1,204 @@
...
+        if Δhumid > 0.0
+            Δflux_humid = ratio_secondary_mass_flux * cloud_base_mass_flux * Δhumid
+            net_flux_humid[k] += Δflux_humid
+            net_flux_humid[nlev] -= Δflux_humid
+        end
+    end
+
+    # 3. Fluxes for top-of-convection layer
+    # Flux of convective precipitation
+    column.precip_convection = max(flux_up_humid - mass_flux * sat_humid_half[cloud_top], 0.0)
```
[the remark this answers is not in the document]
white-alistair: Or just `0`?

src/convection.jl
```diff
@@ -0,0 +1,204 @@
...
+    # 1. Fluxes in the PBL
+    humid_top_of_pbl = min(humid_half[nlev-1], humid[nlev])   # Humidity at the upper boundary of the PBL
+    max_humid_pbl = max(1.01 * humid[nlev], sat_humid[nlev])  # Maximum specific humidity in the PBL
+
+    # Cloud-base mass flux
+    Δp = pres_ref * pres * σ_levels_thick[nlev]  # Pressure difference between bottom and top of PBL
+    mass_flux =
+        Δp / (gravity * 3600humid_relax_time_cnv) *
+        min(1.0, 2.0 * (pres - pres_thresh_cnv) / (1 - pres_thresh_cnv)) *
+        min(5.0, excess_humidity / (max_humid_pbl - humid_top_of_pbl))
```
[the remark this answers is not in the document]
white-alistair: Very interesting, thanks!

src/column_variables.jl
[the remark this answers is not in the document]
milankl: I'd say let's go the safe way and reset everything for the next column. That's the cleanest, probably takes 0 time, and then nothing unexpected happens.

src/convection.jl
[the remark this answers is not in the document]
milankl: Maybe add a comment on that? Not that someone at some points wants to change that and is then surprised?

src/convection.jl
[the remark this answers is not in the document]
milankl: Yes, you can always use integers. Their conversion is even compiled away which you can check via
```julia
julia> f(x::Float32) = 2x
julia> @code_llvm f(1f0)
;  @ REPL[8]:1 within `f`
define float @julia_f_336(float %0) #0 {
top:
; ┌ @ promotion.jl:380 within `*` @ float.jl:404
   %1 = fmul float %0, 2.000000e+00
; └
  ret float %1
}
```
So the compiler decided to do the promotion for you.

src/thermodynamics.jl
milankl: Yes, also fine! Any clue why 0.99? Seems like a weird hard-coded constant to me?!

white-alistair: Yeah, good question, I can't immediately see me why you wouldn't just use `log(σ_levels_half[nlev])`, i.e. `0`.

white-alistair:
@milankl I think this is ready for another review now. If you're happy with the changes then feel free to give a review with approval and I'll merge.

milankl (approved):

src/thermodynamics.jl
milankl: Probably some weird trick that no one bothered commenting...:see_no_evil:

Issue #134, opened 2022-08-31
Output grids


milankl:
Given that we have now the flexibility to use various grids, we need to think about how to output gridded data to netcdf. Several issues

- The octahedral grids cannot be reshaped/raveled into a matrix, although this is possible for the full grids and the healpix grids
- If the latitudes of the compute grid and the output grid don't match then the Legendre polynomials have to be recomputed. That's either expensive or doubles our memory footprint because two sets of Legendre polynomials have to be kept in memory

So I see several possibilities

1) We always output to `FullGaussianGrid` (currently), but that's expensive and considerably slows down the simulation
2) We output to the full grid that corresponds to the compute grid. E.g. `OctahedralGaussianGrid`-`FullGaussianGrid`, `OctahedralClenshawGrid`-`FullClenshawGrid` meaning we would need to define a full version to the HEALPix grid and create a `S_output = SpectralTransform(S,output_grid=...)` which changes parameters like `nlon` but points to the already precomputed Legendre polynomials
3) Like 2) but we output the HEALPix grids as a matrix by concatenating their basepixel squares.
4) We don't do some spectral transform onto the output grid and interpolate on the spatial grid instead, but that requires a `interpolate!(grid::Grid,lat,lon)` function for every `Grid<:AbstractGrid`, which, however, we may need eventually anyway if we want semi-Lagrangian advection at some point

I don't like 1) what we currently do, hence this issue. I do like 2) and 3) and think that's the simplest to implement. 4) is possible but sounds more like a long-term solution. I reckon it's faster than 2) but not than 3) (which is just memory reordering) 

@hottad do you agree?

hottad:
I agree. If the purpose of output is only visualization or casual post-processing, then the option 2) would be most preferable.
Implementing a full grid corresponding to each reduced grid (like octahedral or HEALPix) is quite easy, and maintenance of such code should not be too tedious.
Note that, for this, we do not need `SpectralTransform(S,output_grid=...)` (a full spherical harmonics transform) because doing direct Fourier transform on each reduced latitude circle and then doing inverse transform onto a full circle (which is essentially zonal interpolation) should suffice. If we do like this, then 2) may not be necessarily slower than 4).
For more serious applications beyond visualization, like coupling with data assimilation, we should provide capacity to dump the model state on the native grid.

milankl:
Note to self: Data on an octahedral Clenshaw grid can also (for each of the 4 basepixels) be output like
```julia
julia> A
[...]

PR #138, opened 2022-09-04
add OctahedralClenshawGrid + some refactoring of src/grids.jl

hottad:
@milankl With this PR I suggest some minor refactoring on `src/grids.jl`. I also added support for `OctahedralClenshawGrid`.
I suggest to move the definition of the abstract type `AbstractGrid` to `src/grids_general.jl` and also define interface methods there (which simply raises "undefined" error), followed by methods that apply universally to any subtypes of `AbstractGrid`.
I also reordered the code in `src/grids.jl` so that all the interface methods required for each concrete Grid type are defined immediately after that Grid type's definition.
I thought this will clarify what functions need to be implemented for a Grid and should be helpful when implementing a new concrete Grid type.
The obvious downside is that the code becomes longer.


src/grids_general.jl
```diff
@@ -0,0 +1,95 @@
...
+@inline Base.getindex(G::AbstractGrid,r::AbstractRange) = G.data[r]
+@inline Base.setindex!(G::AbstractGrid,x::AbstractVector,r::AbstractRange) = setindex!(G.data,x,r)
+
+
+# List all the methods required to be implemented for any Grid types here
+# and define the  default behaviour as raising errors
+# This would clarify what interface we need to implement when introducing a new Grid
+
+truncation_order(::Type{G}) where {G<:AbstractGrid} =
+    error("truncation_order(::Type{G}) not implemented for G==$G")
```
milankl, line 29: I'm not sure I like this. I see where you are coming from, but then I'd suggest don't `error("...not implemented")` but use a `throw(MethodError(f,args))` for that (because that's the type of error that's supposed to do that) but then that is exactly what happens anyway if you don't define the method in the first place. So I have another suggestion, which just writes all of that imformation explicitly down as a docstring """....""", that way one can simply do `?AbstractGrid` and gets that information printed into the REPL and it's also automatically included in the documentation.

```julia
"""
    abstract type AbstractGrid{T} <: AbstractVector end

The abstract supertype for all spatial grids on the sphere supported by SpeedyWeather.jl. Every new grid has to be of the form

    abstract type AbstractGridClass{T} <: AbstractGrid{T} end

    struct MyNewGrid{T} <: AbstractGridClass{T}
         data::Vector{T}     # all grid points unravelled into a vector
         nresolution::Int     # resolution parameter, like `nlat_half`, `nside` etc
    end

`MyNewGrid` should belong to a grid class like `AbstractFullGrid`, `AbstractOctahedralGrid` or
`AbstractHEALPixGrid` (that already exist but you may introduce a new class of grids) that share
certain features such as the number of longitude points per latitude ring and indexing, but may
have different latitudes or offset rotations. Each new grid `Grid` (or grid class) then has to
implement the following methods

Fundamental grid properties
    get_nresolution(::Grid)    # returns the resolution parameter
    get_npoints(::Grid)          # total number of grid points
    nlat_odd
    get_nlat
    get_nlat_half
    
Indexing
    get_nlon_max
    get_nlon_per_ring
    each_index_in_ring
    
Coordinates
    get_colat
    get_colatlon

Spectral truncation
    get_truncation
    get_resolution

Quadrature weights and solid angles
    get_quadrature_weights
    get_solid_angle
"""
AbstractGrid
```

Something like this?

src/grids.jl
```diff
@@ -93,19 +238,72 @@ one for each face of the octahedron. E.g. 20,24,28,32,...nlon-4,nlon,nlon,nlon-4
...
+# get_nresolution() is already implemented for AbstractOctahedralGrid
+nlat_odd(::Type{<:OctahedralGaussianGrid}) = false
+# get_nlon() is already implemented for AbstractOctahedralGrid
+# get_nlon_per_ring() is already implemented for AbstractOctahedralGrid
+get_npoints(::Type{<:OctahedralGaussianGrid},nlat_half::Integer) = npoints_octahedral(nlat_half,false)
+get_colat(::Type{<:OctahedralGaussianGrid},nlat_half::Integer) = get_colat(FullGaussianGrid,nlat_half)
+# get_lon() is already implemented for AbstractOctaherdalGrid
+# get_colatlons() is already implemented for AbstractOctaherdalGrid
+# each_index_in_ring() is already implemented for AbstractOctaherdalGrid
+get_quadrature_weights(::Type{<:OctahedralGaussianGrid},nlat_half::Integer) = gaussian_weights(nlat_half)
```
milankl, line 270: I've put a `get_solid_angles` function on top of `get_quadrature_weights`, such that in spectral_transform.jl we have one interface for all grids that give conceptually the same thing back. Before we had to scale some quadrature weights with nlon some not. Hence now, every grid also should provide  a `get_solid_angle` method which is in most cases `get_quadrature ./ nlon` but is directly 4pi/npoints for HEALPix. 

So there'll be some conflicts whichever branch is merged first, but we can resolve that

src/grids_general.jl
hottad: Thank you, Milan! Totally agreed. What you suggest is much more elegant and I like it. Thanks!

src/grids.jl
hottad: Great! This is more elegant, again.

src/grids_general.jl
milankl: Now in [50d47af](https://github.com/milankl/SpeedyWeather.jl/pull/138/commits/50d47af775fca6594085d9bc890409b4aecfcbb1)

src/grids.jl
milankl: merged in

milankl:
Fantastic work, I'll merge this. Daisuke, were you considering to implement the HEALPix4 grid? Because then I'd focus on #134

hottad:
Thank you Milan for merging this PR. I learnt a lot from how you improved my poor coding.
Yes, I am keen on implementing HEALPix4.
I am wondering whether I should define `HEALPix4Grid` as an immediate subtype of `AbstractGrid`, or is there any better way like introducing an intermediate abstract type between `HEALPix4Grid` and `AbstractGrid`.
I initially thought of something like `HEALPix4Grid <: AbstractOcatahedralGrid` (though this naming could be a bit confusing by the way) because HEALPix4 mostly inherits the structure of the octahedral grid, but this is perhaps impossible because `AbstractOctahedralGrid` assumes the `nlon_per_ring` starts from 20 closest to the pole, while in HEALPix4 it starts from just 4 points. Is there a smart way to modify `AbstractOctahedralGrid` so that it can accommodate any number (not restricted to 20) as the first `nlon_per_ring`?
Any thoughts?

milankl:
Anything but poor coding! I wish more people would just create such pull requests and contribue 😄 (there's always little things that get changed. If no one reviews my code, I usually only realise a few pull requests later that some of my commits weren't ideal)

I would probably just branch out another abstract subgroup of `AbstractGrid` like so
```julia
abstract type AbstractHEALPix4Grid{T} <: AbstractGrid{T} end
struct HEALPix4Grid{T} <: AbstractHEALPix4Grid{T}
    ...
end
```
for the time being. When we realise later that we can merge some of these abstract grid groups then we can still do that. But at the moment, I see it as more effort than benefit to try to unify Octahedral and HEALPix4 into one group with `nlon_at_pole` as a distinguishing parameter.

I'll give it a go at implementing `Matrix(::HEALPixGrid)` and Matrix(::HEALPix4Grid)` as I'd love to output directly on the grid in netcdf, which we could do with these conversions.

hottad:
Great! Thank you so much Milan.
I will follow your suggestion and start implementing HEALPix4. Thank you!

> I'll give it a go at implementing Matrix(::HEALPixGrid) and Matrix(::HEALPix4Grid)` as I'd love to output directly on the grid in netcdf, which we could do with these conversions.

Great idea. I would love to see how this can be done!

hottad:
Apologies but I accidentally deleted the branch [hottad:dh/explore-grids](https://github.com/hottad/SpeedyWeather.jl/tree/dh/explore-grids) and couldn't find how I can restore it...

milankl:
But that's fine no? Because its contents were merged to main anyway. I'd just start a new branch like `dh/healpix` on your fork that branches off `milankl:main`?

hottad:
Thank you Milan. Yes, I can just fork your main branch and start anew. I just thought it's better to keep my working branch for traceability, but I realized we can still view the diff on this page even though I deleted the branch. I'm still in the process of familiarizing myself with how github works. Sorry again...

PR #139, opened 2022-09-07
Initialize sigma levels manually

maximilian-gelbrecht:
Enables the possibility to set the vertical sigma levels manually with keyword arguments.

### DIFF
--- a/src/default_parameters.jl
+++ b/src/default_parameters.jl
@@ -46,10 +46,9 @@ The default values of the keywords define the default model setup.
     # of the nlev vertical levels, defined by a generalised logistic function,
     # interpolating ECMWF's L31 configuration
     GLcoefs::GenLogisticCoefs = GenLogisticCoefs{NF}()
-    n_stratosphere_levels::Int=2                                        # number of vertical levels used for the stratosphere
-    initialize_σ_manually::Bool=false                                   # initialize the σ vertical levels manually by providing a σ_levels_half 
-    σ_levels_half::Vector{Real}=Vector{Real}()                          # vector of σ half levels, only if set manually
-    nlev::Int=nlev_default(model, initialize_σ_manually, σ_levels_half) # number of vertical levels 
+    n_stratosphere_levels::Int=2                    # number of vertical levels used for the stratosphere
+    σ_levels_half::Vector{Real}=Vector{Real}()      # vector of σ half levels, only if set manually, otherwise an empty vector
+    nlev::Int=nlev_default(model, σ_levels_half)    # number of vertical levels 
 
     # DIFFUSION AND DRAG
     diffusion_power::Real=4                 # Power n of Laplacian in horizontal diffusion ∇²ⁿ
@@ -115,12 +114,12 @@ The default values of the keywords define the default model setup.
     restart_id::Integer=1           # run_id of restart file in run????/restart.jld2
 end
 
-function nlev_default(model::Symbol, initialize_σ_manually::Bool, σ_half::Vector{T}) where T<:Real
-    if !(initialize_σ_manually)
+function nlev_default(model::Symbol, σ_levels_half::Vector{T}) where T<:Real
+    if length(σ_levels_half) == 0   # choose nlev automatically 
         model == :barotropic && return 1
         model == :shallowwater && return 1
         model == :primitive && return 8
-    else 
-        return length(σ_half) - 1
+    else                            # use manually set levels 
+        return length(σ_levels_half) - 1
     end
 end
--- a/src/default_parameters.jl
+++ b/src/default_parameters.jl
@@ -16,7 +16,6 @@ The default values of the keywords define the default model setup.
     # RESOLUTION AND GRID
     trunc::Int=31                                   # spectral truncation
     Grid::Type{<:AbstractGrid}=FullGaussianGrid     # define the grid type
-    nlev::Int=nlev_default(model)                   # number of vertical levels 
     
     # PHYSICAL CONSTANTS
     radius_earth::Real=6.371e6          # radius of Earth [m]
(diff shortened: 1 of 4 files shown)


maximilian-gelbrecht:
Will something work very unexpectectly when setting the levels manually? Should some kind of warning be included for this case?

src/default_parameters.jl
```diff
@@ -47,7 +46,10 @@ The default values of the keywords define the default model setup.
     # of the nlev vertical levels, defined by a generalised logistic function,
     # interpolating ECMWF's L31 configuration
     GLcoefs::GenLogisticCoefs = GenLogisticCoefs{NF}()
-    n_stratosphere_levels::Int=2        # number of vertical levels used for the stratosphere
+    n_stratosphere_levels::Int=2                                        # number of vertical levels used for the stratosphere
+    initialize_σ_manually::Bool=false                                   # initialize the σ vertical levels manually by providing a σ_levels_half 
+    σ_levels_half::Vector{Real}=Vector{Real}()                          # vector of σ half levels, only if set manually
+    nlev::Int=nlev_default(model, initialize_σ_manually, σ_levels_half) # number of vertical levels
```
milankl, line 52: I'd suggest to remove the `initialize_σ_manually` field, as 0-element vector for `σ_levels_half` could serve that purpose. See suggestion further down

src/default_parameters.jl
```diff
@@ -113,8 +115,12 @@ The default values of the keywords define the default model setup.
...
-    model == :shallowwater && return 1
-    model == :primitive && return 8
+function nlev_default(model::Symbol, initialize_σ_manually::Bool, σ_half::Vector{T}) where T<:Real
+    if !(initialize_σ_manually)
+        model == :barotropic && return 1
+        model == :shallowwater && return 1
+        model == :primitive && return 8
+    else 
+        return length(σ_half) - 1
+    end
```
milankl, line 125: Suggestion would be
```julia
function nlev_default(model::Symbol, σ_levels_half::AbstractVector)
    if length(σ_half)  == 0    # choose nlev automatically
        model == :barotropic && return 1
        model == :shallowwater && return 1
        model == :primitive && return 8
    else     # use manually initialized σ_levels_half instead
        @assert σ_levels_half[1] == 0 "First manually specified σ_levels_half has to be zero 0"
        @assert σ_levels_half[end] == 1 "Last manually specified σ_levels_half has to be 1."
        @assert isincreasing(σ_levels_half) "Manually specified σ_levels_half have to be increasing"
        return length(σ_levels_half) - 1
    end
end
```
That way
- `run_speedy(nlev=4)` would use `nlev=4` and ignore `model` and potentially also `σ_levels_half` if that's specified too.
- `run_speedy(σ_levels_half=[0,0.2,0.5,0.8,1])` would set `nlev=4` through the function above.

src/geometry.jl
```diff
@@ -178,12 +178,14 @@ The first half level is at 0 the last at 1. Evaluate a generalised logistic func
 coefficients in `P` for the distribution of values in between. Default coefficients follow
 the L31 configuration historically used at ECMWF."""
 function vertical_coordinates(P::Parameters)
-    @unpack nlev,GLcoefs = P
+    @unpack nlev,GLcoefs,initialize_σ_manually,σ_levels_half = P
```
milankl, line 181: then remove `initialize_σ_manually` here as that information is stored in the length of `σ_levels_half`

src/geometry.jl
```diff
@@ -178,12 +178,14 @@ The first half level is at 0 the last at 1. Evaluate a generalised logistic func
...
 
-    halflevels_normalised = range(0,1,nlev+1)   # normalised = level/nlev
-    σ_levels_half = generalised_logistic(halflevels_normalised,GLcoefs)
-    σ_levels_half[1] = 0           # topmost half-level is at 0 pressure
-    σ_levels_half[end] = 1         # lowermost half-level is at p=p_surface
+    if !initialize_σ_manually
```
milankl, line 183: `if length(σ_levels_half) == 0`

milankl:
> Will something work very unexpectectly when setting the levels manually? Should some kind of warning be included for this case?

I don't think so, as long as 
```julia

milankl:
Maybe we can add tests in `test/run_speedy.jl`? Like so (commit coming)

milankl:
Also I didn't realise how this works: If you pull request directly from your fork's main branch and I commit to the same pull request then I'm directly committing (without a pull request) to your main (and it doesn't require your confirmation for me to overwrite your main) 😆

milankl:
Ha, stupid me. Tests are currently failing because the suggested changes from my review above haven't (obviously) been committed yet. The tests should pass once that's done, but I'll wait for your response.

src/default_parameters.jl
maximilian-gelbrecht: yep, I agree, that's more elegant

src/default_parameters.jl
maximilian-gelbrecht: if both are specified, it doesn't work anymore, I add an `@assert` for that case in the `vertical coordinates`

maximilian-gelbrecht:
I put the `@asserts` in the `vertical_coordinates` function, but in principle the commit is what you suggested. 
For the tests however, they failed when both `nlev` and `σ_levels_half` were given. I added an `@assert` to ensure that they are the same length.

milankl:
Great, yeah, my suggestion probably wasn't fully consistent. So now one either specifies `nlev` or `sigma_half_levels` and gets automatic, manual, respectively. If both are specified (which is kind of redundant anyway) then they have to match otherwise an `AssertionError` is thrown, correct?

maximilian-gelbrecht:
>  If both are specified (which is kind of redundant anyway) then they have to match otherwise an AssertionError is thrown, correct?

Yes 

So let's merge it?

PR #142, opened 2022-09-09
Inverse Laplace ∇⁻²! via kernel in ∇²!

### LINKED ISSUE #141: Inverse Laplacian
I’d need the inverse laplacian to convert vorticity to streamfunction. I’ve seen that this is available in the code, but commented out in `spectral_gradients!` as `∇⁻²!`. Why is that so? Could we make it functional again? `SpectralTransform` still has the `eigenvalues⁻¹` field that I also only used by `∇⁻²!`

`UV_from_vor!` also somehow inverts the laplacian but I don't understand how. Anyway though, I need to compute the stream function not the wind field.

milankl:
fixes #141

### DIFF
--- a/src/spectral_transform.jl
+++ b/src/spectral_transform.jl
@@ -61,6 +61,7 @@ struct SpectralTransform{NF<:AbstractFloat}
 
     # EIGENVALUES (on unit sphere, no 1/radius²-scaling included)
     eigenvalues  ::Vector{NF}               # = -l*(l+1), degree l of spherical harmonic
+    eigenvalues⁻¹::Vector{NF}               # = -1/(l*(l+1))
 end
 
 """
@@ -181,7 +182,9 @@ function SpectralTransform( ::Type{NF},                     # Number format NF
 
     # EIGENVALUES (on unit sphere, hence 1/radius²-scaling is omitted)
     eigenvalues = [-l*(l+1) for l in 0:lmax+1]
-        
+    eigenvalues⁻¹ = inv.(eigenvalues)
+    eigenvalues⁻¹[1] = 0                    # set the integration constant to 0
+
     # conversion to NF happens here
     SpectralTransform{NF}(  Grid,nresolution,order,
                             lmax,mmax,nfreq_max,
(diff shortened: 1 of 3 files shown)


maximilian-gelbrecht (approved):
Looks good to me!

src/spectral_transform.jl
```diff
@@ -182,8 +181,6 @@ function SpectralTransform( ::Type{NF},                     # Number format NF
 
     # EIGENVALUES (on unit sphere, hence 1/radius²-scaling is omitted)
     eigenvalues = [-l*(l+1) for l in 0:lmax+1]
-    eigenvalues⁻¹ = inv.(eigenvalues)
-    eigenvalues⁻¹[1] = 0                    # set the integration constant to 0
```
milankl, line 186: I just realised that we shouldn't replace `eigenvalues⁻¹` by `1/eigenvalues` as that makes the integration constant infinity. We currently don't test for this (and it's not used by the dynamical core atm) hence the tests still pass
```julia
julia> alms = LowerTriangularMatrix(randn(ComplexF32,33,32));
julia> alms[:,1] .= real.(alms[:,1]);
julia> ∇⁻²alms = copy(alms)
julia> S = SpectralTransform(alms);
julia> SpeedyWeather.∇²!(∇⁻²alms,alms,S,inverse=true)
33×32 LowerTriangularMatrix{ComplexF32}:
   -Inf+NaN*im                0.0+0.0im          …          0.0+0.0im
    -0.369622-0.0im      0.199208-0.408254im                0.0+0.0im
...
```

milankl:
Don't worry, I'll resolve that, just don't merge it yet. Enjoy your vacation 🌴

milankl:
Also @maximilian-gelbrecht, a little change I made in `∇²!` and `∇⁻²!` is that previously it excluded the the last row = highest degree l. This is because scalar quantities have lmax==mmax, but vector quantities use one more degree due to the recursion in the meridional gradient. I changed this to include all degrees and orders, to have a generally valid (inverse) Laplace operator (the higest degree l otherwise wouldn't be touched, and values may carry over which I don't like). I believe this does not have any unforeseen consequences but in case it does, let me know.

maximilian-gelbrecht:
Great, thanks @milankl

PR #143, opened 2022-09-12
Adapt structures for diagnostic variables and prognostic variables

katharinamaetschke:
@maximilian-gelbrecht 
Corrected adapt functions for diagnostic_variables.jl and new adapt functions for prognostic_variables.jl.

### DIFF
--- a/src/prognostic_variables.jl
+++ b/src/prognostic_variables.jl
@@ -96,3 +96,26 @@ function Base.zeros(::Type{PrognosticVariables{NF}},
     pres = zeros(SurfaceLeapfrog{NF},_lmax,_mmax)                     # 2 leapfrog time steps for pres
     return PrognosticVariables(layers,pres,lmax,mmax,N_LEAPFROG,nlev)
 end
+
+
+# GPU methods
+function Adapt.adapt_structure(to, pvl::PrognosticVariablesLayer)
+    PrognosticVariablesLayer(Adapt.adapt(to, pvl.vor), 
+                             Adapt.adapt(to, pvl.div), 
+                             Adapt.adapt(to, pvl.temp),
+                             Adapt.adapt(to, pvl.humid))
+end
+
+function Adapt.adapt_structure(to, p_leap::PrognosticVariablesLeapfrog)
+    PrognosticVariablesLeapfrog(Adapt.adapt(to, p_leap.leapfrog))
+end
+
+function Adapt.adapt_structure(to, s_leap::SurfaceLeapfrog)
+    SurfaceLeapfrog(Adapt.adapt(to, s_leap.leapfrog))
+end
+
+function Adapt.adapt_structure(to , pv::PrognosticVariables)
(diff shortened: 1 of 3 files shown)


milankl:
I had the same doc build fail recently too, just update to v1.8 here
https://github.com/milankl/SpeedyWeather.jl/blob/7715f0a8d0b1c16b7a7e2dab8b48036a3a815228/.github/workflows/Documenter.yml#L15-L17
As we already have it in main

src/prognostic_variables.jl
```diff
@@ -96,3 +96,26 @@ function Base.zeros(::Type{PrognosticVariables{NF}},
...
+    PrognosticVariablesLeapfrog(Adapt.adapt(to, p_leap.leapfrog))
+end
+
+function Adapt.adapt_structure(to, s_leap::SurfaceLeapfrog)
+    SurfaceLeapfrog(Adapt.adapt(to, s_leap.leapfrog))
+end
+
+function Adapt.adapt_structure(to , pv::PrognosticVariables)
+    PrognosticVariables(Adapt.adapt(to, pv.layers),
+                        Adapt.adapt(to, pv.pres)
```
milankl, line 119: `,` missing here

milankl: added with [9e9a452](https://github.com/milankl/SpeedyWeather.jl/pull/143/commits/9e9a452ea9dcbd68ab998a827a598baf91f6c2d1)

maximilian-gelbrecht (commented):
Looks good to me. As mentioned, unfortunately without GPU CI we can't do an automated test for this. 

@katharinamaetschke I try to figure something out, how we can give you access to a HPC GPU to test these things a bit more manually

src/prognostic_variables.jl
maximilian-gelbrecht: Huh, I am surprised that this didn't yield an error with the test in `test/kernelabstractions.jl`. Without GPU CI it can't test if transferring everything to GPU works, but it should yield errors for syntax errors. I'll have to look into this.

milankl:
All of this sounds like we should get the GPU CI going rather sooner than later. I’d be happy to talk to people to get this rolling. We can then just set up GPU CI for certain branches if that’s needed such that all my non-GPU implementations don’t look like we’re clogging up the GPU CI server?

maximilian-gelbrecht:
I didn't look into it with in all detail yet, but the procedure is outlined in https://github.com/JuliaGPU/buildkite 
As mentioned on Zulip, I think we would need to transfer the repo to an organisation instead of a single account. Then it is just a matter of texting one of the admins like Valentin. 
I think we would just write a separate set of tests for the GPU CI that really only tests GPU relevant stuff (which is basically already what I started with `test/kernelabstractions.jl`) and we should also restrict it to the `kernelabstractions_adtest` branch for the development phase.

maximilian-gelbrecht:
I'll merge this now. 

@milankl We can continue the GPU CI topic on Zulip or in a separate issue

PR #145, opened 2022-09-15
Shallow water model and OctahedralClenshawGrid default

### LINKED ISSUE #112: A new grid for SpeedyWeather.jl?

milankl:
Uses OctahedralClenshawGrid to fix #112

### DIFF
--- a/src/default_parameters.jl
+++ b/src/default_parameters.jl
@@ -11,11 +11,11 @@ The default values of the keywords define the default model setup.
     NF::DataType                        # number format (default defined in run_speedy)
 
     # MODEL
-    model::Symbol = :barotropic         # :barotropic, :shallowwater, or :primitive
+    model::Symbol = :shallowwater       # :barotropic, :shallowwater, or :primitive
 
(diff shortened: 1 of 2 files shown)


milankl:
Note: While this got merged, the changes here had already been overwritten by #150 meaning that FullGaussianGrid and BarotropicModel are the default again.

PR #147, opened 2022-09-20
Add kernelabstractions_adtests CI

maximilian-gelbrecht:
We should have CI for the GPU/AD development branch on pull requests. I think this was also why the error wasn't caught. Only the doc was built, but the tests were not run

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -3,6 +3,7 @@ on:
   pull_request:
     branches:
       - main
+      - kernelabstractions_adtests
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


milankl:
Shall we not have a general #gpu branch for this? And then you and Katharina can pull request into that and it's clearer that this is the branch that should do some GPU CI before merging?

maximilian-gelbrecht:
So, just rename the branch? Or do you think about something additional?

milankl:
Given that your branch is up to date with main, just rename I guess? Then all the GPU/AD developments go first in the gpu branch can be tested there and can then be merged into main? Whereas all CPU/architecture independent developments can directly pull request to main? I believe that makes sense...

maximilian-gelbrecht:
Ok, then I'll just rename it to `gpu` 

@katharinamaetschke We rename the `kernelabstractions_adtests` branch into `gpu`, so in the future just do your PRs into that branch

PR #149, opened 2022-09-20
Dense Output

maximilian-gelbrecht:
Adds a kwarg `output_dense::Bool` so that if it is `true` every time step is saved in the netCDF file

### DIFF
--- a/src/constants.jl
+++ b/src/constants.jl
@@ -58,7 +58,7 @@ function Constants(P::Parameters)
 
     # TIME INTEGRATION CONSTANTS
     @unpack robert_filter, williams_filter = P
-    @unpack trunc, Δt_at_T31, n_days, output_dt = P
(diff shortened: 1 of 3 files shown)


milankl:
Not necessary, you can do `output_dt=0` to do the same.

maximilian-gelbrecht:
Ah ok, did not see that

milankl:
But on that note, at the moment we don't call `write_netcdf_output!` after the first two time steps (the first is Euler forward, the second one is an unfiltered leapfrog with dt=dt (and not 2dt as in the main loop). For every `progress!` call here

https://github.com/milankl/SpeedyWeather.jl/blob/65b2bc13399f7d6c59edc40f1fc359e7a37949eb/src/time_integration.jl#L84-L112

We could write 

```julia
progress!(feedback,outputter)
write_netcdf_output!(outputter,time,progn,diagn,M)
```
as in the main loop to also output the first two steps.

PR #150, opened 2022-09-21
FullHEALPixGrid 

milankl:


milankl:
@hottad this implements the full version of the HEALPixGrid to be used for output.

- For the original HEALPix map `nside` has to be a power of 2. However, I'm not sure whether this constraint results only from the nested grid structure with some space-filling curve, meaning we could drop this in our application? Might be nice as it will give us more flexibility in terms of resolution
- Which quadrature weights should be used for the FullHEALPixGrid? If it's only used for output this isn't important, but we may set it up directly with something sensible. Should one use $\sin \theta \Delta \theta$ directly?

hottad:
@mikankl Thank you! Great job, as always.


> * For the original HEALPix map `nside` has to be a power of 2. However, I'm not sure whether this constraint results only from the nested grid structure with some space-filling curve, meaning we could drop this in our application? Might be nice as it will give us more flexibility in terms of resolution

I agree.  We should allow whatever positive integer as `nside`.

> * Which quadrature weights should be used for the FullHEALPixGrid? If it's only used for output this isn't important, but we may set it up directly with something sensible. Should one use sin⁡θΔθ directly?

In my (partial) implementation of HEALPix on ectrans, I tried:
- NH polar cap: w_j = 2*(4*j) /(12*nside^2) where j is the index of the ring (j=1 for the ring closes to the pole)
- Tropic belt: w_j = 2*(4*nside)/(12*nside^2)
- SH: mirror of NH

and I confirmed this gives equally accurate transform for HEALPixGrid and FullHEALPixGrid (and also OctahedralHEALPixGrid).
The source code here:
https://github.com/hottad/ectrans/blob/d4a22d9386419823444c8c43bb21b45eeb860614/src/trans/internal/suleg_mod.F90#L323-L343 (note that in IFS, weight is normalized to give Σ_j w_j =1 instead of 2).

But, yes, you can also try sin⁡θΔθ. It will be interesting to compare the two approaches.

hottad:
Just in case you wonder why the same weights work for the full grid, here is the derivation (excuse my dirty hand-writing...) [FullHEALPixGrid_weight_derivation](https://user-images.githubusercontent.com/88590419/191955621-ba233315-7fc7-4fc9-8f2d-d72cc0d87b6e.png)

milankl:
Looks like all tests pass @hottad 🥳, quick update on what I changed
- `nlat_half` is always the only resolution parameter for every grid (although sometime `nside` is handy for HEALPix, but for HEALPix4 `nlat_half=nside` anyway).
- Everything grid related is now in the folder `Grids/`, sorted into different files.
- The FullHEALPixGrid has been added so that `Grid=HEALPixGrid` can also easily be output into netcdf.

What I haven't done yet
- there's still a bunch of code redundancies that we could merge, but not urgent
- any further tests apart from `run_speedy(Grid=G)` for `G` any of the 8 grids we now support (OctahedralGaussian, OctahedralClenshaw, HEALPix, HEALPix4 + 4xfull equivalents) which looks indistinguishely similar, meaning that after 10 days with T31 no grid is obviously different from the others

So I'll merge this one now, and we can move other issues into new pull requests.

hottad:
Thank you, @milankl! Kudos for the great job! and thanks also summarising what's changed.

Issue #135, opened 2022-08-31
Broadcast operations with LowerTriangularMatrix


milankl:
Broadcasting with LowerTriangularMatrix still returns unexpected results. The biggest problem I found was dotted operations like .*= , see how suddenly a zero appears where it shouldn’t
```julia
julia> L = rand(LowerTriangularMatrix,3,3)
3×3 LowerTriangularMatrix{Float64}:
 0.998746  0.0       0.0
 0.185934  0.783976  0.0
 0.181445  0.950264  0.970376

julia> L .*= 2
3×3 LowerTriangularMatrix{Float64}:
 1.99749   0.0  0.0
 0.371868  0.0  0.0
 0.0       0.0  1.94075
```
This is (I think) because it loops over each index (i,j) with an `@inbounds` such that the `@boundscheck` in `setindex!` is disabled and then converts i,j to a single index k which isn’t safe if `j>i`… The problem is apparently that the broadcasting of `*` creates a `Matrix` which is then fused with `.=` and writes the entries back into a `LowerTriangularMatrix`
```julia
julia> L = rand(LowerTriangularMatrix,3,3)
3×3 LowerTriangularMatrix{Float64}:
 0.0948286  0.0        0.0
 0.10467    0.821121   0.0
 0.971535   0.0221127  0.744497

julia> L .* 2
3×3 Matrix{Float64}:
 0.189657  0.0        0.0
 0.20934   1.64224    0.0
 1.94307   0.0442254  1.48899
```

milankl:
@giordano just raised another related issue
```julia
julia> A = LowerTriangularMatrix(rand(3,3))
3×3 LowerTriangularMatrix{Float64}:
 0.382579  0.0       0.0
 0.554634  0.779331  0.0
 0.453919  0.236713  0.559563

julia> sum(A)
2.966738421294002

julia> prod(A)
0.009942566052321158
```
The sum is correct, but the product should obviously be zero.

maximilian-gelbrecht:
> @giordano just raised another related issue
> 
> ```julia
> julia> A = LowerTriangularMatrix(rand(3,3))
> 3×3 LowerTriangularMatrix{Float64}:
>  0.382579  0.0       0.0
>  0.554634  0.779331  0.0
>  0.453919  0.236713  0.559563
> 
> julia> sum(A)
> 2.966738421294002
> 
> julia> prod(A)
> 0.009942566052321158
> ```
> 
[...]

PR #153, opened 2022-09-26
Vorticity initial conditions

maximilian-gelbrecht:
I first wanted to do it by requiring to hand over an instance of `PrognosticVariables`, but for that you have the problem again that  `PrognosticVariables` are defined at the time the inputs are handles. Changing that would need larger changes, so I am thinking handing over initial conditions as `Vector{LowerTriangluarMatrix}` is fine

### DIFF
--- a/src/default_parameters.jl
+++ b/src/default_parameters.jl
@@ -109,7 +109,8 @@ The default values of the keywords define the default model setup.
 
     # INITIAL CONDITIONS
     seed::Int = abs(rand(Int))          # a random seed that's used in initialize_speedy for the global RNG
-    initial_conditions::Symbol=:barotropic_vorticity    # :rest, :barotropic_vorticity or :restart
+    initial_conditions::Union{Symbol, Vector}=:barotropic_vorticity    # :rest, :barotropic_vorticity, :restart
+                                                                       # or Vector of LowerTriangularMatrix of Vorticity 
 
     # OUTPUT
     verbose::Bool = true            # print dialog for feedback
--- a/src/initial_conditions.jl
+++ b/src/initial_conditions.jl
@@ -25,6 +25,9 @@ function initial_conditions(M::ModelSetup)
                 end
             end
         end
+    
+    elseif typeof(initial_conditions) <: Vector
+        initialize_from_vector!(progn, initial_conditions, M)
 
     elseif initial_conditions == :restart
         initialize_from_file!(progn,M)
@@ -67,6 +70,29 @@ function initialize_from_rest(M::PrimitiveEquationModel)
 
     return progn
 end
+
+function initialize_from_vector!(progn_new::PrognosticVariables{NF}, ic::Vector, M::ModelSetup) where NF
+   
+    @unpack nlev = M.geometry
+    @unpack lmax, mmax = M.spectral_transform
+
+    @assert length(ic) == nlev
+
+    # SPECTRAL TRUNCATION/INTERPOLATION to new resolution and conversion to NF
+    for (layer_new, layer_ic) in zip(progn_new.layers, ic)
+
+        @assert typeof(layer_ic) <: LowerTriangularMatrix 
+
+        layer_new_lf1 = layer_new.leapfrog[1]
+        vor_new = layer_new_lf1.vor
+        lmax,mmax = size(vor_new) .- (2,1)
+
+        vor_ic_trunc = spectral_truncation(layer_ic, lmax+1, mmax)
+        vor_new .=  convert(LowerTriangularMatrix{Complex{NF}}, vor_ic_trunc)
+    end 
+
+    progn_new
+end 
 
 function initialize_from_file!(progn_new::PrognosticVariables{NF},M::ModelSetup) where NF
     @unpack restart_path, restart_id = M.parameters
--- a/test/initialize.jl
+++ b/test/initialize.jl
@@ -59,6 +59,36 @@ end
     """
 end
 
+
+@testset "Initialize from Vorticity IC" begin
+
+    # BAROTROPIC MODEL
+    progn, diagn, model = initialize_speedy(initial_conditions=:rest,model=:barotropic)
+
+    ic = [layer.leapfrog[1].vor for layer in progn.layers]
+    progn, diagn, model = initialize_speedy(initial_conditions=ic,model=:barotropic)
+    for layers in progn.layers
+        for leapfrog in layers.leapfrog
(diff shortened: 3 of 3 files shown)


maximilian-gelbrecht:
There's a Julia subtlety that I don't get, but noticed while doing this PR: 

```Julia
(LowerTriangularMatrix{Complex{NF}} <: LowerTriangularMatrix) == true
```
but
```Julia 
(Vector{LowerTriangularMatrix{Complex{NF}}} <: Vector{LowerTriangularMatrix}) == false
```

Any idea why?

white-alistair:
Isn't this precisely the invariance of Julia types discussed [here](https://docs.julialang.org/en/v1/manual/types/#man-parametric-composite-types)?

maximilian-gelbrecht:
Ah yes, you are right. I also did encounter this before, but somehow wasn't thinking about it in that way.

milankl:
I very much like the idea to have the initial conditions easily choosable with arrays that are defined in the REPL before `run_speedy` is called. However, I'm not sure about the current idea how this PR suggests to implement it. The reason is that conceptually `Parameters` should contain parameters and not entire arrays that contain data. This may sound pendatic, but given the discussion we had about also including the parameters into the restart file, suddenly including a potentially massive array doesn't sound neat. Therefore first question, what's wrong with
```julia
p,d,m = initialize_speedy(kwargs...)
p.layers[1].leapfrog[1].vor .= ...   # change vorticity, or any other prognostic variable
time_stepping!(p,d,m)
```
where the second line could also be replaced by something like
```julia
set_vorticity!(p,some_array...)
set_divergence!(p,some_other_data...)
...
```
which could take `LowerTriangularMatrix` or some gridded data (which is then transformed to spectral space), scalars, etc. The reason I'm suggesting this is also that it feels more modular, as the function how to get the data into the prognostic variables struct doesn't have to be buried in `initialize_speedy!`, and people can easily define their own extensions.

src/initial_conditions.jl
```diff
@@ -68,6 +71,24 @@ function initialize_from_rest(M::PrimitiveEquationModel)
...
+    @assert length(ic) == nlev
+
+    # SPECTRAL TRUNCATION/INTERPOLATION to new resolution and conversion to NF
+    for (layer_new, layer_ic) in zip(progn_new.layers, ic)
+        layer_new_lf1 = layer_new.leapfrog[1]
+        vor_new = layer_new_lf1.vor
+        lmax,mmax = size(vor_new) .- (2,1)
+
+        vor_ic_trunc = spectral_truncation(layer_ic, lmax+1, mmax)
+        vor_new .=  convert(LowerTriangularMatrix{Complex{NF}}, vor_ic_trunc)
```
milankl, line 86: Note that we still have #135 open, meaning that `vor_new .= some_array` may not do what you expect it too. To avoid this broadcasting issue in the first place I've already defined `copyto!` for `:LowerTriangularMatrix` which can handle the type conversion in `convert` implicitly.

test/initialize.jl
```diff
@@ -59,6 +59,36 @@ end
...
+@testset "Initialize from Vorticity IC" begin
+
+    # BAROTROPIC MODEL
+    progn, diagn, model = initialize_speedy(initial_conditions=:rest,model=:barotropic)
+
+    ic = [layer.leapfrog[1].vor for layer in progn.layers]
+    progn, diagn, model = initialize_speedy(initial_conditions=ic,model=:barotropic)
+    for layers in progn.layers
+        for leapfrog in layers.leapfrog
+            @test all(leapfrog.vor .== 0)
```
milankl, line 72: Given the open broadcasting issue related to the comment above, note that this is a redundant test: Replacing zeros with zeros would pass even if nothing is actually done.

test/initialize.jl
```diff
@@ -59,6 +59,36 @@ end
...
+    ic = [layer.leapfrog[1].vor for layer in progn.layers]
+    progn, diagn, model = initialize_speedy(initial_conditions=ic,model=:shallowwater)
+    for layers in progn.layers
+        for leapfrog in layers.leapfrog
+            @test all(leapfrog.vor .== 0)
+            @test all(leapfrog.div .== 0)
+        end
+    end
+    @test all(progn.pres.leapfrog[1] .== 0)
+    @test all(progn.pres.leapfrog[2] .== 0)
```
milankl, line 88: Same here.

src/default_parameters.jl
```diff
@@ -109,7 +109,8 @@ The default values of the keywords define the default model setup.
 
     # INITIAL CONDITIONS
     seed::Int = abs(rand(Int))          # a random seed that's used in initialize_speedy for the global RNG
-    initial_conditions::Symbol=:barotropic_vorticity    # :rest, :barotropic_vorticity or :restart
+    initial_conditions::Union{Symbol, Vector}=:barotropic_vorticity    # :rest, :barotropic_vorticity, :restart
```
milankl, line 112: Another problem I see here is that with `output=true` the parameters struct is printed to `parameters.txt` such that one can easily trace back what parameters were used in a given simulation. I'm not sure what happens if a big array is part of the parameters struct. This also applies to setting the vertical levels manually, which I haven't checkedn. Those ones, ideally, should be written into that .txt file (as just a vector) but entire vectors or matrices obviously not...

src/initial_conditions.jl
maximilian-gelbrecht: This is just copied from the `initialize_from_file!` function. Then that should also be looked at, but we can also replace that function with `set_vorticity!` etc as well

test/initialize.jl
maximilian-gelbrecht: Same here as well, that's essentially the same test as already done for the other initialisations

maximilian-gelbrecht:
Mh, initial conditions do belong to a well defined differential equation problem, so I am not fundamentally opposed to including them in the Parameters (as this is also currently the only way input arguments are handled). 

However, I also like your idea of having flexible `set_vorticity!(p,some_array...)` etc. functions. This does look like a clean solution.

maximilian-gelbrecht:
I am working on a new draft for this

src/initial_conditions.jl
milankl: Yes, good point!

test/initialize.jl
milankl: Haha, boooh past me setting up shitty tests 😆

maximilian-gelbrecht:
This is closed in favour of https://github.com/milankl/SpeedyWeather.jl/pull/154

PR #154, opened 2022-09-27
set_vars! function to easily set variables in PrognosticVariables

maximilian-gelbrecht:
This is a currently a draft, I want to get a review on it before I fully finish it. 

* I implemented `set....!(prong,...)` functions for all variables for spectral, grid and matrix input. These inputs are currently all `Vector{...}`

* no doc strings yet 
* initializiation routines could be reworked with this as well 
* maybe there are some other parts of the code that could use this @milankl ?

I was also thinking of implementing the opposing `get....(progn...)` functions that return the `Vectors{...}`. What do you think? 

A small observation: for the `ModelSetup` there should probably be a `has(M::ModelSetup, var_name::Symbol)` function, that could make a small minor things a bit more transparent.

### DIFF
--- a/src/initial_conditions.jl
+++ b/src/initial_conditions.jl
@@ -75,7 +75,7 @@ function initialize_from_file!(progn_new::PrognosticVariables{NF},M::ModelSetup)
     version = restart_file["version"]   # currently unused, TODO check for compat with version
     time = restart_file["time"]         # currently unused
 
-    var_names = propertynames(progn_old.layers[1].leapfrog[1])
+    var_names = [propertynames(progn_old.layers[1].leapfrog[1])..., :pres]
 
     for var_name in var_names
         if has(progn_new, var_name) 
--- a/src/initial_conditions.jl
+++ b/src/initial_conditions.jl
@@ -75,11 +75,13 @@ function initialize_from_file!(progn_new::PrognosticVariables{NF},M::ModelSetup)
     version = restart_file["version"]   # currently unused, TODO check for compat with version
     time = restart_file["time"]         # currently unused
 
-    var_names = [propertynames(progn_old.layers[1].leapfrog[1])]
+    var_names = propertynames(progn_old.layers[1].leapfrog[1])
 
-    for var_name in var_names 
-        var = get_var(progn_old, var_name) 
-        set_var!(progn_new, var_name, var)
+    for var_name in var_names
+        if has(progn_new, var_name) 
+            var = get_var(progn_old, var_name) 
+            set_var!(progn_new, var_name, var)
+        end
     end 
 
     return progn_new
--- a/src/initial_conditions.jl
+++ b/src/initial_conditions.jl
@@ -75,27 +75,12 @@ function initialize_from_file!(progn_new::PrognosticVariables{NF},M::ModelSetup)
     version = restart_file["version"]   # currently unused, TODO check for compat with version
     time = restart_file["time"]         # currently unused
 
-    # SPECTRAL TRUNCATION/INTERPOLATION to new resolution and conversion to NF
-    for (layer_old,layer_new) in zip(progn_old.layers,progn_new.layers)
+    var_names = [propertynames(progn_old.layers[1].leapfrog[1])]
 
-        layer_old_lf1 = layer_old.leapfrog[1]
-        layer_new_lf1 = layer_new.leapfrog[1]
-
-        vars_old = (getproperty(layer_old_lf1,prop) for prop in propertynames(layer_old_lf1))
-        vars_new = (getproperty(layer_new_lf1,prop) for prop in propertynames(layer_new_lf1))
-    
-        for (var_old,var_new) in zip(vars_old,vars_new)
-            lmax,mmax = size(var_new) .- (2,1)
-            var_old_trunc = spectral_truncation(var_old,lmax+1,mmax)
-            var_new .= convert(LowerTriangularMatrix{Complex{NF}},var_old_trunc)
-        end
-    end
-
-    # same for surface pressure
-    pres = progn_new.pres.leapfrog[1]
-    lmax,mmax = size(pres) .- (2,1)
-    pres .= convert(LowerTriangularMatrix{Complex{NF}},
-                                    spectral_truncation(progn_old.pres.leapfrog[1],lmax+1,mmax))
+    for var_name in var_names 
+        var = get_var(progn_old, var_name) 
+        set_var!(progn_new, var_name, var)
+    end 
 
     return progn_new
 end
--- a/src/models.jl
+++ b/src/models.jl
@@ -20,6 +20,8 @@ struct BarotropicModel{NF<:AbstractFloat, D<:AbstractDevice} <: ModelSetup{D}
     device_setup::DeviceSetup{D}
 end
 
+has(::Type{BarotropicModel{NF,D}}, var_name::Symbol) where {NF,D} = var_name in [:vor]
+
 """
     M = ShallowWaterModel(  ::Parameters,
                             ::Constants,
@@ -40,6 +42,9 @@ struct ShallowWaterModel{NF<:AbstractFloat, D<:AbstractDevice} <: ModelSetup{D}
     implicit::Implicit{NF}
     device_setup::DeviceSetup{D}
 end
+
+has(::Type{ShallowWaterModel{NF,D}}, var_name::Symbol) where {NF,D} = var_name in [:vor, :div, :pres]
+
 
 """
     M = PrimitiveEquationModel( ::Parameters,
@@ -62,3 +67,13 @@ struct PrimitiveEquationModel{NF<:AbstractFloat,D<:AbstractDevice} <: ModelSetup
     implicit::Implicit{NF}
     device_setup::DeviceSetup{D}
 end
+
+has(::Type{PrimitiveEquationModel{NF,D}}, var_name::Symbol) where {NF,D} = var_name in [:vor, :div, :temp, :humid, :pres]
+
+"""
+    has(::ModelSetup, var_name::Symbol)
+
+Returns true if the model `M` has a prognostic variable `var_name`, false otherwise. The default fallback is that all variables are included. 
+"""
+has(::Type{ModelSetup{D}}, var_name::Symbol) where D = var_name in [:vor, :div, :temp, :humid, :pres] 
+has(M::ModelSetup, var_name) = has(typeof(M), var_name)
--- a/src/prognostic_variables.jl
+++ b/src/prognostic_variables.jl
@@ -105,7 +105,7 @@ has(progn::PrognosticVariables{NF,M}, var_name::Symbol) where {NF,M} = has(M, va
              var::Vector{<:LowerTriangularMatrix};
              lf::Integer=1) where NF
 
-Sets the prognostic variable with the name `varname` in all layers at leapfrog index `lf` with values given in `var` a vector with all information for all layers.
+Sets the prognostic variable with the name `varname` in all layers at leapfrog index `lf` with values given in `var` a vector with all information for all layers in spectral space.
 """
 function set_var!(progn::PrognosticVariables{NF}, 
                   varname::Symbol, 
@@ -128,6 +128,14 @@ function _set_var_core!(var_old::LowerTriangularMatrix{T}, var_new::LowerTriangu
     copyto!(var_old, var_new_trunc)
 end 
 
+"""
+    set_var!(progn::PrognosticVariables{NF},        
+             varname::Symbol, 
+             var::Vector{<:AbstractGrid};
+             lf::Integer=1) where NF
+
+Sets the prognostic variable with the name `varname` in all layers at leapfrog index `lf` with values given in `var` a vector with all information for all layers in grid space.
+"""
 function set_var!(progn::PrognosticVariables{NF}, 
                   varname::Symbol, 
                   var::Vector{<:AbstractGrid};
@@ -140,6 +148,15 @@ function set_var!(progn::PrognosticVariables{NF},
     return set_var!(progn, varname, var_sph; lf=lf)
 end 
 
+"""
+    set_var!(progn::PrognosticVariables{NF}, 
+             varname::Symbol, 
+             var::Vector{<:AbstractGrid}, 
+             M::ModelSetup;
+             lf::Integer=1) where NF
+
+Sets the prognostic variable with the name `varname` in all layers at leapfrog index `lf` with values given in `var` a vector with all information for all layers in grid space.
+"""
 function set_var!(progn::PrognosticVariables{NF}, 
                   varname::Symbol, 
                   var::Vector{<:AbstractGrid}, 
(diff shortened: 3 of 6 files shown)


milankl:
>     * I implemented `set....!(prong,...)` functions for all variables for spectral, grid and matrix input. These inputs are currently all `Vector{...}`

Maybe we could even add methods like `set_vorticity!(progn,L,lf=1)` such that for a `L` being a LowerTriangularMatrix, or Matrix, or Gridded, (and not a vector of it) one can easily set the right leapfrog time step `lf` or layer `layer` etc.

>     * initializiation routines could be reworked with this as well
>     * maybe there are some other parts of the code that could use this @milankl ?

I think only the initial conditions actually `set_...!` the prognostic variables.

> I was also thinking of implementing the opposing `get....(progn...)` functions that return the `Vectors{...}`. What do you think?

Yes, the time integration has lines like
```
vor_old  = progn.leapfrog[1].vor
vor_new  = progn.leapfrog[2].vor
```
which could be wrapped more generally into `vor_old, vor_new = get_vorticity(progn,lf=(1,2))` but maybe it's not worth it?

> A small observation: for the `ModelSetup` there should probably be a `has(M::ModelSetup, var_name::Symbol)` function, that could make a small minor things a bit more transparent.

Yes, at the moment, we set up all prognostic variables as a dummy 0-element array when they are not used in a given ModelSetup. E.g. in the BaroptropicModel
```julia
julia> p.layers[1].leapfrog[1].div
0×0 LowerTriangularMatrix{ComplexF64}
```
So if it's useful for you please implement `has(M::ModelSetup, var_name::Symbol)`. At the moment we have lines like
```julia
M isa BarotropicModel ? nothing : gridded!(diagn.surface.pres_grid,progn.pres.leapfrog[lf],S)
```
which could then take the form like
```julia
has(M,:pres) ? nothing : gridded!(diagn.surface.pres_grid,progn.pres.leapfrog[lf],S)
```
I find both readable, but your suggestions externalises that property of M to a `has` function which we can define in `models.jl` to allow them to be more easily extended? I like that.

src/prognostic_variables.jl
```diff
@@ -96,3 +96,85 @@ function Base.zeros(::Type{PrognosticVariables{NF}},
...
+
+    for (progn_layer, var_layer) in zip(progn.layers, var)
+        _set_var_core!(getfield(progn_layer.leapfrog[lf], varname), var_layer)
+    end 
+
+    return progn 
+end 
+
+function _set_var_core!(var_old::LowerTriangularMatrix{T}, var_new::LowerTriangularMatrix{R}) where {T,R}
+    @assert size(var_old) != (0,0)
```
milankl, line 117: I think this could just `return nothing`?

src/prognostic_variables.jl
```diff
@@ -96,3 +96,85 @@ function Base.zeros(::Type{PrognosticVariables{NF}},
     pres = zeros(SurfaceLeapfrog{NF},_lmax,_mmax)                     # 2 leapfrog time steps for pres
     return PrognosticVariables(layers,pres,lmax,mmax,N_LEAPFROG,nlev)
 end
+
+# SET_VAR FUNCTIONS TO ASSIGN NEW VALUES TO PrognosticVariables
+
+function set_var!(progn::PrognosticVariables{NF}, 
+                  varname::Symbol, 
+                  var::Vector{<:LowerTriangularMatrix},
+                  lf::Integer=1) where NF
```
milankl, line 105: If these lines are changed to
```julia
                 var::LowerTriangularMatrix...;
                 lf::Integer=1) where NF
```
then both `L` and `(L1,L2)` could be used as input? (as ... just unravels the tuple?) Is that helpful or would that just complicate things further down?

src/prognostic_variables.jl
```diff
@@ -96,3 +96,85 @@ function Base.zeros(::Type{PrognosticVariables{NF}},
...
+
+    @assert length(var) == length(progn.layers)
```
milankl, line 107: Do you know whether setting vorticity in a single layer only would make sense?

src/prognostic_variables.jl
```diff
@@ -96,3 +96,85 @@ function Base.zeros(::Type{PrognosticVariables{NF}},
...
+
+    var_sph = [spectral(var_layer, M.spectral_transform) for var_layer in var]
+
+    return set_var!(progn, varname, var_sph, lf)
+end 
+
+function set_var!(progn::PrognosticVariables{NF}, 
+                  varname::Symbol, 
+                  var::Vector{<:AbstractMatrix}, 
+                  Grid::Type{<:AbstractGrid}=FullGaussianGrid,
```
milankl, line 151: Same ... consideration as above

src/prognostic_variables.jl
```diff
@@ -96,3 +96,85 @@ function Base.zeros(::Type{PrognosticVariables{NF}},
...
+    @assert length(var) == length(progn.layers)
+
+    for (progn_layer, var_layer) in zip(progn.layers, var)
+        _set_var_core!(getfield(progn_layer.leapfrog[lf], varname), var_layer)
+    end 
+
+    return progn 
+end 
+
+function _set_var_core!(var_old::LowerTriangularMatrix{T}, var_new::LowerTriangularMatrix{R}) where {T,R}
```
milankl, line 116: Should be also allow `Matrix{R}` here? In case someone doesn't bother doing `LowerTriangularMatrix(::Matrix)`?

src/prognostic_variables.jl
milankl: The reason I'm asking is that this removes the need for a `has(M,:var_name)` function as 0-element arrays will just be skipped with `return nothing`

src/prognostic_variables.jl
```diff
@@ -96,3 +96,85 @@ function Base.zeros(::Type{PrognosticVariables{NF}},
...
+function set_pressure!(progn::PrognosticVariables{NF}, 
+                       pressure::LowerTriangularMatrix,
+                       lf::Integer=1) where NF
+
+    _set_var_core!(progn.pres.leapfrog[lf], pressure)
+
+    return progn
+end
+
+set_pressure!(progn::PrognosticVariables, pressure::AbstractGrid, M::ModelSetup, lf::Integer=1) = set_pressure!(progn, spectral(pressure, M.spectral_transform), lf)
```
milankl, line 175: I'd suggest to make `lf` a keyword argument? if we want to add a layer keyword too, then `set_pressure!(progn,pres,1,1)` might be confusing, but `set_pressure!(progn,pres,lf=1,layer=1)` more intuitive?

maximilian-gelbrecht:
I would not add functions/functionality to set individual layers. It does make the functions more complicated (or needs a separate set of functions). I don't see a super important purpose for it, and with the `get...` functions it is also very easy to emulate that behaviour

```Julia 
vor = get_vorticity(progn)
vor[3] = ...
set_vorticity!(progn, vor) 
```

maximilian-gelbrecht:
Having all of these function, the output (to netcdf) could be rewritten, but I am a bit too lazy to this now right away.

src/prognostic_variables.jl
maximilian-gelbrecht: That's what the other functions are for, they handle grid input. Or do you mean `Matrix` to input SPHs? That could get a bit messy because in other parts we also assume that Matrices are in grid space

src/prognostic_variables.jl
maximilian-gelbrecht: I removed this in favour of the `has` function now. With it, it should always work and no assert is needed

src/prognostic_variables.jl
milankl: Yes, true. No, users can do `LowerTriangularMatrix(::Matrix)` manually then for the spherical harmonic coefficients

maximilian-gelbrecht:
From my side this is finished. 
One important change: I made `PrognosticVariables` parametric on the `ModelSetup`. If no ModelSetup is given, then it defaults to the abstract type. I've done this for the `has` function, so that one can easily return which variables are stored within an instance of `PrognosticVariables`. Maybe this could also be useful for other stuff in the future.

src/initial_conditions.jl
```diff
@@ -75,27 +75,14 @@ function initialize_from_file!(progn_new::PrognosticVariables{NF},M::ModelSetup)
...
-        vars_new = (getproperty(layer_new_lf1,prop) for prop in propertynames(layer_new_lf1))
-    
-        for (var_old,var_new) in zip(vars_old,vars_new)
-            lmax,mmax = size(var_new) .- (2,1)
-            var_old_trunc = spectral_truncation(var_old,lmax+1,mmax)
-            var_new .= convert(LowerTriangularMatrix{Complex{NF}},var_old_trunc)
+    for var_name in var_names
+        if has(progn_new, var_name) 
+            var = get_var(progn_old, var_name) 
+            set_var!(progn_new, var_name, var)
```
milankl, line 83: Neat. I like it.

src/models.jl
```diff
@@ -20,6 +20,8 @@ struct BarotropicModel{NF<:AbstractFloat, D<:AbstractDevice} <: ModelSetup{D}
     device_setup::DeviceSetup{D}
 end
 
+has(::Type{BarotropicModel{NF,D}}, var_name::Symbol) where {NF,D} = var_name in [:vor]
```
milankl, line 23: I believe the `,D}` can be dropped as it's not used (I had to remind myself what D stands for again...)? As far as I understand multi-parametric types in Julia they work like type{T1,T2,T3,T4} to be like type{T1}{T2}{T3}
```julia
julia> struct ABC{A,B,C}
           a::Int
           b::Int
       end

julia> abc = ABC{Float64,Float32,Float16}(1,2)
ABC{Float64, Float32, Float16}(1, 2)

julia> t(abc::ABC{T}) where T = T(abc.a)
julia> t(abc)
1.0
```

milankl: So dispatch is only performed following the first type if the others are not specified.

src/models.jl
```diff
@@ -41,6 +43,9 @@ struct ShallowWaterModel{NF<:AbstractFloat, D<:AbstractDevice} <: ModelSetup{D}
     device_setup::DeviceSetup{D}
 end
 
+has(::Type{ShallowWaterModel{NF,D}}, var_name::Symbol) where {NF,D} = var_name in [:vor, :div, :pres]
```
milankl, line 46: Same here

src/models.jl
```diff
@@ -62,3 +67,13 @@ struct PrimitiveEquationModel{NF<:AbstractFloat,D<:AbstractDevice} <: ModelSetup
     implicit::Implicit{NF}
     device_setup::DeviceSetup{D}
 end
+
+has(::Type{PrimitiveEquationModel{NF,D}}, var_name::Symbol) where {NF,D} = var_name in [:vor, :div, :temp, :humid, :pres]
```
milankl, line 71: And here

src/models.jl
```diff
@@ -62,3 +67,13 @@ struct PrimitiveEquationModel{NF<:AbstractFloat,D<:AbstractDevice} <: ModelSetup
...
+
+"""
+    has(::ModelSetup, var_name::Symbol)
+
+Returns true if the model `M` has a prognostic variable `var_name`, false otherwise. The default fallback is that all variables are included. 
+"""
+has(::Type{ModelSetup{D}}, var_name::Symbol) where D = var_name in [:vor, :div, :temp, :humid, :pres]
```
milankl, line 78: It should also be possible to drop it here? Or do you see any advantage of keeping it?

src/prognostic_variables.jl
```diff
@@ -44,12 +44,11 @@ end
...
-    vor   = zeros(LowerTriangularMatrix{Complex{NF}},lmax+2,mmax+1)
-    lmax,mmax = model isa BarotropicModel   ? (-2,-1) : (lmax,mmax)   # other variables not needed for BarotropicModel
-    div   = zeros(LowerTriangularMatrix{Complex{NF}},lmax+2,mmax+1)
-    lmax,mmax = model isa ShallowWaterModel ? (-2,-1) : (lmax,mmax)   # other variables not needed for ShallowWaterModel
-    temp  = zeros(LowerTriangularMatrix{Complex{NF}},lmax+2,mmax+1)
-    humid = zeros(LowerTriangularMatrix{Complex{NF}},lmax+2,mmax+1)
+    vor = has(model, :vor) ? zeros(LowerTriangularMatrix{Complex{NF}},lmax+2,mmax+1) : LowerTriangularMatrix{Complex{NF}}(undef, 0, 0)
+    div = has(model, :div) ? zeros(LowerTriangularMatrix{Complex{NF}},lmax+2,mmax+1) : LowerTriangularMatrix{Complex{NF}}(undef, 0, 0)
+    temp = has(model, :temp) ? zeros(LowerTriangularMatrix{Complex{NF}},lmax+2,mmax+1) : LowerTriangularMatrix{Complex{NF}}(undef, 0, 0)
+    humid = has(model, :humid) ? zeros(LowerTriangularMatrix{Complex{NF}},lmax+2,mmax+1) : LowerTriangularMatrix{Complex{NF}}(undef, 0, 0)
```
milankl, line 50: Yeah neat.

src/prognostic_variables.jl
```diff
@@ -92,7 +91,199 @@ function Base.zeros(::Type{PrognosticVariables{NF}},
...
+Sets the prognostic variable with the name `varname` in all layers at leapfrog index `lf` 
+with values given in `var` a vector with all information for all layers in spectral space.
+"""
+function set_var!(progn::PrognosticVariables{NF}, 
+                  varname::Symbol, 
+                  var::Vector{<:LowerTriangularMatrix};
+                  lf::Integer=1) where NF
+
+    @assert length(var) == length(progn.layers)
+    @assert has(progn, varname)
```
milankl, line 117: Reminder to self that we should do something about error messages in general. I like the `@assert` here as they actually print what's wrong. However, often I ended up using `throw(BoundsError)` which is kind of clear if there's only one `@boundscheck` before a loop in a function but if there's several ones, you don't know which array sizes don't match. Maybe we can make them somehow more telling...

src/models.jl
maximilian-gelbrecht: I don't understand why exactly, but it is required, otherwise it won't work. I did try it before. Just again: 
```Julia
 MethodError: no method matching has(::Type{SpeedyWeather.PrimitiveEquationModel{Float64, SpeedyWeather.CPUDevice}}, ::Symbol)
  Closest candidates are:
    has(::Type{SpeedyWeather.BarotropicModel{NF}}, ::Symbol) where NF at ~/Nextcloud/SpeedyFork/SpeedyWeather.jl/src/models.jl:23
    has(::Type{SpeedyWeather.ShallowWaterModel{NF}}, ::Symbol) where NF at ~/Nextcloud/SpeedyFork/SpeedyWeather.jl/src/models.jl:46
    has(::Type{SpeedyWeather.PrimitiveEquationModel{NF}}, ::Symbol) where NF at ~/Nextcloud/SpeedyFork/SpeedyWeather.jl/src/models.jl:71
```

src/prognostic_variables.jl
maximilian-gelbrecht: I didn't know that before, but `@assert` let's you specify error messages: 
`@assert has(progn, varname) "No variable with this name found."`

maximilian-gelbrecht:
> Last question, while we are at it: Should we also include `set_vorticity!(progn,0)` to just set the vorticity to zero (or any other scalar? This could be done via
> 
> ```julia
> function set_var!(progn::PrognosticVariables{NF}, 
>                   varname::Symbol, 
>                   s::Number;
>                   lf::Integer=1) where NF
> 
>     for progn_layer in zip(progn.layers
>         fill!(getfield(progn_layer.leapfrog[lf], varname), s)
>     end 
> 
>     return progn 
> end 
> ```
> 
> functions like `set_vorticity!` should just pass on `s::Number` etc?

Yeah, why not, I'll add it

PR #155, opened 2022-09-29
fix get_var for pressure

maximilian-gelbrecht:
A small bug made it through the PR

With this the `initialize_from_file!` will work also for non-barotropic models

### DIFF
--- a/src/initial_conditions.jl
+++ b/src/initial_conditions.jl
@@ -26,16 +26,16 @@ function initial_conditions(M::ModelSetup)
             end
         end
 
+        # SCALING
+        @unpack radius_earth = M.geometry
+        scale!(progn,:vor,radius_earth)
+        scale!(progn,:div,radius_earth)
+
     elseif initial_conditions == :restart
         initialize_from_file!(progn,M)
     else
         throw(error("Incorrect initialization option, $initial_conditions given."))
     end
-
-    # SCALING
-    @unpack radius_earth = M.geometry
-    scale!(progn,:vor,radius_earth)
-    scale!(progn,:div,radius_earth)
 
     return progn
 end
@@ -75,7 +75,7 @@ function initialize_from_file!(progn_new::PrognosticVariables{NF},M::ModelSetup)
     version = restart_file["version"]   # currently unused, TODO check for compat with version
     time = restart_file["time"]         # currently unused
 
-    var_names = [propertynames(progn_old.layers[1].leapfrog[1])..., :pres]
+    var_names = propertynames(progn_old.layers[1].leapfrog[1])
 
     for var_name in var_names
         if has(progn_new, var_name) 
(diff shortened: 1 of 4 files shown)


maximilian-gelbrecht:
I ran out of time for today, I'll write a slightly better test tomorrow and then merge it directly

maximilian-gelbrecht:
@milankl While working on this I noticed that the `PrognosticVariables` instance that is saved to JLD2 is also unscaled. This lead to a wrong initialisation from file both in the old and new version of the `initialization_from_file!` routine. 

It is clear that for the netCDF output units have to be in SI standard. But for saving the Julia structs I would just skip the unscaling, and save them as they are internally used.

milankl:
I believe that can lead to more problems in the future as we have to make sure that possibly different scalings are compatible. The radius scaling probably will always stay as it is, but we may need to introduce further scalings. Is there a big disadvantage skipping the scaling?

milankl:
Alternatively we can just add information about the scaling in the restart file if we decide to keep the variables scaled?

maximilian-gelbrecht:
I see no disadvantage skipping this scaling for the JLD2 file. For me it intuitively makes sense when the `PrognesticVariables` instance is just saved as it is at the last time step without making any adjustments for the scaling. 

NetCDF output of course remains in SI.

milankl:
Yes, shall we add fields like `"vorticity_scaling" = radius_earth`, `"divergence_scaling" = radius_earth` to make this clearer? in the restart file?

maximilian-gelbrecht:
One counterargument would be that an instance of `PrognosticVariables` is also returned to the user when using `run_speedy`, then that would be non-SI as well

maximilian-gelbrecht:
Ah, but that's only written when `output==true`. So currently the `p` that gets returned from a `run_speedy` is in SI when `output==true` and scaled when `output==false`. Yeah, that's inconsistent. So I'd still say, it's okay to non unscale the `PrognosticVariables`

maximilian-gelbrecht:
Okay, I am confused by the whole scaling thing. There actually was an additional scaling of the initial conditions when loading them. Anyway, now I changed it so that all tests are working. I'd merge this PR, but maybe we should take a look at the scaling and make this a bit clearer. 

E.g. `PrognosticVariables` could have fields `vorticity_scale`, ... 
and then one would define `scale(progn)`, `unscaled(progn)` directly and agree that really only the netCDF output is in SI.

milankl:
Thanks for merging, yeah generally I think it would be nice and clear to just carry around the scale within the `PrognosticVariables` such that at any point in time one can unscale (or scale more). However, I'd leave that for now, as I believe this becomes only really important when we move towards a 16-bit dynamical core. For that we may want to have a wider range of possible scalings (per layer, per variable, per spectral mode). However, I don't know yet what makes sense for 16-bit arithmetic and what would just unnecessarily complicate things.

When I said to add a field "vorticity_scale" I meant when writing the restart file here

https://github.com/milankl/SpeedyWeather.jl/blob/cda544f1cf9e5af579d051d4d09600443a2ddd29/src/output.jl#L373-L378

Clearly, when the scale is also a field in `PrognosticVariables` that ☝🏼 isn't necessary, such a field could be added here

https://github.com/milankl/SpeedyWeather.jl/blob/cda544f1cf9e5af579d051d4d09600443a2ddd29/src/prognostic_variables.jl#L21-L31

Issue #131, opened 2022-08-26
Shortening the Legendre Transform over m


milankl:
This is an issue to collect ideas around the linear vs quadratic vs cubic truncation when looping over the zonal wavenumbers $m$. With #127 we'll do 
```julia
for m in 1:min(nfreq:mmax+1)
```
with `nfreq  = nlon÷2 + 1` and `nlon` the number of longitude points at a given latitude ring. So for full grids, the loop goes over all `m` but for reduced grids this can be shortened. IFS seems to be doing the following (`j` `nlon` `nfreq`) at Tco79
```
 (   1  20   8) (   2  24  10) (   3  28  12) (   4  32  14) (   5  36  16) (   6  40  18) (   7  44  20) (   8  48  22)
 (   9  52  24) (  10  56  26) (  11  60  27) (  12  64  29) (  13  68  31) (  14  72  33) (  15  76  35) (  16  80  36)
 (  17  84  38) (  18  88  40) (  19  92  41) (  20  96  43) (  21 100  44) (  22 104  46) (  23 108  47) (  24 112  49)
 (  25 116  50) (  26 120  52) (  27 124  53) (  28 128  55) (  29 132  56) (  30 136  57) (  31 140  58) (  32 144  60)
 (  33 148  61) (  34 152  62) (  35 156  63) (  36 160  64) (  37 164  65) (  38 168  67) (  39 172  68) (  40 176  69)
 (  41 180  70) (  42 184  71) (  43 188  72) (  44 192  73) (  45 196  74) (  46 200  75) (  47 204  76) (  48 208  77)
 (  49 212  78) (  50 216  79) (  51 220  79) (  52 224  79) (  53 228  79) (  54 232  79) (  55 236  79) (  56 240  79)
 (  57 244  79) (  58 248  79) (  59 252  79) (  60 256  79) (  61 260  79) (  62 264  79) (  63 268  79) (  64 272  79)
 (  65 276  79) (  66 280  79) (  67 284  79) (  68 288  79) (  69 292  79) (  70 296  79) (  71 300  79) (  72 304  79)
 (  73 308  79) (  74 312  79) (  75 316  79) (  76 320  79) (  77 324  79) (  78 328  79) (  79 332  79) (  80 336  79)
 (  81 336  79) (  82 332  79) (  83 328  79) (  84 324  79) (  85 320  79) (  86 316  79) (  87 312  79) (  88 308  79)
[...]

PR #156, opened 2022-10-04
HEALPix4Grid and FullHEALPix4Grid implementation

hottad:
@milankl  As I promised before, I implemented HEALPix4 grid that we proposed in issue #112 (along with its FullGrid version).
Note, however, that, like our current HEALPixGrid implementation, zonal shifts of the first points on each latitude ring is not turned on yet.

This pull request should pass all the tests in test/runtests.jl.

Results of 10-day integrations visually look very similar to the established OctahedralGaussianGrid.

10day-iteg-OctahedralGaussian-HEALPix
10day-iteg-HEALPix4-FullHEALPix4

I also did some convergence assessment of transform errors by comparing spectral coefficients before and after round-trip transforms. Results attached below:
Double precision:
 transform_convergence_dp
Single precision:
transform_convergence_sp

From these I believe HEALPix4 is at least as accurate as HEALPix.

### DIFF
--- a/src/grids.jl
+++ b/src/grids.jl
@@ -604,7 +604,7 @@ matrix_size(::Type{HEALPix4Grid},nside::Integer)=(2nside,2nside)
 Base.Matrix(G::HEALPix4Grid{T};kwargs...) where T = Matrix!(zeros(T,matrix_size(G)...),G;kwargs...)
 """
     Matrix!(M::AbstractMatrix,
-            G::OctahedralClenshawGrid;
+            G::HEALPix4Grid;
             quadrant_rotation=(0,1,2,3),
             matrix_quadrant=((2,2),(1,2),(1,1),(2,1)),
             )
@@ -615,7 +615,7 @@ Every quadrant of the grid `G` is rotated as specified in `quadrant_rotation`,
 eastward starting from 0˚E. The grid quadrants are moved into the matrix quadrant
 (i,j) as specified. Defaults are equivalent to centered at 0˚E and a rotation
 such that the North Pole is at M's midpoint."""
-Matrix!(M::AbstractMatrix,G::OctahedralClenshawGrid;kwargs...) = Matrix!((M,G);kwargs...)
+Matrix!(M::AbstractMatrix,G::HEALPix4Grid;kwargs...) = Matrix!((M,G);kwargs...)
 
 """
     Matrix!(MGs::Tuple{AbstractMatrix{T},HEALPix4Grid}...;kwargs...)
--- a/src/grids.jl
+++ b/src/grids.jl
@@ -174,6 +174,68 @@ function each_index_in_ring(::Type{<:AbstractHEALPixGrid},  # function for HEALP
 end
 
 """
+    abstract type AbstractHEALPix4Grid{T} <: AbstractGrid{T} end
+
+An `AbstractHEALPix4Grid` is a horizontal grid similar to the standard OctahedralGrid,
+but the number of points in the ring closest to the Poles starts from 4 instead of 20,
+and the longitude of the first point in each ring is shifted as in HEALPixGrid.
+Also, different latitudes can be used."""
+abstract type AbstractHEALPix4Grid{T} <: AbstractGrid{T} end
+
+get_nresolution(G::AbstractHEALPix4Grid) = G.nside
+get_nlat_half(::Type{<:AbstractHEALPix4Grid},nside::Integer) = nside
+nlat_odd(::Type{<:AbstractHEALPix4Grid}) = true
+get_nlon_max(::Type{<:AbstractHEALPix4Grid},nside::Integer) = 4nside
+
+function get_nlon_per_ring(::Type{<:AbstractHEALPix4Grid},nside::Integer,j::Integer)
+    nlat = 2nside-1
+    @assert 0 < j <= nlat "Ring $j is outside J$nside grid."
+    nlat_half = (nlat+1)÷2
+    j = j > nlat_half ? nlat - j + 1 : j      # flip north south due to symmetry
+    return nlon_healpix4(nside,j)
+end
+
+get_npoints(::Type{<:AbstractHEALPix4Grid},nside::Integer) = npoints_healpix4(nside)
+get_lon(::Type{<:AbstractHEALPix4Grid},nside::Integer) = Float64[]    # only defined for full grids
+
+function get_colatlons(Grid::Type{<:AbstractHEALPix4Grid},nside::Integer)
+    npoints = get_npoints(Grid,nside)
+    colat = get_colat(Grid,nside)
+
+    colats = zeros(npoints)
+    lons = zeros(npoints)
+
+    ij = 1
+    for j in 1:2nside-1
+        nlon = get_nlon_per_ring(Grid,nside,j)
+        lon = collect(π/nlon:2π/nlon:2π)
+
+        colats[ij:ij+nlon-1] .= colat[j]
+        lons[ij:ij+nlon-1] .= lon
+
+        ij += nlon
+    end
+    return colats, lons
+end
+
+function each_index_in_ring(::Type{<:AbstractHEALPix4Grid}, # function for HEALPix4 grids
+                            j::Integer,                     # ring index north to south
+                            nside::Integer)                 # resolution param
+
+    @boundscheck 0 < j < 2nside || throw(BoundsError)       # ring index valid?
+    if j <= nside                                           # northern hemisphere incl Equator
+        index_1st = 2j*(j-1) + 1                            # first in-ring index i
+        index_end = 2j*(j+1)                                # last in-ring index i
+    else                                                    # southern hemisphere 
+        n = 4nside^2                                        # total number of points
+        j = 2nside - j                                      # count ring index from south pole
+        index_1st = n - 2j*(j+1) + 1                        # count backwards
+        index_end = n - 2j*(j-1)
+    end
+    return index_1st:index_end                              # range of i's in ring
+end
+
+"""
     G = FullClenshawGrid{T}
 
 A FullClenshawGrid is a regular latitude-longitude grid with an odd number of `nlat` equi-spaced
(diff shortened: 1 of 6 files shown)


milankl:
Thanks Daisuke, this is amazing! Wonderful work, I'll review it in the next few days. Regarding the convergence tests, many thanks for testing these, this was on my todo list for a long time, and glad you had the time to do it. So if I may try to summarize these findings

- The Gaussian and Clenshaw-Curtis grids are exact up to a multiple of the machine precision, but for some reason the error increases slightly (nothing I'm actually worried about) with resolution. This could be because of an increasing inaccuracy of the Legendre polynomials but could also have other reasons. Anyway, I'm happy with those grids, and don't think they need further accuracy improvement for the time being.
- The HEALPix grids increase in accuracy with higher resolution, which might just be because the missing longitude-offset rotation is less wrong at higher resolution. I still don't actually understand why the longitude-offset rotation wasn't working when I last implemented it. Reminder to self to work on this, we clearly need this.

milankl:
Unfortunately #150 also includes some small restructuring: I've decided to use `nlat_half` for every grid for clearity. Meaning `n_resolution` shouldn't be used anymore, that doesn't change much, but we should probably merge that pull request before this one here, such that the conflicts are flagged and we can adapt the new structure.

hottad:
Thank you Milan for summarizing what we can learn from this convergence test.
I forgot to attach the script I used to produce these figures. This is what I used: [transform_convergence.jl.txt](https://github.com/milankl/SpeedyWeather.jl/files/9708295/transform_convergence.jl.txt)

> * The Gaussian and Clenshaw-Curtis grids are exact up to a multiple of the machine precision, but for some reason the error increases slightly (nothing I'm actually worried about) with resolution. This could be because of an increasing inaccuracy of the Legendre polynomials but could also have other reasons. Anyway, I'm happy with those grids, and don't think they need further accuracy improvement for the time being.

Yes, it is likely that increasing inaccuracy of the Legendre polynomials is the main reason, but accumulation of rounding errors may also contribute.

I did some literature review on this before and it seems that increase of transform errors with increased resolution inflicts other high-quality SHT implementations as well, as I expound below, so for now we can just accept this as is, as you say.

* Ishioka (2018, JMSJ) https://doi.org/10.2151/jmsj.2018-019 has the following table, which shows increase of errors (both L2-norm and inf-norm), both for ISPACK (Japanese package) and SHTns (French package)
image

* Another similar results from Enomoto (2015, SOLA) https://doi.org/10.2151/sola.2015-033 which documents SHT used in AFES. This one is orthonormality and identity errors of 1D associated Legendre transform only (not 2D spherical harmonics transform) and L2-norm only.
image
* Yet another similar results from Reinecke and Seljbotn (2013) Libsharp paper https://arxiv.org/abs/1303.4945
image
* Round-trip transform errors for Gaussian quadrature on reduced Gaussian grid, double precision. L2 error (rms error) increase linearly with resolution, while max error (inf-norm error) scales as (lmax)^3/2



> * The HEALPix grids increase in accuracy with higher resolution, which might just be because the missing longitude-offset rotation is less wrong at higher resolution. 
Diminished impact from missing longitude-offset rotation is certainly one factor, but I assume what is more important is quadrature error of the Riemann summation (which is only first-order accurate), so it is natural that the accuracy improves in proportion to the number of quadrature points (latitude nodes). I expect that, even when we fix the longitude-shift issue, we will see convergence property similar to the one we have here.

> I still don't actually understand why the longitude-offset rotation wasn't working when I last implemented it. Reminder to self to work on this, we clearly need this.

Thank you. I agree we would have to fix this issue at some point. I will also try to find where we went wrong.

Sorry with the conflict with #150. We wouldn't have to deal with this had I done HEALPix4 earlier. Apologies...

milankl:
> Sorry with the conflict with https://github.com/milankl/SpeedyWeather.jl/pull/150. We wouldn't have to deal with this had I done HEALPix4 earlier. Apologies...

No need to apologize, and I actually think it might be better to merge this first and then we can adapt the new structure in #150 by resolving conflicts therein. I should have just merged things earlier 😉

milankl:
Just ran a T255 simulation using HEALPix4, looks good to me, no obvious artefacts that pop up

image

milankl:
Somehow I get this info popping up with `HEALPix4Grid`, but no stacktrace...
```julia
julia> run_speedy(n_days=10,Grid=HEALPix4Grid,model=:shallowwater,implicit_α=1,output=false,trunc=255,initial_conditions=:restart)
┌ Info: chunk dims
│   tuple(chunk_dimensions...) = (0x000000000020c000, 0x0000000000001000)
│   dimensionality = 0x02
│   dimensionality_size = 0x08
└   chunk_indexing_type = 0x01
```
For both `output=true/false`, only at T255 not at lower resolutions and in fact 6 of these infos at once and that's it.

src/grids.jl
```diff
@@ -466,6 +565,126 @@ get_colat(G::Type{<:HEALPixGrid},nside::Integer) =
...
+            matrix_quadrant=((2,2),(1,2),(1,1),(2,1)),
+            )
+
+Sorts the gridpoints in `G` into the matrix `M` without interpolation.
+Every quadrant of the grid `G` is rotated as specified in `quadrant_rotation`,
+0 is no rotation, 1 is 90˚ clockwise, 2 is 180˚ etc. Grid quadrants are counted
+eastward starting from 0˚E. The grid quadrants are moved into the matrix quadrant
+(i,j) as specified. Defaults are equivalent to centered at 0˚E and a rotation
+such that the North Pole is at M's midpoint."""
+Matrix!(M::AbstractMatrix,G::HEALPix4Grid;kwargs...) = Matrix!((M,G);kwargs...)
```
milankl, line 618: We can probably define this for any `AbstractGrid`?

src/grids.jl
```diff
@@ -475,9 +694,15 @@ function clenshaw_curtis_weights(nlat_half::Integer)
...
+healpix4_weights(nside::Integer) =
+    [nlon_healpix4(nside,j) for j in 1:nside].*(2/npoints_healpix4(nside))
 
 # SOLID ANGLES ΔΩ = sinθ Δθ Δϕ
 get_solid_angles(Grid::Type{<:AbstractGrid},nlat_half::Integer) = 
     get_quadrature_weights(Grid,nlat_half) .* (2π./get_nlons(Grid,nlat_half))
 get_solid_angles(Grid::Type{<:HEALPixGrid},nside::Integer) =
     4π/get_npoints(Grid,nside)*ones(get_nlat_half(Grid,nside))
+get_solid_angles(Grid::Type{<:HEALPix4Grid},nside::Integer) =
+    4π/get_npoints(Grid,nside)*ones(get_nlat_half(Grid,nside))
```
milankl, line 708: We could just `Union{HEALPixGrid,HEALPix4Grid}` here, no? I believe in general it might make sense to add a bit more structure into the grid type hierarchy, what about

abstract type AbstractGrid <: AbstractVector end
abstract type AbstractHEALPixGrid <: AbstractGrid end
abstract type AbstractHEALPix12Grid <: AbstractHEALPixGrid end
abstract type AbstractHEALPix4Grid <: AbstractHEALPixGrid end
```
and the concrete types as subtypes of those?

File: test/spectral_transform.jl [hunk 3]
```diff
@@ -90,6 +93,34 @@ end
     end
 end
 
+@testset "Transform: Individual Legendre polynomials (inexact transforms)" begin
+    @testset for trunc in spectral_resolutions_inexact
```
milankl, line 97: Yeah, really like the idea to split between exact grids and inexact for testing!

test/spectral_transform.jl
```diff
@@ -98,7 +129,7 @@ end
             for Grid in (   FullGaussianGrid,
                             FullClenshawGrid,
                             OctahedralGaussianGrid,
-                            OctahedralClenshawGrid,)
+                            OctahedralClenshawGrid)
                             # HEALPixGrid)
                             
                 P = Parameters(;NF,trunc,model=:shallowwater)
```
milankl, line 135: This line still has `;Grid` missing... at the moment it just tests the default grid over and over again. But somehow there's some errors thrown otherwise, sounds like something for another issue/pull request though...

milankl (commented):
Overall: I'm tempted to merge the pull requests as the remaining issues aren't new issues and then we have the HEALPix4Grid already in, and can address conflicts with #150 afterwards in therein. Any objections @hottad ?

milankl:
As a qualitative check, I've been outputting as `output_grid=:matrix` and I cannot see any issues around the poles

image

I remember that we discussed whether 4 grid points aren't enough at the highest latitude ring. This eddy that's close to the pole here circles smoothly around the poles across all 4 quadrants without experiencing any visual distortion. Sure, one can see the line artefacts between the quadrants because that's where angles suddenly take a 90˚ turn, but that's just the projection.

src/grids.jl
```diff
@@ -173,6 +173,68 @@ function each_index_in_ring(::Type{<:AbstractHEALPixGrid},  # function for HEALP
...
+        nlon = get_nlon_per_ring(Grid,nside,j)
+        lon = collect(π/nlon:2π/nlon:2π)
+
+        colats[ij:ij+nlon-1] .= colat[j]
+        lons[ij:ij+nlon-1] .= lon
+
+        ij += nlon
+    end
+    return colats, lons
+end
```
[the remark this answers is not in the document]
hottad: I agree. Let's think about how best we can do this, later on.

src/grids.jl
```diff
@@ -251,6 +313,43 @@ get_colat(::Type{<:FullGaussianGrid},nlat_half::Integer) =
...
+and the same `nlon` longitudes for every latitude ring. The grid points are closer in zonal direction
+around the poles. The values of all grid points are stored in a vector field `v` that unravels
+the data 0 to 360˚, then ring by ring, which are sorted north to south."""
+struct FullHEALPix4Grid{T} <: AbstractFullGrid{T}
+    data::Vector{T}    # data vector, ring by ring, north to south
+    nlat_half::Int  # number of latitudes on one hemisphere
+
+    FullHEALPix4Grid{T}(data,nlat_half) where T = length(data) == npoints_fullhealpix4(nlat_half) ?
+    new(data,nlat_half) : error("$(length(data))-element Vector{$(eltype(data))} cannot be used to create a "*
+        "F$nlat_half ($(4nlat_half)x$(2nlat_half - 1)) FullHEALPix4Grid{$T}.")
```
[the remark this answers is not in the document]
hottad: Agreed. As we expand the number of grids, there is clearly not obvious way to match each grid with one letter.
When I used J for HEALPix4, I had J for [juggling bean bag](https://commons.wikimedia.org/wiki/File:Juggling_balls.jpg) in my mind.
I'm tempted to group HEALPix4Grid with OctahedralGrid rather than with HEALPix12, though.

src/grids.jl
hottad: If we can, that would be great. I don't see how we can do this for HEALPix12. Any ideas?

src/grids.jl
hottad: Agreed. Let's refactor the code like this when we add another HEALPix-type grid.

test/spectral_transform.jl
hottad: Glad you like it!

test/spectral_transform.jl
hottad: Thank you for letting me know. I didn't notice this. So I was just looping over the same code...

hottad:
Hi Milan, thank you so much for reviewing and merging this PR.
Apologies that I couldn't respond to you earlier...

Your plots with T255 HEALPix4 is just fantastic! and the pole-centered matrix plot is amazing.
Clearly there is no instability or noise around the pole.
This contradicts with  is Courtier and Naughton (1994) that we discussed at #131, which is amazing and worth investigating in detail (could lead to a publication on its own). 

Thank you once again!

milankl:
On the topic of publication: I believe there's already a bunch of things that would be worth considering to be part of a publication:
- Testing the HEALPix model in global atmospheric model
- Introducing the HEALPix4 grid as an alternative that resembles the octahedral grid while being equal-area
- The implementation of a grid-flexible spectral transform
- Comparing different grids with respect to convergence, accuracy and performance
- The need for 20 points around the pole (although this is probably best tested with an actual octahedral grid that changes nlon from 16+4j  to 4j)
- Different truncations discussed in #131 

I think this would make a fabulous paper. I'd be very happy to run simulations for it, produce graphs and write half the paper. I believe you should suggest a general outline though, as you are much more aware than me what's been published on this in the past and what's brand new.

hottad:
Thank you, Milan. I agree we've already got a lot that's worth writing a paper about.
I believe all the items in the list you gave above are new findings. I will try to figure out an outline in due time.

hottad:
Thank you @milankl for these beautiful plots!
> This actually yields funky pentagons around the Equator for the OctahedralGaussianGrids and the octahedral faces don't align with grid faces. I believe the faces need a clear definition in IFS for the semi-Lagrangian advection scheme do you know how this is done?

Semi-Lagrangian advection for IFS uses the same code for Octahedral Gaussian Grid (what they call O-grid) and the previously-used reduced Gaussian Grid (what they call N-grid) and the code doesn't even define the faces (or the edges).
It does zonal interpolation along the latitude circles first so that points in adjacent latitudes align along the same meridian, and then does meridional interpolation. 

IFS also has a finite-volume version. I do not know much about how it is implemented. From a quick look at this published documentation:
https://gmd.copernicus.org/articles/12/651/2019/
It seems they first construct the primary mesh by connecting the nodes as in their Figure 2 and then derives the dual mesh using a generic method illustrated in their Figure 1.

From their Figure 2, it appears like they do some tricky thing along the Equator, like making triangular meshes by dividing the squares, which is not necessary if we use Clenshaw-Curtis or HEALPix/HEALPix4.

PR #157, opened 2022-10-07
Scaling Prognostic Variables

maximilian-gelbrecht:
Initial draft 

* Does only apply to prognostic variables, not diagnostic variables 
* Is not yet applied to the netCDF output 
* Still need to add some tests

* For future Float16, one could just add a similar scale field to the `PrognosticVariablesLayer` if that is needed per layer and this one keeps to be the global scale

### DIFF
--- a/src/models.jl
+++ b/src/models.jl
@@ -1,4 +1,4 @@
-abstract type ModelSetup{D} end
+abstract type ModelSetup{NF,D} end
 
 """
     M = BarotropicModel(::Parameters,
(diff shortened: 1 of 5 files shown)


milankl:
Thanks yeah I think your idea is right. At the moment the scaling is done a bit ad-hoc in random places in the code, whereas with your idea this could be more centrally in functions and clearer structured. Regarding low precision scaling: Yes, while it's nice to think about this now how this could be done, we shouldn't base any design now on this: We may need it, we may not need it, or we may need it in a different form as we currently think. At some point I'll track precision and scale with tools like [Sherlogs.jl](https://github.com/milankl/Sherlogs.jl) and then we can make a better informed choice.

Issue #158, opened 2022-10-10
BarotropicModel and ShallowWaterModel with several layers


milankl:
At the moment the number of vertical levels `nlev` can be chosen for `BarotropicModel` and `ShallowWaterModel` but the layers aren't actually coupled (BarotropicModel: independently executed, ShallowWaterModel: Only the first layer is part of the time integration). I believe for the BarotropicModel we can do [Charney and Phillips, 1953](https://journals.ametsoc.org/view/journals/atsc/10/2/1520-0469_1953_010_0071_niotqg_2_0_co_2.xml)

![image](https://user-images.githubusercontent.com/25530332/194897575-b6e3e7e5-9b82-42ed-99cd-c2374f8d62f6.png)

for every layer. This is interesting because it suddenly couples the layers through stream function, although vorticity remains the only prognostic variable in each layer. I believe that could be easily set up, even for the n-layer system. At the moment we don’t calculate the stream function explicitly, as it’s not used anywhere else, but thanks to #142 the operators are easily available. We could cut the algorithm to get from vorticity to stream function and from stream function to velocities into two parts to get the numerator. That looks like what’s described in section 3 as method B.

@maximilian-gelbrecht

PR #161, opened 2022-10-12
Add development notes

dmey:
This PR adds information on how to set up SpeedyWeather.jl for local development

@milankl hope this helps. I was missing something like this when starting changing a few things today. Can you please check that this makes sense and suggest changes?


milankl:
Sure, I'm happy to add anything that helps people to use this software. But enlighten me: Why would someone use conda and not Julia's package manager?

DEVELOP.md
```diff
@@ -0,0 +1,43 @@
...
+
+
+## Develop
+
+Make sure to load Revise before importing anything else. For example:
+
+```julia
+using Revise, SpeedyWeather
+run_speedy( output=true)
+run_speedy(n_days=30, trunc=63, Grid=OctahedralGaussianGrid, model=:shallowwater, nlev=8, σ_levels_half=[0.000, 0.050, 0.140, 0.260, 0.420, 0.600, 0.770, 0.900, 1.000], output=true)
```
milankl, line 37: This example currently does not make sense as the ShallowWaterModel ignores vertical levels because a stacked shallow water model is not implemented at the moment. Also I'm not aware of a shallow water model with sigma coordinates in the vertical. I believe the stacked shallow water model uses layer thicknesses per layer as vertical coordinate.

milankl:
`DEVELOP.md` is just a file that people are supposed to read thought, right?

DEVELOP.md
```diff
@@ -34,7 +34,7 @@ Make sure to load Revise before importing anything else. For example:
 ```julia
 using Revise, SpeedyWeather
 run_speedy( output=true)
-run_speedy(n_days=30, trunc=63, Grid=OctahedralGaussianGrid, model=:shallowwater, nlev=8, σ_levels_half=[0.000, 0.050, 0.140, 0.260, 0.420, 0.600, 0.770, 0.900, 1.000], output=true)
+run_speedy(n_days=30, trunc=63, Grid=OctahedralGaussianGrid, nlev=8, σ_levels_half=[0.000, 0.050, 0.140, 0.260, 0.420, 0.600, 0.770, 0.900, 1.000], output=true)
```
milankl, line 37: Sorry, neither this makes currently sense as `model=:barotropic` by default at the moment with no physically sensible treatment of the vertical, see #158. Also choosing `σ_levels_half` will choose `nlev` automatically.

milankl:
Shall we not move that into the documentation instead of having an otherwise unlinked file in the toplevel folder?

milankl:
I'd move it to `/docs/src` and then add "Development" as a new page in `/docs/make.jl`

vchuravy:
I don't think recommending `conda` to install Julia is the best way. I recommend either downloadinf it directly from the website or using juliaup (https://github.com/JuliaLang/juliaup)

milankl:
I don't develop with jupyter notebooks as I find their easily inconsistent state difficult to work with when I develop packages. For plotting, and doing analyses, sure, but otherwise I use VS Code and the REPL therein.

milankl:
I use the REPL. I mean in the end, develop as you like, I feel like if you are at that state you probably don't really need any guidance how to install X? Or you rather check the documentation in your IDE than here

docs/src/develop.md
```diff
@@ -0,0 +1,58 @@
...
+```
+git clone https://github.com/milankl/SpeedyWeather.jl.git
+cd SpeedyWeather.jl
+```
+
+To install Julia, manually download the latest version from the [Julia release page](https://julialang.org/downloads/) or, alternately, install it using [juliaup](https://github.com/JuliaLang/juliaup), the official Julia installer and version multiplexer. For Linux, macOS, and WSL run:
+
+```
+curl -fsSL https://install.julialang.org | sh
+```
```
milankl, line 16: I'm still not convinced that our documentation is the place to describe how to install Julia. I believe most users would go to julialang.org or to its documentation

docs/src/develop.md
```diff
@@ -0,0 +1,58 @@
+# Development notes
+
+## Set up
+
+Clone SpeedyWeather.jl from GitHub and change the current working directory:
+
+```
+git clone https://github.com/milankl/SpeedyWeather.jl.git
+cd SpeedyWeather.jl
+```
```
milankl, line 10: Note that you can also do `pkg> develop SpeedyWeather` in the package manager

docs/src/develop.md
```diff
@@ -0,0 +1,58 @@
...
+
+
+## Install
+
+To install SpeedyWeather.jl as a development package and [Revise.jl](https://github.com/timholy/Revise.jl) to automatically update function definitions in a running session, run:
+
+```
+julia -e 'import Pkg; Pkg.develop(path="."); Pkg.add("Revise")'
+```
```
milankl, line 25: I personally don't use Revise.jl, to update function definitions I execute them in VS Code. If we want to give people advice here then maybe we can choose a more inclusive language, pointing them to different possibilities they have?

docs/src/develop.md
```diff
@@ -0,0 +1,58 @@
...
+```  
+
+## Documentation
+
+To generate the documentation, run:
+
+```
+julia --project=docs -e 'import Pkg; Pkg.instantiate()'
+julia --project=docs docs/make.jl
+```
```
milankl, line 44: Again, this is probably also something people look up in Documenter.jl?

docs/src/develop.md
```diff
@@ -0,0 +1,58 @@
...
+
+
+## Develop
+
+Make sure to load Revise before importing anything else. For example:
+
+```julia
+using Revise, SpeedyWeather
+run_speedy(n_days=30, trunc=63, Grid=OctahedralGaussianGrid, model=:shallowwater, output=true)
+```
```
milankl, line 35: Only if you use Revise ...

docs/src/develop.md
```diff
@@ -0,0 +1,58 @@
...
+
+
+## Testing
+
+To run tests, run: 
+
+```
+julia -e 'import Pkg; Pkg.test("SpeedyWeather")'
+```
```
milankl, line 53: Also something I personally just execute the tests inside VS Code and then make sure I didn't do something stupid with CI. I believe for this info people would check Test in Base and not the documentation here. Maybe I just don't want to be patronizing and let people develop as they want...

PR #164, opened 2022-10-14
LowerTriangularMatrix Broadcast fix

maximilian-gelbrecht:
I lied. I didn't give up. 

I found some code in LinearAlgebra.jl that was helpful. 

However, there's still an error in this draft. I doesn't work for non-square matrices. It's about the indexing in the `copyto!` function. That's definitely fixeable. I didn't want to do a "hacky" solution in which I distinguish between square matrices and those with an added row, it would be nicer to have something more elegant and general.

### DIFF
--- a/src/lower_triangular_matrix.jl
+++ b/src/lower_triangular_matrix.jl
@@ -285,5 +285,37 @@ function scale!(L::LowerTriangularMatrix{T},s::Number) where T
     L
 end
 
+# Broadcast (more or less copied and adjusted from LinearAlgebra.jl)
+import Base: similar, copyto!
+import Base.Broadcast: BroadcastStyle, Broadcasted, DefaultArrayStyle
+import LinearAlgebra: isstructurepreserving, fzeropreserving
+
+struct LowerTriangularStyle <: Broadcast.AbstractArrayStyle{2} end
+
+Base.BroadcastStyle(::Type{<:LowerTriangularMatrix}) = LowerTriangularStyle()
+
+function Base.similar(bc::Broadcasted{LowerTriangularStyle}, ::Type{NF}) where NF
+    inds = axes(bc)
+    if isstructurepreserving(bc) || fzeropreserving(bc) 
+        return LowerTriangularMatrix{NF}(undef, inds[1][end], inds[2][end])
+    end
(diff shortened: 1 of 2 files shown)


src/lower_triangular_matrix.jl
```diff
@@ -285,5 +285,37 @@ function scale!(L::LowerTriangularMatrix{T},s::Number) where T
...
+
+LowerTriangularStyle(::Val{0}) = LowerTriangularStyle()
+LowerTriangularStyle(::Val{1}) = LowerTriangularStyle()
+LowerTriangularStyle(::Val{2}) = LowerTriangularStyle()
+
+function Base.copyto!(dest::LowerTriangularMatrix{T}, bc::Broadcasted{<:LowerTriangularStyle}) where T
+    axs = axes(dest)
+    axes(bc) == axs || Broadcast.throwdm(axes(bc), axs)
+    for i in axs[1]
+        for j in 1:i
```
maximilian-gelbrecht, line 313: This indexing here results in an error for non square matrices. @milankl what's the most elegant way to do this here?

maximilian-gelbrecht: It needs to iterate over all nonzero elements and also have their 2d index, alternatively we could also write a `k2ij` function and use it here, maybe that would be an elegant solution

maximilian-gelbrecht: Never mind, I found something

maximilian-gelbrecht:
Okay, I fixed that issue. The only thing remaining is to add tests. What kind of broadcasts shall we test?

maximilian-gelbrecht:
That's odd. I added another commit, but it is not showing up here. https://github.com/maximilian-gelbrecht/SpeedyWeather.jl/commit/578295e83c3768209eb933db418c3c32804bac53

milankl:
I believe the ones that were already raised in #135 
```julia
julia> L = rand(LowerTriangularMatrix,3,3)
3×3 LowerTriangularMatrix{Float64}:
 0.998746  0.0       0.0
 0.185934  0.783976  0.0
 0.181445  0.950264  0.970376

julia> L .*= 2
3×3 LowerTriangularMatrix{Float64}:
 1.99749   0.0  0.0
 0.371868  0.0  0.0
 0.0       0.0  1.94075
```
should pass for example/

PR #167, opened 2022-10-14
No [l,m] access to LTM in gradients

milankl:


milankl:
@maximilian-gelbrecht I realised we didn't propagate the `@inbounds` for `setindex!(::LowerTriangularMatrix,...)` -- the last commit does that. However, I'm not sure how to best test that because it can create a segfault to write outside the data range. I've done it once on my laptop and killed the session, so I believe that's a good sign 🤣 Uncommented these tests as they obviously shouldn't run regularly with CI...

PR #169, opened 2022-10-27
Add shortwave radiation parametrization

dmey:
This PR adds SPEEDY shortwave radiation parametrization. The code is kept as close as possible to [its original](https://github.com/samhatfield/speedy_physics.f90) but will refactor at a later stage after this PR is merged to keep comparison straightforward. I carry out two type of evaluations. First I compare fortran and julia functions with [speedy_physics_compare](https://github.com/dmey/speedy_physics_compare/blob/44ce685c148ff3208ea98f9216c7b88b4ca41266/evaluate_shortwave_radiation.ipynb).  Second, I evaluate each function for ranges in speedy_physics_comapre. All tests pass.

There are also some general questions and issues common to other parametrization schemes (e.g. overloading of physical quantities such as 'pres' for both atmospheric  surface pressure and its normalized counterpart) but we can discuss them somewhere else.

### DIFF
--- a/src/default_parameters.jl
+++ b/src/default_parameters.jl
@@ -35,6 +35,7 @@ The default values of the keywords define the default model setup.
                                         # specific humidity [g/Kg]
     alhs::Real=2801                     # latent heat of sublimation [?]
     sbc::Real=5.67e-8                   # stefan-Boltzmann constant [W/m^2/K^4]
+    p0::Real=1.e+5                      # Reference pressure (Pa)
 
     # STANDARD ATMOSPHERE
     lapse_rate::Real=6                  # reference temperature lapse rate -dT/dz [K/km]
@@ -86,6 +87,42 @@ The default values of the keywords define the default model setup.
     humid_relax_time_cnv::Real = 6.0        # Relaxation time for PBL humidity (hours)
     max_entrainment::Real = 0.5             # Maximum entrainment as a fraction of cloud-base mass flux
     ratio_secondary_mass_flux::Real = 0.8   # Ratio between secondary and primary mass flux at cloud-base
+
+    # Shortwave radiation: sol_oz
+    solc::Real = 342.0 ## Solar constant (area averaged) in W / m^2
+    epssw::Real = 0.020 ## Fraction of incoming solar radiation absorbed by ozone
+
+    # Shortwave radiation: cloud
+    rhcl1::Real = 0.30 # Relative humidity threshold corresponding to cloud cover = 0
+    rhcl2::Real = 1.00 # Relative humidity correponding to cloud cover = 1
+    rrcl::Real = 1.0 / (rhcl2 - rhcl1)
+    qcl::Real = 0.20 # Specific humidity threshold for cloud cover
+    pmaxcl::Real = 10.0 # Maximum value of precipitation (mm/day) contributing to cloud cover
+    wpcl::Real = 0.2  # Cloud cover weight for the square-root of precipitation (for p = 1 mm/day)
+    gse_s1::Real = 0.40 # Gradient of dry static energy corresponding to stratiform cloud cover = 1
+    gse_s0::Real = 0.25 # Gradient of dry static energy corresponding to stratiform cloud cover = 0
+    clsmax::Real = 0.60 # Maximum stratiform cloud cover
+    clsminl::Real = 0.15 # Minimum stratiform cloud cover over land (for RH = 1)
+
+    # Shortwave radiation: radsw
+    albcl::Real = 0.43  # Cloud albedo (for cloud cover = 1)
+    albcls::Real = 0.50  # Stratiform cloud albedo (for st. cloud cover = 1)
+    abscl1::Real = 0.015 # Absorptivity of clouds (visible band, maximum value)
+    abscl2::Real = 0.15  # Absorptivity of clouds (visible band, for dq_base = 1 g/kg)
+
+    absdry::Real = 0.033 # Absorptivity of dry air (visible band)
+    absaer::Real = 0.033 # Absorptivity of aerosols (visible band)
+    abswv1::Real = 0.022 # Absorptivity of water vapour (visible band, for dq = 1 g/kg)
+    abswv2::Real = 15.0  # Absorptivity of water vapour (near IR band, for dq = 1 g/kg)
+
+    ablwin::Real = 0.3   # Absorptivity of air in "window" band
+    ablco2::Real = 6.0   # Absorptivity of air in CO2 band
+    ablwv1::Real = 0.7   # Absorptivity of water vapour in H2O band 1 (weak), (for dq = 1 g/kg)
+    ablwv2::Real = 50.0   # Absorptivity of water vapour in H2O band 2 (strong), (for dq = 1 g/kg)
+    ablcl2::Real = 0.6   # Absorptivity of "thin" upper clouds in window and H2O bands
+    ablcl1::Real = 12.0   # Absorptivity of "thick" clouds in window band (below cloud top)
+    epslw::Real = 0.05    # Fraction of blackbody spectrum absorbed/emitted by PBL only
+
 
     # TIME STEPPING
     startdate::DateTime = DateTime(2000,1,1)# time at which the integration starts
--- a/src/default_parameters.jl
+++ b/src/default_parameters.jl
@@ -35,7 +35,6 @@ The default values of the keywords define the default model setup.
                                         # specific humidity [g/Kg]
     alhs::Real=2801                     # latent heat of sublimation [?]
     sbc::Real=5.67e-8                   # stefan-Boltzmann constant [W/m^2/K^4]
-    p0::Real=1.e+5                      # Reference pressure (Pa)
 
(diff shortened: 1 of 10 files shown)


milankl:
This is fantastic! Thanks a million. Will give it a proper review next week but already upfront

- please avoid hardcoding Float64s in functions, write a = convert(NF,1.23) instead to minimise type promotions and avoid a cascading away from NF. Note that whenever ::NF is annotated this happens automatically. 
- would it be possible to add an optional daily cycle as well? We already have a seasonal cycle boolean in default_parameters.jl and this could be added
- Could we group the parameters for radiation similar to how Alistair already did that for humidity?
- As you are confident that this PR replicates Fortran speedy already, feel free to diverge with the variable naming. I’m zero fan of these cryptic names.
- Fortran-SPEEDY has the number of stratospheric levels hardcoded (that’s why vertical loops sometimes start with 2 or 3). Could you reuse that parameter from default_parameters.jl instead?

milankl:
> Yes but not now as it's not in the repo with MIT licence so will need to do it differently and blindfolded! I've already found https://github.com/CliMA/Insolation.jl we can use for that...

Great thanks! No worries, I don't think this is really needed right now, but just so that we keep this in mind. Also there's a `time::DateTime` variable that is already passed around in the time stepping, but didn't make it's way yet into the parameterizations. @white-alistair maybe we can adapt tendencies.jl etc to pass on that variable as well. It always represents the central time in the leapfrog time stepping, the one that should be used to evaluate all tendencies (not sure about radiation which I believe is evaluated less often we can adapt that accordingly).

milankl:
> sorry I don't understand, can you point me to that?

We started something that tries to address this problem where suddenly many many parameters flood into `Parameters` that, however, should be grouped for clarity. Effectively something like
```julia
julia> using SpeedyWeather
julia> p,d,m = initialize_speedy();
julia> m.parameters.magnus_coefs
SpeedyWeather.MagnusCoefs{Float64}
  e₀: Float64 6.108
  T₀: Float64 273.16
  T₁: Float64 35.86
  T₂: Float64 7.66
  C₁: Float64 17.269
  C₂: Float64 21.875


julia> m.parameters.magnus_coefs.T₀
273.16
```
Those parameters then go in `parameter_structs` with some default values, and a instance is initialised when `Parameters` is called. I hope that makes sense. It's not perfect yet, but the first approach to group parameters, especially those for the parameterizations are not used in large parts of the code, hence the idea.

milankl:
> sure, me neither but I don't want to deviate from the original code with this PR given that there is a lot of code that will need to be refactored anyway which will lead to some of those vars being dropped. The remaining we can change then!?

Absolutely, thanks for your effort in the first place! It's not required for now but just a suggestion/inspiration for the future!

white-alistair:
> 

Yes, I am still planning to tackle this very soon!

white-alistair:
> > sorry I don't understand, can you point me to that?
> 
> We started something that tries to address this problem where suddenly many many parameters flood into `Parameters` that, however, should be grouped for clarity. Effectively something like
> 
> ```julia
> julia> using SpeedyWeather
> julia> p,d,m = initialize_speedy();
> julia> m.parameters.magnus_coefs
> SpeedyWeather.MagnusCoefs{Float64}
>   e₀: Float64 6.108
>   T₀: Float64 273.16
>   T₁: Float64 35.86
>   T₂: Float64 7.66
>   C₁: Float64 17.269
>   C₂: Float64 21.875
> 
> 
> julia> m.parameters.magnus_coefs.T₀
> 273.16
> ```
> 
> Those parameters then go in `parameter_structs` with some default values, and a instance is initialised when `Parameters` is called. I hope that makes sense. It's not perfect yet, but the first approach to group parameters, especially those for the parameterizations are not used in large parts of the code, hence the idea.

Yes, I am still planning to tackle this very soon!

white-alistair:
@dmey great work, I especially like the idea of doing a direct comparison with the Fortran Speedy functions. 

As long as we keep the old and obscure variable names, then, could we perhaps make sure that they are all commented with their meaning until we refactor them? Exactly like you've done for the parameters in `default_parameters.jl`, which I like.

test/shortwave_radiation.jl
```diff
@@ -0,0 +1,149 @@
...
+                0.9526258427031802 0.3788324531779398 1.0 1.0;
+                0.9286716429916902 0.2276374390970355 0.994618802250959 0.6801722651144487;
+                0.02148265489609933 0.12596241035550795 0.7900788064253661 4.906182803130207e-8;
+                0.9287847242363858 0.22819245387064319 0.31050206404533354 5.234705430612321e-37
+                ], rtol=0.1)
+            @test isapprox(column.tend_t_rsw, [
+                11.947794979452055, 27.611640775148402, 42.565299927060124, 40.46336031165737
+                ], rtol=0.1)
+            # Range is generally order 10^-5
+            @test column.temp_tend ≈ [7.189150408440864e-6, 1.2141337616807728e-5, 1.3196339560843619e-5, 1.5994200814647754e-5]
```
white-alistair, line 146: Where do all of the numbers in these tests come from? From the Fortran Speedy functions?

src/shortwave_radiation.jl
```diff
@@ -0,0 +1,345 @@
...
+    humid, icltop, cloudc, clstr, ozupp, ozone, zenit, stratz, fsol, qcloud, albsfc,
+        nlev = column
+    @unpack σ_levels_full, σ_levels_thick = model.geometry
+
+    @unpack albcl, albcls, abscl1, abscl2, absdry, absaer, abswv1, abswv2, ablwin, ablco2, ablwv1,
+        ablwv2, ablcl2, ablcl1, epslw = model.parameters
+
+    # Locals variables
+    flux::Vector{NF} = fill(NaN, 2)
+    stratc::Vector{NF} = fill(NaN, 2)
```
milankl, line 212: This first creates a `Vector{Float64}` and then converts it to `Vector{NF}` and the previous vector is eventually garbage collected. You can do directly `fill(convert(NF,NaN),2)` instead. However, is it possible to allocate arrays in `column::ColumnVariables`? Or why does a 2-element vector need to be a vector at all?

src/shortwave_radiation.jl
```diff
@@ -0,0 +1,345 @@
...
+    flux[1] = column.tau2[k, 1] * (flux[1] - ozupp * norm_pres)
+    column.tend_t_rsw[k] = column.tend_t_rsw[k] - flux[1]
+
+    k = 2
+    column.tend_t_rsw[k] = flux[1]
+    flux[1] = column.tau2[k, 1] * (flux[1] - ozone * norm_pres)
+    column.tend_t_rsw[k] = column.tend_t_rsw[k] - flux[1]
+
+    # 3.3  Absorption and reflection in the troposphere
+    for k in 3:nlev
```
milankl, line 274: hardcoded `n_stratosphere_levels`?

src/default_parameters.jl
```diff
@@ -87,6 +88,42 @@ The default values of the keywords define the default model setup.
...
+    abswv1::Real = 0.022 # Absorptivity of water vapour (visible band, for dq = 1 g/kg)
+    abswv2::Real = 15.0  # Absorptivity of water vapour (near IR band, for dq = 1 g/kg)
+
+    ablwin::Real = 0.3   # Absorptivity of air in "window" band
+    ablco2::Real = 6.0   # Absorptivity of air in CO2 band
+    ablwv1::Real = 0.7   # Absorptivity of water vapour in H2O band 1 (weak), (for dq = 1 g/kg)
+    ablwv2::Real = 50.0   # Absorptivity of water vapour in H2O band 2 (strong), (for dq = 1 g/kg)
+    ablcl2::Real = 0.6   # Absorptivity of "thin" upper clouds in window and H2O bands
+    ablcl1::Real = 12.0   # Absorptivity of "thick" clouds in window band (below cloud top)
+    epslw::Real = 0.05    # Fraction of blackbody spectrum absorbed/emitted by PBL only
```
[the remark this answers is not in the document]
milankl: I'll add a commit to address that.

milankl:
The last commit adds
```julia
@with_kw struct RadiationCoefs{NF<:Real} <: Coefficients
```
with all the parameters that you put into `default_parameters.jl`. Hence I changed the `@unpack`s to
```julia
@unpack cp, p0 = model.parameters.radiation_coefs
```
and in `default_parameters.jl` we simply do
```julia
    # Radiation
    radiation_coefs::RadiationCoefs = RadiationCoefs{NF}()
```
now. Technically, you can even select a few parameters as "this is what a user may want to change" (I assume many of them do not fall into this category?) and pass them on as an argument.

milankl:
To illustrate the latter, one can do
```julia
using SpeedyWeather
my_radiation = SpeedyWeather.RadiationCoefs(p123=12.345)
run_speedy(radiation_coefs=my_radiation)
```
which groups the radiation coefficients, and moves their access into a struct that can be defined in global scope.

milankl:
Is there anything missing to merge this pull request?

milankl:
@dmey ?

milankl (approved):

PR #173, opened 2022-11-01
Adapt functions

katharinamaetschke:
Adapt functions for for boundaries.jl and column_variables.jl, already tested with ]test SpeedyWeather.


maximilian-gelbrecht:
@katharinamaetschke Can you please choose slightly more specific names for the PRs? Then it is easier to see what you did in hindsight when just looking at the list of PRs. E.g. in this case: "Adapt ColumnVariables and Boundaries"

PR #176, opened 2022-11-03
Use BarotropicModel instead of :barotropic etc

milankl:
Same for `ShallowWaterModel`, `PrimitiveEquationModel`.

### DIFF
--- a/src/define_diffusion.jl
+++ b/src/define_diffusion.jl
@@ -86,14 +86,14 @@ function HorizontalDiffusion(   P::Parameters,          # Parameter struct
         end
     end
 
-    if P.model <: BarotropicModel || P.model <: ShallowWaterModel       # orographic correction not needed
+    if P.model <: Barotropic || P.model <: ShallowWater                 # orographic correction not needed
         
         temp_correction_vert        = zeros(0)                          # create dummy arrays
         humid_correction_vert       = zeros(0)
         temp_correction_horizontal  = zeros(LowerTriangularMatrix{Complex{P.NF}},0,0) 
         humid_correction_horizontal = zeros(LowerTriangularMatrix{Complex{P.NF}},0,0) 
 
-    else    # P.model <: PrimitiveEquationModel, orographic correction only needed then
+    else    # P.model <: PrimitiveEquation, orographic correction only needed then
 
         temp_correction_vert        = zeros(0)                          # create dummy arrays
         humid_correction_vert       = zeros(0)
--- a/src/define_diffusion.jl
+++ b/src/define_diffusion.jl
@@ -86,14 +86,14 @@ function HorizontalDiffusion(   P::Parameters,          # Parameter struct
         end
     end
 
-    if P.model == :barotropic || P.model == :shallowwater               # orographic correction not needed
+    if P.model <: BarotropicModel || P.model <: ShallowWaterModel       # orographic correction not needed
         
         temp_correction_vert        = zeros(0)                          # create dummy arrays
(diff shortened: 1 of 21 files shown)


milankl:
@maximilian-gelbrecht I thought it would be more consistent to use `model::Type{<:ModelSetup}` throughout, meaning that we do
```julia
run_speedy(model=ShallowWaterModel)
```
instead of `model=:shallowwater`. This might give us more scope in the future for externally defined `ModelSetup`s and (most important ofc) it works with auto-completion 😉

milankl:
Ah same problem as always, we cannot define Parameters without having defined BarotropicModel and vice versa 💩

maximilian-gelbrecht:
But here we have an abstract type. Shouldn't this just work when defining `abstract type ModelSetup{D} end` e.g. directly in the SpeedyWeather.jl file (or somewhere else where it is loaded before all other files).

milankl:
Yeah, but we then also would need a `AbstractBarotropicModel`... which can be defined before Parameters and used therein

maximilian-gelbrecht:
Yeah, use in `Parameters` wouldn't be a problem, one could do the default argument also by calling a function at run time. But you are also using the model types in `define_diffusion.jl` which always needs to be called before defining the concrete model types. 

Mh, ... so either introducing all the abstract types, or maybe we just keep it as it is?

Long-term, an alternative could also be to do something a bit more akin of the syntax of DiffEq.jl etc: 
```Julia 
model = ShallowWaterModel(kwargs...)
p, d = solve(model)
```

src/define_diffusion.jl
```diff
@@ -86,14 +86,14 @@ function HorizontalDiffusion(   P::Parameters,          # Parameter struct
         end
     end
 
-    if P.model == :barotropic || P.model == :shallowwater               # orographic correction not needed
+    if P.model <: BarotropicModel || P.model <: ShallowWaterModel       # orographic correction not needed
```
maximilian-gelbrecht, line 89: I'd assume that use of the model types here, would always require the added abstract types. This file needs to be called before the file that defines the models, as it needs `HorizontalDiffusion`

milankl:
What about
```julia
abstract type ModelSetup{D} end
abstract type Barotropic{D} <: ModelSetup end
abstract type ShallowWater{D} <: ModelSetup end
abstract type PrimitiveEquation{D} <: ModelSetup end
```
and then subtype `BarotropicModel` etc from there?

maximilian-gelbrecht:
My impression is that other libraries also use more parametric types in the struct because of this problem, and specify the types of the fields in the structs less often. This does definitely come as the cost of less readable code though

maximilian-gelbrecht:
> What about
> 
> ```julia
> abstract type ModelSetup{D} end
> abstract type Barotropic{D} <: ModelSetup end
> abstract type ShallowWater{D} <: ModelSetup end
> abstract type PrimitiveEquation{D} <: ModelSetup end
> ```
> 
> and then subtype `BarotropicModel` etc from there?

Yeah, I guess that's the easiest way to do it.

milankl:
Okay that's how it's now!
```julia
abstract type ModelSetup end
abstract type Barotropic <: ModelSetup end
abstract type ShallowWater <: ModelSetup end
abstract type PrimitiveEquation <: ModelSetup end
```
and
```julia
struct BarotropicModel <: Barotropic
struct ShallowWaterModel <: ShallowWater
struct PrimitiveEquationModel <: PrimitiveEquation
```
so without `Model` it's the abstract supertype that can be used early on, with `Model` it's the concrete type.

PR #177, opened 2022-11-07
Add longwave radiation parametrization

dmey:
This PR adds SPEEDY longwave radiation parametrization. The code is kept as close as possible to [its original](https://github.com/samhatfield/speedy_physics.f90) but will refactor at a later stage after this PR is merged to keep comparison straightforward. I carry out two type of evaluations. First I compare fortran and julia functions with [speedy_physics_compare](https://github.com/dmey/speedy_physics_compare/blob/44ce685c148ff3208ea98f9216c7b88b4ca41266/evaluate_longwave_radiation.ipynb).  Second, I evaluate each function for ranges in speedy_physics_comapre. All tests pass.

### DIFF
--- a/src/longwave_radiation.jl
+++ b/src/longwave_radiation.jl
@@ -5,10 +5,9 @@
 
 Initialise variables and parameters used by the longwave radiation parametrization 
 """
-function initialise_longwave_radiation!(
-    P::Parameters, K::ParameterizationConstants
-)
-    radset!(P, K)
+function initialise_longwave_radiation!(K::ParameterizationConstants,
+                                        P::Parameters) 
+    radset!(K,P)
 end
 
 """
@@ -19,9 +18,8 @@ end
 Compute air temperature tendencies from longwave radiation for an atmospheric column.
 For more details see http://users.ictp.it/~kucharsk/speedy_description/km_ver41_appendixA.pdf
 """
-function longwave_radiation!(
-    column::ColumnVariables{NF}, model::PrimitiveEquationModel
-) where {NF<:AbstractFloat}
+function longwave_radiation!(   column::ColumnVariables,
+                                model::PrimitiveEquationModel)
 
     radlw_down!(column, model)
     compute_bbe!(column, model)
@@ -34,7 +32,9 @@ end
 Compute energy fractions in four longwave bands as a function of temperature.
 
 """
-function radset!(P::Parameters, K::ParameterizationConstants)
+function radset!(   K::ParameterizationConstants,
+                    P::Parameters)
+
     @unpack NF, nband = P
     @unpack epslw = P.radiation_coefs
     @unpack fband = K
@@ -56,7 +56,6 @@ function radset!(P::Parameters, K::ParameterizationConstants)
             fband[jtemp, jb] = fband[320, jb]
         end
     end
-    return fband
 end
 
 
--- a/src/parameterization_constants.jl
+++ b/src/parameterization_constants.jl
@@ -9,8 +9,7 @@ function ParameterizationConstants(P::Parameters)
     @unpack NF, nband = P
     fband = nans(NF,400,nband)   # Energy fraction emitted in each longwave band = f(T)
 
-    # conversion to number format NF happens here
-    K = ParameterizationConstants{NF}(fband)
-    initialise_longwave_radiation!(P, K)
+    K = ParameterizationConstants(fband)
+    initialise_longwave_radiation!(K,P)
     return K
 end


src/longwave_radiation.jl
```diff
@@ -0,0 +1,203 @@
...
+    for jb = 1:4
+        for jtemp = 100:199
+            fband[jtemp, jb] = fband[200, jb]
+        end
+        for jtemp = 321:400
+            fband[jtemp, jb] = fband[320, jb]
+        end
+    end
+    return fband
+end
```
milankl, line 29: Do I understand this right that `radset!` is a function that is only executed once as a precomputing step? Cause then I would suggest to add a `ParameterizationConstants` field to `PrimitiveEquationModel` and execute `radset!` in `initialize_speedy`.

milankl: (and move `fband` from `Parameters` into `ParameterizationConstants`, cause it's not a parameter a user would choose but a precomputed array with constant entries ...)

src/longwave_radiation.jl
```diff
@@ -0,0 +1,203 @@
...
+    end
+    
+    # Mean temperature in stratospheric layers
+    for k=1:n_stratosphere_levels-1
+        column.st4a[k, 2] = NF(0.75) * temp[k] + NF(0.25) * column.st4a[k, 1]
+        column.st4a[k+1, 2] = NF(0.50) * temp[k+1] + NF(0.25) * (column.st4a[k, 1] + column.st4a[k+1, 1])
+    end
+
+    # Temperature gradient in tropospheric layers
+    anish  = NF(0.5 * 1.)
```
milankl, line 65: what? 😄

src/longwave_radiation.jl
```diff
@@ -0,0 +1,203 @@
...
+    # 3.0 Emission ad absorption of longwave downward flux.
+    #    For downward emission, a correction term depending on the
+    #    local temperature gradient and on the layer transmissivity is
+    #    added to the average (full - level) emission of each layer.
+
+    # 3.1  Stratosphere
+    k = n_stratosphere_levels-1
+    for jb = 1:nband-2
+        emis = 1. - tau2[k, jb]
+        brad = fband[convert(Int, temp[k]), jb] * (column.st4a[k, 1] + emis * column.st4a[k, 2])
```
milankl, line 97: Okay now I see how `fband` is supposed to be used. Interesting! So on the computational side we have currently 1 conversion to integer then 1 pull from memory. Alternatively we can at some point think about recomputing the mahts behind fband on the fly: Something like `(NF(0.148) - NF(3.0e-6) * (jtemp - 247)^2) * eps1` looks like 4 arithmetic operations to me, which can even be fused.

milankl:
Reminder to self to solve the `fband` in `Parameters` issue then merge.

milankl:
Yes, something like looks good! I'm wondering whether we should at this stage also just define
```julia
struct Contants
     dynamics::DynamicsConstants
     parameterizations::ParameterizationConstants
end

with the `dynamics` field the current `Constants` struct, which mostly contains time stepping and stuff used in the dynamial core. Then for

```julia
struct ParameterizationConstants
    radiation::RadiationConstants
    clouds::CloudsConstants
    ...
end
```

such that we define in speedy internally `SpeedyRadiationConstants <: RadiationConstants` but every other scheme can define their own. Such a scheme would then only need to provide the function `initialize_radiation_constants(::RadiationConstants,::Parameters)` and can do that whichever way they want. I'm still happy to merge things right now as they are, as it's good enough, and we'll rework things soon anyway, but just as a thought going forward!

src/parameterization_constants.jl
```diff
@@ -0,0 +1,16 @@
+"""
+Parameters for radiation parameterizations.
+"""
+@with_kw struct ParameterizationConstants{NF<:AbstractFloat}
+    fband::Matrix{Real}
+end
```
milankl, line 6: without default value skip the `@with_kw` constructor macro. We also should have `fband::Matrix{NF}` for type stability. And I would put everything else that's a constant in our way of thinking about it in here too.

src/parameterization_constants.jl
```diff
@@ -0,0 +1,16 @@
+"""
+Parameters for radiation parameterizations.
+"""
+@with_kw struct ParameterizationConstants{NF<:AbstractFloat}
+    fband::Matrix{Real}
+end    
+
+function ParameterizationConstants(P::Parameters)
+    @unpack NF, nband = P
+    fband = fill(NF(NaN), 400, nband)   # Energy fraction emitted in each longwave band = f(T)
```
milankl, line 10: This can now be written as `fband = nans(NF,400,nband)` also the 400 looks weirdly hardcoded, any clue with 400 and not, say, 500?

milankl:
Yes, absolutely! This is fantastic. Just changed the order of some arguments in bang functions `f!` such that the first argument is the one being mutated. I know it's not written down anywhere, but this is a Julia convention. So like `f!(argument_that_will_be_changed,...,read_only_argument,....)`

PR #178, opened 2022-11-07
Remove input data folder from SpeedyWeather.jl

dmey:
This PR removes the input_data folder from SpeedyWeather.jl and adds functions to check and download input data at the first run of  SpeedyWeather.jl. As more input data are likely to be added soon (e.g. surface fluxes param), I thought this is a good time to add something like this before this repo gets too bloated. For now, I use the same licence as in SpeedyWeather.jl  but we should eventually correct the copyright and licencing information to that of the data. For now, and for simplicity,  the input_data archive is stored on GitHub at https://github.com/dmey/speedy-weather-input-data but we should eventually move it onto HTTP storage. @milankl more immediately, if you like the sound of this, I would like to move the input_data repo under your username or on org for SpeedyWeather.jl.

### DIFF
--- a/src/utility_functions.jl
+++ b/src/utility_functions.jl
@@ -176,3 +176,18 @@ function download_input_data(input_data_path)
     close(r)
     mv(dir_names[1], input_data_path, force=true);
 end
+
+# NAN initialisation
+"""
+    A = nans(T,dims...)
+
+Allocate array A with NaNs of type T. Similar to zeros(T,dims...)."""
+function nans(::Type{T},dims...) where T
+    return fill(convert(T,NaN),dims...)
+end
+
+"""
+    A = nans(dims...)
+
+Allocate A::Array{Float64} with NaNs."""
+nans(dims...) = nans(Float64,dims...)
--- a/test/utility_functions.jl
+++ b/test/utility_functions.jl
@@ -46,3 +46,20 @@ end
     @test SpeedyWeather.readable_secs(1.23) == CompoundPeriod(Second(1), Millisecond(230))
     @test SpeedyWeather.readable_secs(0.123) == CompoundPeriod(Second(0), Millisecond(120))
 end
+
+@testset "nans" begin
+    for s in ((3,),(3,4),(3,4,5))
+        for T in (Float16,Float32,Float64)
+            A = SpeedyWeather.nans(T,s...)
+            for a in A
+                @test isnan(a)
+            end
+            @test size(A) == s
+        end
+
+        A = SpeedyWeather.nans(s...)
+        for a in A
+            @test isnan(a)
+        end
+    end
+end


milankl:
Just set up github.com/SpeedyWeather and therein a repository called [SpeedyWeatherInputData](https://github.com/SpeedyWeather/SpeedyWeatherInputData)

milankl:
I think we should agree on an interface between SpeedyWeather.jl and SpeedyWeatherInputData.jl, because this might be also the time to think about what to do if someone wants to use their own orography for example. One way could be do officially register SpeedyWeatherInputData.jl and then have as a default
```julia
run_speedy(input_data=SpeedyWeatherInputData)
```
but someone could just load another module
```julia
import MyInputData
using SpeedyWeather
run_speedy(input_data=MyInputData)
```
For orography for example we need just an array on a grid that is supported by the spectral transform in SpeedyWeather.jl, so SpeedyWeatherInputData could provide a function like `get_orography` and the transform is then done in SpeedyWeather.jl. What do you think?

src/utility_functions.jl
```diff
@@ -152,6 +152,31 @@ end
     r == 3 && return rotate_matrix_indices_270(i,j,s)
 end
 
+"""
+Downloads input files as required by SpeedyWeather.jl
+"""
+function download_input_data(input_data_path)
+    input_data_url = "https://github.com/dmey/speedy-weather-input-data/archive/refs/heads/main.zip"
```
milankl, line 159: Yeah, we shouldn't hardcode the url. But I'm not sure what the best way is to provide a nice interface, which remains compatible or at least provides some control between input data version and SpeedyWeather.jl version. I like the idea to just create a download some netcdf files and create a folder once. But not sure what undesired consequences of that are compared to having a registered package with version and we can then adapt and adjust version compatibility in SpeedyWeather.jl.

milankl:
Regarding orga see #182

milankl:
I'm honestly thinking to postpone this idea. While I agree with it in general, moving the data somewhere else and finding a good interface should probably only be done once we've had the chance to play around with it more. This is to understand what we would want from such a code/input data separation and what interface would make most sense. At the moment, I wanted to have primitive equations (but not necessarily all the parameterizations) for v0.5 and only if the surface parameterizations are implemented we will know better how to structure this separation.

Does that make sense?

milankl:
Another point is reproducibility. If we turn some SpeedyWeather.jl release into a doi (say with zenodo) in the future then we would have a dependency on another repository which should be clearly tagged with a version number. What I like about having the smallest default data set within the repository itself is that we don't have to deal with that.

milankl:
Alright, I give this PR a v0.6 milestone for now and we can pick that up again once the primitive equation model is up and running and we actually use all that input data.

milankl:
Superseded by #983

PR #179, opened 2022-11-07
Update copyright and licensing information

dmey:
This PR adds copyright and licensing information for parametrization schemes used in SpeedyWeather.jl.


milankl:
At what point does our implementation count as copyright-independent from Fortran-SPEEDY? Of course, if we copy the algorithm exactly but in another language then it's still Molteni et al's intellectual property, but I doubt that they have invented all the physics and maths behind the algorithms. Just asking when we can call our implementation "our" implementation?

Issue #185, opened 2022-11-09
Solution struct / return the Output struct


maximilian-gelbrecht:
While working with SpeedyWeather.jl now for a bit, I noticed something that I would consider unpractical: When running Speedy the output is only to NetCDF in a path that is set during runtime (and not saved somewhere in a struct that is returned to the user). 

A practical example: 

```Julia 
using SpeedyWeather 
p = run_speedy(Float32, output=true)
```
Now I want to continue to work with this simulation run in the same script. Just from running it like this there's no way of knowing what the path is. If it is the first time I've run Speedy in this folder it is `run0001`, but maybe I've already run it a few times, then it is something else. 

A way of fixing this would be to either directly return the `Output` object as well (maybe only optionally via a kwarg) or to create a separate `Solution` object that can hold all kind of information that is only set during run time / time stepping. A separate Solution object could e.g. also have some wrapper functions around NetCDF.jl that make it mimic the solution objects from DiffEq.jl a bit. 

This would also make writing tests for the netCDF output more reliable.

milankl:
Maybe an easier solution: We could implement a `run_id::Union{Integer,String}` keyword that let's you manually specify the run id (possibly overwriting an existing folder) that way you could do (integer version)
```julia
using SpeedyWeather, Printf
run_id = 1
p = run_speedy(Float32, output=true;run_id)
path = @sprintf("run%04d/output.nc")
vor =  ncread(path, "vor")
```
or
```julia
run_id = "0001"
p = run_speedy(Float32, output=true;run_id)
path = "run$run_id/output.nc"
vor =  ncread(path, "vor")
```
or even
```julia
run_id = "coolest_experiment_ever"
p = run_speedy(Float32, output=true;run_id)
path = "run$run_id/output.nc" # this is now "runcoolest_experiment_ever/output.nc"
vor =  ncread(path, "vor")
```

maximilian-gelbrecht:
Okay, yeah, I am also fine with a slightly more direct solution. 

So, with the keywords that are already there and how this is currently set up, I think we could just add a kwarg `run_id` as Milan suggested and then if `run_id` is empty `get_run_id_path` returns the same id as before and otherwise whatever the user provides is just taken. 

I can do a PR later today or tomorrow that does that.

maximilian-gelbrecht:
[...]

PR #186, opened 2022-11-09
Output Time Int64 and Seconds

maximilian-gelbrecht:
I tested it manually, and it seems to work. 

Depending on how we feel about #185, I'll add a fitting test for the CI.

### DIFF
--- a/src/output.jl
+++ b/src/output.jl
@@ -84,9 +84,9 @@ function initialize_netcdf_output(  progn::PrognosticVariables,
 
     # DEFINE NETCDF DIMENSIONS TIME
     @unpack startdate = M.parameters
-    time_string = "hours since $(Dates.format(startdate, "yyyy-mm-dd HH:MM:0.0"))"
+    time_string = "seconds since $(Dates.format(startdate, "yyyy-mm-dd HH:MM:0.0"))"
     dim_time = NcDim("time",0,unlimited=true)
-    var_time = NcVar("time",dim_time,t=Int32,atts=Dict("units"=>time_string,"long_name"=>"time"))
+    var_time = NcVar("time",dim_time,t=Int64,atts=Dict("units"=>time_string,"long_name"=>"time"))
 
     # DEFINE NETCDF DIMENSIONS SPACE
     @unpack output_grid, nlev = M.parameters
@@ -212,8 +212,8 @@ function write_netcdf_time!(outputter::Output,
     @unpack netcdf_file, startdate = outputter
     i = outputter.output_counter
 
-    time_hrs = [round(Int32,Dates.Second(time-startdate).value/3600)]
-    NetCDF.putvar(netcdf_file,"time",time_hrs,start=[i])    # write time [hrs] of next output step
+    time_sec = [round(Int64,Dates.value(Dates.Second(time-startdate)))]
+    NetCDF.putvar(netcdf_file,"time",time_sec,start=[i])    # write time [hrs] of next output step
     NetCDF.sync(netcdf_file)                                # sync to flush variables to disc
 
     return nothing


milankl:
Okay I just tested your branch, I now remember why I haven't done the transition from int32 and hrs since X to int64 and seconds since X yet: Because ncview (that amazing but crappy software to quickly view videos of nc files) cannot handle it
```
ncview: netcdf_dim_value: unknown data type (10) for
dimension time
```
Anyway, your use case is much more important than me using ncview with a correct time axis (it now just counts up individual seconds for each time step)

maximilian-gelbrecht:
> Okay I just tested your branch, I now remember why I haven't done the transition from int32 and hrs since X to int64 and seconds since X yet: Because ncview (that amazing but crappy software to quickly view videos of nc files) cannot handle it
> 
> ```
> ncview: netcdf_dim_value: unknown data type (10) for
> dimension time
> ```
> 
> Anyway, your use case is much more important than me using ncview with a correct time axis (it now just counts up individual seconds for each time step)

Ncview is practical though! I also love it. But yeah, as long as it still displays something, I would be okay with it. 

I don't see a reason why Float64 shoudn't work, but I didn't test it yet. I would add the thing about the `run_id` as discussed in #185, and then we can also have some consistent tests of the netCDF files more easily then before, which I then also add

maximilian-gelbrecht:
I've added the `run_id` bit from #185. 

Now netCDF output can be tested better through this as well. Currently I set up just two tests, both testing the time axis. 
The tests compare computing the time axis in Float64 to the output in Int64 both for a short dense simulation and for a 1kyr simulation.

src/default_parameters.jl
```diff
@@ -112,14 +112,15 @@ The default values of the keywords define the default model setup.
...
-    output_path::String = pwd()         # path to output folder
-    output_filename::String="output.nc"     # name of the output netcdf file
-    output_vars::Vector{Symbol}=[:vor]      # variables to output: :u, :v, :vor, :div, :
-    compression_level::Integer = 3      # 1=low but fast, 9=high but slow
-    keepbits::Integer = 7               # mantissa bits to keep for every variable
+    verbose::Bool = true            # print dialog for feedback
+    output::Bool = false            # Store data in netCDF?
+    output_dt::Real = 6             # output time step [hours]
+    output_path::String = pwd()     # path to output folder
+    run_id::String = get_run_id(output, output_path)    # name of the output folder, defaults to 4-digit number counting up from 0001
```
milankl, line 119: Do you think it's possible to make `run_id::Union{String,Integer}`? It might just be more convenient to write `run_id=1` instead of always going through `@sprintf("%04d",i)`. Meaning that if people want to use a string, fine, but if someone just wants to number runs, they can write `run_id = 1` instead of `"0001"`

src/default_parameters.jl
```diff
@@ -133,15 +134,16 @@ The default values of the keywords define the default model setup.
     # RESTART
     write_restart::Bool = output        # also write restart file if output==true?
     restart_path::String = output_path  # path for restart file
-    restart_id::Integer = 1             # run_id of restart file in run????/restart.jld2
+    restart_id::String = "0001"         # run_id of restart file in run????/restart.jld2
```
milankl, line 137: Same here. We could use the format "%04d" under the hood, but also allow for integers that are converted like this.

src/output.jl
```diff
@@ -37,34 +37,50 @@ end
...
 
-    if output
+    if output 
         # pull list of existing run???? folders via readdir
-        pattern = r"run\d\d\d\d"                # run???? in regex
+        pattern = r"run-\d\d\d\d"                # run-???? in regex
         runlist = filter(x->startswith(x,pattern),readdir(output_path))
         runlist = filter(x->endswith(  x,pattern),runlist)
-        existing_runs = [parse(Int,id[4:end]) for id in runlist]
+        existing_runs = [parse(Int,id[5:end]) for id in runlist]
```
milankl, line 51: I believe this one will ignore any folders that don't follow the `run-????` with `?` being digits pattern. I guess that is intended?

src/output.jl
```diff
@@ -37,34 +37,50 @@ end
...
 
         # get the run id from existing folders
         if length(existing_runs) == 0           # if no runfolder exists yet
             run_id = 1                          # start with run0001
         else
             run_id = maximum(existing_runs)+1   # next run gets id +1
```
milankl, line 57: Btw, if someone uses "9999" as run_id then the next run_id would be `10000` which is then converted to `"10000"` which however is ignored for `runlist`, so if "9999" exists all runs after that will overwrite into "run-10000"

test/netcdf_output.jl
```diff
@@ -0,0 +1,32 @@
...
+
+        p, d, m = initialize_speedy(Float32, output=true, n_days=1, output_dt=0, run_id="dense-output-test")
+        SpeedyWeather.time_stepping!(p, d, m)
+
+        t = NetCDF.ncread("run-dense-output-test/output.nc", "time")
+        @test t ≈ Int64.(manual_time_axis(m.constants.Δt_sec, m.constants.n_timesteps))
+
+        # this is a nonsense simulation with way too large timesteps, but it's here to test the time axis output
+        # 1kyrs simulation         
+        p, d, m = initialize_speedy(Float32, output=true, n_days=365000, Δt_at_T31=60*24*365*10, output_dt=24*365*10, run_id="long-output-test")
```
milankl, line 23: Yo, a 1kyr simulation in the default configuration takes 45min on my laptop to complete. I think we should try to make tests that one can also run locally in a reasonable amount of time?

maximilian-gelbrecht: Look at it closely, it does 10 year time steps. The whole simulation only takes milliseconds.

src/default_parameters.jl
maximilian-gelbrecht: Yeah, I can add that on Monday.

src/output.jl
maximilian-gelbrecht: Yes, that's intended there. That's only for the automatically chosen names

src/output.jl
maximilian-gelbrecht: Yes, I know. But I would be okay with that. I would say for the things that we are doing, that we can trust the users to not mess the naming like this.

src/default_parameters.jl
maximilian-gelbrecht: Added

maximilian-gelbrecht:
The `run_id` can now be `Integer` or `String`. With this added, I'd merge it if there's nothing else. I'll wait until tomorrow.

PR #187, opened 2022-11-09
Adapt functions DefineDiffusion, DefineImplicit and Geometry

katharinamaetschke:
Adapt functions for define_diffusion, define_implicit and geometry, tested with ]test SpeedyWeather.


maximilian-gelbrecht:
Thanks Katharina. With the move a GitHub orga, we'll also have proper GPU CI soon. Then the whole testing will be easier and directly possible on GitHub as well.

PR #189, opened 2022-11-14
Surface pressure tendency

milankl:
The surface pressure tendency is calculated as

$\frac{\partial \ln p_s}{\partial t}  = -\frac{1}{p_s} \sum_k \nabla \cdot (\mathbf{u}_k \Delta p_k) = -\mathbf{\bar{u}} \cdot \nabla \ln p_s$

with the vertical mean velocities $\mathbf{\bar{u}} = (\bar{u},\bar{v})$

$\bar{u} = \sum_k \sigma_k u_k$

and similarly for $\bar{v}$.

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -196,24 +196,6 @@
 #     end
 # end
 
-# """gradient_latitude! but precalculate the recursion factors `ϵlms` in case they are not provided."""
-# function gradient_latitude!(coslat_u::AbstractMatrix{Complex{NF}},  # output: cos(lat)*u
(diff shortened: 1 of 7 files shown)


milankl:
> ∂ln⁡ps∂t=−1ps∑k∇⋅(ukΔpk)=−u¯⋅∇ln⁡ps

The last equality uses $\Delta p_k = \Delta \sigma_k p_s$ ($\sigma$-coordinates) but when I do this calculation I end up with an additional $-\bar{D}$ term, i.e. the vertically averaged divergence. As speedy.f90 does not use this term I've also left it out and assumed that it's my mistake. But just to keep track of the underlying math this note.

PR #190, opened 2022-11-16
Adapt functions Grids, Models, ParameterStructs and SpectralTransform

katharinamaetschke:
Adapt functions Grids, Models, ParameterStructs and SpectralTransform. I still had an error while testing, but wanted to upload these. I will try to fix the error tomorrow. 

Of the src files that contain structs DefaultParameters, Feedback, GPU and Output are still missing. I'll also upload them the next days.


milankl:
I also often push commits although I know they are failing, that's fine. 

Note that the code structure of grids.jl considerably changed since you forked. I don't think you'd need to adapt the adapt functions (no pun intended 😉 ) but there will be a version conflict. Also we have added `HEALPix4Grid` since 😄

src/grids.jl
```diff
@@ -481,3 +516,11 @@ get_solid_angles(Grid::Type{<:AbstractGrid},nlat_half::Integer) =
     get_quadrature_weights(Grid,nlat_half) .* (2π./get_nlons(Grid,nlat_half))
 get_solid_angles(Grid::Type{<:HEALPixGrid},nside::Integer) =
     4π/get_npoints(Grid,nside)*ones(get_nlat_half(Grid,nside))
+
+
+# GPU methods HEALPixGrid
+
+function Adapt.adapt_structure(to, hg::HEALPixGrid)
+    HEALPixGrid(Adapt.adapt(to, hg.data),
+                     hg.nside)
```
milankl, line 525: `hg.nlat_half` since v0.4

src/grids.jl
```diff
@@ -212,6 +212,15 @@ get_colat(::Type{<:FullClenshawGrid},nlat_half::Integer) = [j/(2nlat_half)*π fo
 get_quadrature_weights(::Type{<:FullClenshawGrid},nlat_half::Integer) = clenshaw_curtis_weights(nlat_half)
 full_grid(::Type{<:FullClenshawGrid}) = FullClenshawGrid    # the full grid with same latitudes
 
+
+# GPU methods FullClenshawGrid
+
+function Adapt.adapt_structure(to, fcg::FullClenshawGrid)
+    FullClenshawGrid(Adapt.adapt(to, fcg.data),
+                     fcg.nlat_half)
+end
```
milankl, line 221: I don't know whether there's anything special within the `Adapt` framework that wouldn't allow this, but one should be able to dispatch that with a single function
```julia
function Adapt.adapt_structure(to, grid::Grid) where {Grid<:AbstractGrid}
    Grid(Adapt.adapt(to, grid.data), grid.nlat_half)
end
```
all `Grid<:AbstractGrid` now use `nlat_half` as resolution parameter.

katharinamaetschke:
> Note that the code structure of grids.jl considerably changed since you forked. I don't think you'd need to adapt the adapt functions (no pun intended 😉 ) but there will be a version conflict. Also we have added `HEALPix4Grid` since 😄

Thanks for the heads up! I noticed now what I was doing wrong: I'm always just syncing my forked gpu branch with the speedy weather gpu branch, but that does not have the newest commits of the main branch of course other than I thought. I'm just updating the files.

src/grids.jl
katharinamaetschke: So, do I understand this correctly: I would just put what you wrote here into the (new) Grid.jl file and then it should be dispatched to all other files in the Grids folder (e.g. full_grids.jl), since there, all the structs are for AbstractGrid types?

milankl: Yes, this should work for all `Grid<:AbstractGrid` as we changed to a uniform interface: The grid points are always in `grid.data` stored as `::Vector` and the resolution parameter is always half the number of latitude rings `nlat_half`. So one function should be good for all grids, but test it, I don't know whether the Adapt framework requires anything special.

katharinamaetschke: Okay, thanks!

maximilian-gelbrecht: I like Milan's suggestion. Could you change it like that, and then I'll also have a look into solving the merge conflicts that are now present due to the updated `gpu` branch

PR #193, opened 2022-11-21
Keep GPU branch up-to-date with Main

maximilian-gelbrecht:
This PR keeps the GPU branch up to date with main. 

Most notably this includes the revised code structure for `Grids`. 

@katharinamaetschke Unfortunately you'll also have to redo the `adapt` functions for `SurfaceVariables`, as this got some changes as well.


milankl:
Also a heads up, as I'm extending the dynamical core at the moment `DiagnosticVariables` is subject to change, not just at the surface but also the `dynamic_variables` on layers.

maximilian-gelbrecht:
Thanks for the heads up. 

CI Test will still fail as long as the adapt functions are not complete. This will be soon however. 

@milankl Katharina will start with writing the actual kernels soon, let's set up GPU CI soon as well. So the kernels can be tested on GPU.

PR #202, opened 2022-12-12
Fix typo

navidcy:


milankl:
Thanks for the efforts, I know the documentation is a bit behind 🙈 But once the primitive equations are up and running and have a definite form that will get a bit more care!

PR #206, opened 2022-12-16
temp, pres initial conditions at rest

milankl:
Initialises temperature $T$ as
$T_s = T_{ref} - \frac{\Gamma \Phi_s}{g}\quad$ (1)
at the surface (`k = nlev`) with lapse rate $\Gamma$, gravity $g$ and surface geopotential $\Phi_s$ and above
$T_k = T_s \sigma_k^{\frac{R\Gamma}{g}}\quad$ (2)
with $\sigma_k$ the sigma vertical coordinate and $R$ the specific gas constant for dry air. In the stratosphere (determined by `n_stratosphere_levels`) $T_k = T_{top}$, a constant.

Logarithm of surface pressure is initialised via lapse rate and surface geopotential as
$\ln p_s = \ln p_{ref} + \frac{g}{R\Gamma}\ln \left(1 - \frac{\Gamma \Phi_s}{gT_{ref}} \right) \quad$      (3)

which derives from the hydrostatic equation
$\frac{\partial p}{\partial z} = -g\rho$
with a change of coordinates from $z$ to $\Phi = gz$, the ideal gas law $p = \rho R T$ with $T = T_{ref} - \Gamma z$ and then integrating over $\Phi$ from 0 (where $p = p_{ref}$) to $\Phi_s$ where $p=p_s$.

@white-alistair While I reverse engineered (2) from Fortran SPEEDY's implementation, do you know how to derive (1)? (I just write these things down here so they can be later copied into the documentation)

### DIFF
--- a/src/boundaries.jl
+++ b/src/boundaries.jl
@@ -34,7 +34,7 @@ function Boundaries(P::Parameters,S::SpectralTransform{NF}) where NF
 
     elseif P.model <: ShallowWater || P.model <: PrimitiveEquation
 
-        # OROGRAPHY
+        # OROGRAPHY (on grid)
         orography_highres = ncfile.vars["orog"][:,:]        # height [m]
 
         # Interpolate/coarsen to desired resolution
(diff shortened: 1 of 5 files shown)


milankl:
Okay just realise it's the same thing. Integrate $\partial_zp = -g\rho$ from $z=0$ to $z=z$, replace $\rho = \frac{p}{RT}$ and $T = T_{ref} - \Gamma z$ and then you can get $T$ back after the integration.

Issue #214, opened 2023-01-06
Merging strategy


dmey:
@milankl I noticed that the last commits were not squashed in #177. I am not sure if this was on purpose but in any case it would be good to discuss this for the future. In my other projects I tend to keep `main` as clean as possible and squash any commit from PRs but you may have other reasons for not doing this. Do you have any preference that we should follow from this point forward when merging PRs?

milankl:
I honestly don't mind that much. I usually try to give telling messages to commits so that it's clear which model part is affected by one. I agree that your commit "remove newline" isn't and so should have been squashed. But "fband" is almost okay, but in general I'd rather have something like "remove xyz in radset!" so that it's clearer from the commit history which parts of the model where affected. Honestly, I often end up tweaking several corners of the model when I implement something, such that I'd rather not have them squashed to better record which model parts where affected. Maybe in short:

- if pull requests affect isolated parts of the model; you can use short, generic commit messages and squash
- else: rather use more specific commit messages and don't squash

Does that make sense?

milankl:
I'm happy to make that decision upon merging, but I'd rather keep the labels topical, meaning that if I want to find all issues/pull requests on parameterizations then I klick on the label parameterizations.

PR #217, opened 2023-01-06
Parameters{Model<:ModelSetup} and orography abstracted

milankl:


milankl:
@dmey regarding squashing in #214, here ☝🏼(look at the commit messages)  I don't find it useful to squash as I feel a lot of information in the git history would be lost if I did. Do you agree?

milankl:
@maximilian-gelbrecht maybe you have an idea: I get a weird error with
```julia
julia> using SpeedyWeather
julia> Parameters()
ERROR: UndefVarError: Model not defined
Stacktrace:
 [1] Parameters()
   @ SpeedyWeather ./util.jl:504
 [2] top-level scope
   @ REPL[2]:1
 [3] top-level scope
   @ ~/.julia/packages/CUDA/Ey3w2/src/initialization.jl:52
```
(note how the error originates from CUDA!?!) The underlying reason is that while previously the number format NF had to be provided, so v0.4 we had
```julia
julia> Parameters()
ERROR: Field 'NF' has no default, supply it with keyword.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] Parameters()
   @ SpeedyWeather ~/.julia/packages/Parameters/MK0O4/src/Parameters.jl:493
 [3] top-level scope
   @ REPL[2]:1
 [4] top-level scope
   @ ~/.julia/packages/CUDA/Ey3w2/src/initialization.jl:52
```
Now the model has to be provided, so something like
```julia
julia> Parameters{BarotropicModel}()
```
works fine. So I went on to define a default model
```julia
Parameters(;kwargs...) = Parameters{default_concrete_model(DEFAULT_MODEL)}(;kwargs...)
```
but somehow the error at the top is triggered. If I define this manually in global scope afterwards like
```julia
julia> SpeedyWeather.Parameters(;kwargs...) = SpeedyWeather.Parameters{SpeedyWeather.default_concrete_model(SpeedyWeather.DEFAULT_MODEL)}(;kwargs...)

julia> Parameters()
Parameters{BarotropicModel}(Float64,...
```
it works fine.

milankl:
I'll merge this still to move forward, but at the moment we'll need to add the `{Model}` parameter to `Parameters` unfortunately.

PR #218, opened 2023-01-07
Linting (format)

dmey:
I propose to start discussing the code-style to use in future commits. For now I have used default from the sciml style in JuliaFormatter but we can change the config to whatever you prefer. You can run the formatter locally with `julia --project=development development/format.jl`. The CI will now fail if the style declared in `.JuliaFormatter.toml` is not respected.


white-alistair:
As I mentioned on https://github.com/SpeedyWeather/SpeedyWeather.jl/issues/220, I am a big fan of using a consistent style. I would also be in favour of developing our own style instead of using sciml.

In case anyone doesn't know, you can also disable JuliaFormatter.jl for certain blocks of code by doing:

```julia
#! format: off
# Your custom formatted Julia here.
#! format: on
```

milankl:
> As I mentioned on #220, I am a big fan of using a consistent style. I would also be in favour of developing our own style instead of using sciml.

To be clear: I'm not against a consistent style, I'm not even consistent with my own style. I'm happy to merge a "this is now everything converted to style XYZ" pull request and then I'll stick to whatever that style says. What would you do differently from SciML's style?

milankl:
I don't feel that linting/formatting is necessary for v0.5 but we should keep the discussion going on what our style should be. Happy to stick to any suggestions, in the meantime I'll just keep doing my somewhat inconsistent style.

white-alistair:
> What would you do differently from SciML's style?

I wouldn't say I feel very strongly about that. I just don't like the idea of some upstream style change resulting hitting every file in this repo. But not that bothered either way.

milankl:
I'd definitely back that position.

milankl:
I'll close this PR as there's now tons of conflicts and I believe it's more hassle to merge them into here than to start afresh with some automatic linting based on the current main branch. But feel free to reopen if that's not correct.

PR #219, opened 2023-01-07
Move parameterisations into folder (closes #183)

### LINKED ISSUE #183: Move parameterisations into folder
Now as there are more and more parameterisations coming in I would suggest to move every parameterization-related to `src/parameterizations` eventually this could even become it's own module, but I don't think this is needed now.

At the moment I think to move in there
- column_variables.jl
- convection.jl
- large_scale_condensation.jl
- thermodynamics.jl
- shortwave_radiation.jl
- longwave_radiation.jl

so basically everything that is called within `parameterization_tendencies!` (but I would leave that one as it defines the interface on the dynamics side).

@white-alistair @dmey

dmey:


milankl:
Yep, happy with that! Btw, you have to write something like `fixes #183` into the description, not into the title so that the issue gets automatically linked.

milankl:
> Thanks and it should be linked directly from the commit.

Doesn't have to be. I rarely can squeeze all changes needed to address an issue into a single commit (and doing so wouldn't allow any reasonable commit message). But if you can address an issue in a single commit, please do so, yes.

white-alistair (approved):
Hey, looks good to me!

Minor suggestion: lower case "parameterizations" for the folder name instead of "Parameterizations"? I think this is more consistent with general Julia style, and the capitalised version makes me think it's a module/submodule, which it's not.

milankl:
Happy with eithe lower or uppercase: You are right about lowercase, but with the anticipation of it being its own module at some point we could already uppercase it. But maybe that's way too soon, so lowercase it should be for now I guess.

However, I'm considering to move everything dynamics (once that's a bit more mature) into `/dynamics`, so we could also just call the parameterizations `/physics` as quite often done in NWP.

milankl:
closing as superceded by #241

PR #223, opened 2023-01-10
ZonalRidge orography added, Geometry with lond,latd

milankl:


milankl:
@dmey the `lat`,`lon` in ColumnVariables was pulled from Geometry, but it turned out they were actually colatitudes and longitudes in radians, not latitude and longitudes. I've changed those to be consistently latitudes and longitudes in ˚N,˚E and therefore also swapped `cos,sin` to `cosd,sind`. However, now there are a bunch of tests failing in shortwave radiation. Could you check what's going on there?

milankl:
Absolutely no problem. For now I'll set the tests to breaking, and merge, then you can check what's going on whenever you got time for this. In general I found tests that are formulated as exact as yours less helpful than more qualitative checks. E.g. you check the results to an accuracy that's many orders of magnitude higher than what we require for the spectral transforms, although they have some exact mathematics in the continuous limit behind them! I've spent to much time in the last year updating tests because they were formulated unnecessarily precise. Given that SpeedyWeather isn't mature nor it's structure very stable yet I'd go for more qualitatitive tests, meaning tests that immediately flag if things go horribly wrong. Does that make sense?

PR #227, opened 2023-01-12
Obtain n_stratosphere_levels from default parameter sigma_tropopause

milankl:
Instead of determining `n_stratosphere_levels` in default_parameters.jl, we now choose `σ_tropopause` the sigma coordinate at which the tropopause starts. This is dynamically more consistent as it will automatically adjust to higher vertical resolutions. The default is 0.2, in which case we'll still have 2 stratospheric levels (as far as I understand it, they conceptually include the tropopause?), but this is now calculated in Geometry(::Parameters) after the sigma coordinates have been calculated.

So most changes here are from `@unpack n_stratosphere_levels = P` with `P` the parameter struct, to `@unpack n_stratosphere_levels = G` with `G` the Geometry struct

@dmey @white-alistair


milankl:
@dmey sorry I have to skip more radiation tests. You formulated the tests for `nlev=4` vertical levels, of which originally 2 (because always 2) were considered stratosphere. Now we would have
```julia
julia> Geometry(Parameters{PrimitiveDryCore}(nlev=4)).n_stratosphere_levels
1
```
one level for the stratosphere/tropopause with 4 vertical levels. In contrast, 60 levels will have
```julia
julia> Geometry(Parameters{PrimitiveDryCore}(nlev=60)).n_stratosphere_levels
18
```
18 levels in the stratosphere/tropopause.

PR #229, opened 2023-01-13
add convenience interpolate method, show ring info for grids

milankl:
We now can omit `nlat_half` and it'll just use the same as from the input grid
```julia
julia> A = randn(OctahedralClenshawGrid,24)             # create data on some grid
julia> SpeedyWeather.interpolate(FullGaussianGrid,A)    # interpolate on another with same nlat_half
4608-element, 48-ring FullGaussianGrid{Float64}:
 -0.6548842797988247
 -0.6404409999872624
  ⋮
```
interpolating onto a lower or higher resolution grid? Still works as before:
```julia
julia> SpeedyWeather.interpolate(HEALPixGrid,256,A)
196608-element, 511-ring HEALPixGrid{Float64}:
  0.032597552308651516
  0.06585833004005627
  ⋮

julia> SpeedyWeather.interpolate(OctahedralGaussianGrid,8,A)
544-element, 16-ring OctahedralGaussianGrid{Float64}:
  0.16641055238859648
 -0.9763586037633543
  ⋮

### DIFF
--- a/src/Grids/Grids.jl
+++ b/src/Grids/Grids.jl
@@ -1,6 +1,7 @@
 #TODO turn into module?
 
 include("grids_general.jl") # def AbstractGrid
+include("show.jl")
 
 include("full_grids.jl")    # def AbstractFullGrid, Full*Grid, * = Gaussian, Clenshaw, HEALPix, HEALPix4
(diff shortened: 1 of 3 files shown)


milankl:
@hottad we now have a really convenient interpolation library too!

PR #230, opened 2023-01-13
InitialConditions abstracted

milankl:
@maximilian-gelbrecht this changes the interface from `initial_condtions=:something` to `initial_conditions=SomeThing` with the latter being an initial condition `<:InitialConditions`. This allows us easier to formulate other initial conditions, even externally.

### DIFF
--- a/src/initial_conditions.jl
+++ b/src/initial_conditions.jl
@@ -32,7 +32,7 @@ end
 function initial_conditions!(   ::Type{StartFromRest},
                                 progn::PrognosticVariables,
                                 model::PrimitiveEquation)
-    homogenuous_temperature!(progn,model)
+    homogeneous_temperature!(progn,model)
     pres_grid = pressure_on_orography!(progn,model)
     # TODO initialise humidity
 end
@@ -128,7 +128,7 @@ function initial_conditions!(   ::Type{StartFromFile},
     return progn_new
 end
 
-function homogenous_temperature!(   progn::PrognosticVariables,
+function homogeneous_temperature!(  progn::PrognosticVariables,
                                     model::PrimitiveEquation)
 
     P = model.parameters
@@ -174,7 +174,7 @@ function homogenous_temperature!(   progn::PrognosticVariables,
     end
 end
 
-function initialize_pressure!(  progn::PrognosticVariables,
+function pressure_on_orography!(progn::PrognosticVariables,
                                 model::PrimitiveEquation)
     
     P = model.parameters
--- a/src/initial_conditions.jl
+++ b/src/initial_conditions.jl
@@ -1,52 +1,100 @@
+abstract type ZonalWind <: InitialConditions end
+abstract type StartFromRest <: InitialConditions end
+abstract type StartFromFile <: InitialConditions end
+abstract type StartWithVorticity <: InitialConditions end
+
 """
     prognostic_variables = initial_conditions(M::ModelSetup)
 
 Initialize the prognostic variables from rest, with some initial vorticity,
 or restart from file."""
-function initial_conditions(M::ModelSetup)
+function initial_conditions(model::ModelSetup)
 
-    progn = initialize_from_rest(M)             # allocate variables in any case
-    @unpack initial_conditions = M.parameters
-    @unpack radius_earth = M.geometry
+    progn = allocate_prognostic_variables(model)    # allocate variables in any case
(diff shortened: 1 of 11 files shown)


maximilian-gelbrecht:
Looks good. I also looked into the error, but I had to stop no before I could make a PR. The error is is due to the scaling with the radius. There's an inconsistency somewhere.

src/initial_conditions.jl
```diff
@@ -1,86 +1,114 @@
...
-        nothing 
+    progn = allocate_prognostic_variables(model)    # allocate variables in any case
+    IC = model.parameters.initial_conditions        # type of initial conditions
 
-    elseif initial_conditions == :barotropic_vorticity              # zonal wind with perturbation
+    initial_conditions!(IC,progn,model)     # dispatch to the type of initial conditions
 
-        mmax = min(15,progn.mmax+1)     # perturb only larger modes
-        lmax = min(15,progn.lmax+1)
+    # SCALING: we use vorticity*radius,divergence*radius in the dynamical core
```
maximilian-gelbrecht, line 18: Scaling is applied to all initial conditions, this means that all input is assumed to be in SI units. I think that's fine, but it should be mentioned in the docstring

test/run_speedy_with_output.jl
```diff
@@ -23,12 +23,12 @@ end
 
 @testset "Restart from output file" begin
     
-    p = run_speedy(Float32, initial_conditions=:restart, restart_id=1)
+    p = run_speedy(Float32, initial_conditions=StartFromFile, restart_id=1)
```
maximilian-gelbrecht, line 26: This line is redundant anyway, we can just delete it.

test/run_speedy_with_output.jl
```diff
@@ -23,12 +23,12 @@ end
...
 
     p, d, m = initialize_speedy(Float32, ShallowWater, output=true, run_id="restart-test")
     SpeedyWeather.time_stepping!(p, d, m)
  
-    progn, diagn, model = initialize_speedy(Float32, ShallowWaterModel, initial_conditions=:restart, restart_id="restart-test")
+    progn, diagn, model = initialize_speedy(Float32, ShallowWaterModel, initial_conditions=StartFromFile, restart_id="restart-test")
 
     for varname in propertynames(progn.layers[1].leapfrog[1])
         if SpeedyWeather.has(progn, varname)
```
maximilian-gelbrecht, line 34: It would suggest to just add the scaling to the test. `p` will always be in model units, and setting the initial conditions in SI units also makes sense.

maximilian-gelbrecht (commented):
With these changes, there is still no way to start it from an arbitrary field, right? Maybe we could add a `StartFromProgn` so that you can hand over an instance of `PrognosticVariables`. Then `StartFromFile` can also use this.

test/run_speedy_with_output.jl
milankl: I decided to move all scaling into time_stepping!. This way all prognostic variables outside the dynamical core aren't scaled, and then within one function, before the main time loop starts things are scaled and unscaled afterwards. The netcdf output of course still countains unscaling too. But I hope that makes things clearer. That way the user shouldn't come in contact with scaled prognostic variables and we shield this property better in the dynamical core.

test/run_speedy_with_output.jl
milankl: done

src/initial_conditions.jl
milankl: Added a line to the docstring for that.

milankl:
> With these changes, there is still no way to start it from an arbitrary field, right? Maybe we could add a `StartFromProgn` so that you can hand over an instance of `PrognosticVariables`. Then `StartFromFile` can also use this.

You can still do
```julia
p,d,m = initialize_speedy(...)
# do something with p or obtain another p from somewhere, then
run_speedy!(p,d,m)
```

but you are asking more for something like `p,d,m = initialize(p_ini)` right? The diagnostic variables could be reconstructed from the prognostic variables, but at the moment the model couldn't (because it contains parameters that aren't stored in the prognostic variables. So at the moment we can construct
- Parameters <-> ModelSetup -> PrognosticVariables, DiagnosticVariables
- technically, one can also do PrognosticVariables -> DiagnosticVariables but currently not implemented or needed
but other directions are not possible (as information is missing).

PR #234, opened 2023-01-21
Rename to OctaHEALPixGrid

milankl:
@hottad I always found HEALPix4 a bulky term and not really telling. So I thought about renaming it to OctaHEALPix, as it is in the end another grid that looks like an octahedron projected on a sphere. This might also help to communicate that the OctaHEALPix grid is a hybrid between the other octahedral grids and the HEALPix grid. What do you think?


hottad:
Hi @milankl ! I totally agree and I like the new name OctaHEALPix. It's more descriptive and much less confusing. Thank you!

PR #235, opened 2023-01-24
started grid documentation

milankl:


milankl:
@serjaaa I've documented the grids, including the OctaHEALPix grid we spoke about. I believe to get the grid cell which includes some coordinates $\phi, \theta$ you'd need to invert the boundary equation for $k,k'$ and then obtain the ring index $j$ and the in-ring index $i$ from it. I haven't thought that through but I can probably implement that fairly easily given that we already have loads of functionality on HEALPix/OctaHEALPix grids implemented, but you'd need to find a way to interface to Julia if that was in your interest!

milankl:
For some reason the documentation is currently not buildin, check here in the mean time

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/main/docs/src/grids.md

serjaaa:
Thanks a lot!
enrico

Il giorno gio 26 gen 2023 alle ore 19:37 Milan Klöwer <
***@***.***> ha scritto:

PR #239, opened 2023-01-27
Galewsky 2004 initial conditions for shallow water

milankl:
Based on [Galewsky, Scott, Polvani, 2004](https://a.tellusjournals.se/article/10.340/tellusa.v56i5.14436/), with the new initial conditions `NorthMitlatitudeJet <: InitialConditions` we should get something similar to their solutions

image

Not everything is exactly the same in our setup vs theirs so qualitative comparison should be good enough for now.

### DIFF
--- a/src/SpeedyWeather.jl
+++ b/src/SpeedyWeather.jl
@@ -67,7 +67,7 @@ module SpeedyWeather
     # EXPORT INITIAL CONDITIONS
     export  StartFromFile,
             StartFromRest,
-            ZonalWind,
+            NorthMidlatitudeJet,
             StartWithVorticity
 
     # EXPORT INTERPOLATION FOR GRIDS
--- a/src/default_parameters.jl
+++ b/src/default_parameters.jl
@@ -119,7 +119,7 @@ Base.@kwdef struct Parameters{Model<:ModelSetup} <: AbstractParameters{Model}
 
     # INITIAL CONDITIONS
     seed::Int = abs(rand(Int))          # random seed for the global random number generator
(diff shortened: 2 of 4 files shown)


milankl:
FullGaussianGrid at T255 with a 218s time step (7x as long as theirs) and despite the lower resolution their solutions looks so much noisier, but maybe that's just the visualisation

image

and the same with the OctaHEALPixGrid @hottad, looking good!

image

milankl:
And sanity check, T85 (time step 656s, ~11min) on the OctaHEALPixGrid but using Float32

image

hottad:
Great work, Milan! I am surprised (and pleased) to see the OctaHEALPix result so similar to the canonical FullGaussian despite Legendre transform not being exact!
Maybe hte use of cubic truncation is one factor making your result looking less jagged than Galewsky et al?

milankl:
I don't think so as I'm still using quadratic truncation for the FullGaussianGrid vs cubic for the HEALPixGrids. Maybe it's the implicit diffusion we're using? Galewsky talk about no diffusion vs explicit diffusion for which the solutions become way smoother to the point where they actually kill part of the turbulence. Anyway, just created an overview, Float64 for now, but I want to rerun the HEALPix simulations with Float32 just to make the point that precision-wise it doesn't matter what you use.

image

hottad:
@milankl thank you! This is great! And apologies for my very late response. It is really nice that OctaHEALPixGrid is as good as HEALPixGrid despite using much less grid points for the same truncation wavenumber. Amazing!

PR #248, opened 2023-02-09
CompatHelper: bump compat for CUDA to 4, (keep existing compat)

milankl:
@maximilian-gelbrecht any reason not to bump compatibility with CUDA.jl to v4?


maximilian-gelbrecht:
No, we should do it, sticking to v3 restricts some other packages in the long run. I don't know why CompatHelper suggested this in the first place

milankl:
I always keeps the existing compatibility. If we now we shouldn't support CUDA.jl 3 then we can still remove it.

PR #258, opened 2023-03-20
Utility function load_trajectory added

maximilian-gelbrecht:
This adds a utility function to quickly load a saved trajectory. I use this when working with Speedy. If you don't find it useful, I can keep it to myself. 

This also makes potentially fully testing the netCDF output in the tests a bit easier. But I didn't add all of those tests now.

### DIFF
--- a/src/output/output.jl
+++ b/src/output/output.jl
@@ -409,4 +409,7 @@ get_full_output_file_path(P::Parameters) = joinpath(P.output_path, string("run-"
 
 Loads a `var_name` trajectory of the model `M` that has been saved in a netCDF file during the time stepping.
 """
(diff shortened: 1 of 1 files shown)


milankl:
Sounds great! But why is it called a trajectory and not just a path?

maximilian-gelbrecht:
What exactly do you mean? 

There's one function that returns the full path of the netCDF file to load, and one function that actually loads a trajectory / variable from that file.

milankl:
Sorry, now I understand you want to load all time steps of a variable that was output into netcdf! Yes, "all time steps from one variable" = "trajectory" = sounds good to me!

milankl (approved):

PR #260, opened 2023-03-20
Add clean up of run folders after tests

maximilian-gelbrecht:
When testing SpeedyWeather locally, testing it a second time will fail as the named run folders already exist. This PR adds a clean up after testing that deletes all run folders

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -44,3 +44,4 @@ include("netcdf_output.jl")
  run_folders = filter(x->startswith(x, "run-"), readdir("."))
  for folder in run_folders
      rm(folder, recursive=true)
+ end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -39,3 +39,8 @@ include("run_speedy_with_output.jl")
 
 # OUTPUT 
 include("netcdf_output.jl")
+
+ # CLEAN UP 
+ run_folders = filter(x->startswith(x, "run-"), readdir("."))
+ for folder in run_folders
+     rm(folder, recursive=true)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -43,5 +43,9 @@ include("netcdf_output.jl")
  # CLEAN UP 
  run_folders = filter(x->startswith(x, "run-"), readdir("."))
  for folder in run_folders
-     rm(folder, recursive=true)
+    if isdir(folder)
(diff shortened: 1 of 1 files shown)


maximilian-gelbrecht:
Then you'd have to add an `output_path` kwarg to every test with output. 

I just added a check to see if it is really is a folder containing the output files

white-alistair:
You could avoid doing the cleanup manually by using the [mktempdir](https://docs.julialang.org/en/v1/base/file/#Base.Filesystem.mktempdir-Tuple{AbstractString}) method for temporary folders created by tests? There are similar methods for files as well.

maximilian-gelbrecht:
I'll add an output folder for the tests tomorrow

white-alistair:
> No, this would make things unnecessarily complicated

Unnecessarily complicated how? In my opinion it's the cleanest solution by far to this kind of problem, and it's a common unit testing pattern (hence why the method exists in the first place, I assume).

> and wouldn't work if you test twice (or more) in the same process which is when I typically encounter the error.

Why, exactly, wouldn't it work in that case? I'm guessing you're running the tests multiple times in a single Julia instance during development, which is fair enough. In that case it'll create a temporary output folder with a unique name every time you run the tests, ensuring that successive test runs, in the same instance or otherwise, will be independent. No manual cleanup required.

maximilian-gelbrecht:
> > No, this would make things unnecessarily complicated
> 
> Unnecessarily complicated how? In my opinion it's the cleanest solution by far to this kind of problem, and it's a common unit testing pattern (hence why the method exists in the first place, I assume).

You'd need an extra kwarg and extra checks during setting up the output just for the test case. I find that unnecessarily complicated. You normally want to keep these folders, only when running the unit test you want to delete them afterwards. 

> 
> > and wouldn't work if you test twice (or more) in the same process which is when I typically encounter the error.
> 
> Why, exactly, wouldn't it work in that case? I'm guessing you're running the tests multiple times in a single Julia instance during development, which is fair enough. In that case it'll create a temporary output folder with a unique name every time you run the tests, ensuring that successive test runs, in the same instance or otherwise, will be independent. No manual cleanup required.

All folders created with `mktempdir` are only deleted after the Julia process ends, but you might want to run the test again in the same Julia process. I guess one could add a random number to those runs that do have a specific output name (which is only 3 out of ~10), but still prefer to just clean up after the tests are done.

PR #261, opened 2023-03-20
Use Temporary Files for Test Output

white-alistair:
@maximilian-gelbrecht since we seem to be talking past each other on the other PR, I thought it would be easier to just show you what I mean here.

In my opinion this is the most correct, clean, and safe way of dealing with files and folders that are created by tests.

Whether you're running the tests locally using `] test` or `include("path_to_test")`, you can run the tests repeatedly without any issues, ensuring that successive test runs are independent.

It also avoids the dangerous use of `rm` on all files matching a certain pattern.

### DIFF
--- a/test/netcdf_output.jl
+++ b/test/netcdf_output.jl
@@ -12,7 +12,7 @@
             time_axis 
         end 
 
-        tmp_output_path = mktempdir(pwd(), prefix = "tmp_test_output_")  # Cleaned up when the process exits
+        tmp_output_path = mktempdir(pwd(), prefix = "tmp_testruns_")  # Cleaned up when the process exits
         
         p, d, m = initialize_speedy(Float32, output=true, output_path=tmp_output_path, n_days=1, output_dt=0, run_id="dense-output-test")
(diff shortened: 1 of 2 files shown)


white-alistair:
@milankl

milankl:
I like this, would probably call that folder `tmp_testruns` to stick to the terminology that a *run* is one simulation which can have several output files

white-alistair:
Before merging I would wait for @maximilian-gelbrecht to confirm this actually solves his problem.

white-alistair:
> One annoying thing though is that `mktempdir` does not work when you kill the process. And in VSCode this is how I end Julia processes usually. Which at least for me, makes this not very useful, and I would still manually remove those folders in the end after all tests.

Just `ctrl + d` the Julia process.

maximilian-gelbrecht:
I really don't get what's wrong with just deleting these temp folders in one line at the end of the test, and then it works in all cases. When the folders are created during the test, then it is definitely save to delete them. `mktempdir` does nothing else when the process is closed.

Issue #255, opened 2023-02-22
Renaming in Prognostic/DiagnosticVariables


milankl:
The following naming wasn't very clever of me, and I'd love this to be more consistent across Prognostic/Diagnostic variables
```
julia> p,d,m = initialize_speedy(PrimitiveDryCore);

julia> typeof(p.layers[1])
Main.SpeedyWeather.PrognosticVariablesLeapfrog{Float64}

julia> typeof(d.layers[1])
Main.SpeedyWeather.DiagnosticVariablesLayer{Float64, FullGaussianGrid{Float64}}
```

PR #262, opened 2023-03-21
copy! for PrognosticVariables

maximilian-gelbrecht:
I took the copy operation for PrognosticVariables that was already implemented when restarting from the restart file and made it a separate function. With that it is really easy to restart from Prognostic Variables, e.g. when spinning up the model 

You could do the `set_var` before, but with `copy!` you do all variabels at once. 

```Julia 

# spin up 
 p1, d1, m1 = initialize_speedy(Float32, ShallowWater)
 run_speedy!(p1, d1, m1)
 
# proper run (e.g. with output on) 
p2, d2, m2 = initialize_speedy(Float32, ShallowWater)
copy!(p2, p1)
```

### DIFF
--- a/src/dynamics/initial_conditions.jl
+++ b/src/dynamics/initial_conditions.jl
@@ -264,18 +264,7 @@ function initial_conditions!(   ::Type{StartFromFile},
     version = restart_file["version"]   # currently unused, TODO check for compat with version
     time = restart_file["time"]         # currently unused
 
-    var_names = propertynames(progn_old.layers[1].leapfrog[1])
-
-    for var_name in var_names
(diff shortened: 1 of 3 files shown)


milankl:
Yep, I like it! Note that I want to address #255 in the next days to be more consistent and allow something like sea surface temperatures to be a prognostic variable on the surface for a slab-ocean model. But that'll only affect lower level functions so happy to merge this now

milankl (approved):

maximilian-gelbrecht:
Okay, tag me in the PR that I don't miss this, then I am happy to help if there are any problems

Issue #20, opened 2021-12-14
Parallelism in the vertical


milankl:
One easy way to parallelise speedy might be to distribute the calculation of the spectral transform across `n` workers in the vertical, using [SharedArrays](https://docs.julialang.org/en/v1/stdlib/SharedArrays/) ([documentation here](https://docs.julialang.org/en/v1/manual/distributed-computing/#man-shared-arrays)) from Julia's standard library. While this limits us to `n`x speedups from parallelisation, which might be fine as SpeedyWeather.jl will probably run on small clusters only anyway. Such that for T30 and `n=8` levels we can (hopefully) efficiently run on 8 cores, but for `n=48,64` levels we could get significant speedups for higher resolution versions of SpeedyWeather.jl ([T100-T500](https://confluence.ecmwf.int/display/OIFS/4.3+OpenIFS%3A+Horizontal+Resolution+and+Configurations)). Given the shared memory of this approach, we'll be limited by the [48 cores on A64FX](https://www.fujitsu.com/downloads/SUPER/a64fx/a64fx_datasheet_en.pdf), but that might be absolutely sufficient for now.

milankl:
As outlined here, written with `@eval` metaprogramming style one could similar collect all functions that need a vertical loop and think about adding an `@distributed` here once SharedArrays are set up
https://github.com/milankl/SpeedyWeather.jl/blob/f21b69eae24f9be0b0667ebe381671d994b9b14f/src/distributed_vertical.jl#L1-L15

milankl:
With #117 the `@eval`-based looping on the level of individiual functions has been removed. We now have, e.g. for the barotropic model a single place in `timestep!` where the looping over the vertical is happening, hence we moved that loop as far up as possible
https://github.com/milankl/SpeedyWeather.jl/blob/4a3ad4a7659bd1ae8437b15617781420b396678e/src/time_integration.jl#L246-L252
this is where any `@distributed`-like parallelism should be applied, as the evaluation within the loop is conceptually independent across layers.

milankl:
One thread
```julia
julia> run_speedy(Float32,PrimitiveDryCore,trunc=127,nlev=26,n_days=1,orography=ZonalRidge)
475.47 days/day
```
vs 8 threads (4.1x)
```julia
julia> run_speedy(Float32,PrimitiveDryCore,trunc=127,nlev=26,n_days=1,orography=ZonalRidge)
5.50 years/day
```
[...]

PR #264, opened 2023-03-22
Multi-threading added

milankl:
A first, easy attempt at #20 using threads for the parallelisation across the vertical.


milankl:
@white-alistair note: this still doesn't scale well, I only get 4-6x even for 16 threads at T127 and 31 levels. But good enough for now to do some experiments at reasonable resolution.

PR #267, opened 2023-03-23
CompatHelper: bump compat for KernelAbstractions to 0.9, (keep existi…

milankl:
…ng compat)


vchuravy:
Sadly not as easy. KA 0.9 will require a bit of a rewrite. We  dropped the event systems (which simplifies life a fair bit)

milankl:
superceded by #357

PR #269, opened 2023-03-27
Boundary layer scheme from Held and Suarez

milankl:
This introduces the simple boundary layer scheme from [Held and Suarez](https://journals.ametsoc.org/view/journals/bams/75/10/1520-0477_1994_075_1825_apftio_2_0_co_2.xml?tab_body=pdf) 

image

and also introduces `abstract type BoundaryLayer end` with

```julia
struct NoBoundaryLayer end
struct LinearDrag ... end
```
such that `run_speedy(boundary_layer=NoBoundaryLayer())` switches it off but

```julia
run_speedy(boundary_layer=LinearDrag(kwargs...))
```
uses that scheme and in general one can pass on any struct `<: BoundaryLayer` when also defining `boundary_layer!(column,::MyBoundaryLayer,model)` to introduce a new scheme.

@white-alistair this is the first time the `parameterization_tendencies!` loop actually runs through and executes things, but other schemes are currently commented. I want to test them one by one.

### DIFF
--- a/src/SpeedyWeather.jl
+++ b/src/SpeedyWeather.jl
@@ -1,143 +1,151 @@
 module SpeedyWeather
 
-    # STRUCTURE
-    import Parameters: @unpack
+# STRUCTURE
+import Parameters: @unpack
 
-    # NUMERICS
-    import Random
-    import FastGaussQuadrature
-    import LinearAlgebra: LinearAlgebra, Diagonal
+# NUMERICS
+import Random
+import FastGaussQuadrature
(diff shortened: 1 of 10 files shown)


white-alistair:
@milankl testing the parameterisations one by one also seems like the right time to refactor them. I think @dmey and I took quite different yet complementary approaches, which I like, and it would be good to keep the best of both.

milankl:
Yes, please let me know if you have any opinion on "best of both".

PR #270, opened 2023-03-28
Held Suarez temperature forcing

milankl:
In addition to #269 also implements the temperature relaxation

### DIFF
--- a/src/RingGrids/RingGrids.jl
+++ b/src/RingGrids/RingGrids.jl
@@ -36,6 +36,7 @@ module RingGrids
             each_index_in_ring!,
             eachgridpoint,
             eachring,
+            whichring,
             get_nlons,
             get_nlon_max,
             get_quadrature_weights,
--- a/src/RingGrids/grids_general.jl
+++ b/src/RingGrids/grids_general.jl
@@ -199,6 +199,20 @@ function eachring(grid1::Grid,grids::Grid...) where {Grid<:AbstractGrid}
 end
 
 """
+    j = whichring(ij,rings)
+
+Obtain ring index j from gridpoint ij and Vector{UnitRange} describing rind indices
+as obtained from eachring(::Grid)"""
+function whichring(ij::Integer,rings::Vector{<:UnitRange})
+    @boundscheck 0 < ij <= rings[end][end] || throw(BoundsError)
+    j = 1
+    @inbounds while ij > rings[j][end]
+        j += 1
+    end
+    return j
+end
+
+"""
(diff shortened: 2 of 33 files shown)


milankl:
The last commits reduce drastically the time spend in the parameterizations, which was (boundary layer drag only) about 2x slower than for `physics=false`

image

milankl:
@maximilian-gelbrecht the last commits remove the idea to use abstract types in default_parameters.jl to dispatch initial conditions, orography, etc and uses concrete structs instead so that they directly can hold associated parameters. E.g.
```julia
run_speedy(initial_conditions=ZonalWind(u₀=25))
run_speedy(orography=ZonalRidge())
run_speedy(boundary_layer=LinearDrag(drag_time=12))
```
meaning that `orography=ZonalRidge` will now throw an error because it passes on the type and not an instance of it. Adding the `()` will, however, initialize with default values.

milankl:
T63, 8 levels, surface temperature with Held Suarez forcing and full orography from the last commit 

https://user-images.githubusercontent.com/25530332/228679180-62e1b4a3-d4d5-4826-ad88-8b0a478ca568.mp4

maximilian-gelbrecht:
> @maximilian-gelbrecht the last commits remove the idea to use abstract types in default_parameters.jl to dispatch initial conditions, orography, etc and uses concrete structs instead so that they directly can hold associated parameters. E.g.
> 
> ```julia
> run_speedy(initial_conditions=ZonalWind(u₀=25))
> run_speedy(orography=ZonalRidge())
> run_speedy(boundary_layer=LinearDrag(drag_time=12))
> ```
> 
> meaning that `orography=ZonalRidge` will now throw an error because it passes on the type and not an instance of it. Adding the `()` will, however, initialize with default values.

Nice! I was already thinking about suggesting that change.

milankl:
@white-alistair I had to adapt the thermodynamics tests as we now centrally calculate the pressure (and log pressure) on every level in `get_column!` so that it doesn't have to be recalculated for every parameterization. Meaning column.pres is now a vector of nlev+1 with the last element being the surface

Will also eventually overhaul the vertical interpolation as recalculating the interpolation weights for every grid cell can be avoided. That actually sounds like it should be part of `Geometry` (in contrast to a lot of other things in there, that don't seem like "Geometry"...)

milankl:
And just to add: I believe at some point it makes sense to implement hybrid $\sigma$-pressure coordinates, because it seems to be widely used standard. It's not suuuper difficult to implement for the dynamics, but it obviously interferes with the physics so want to put this off for now, but it's on my todo wishlist.

Issue #272, opened 2023-03-30
Not thread-safe / bit reproducible ?!


milankl:
When executing the current version in `mk/heldsuarez` (#270) with several threads, results are not bit reproducible, although we don't have any multi-threaded reduction operation (meaning bit reproducibility should be achievable)

![image](https://user-images.githubusercontent.com/25530332/228887936-4fcd8b99-7822-4355-a639-d964d3c82ba3.png)

using only one thread it I get exactly the same answer though.

@christophernhill you were right, SpeedyWeather is currently not bit reproducible on several threads...
@white-alistair As you worked with Julia's multi-threading before, any idea how to address this best other than serialising parallel loops one by one and checking when the results become bit reproducible?

milankl:
So you removed the `@floop`s in `dynamics_tendencies!`? And then you got it reproducible again? Okay that's good, that excludes (assuming you did that from the current state of `mk/heldsuarez`)
- time stepping
- physics
- semi-implicit corrections

white-alistair:
That's exactly what I did, except on `main`. Will do it again on `mk/heldsuarez` and report back...

white-alistair:
Okay yep, this is on `mk/heldsuarez`, I think we can narrow it down to `dynamics_tendencies!`.

![image](https://user-images.githubusercontent.com/59571934/228923602-65110881-4140-4c30-ad37-c3aa98230706.png)

milankl:
Okay I think I know what's happening: In here
```julia
    @floop for layer in diagn.layers
        vertical_velocity!(layer,surface,model)     # calculate σ̇ for the vertical mass flux M = pₛσ̇
        ...
        vertical_advection!(layer,diagn,model)      # use σ̇ for the vertical advection of u,v,T,q
        ...
    end
```
the advection uses the velocity from the layer above, which, if the thread above didn't finish yet, may use the vertical velocity from the previous time step. It's not a dramatic race condition because the vertical velocity is zero to start with anyway. So solution would be to separate this out, to guarantee that all threads are done computing the vertical velocity before the the advection is multi-threaded again.

milankl:
Ouh yeah! 16 threads at the top, 1 thread bottom, that looks very good

![image](https://user-images.githubusercontent.com/25530332/228960371-a84c0e38-1bc8-45ef-8031-9d43361a29ac.png)

Is indeed bit reproducible, single threaded
```julia
[...]

PR #274, opened 2023-03-30
Make default RNG seed deterministic

white-alistair:
This seems sensible...?


white-alistair:
Spotted this while investigating #272

milankl (approved):

PR #280, opened 2023-04-04
TemperatureRelaxation, BoundaryLayer type stable

milankl:
The evaluation of the parameterizations wasn't type stable. This is addressed now. Problem solved with

- Any parameters in `Parameters` should have concrete types. E.g. `a::Float64 = 1` or `b::SomeParametricType{Float64}`. Use Float64 as default here.
- Precalculated values/arrays needed should go in `DynamicsConstants`/`ParameterizationConstants` where they are converted to the number format `NF`.
- Alternatively values from `Parameters` can be used directly in the time integration by converting them on the fly to `NF` which in many situations might be more convenient than introducing a new field in `DynamicsConstants`/`ParameterizationConstants`

@white-alistair 

For the record, with Held-Suarez forcing and default resolutoin (T31, 8 levels) with (maybe experimentally) an immutable `ColumnVariables` struct

```julia
julia> @btime SpeedyWeather.parameterization_tendencies!($d,$m);
  2.375 ms (1 allocation: 896 bytes)
```

the remaining allocation is from `eachring(::Grid)` and intentional. Timing was

```julia
julia> @btime SpeedyWeather.parameterization_tendencies!($d,$t0,$m);
  4.813 ms (4639 allocations: 606.05 KiB)
```

So 2x faster 🎉

### DIFF
--- a/src/RingGrids/grids_general.jl
+++ b/src/RingGrids/grids_general.jl
@@ -203,7 +203,7 @@ end
 
 Obtain ring index j from gridpoint ij and Vector{UnitRange} describing rind indices
 as obtained from eachring(::Grid)"""
-function whichring(ij::Integer,rings::Vector{<:UnitRange})
+function whichring(ij::Integer,rings::Vector{UnitRange{Int}})
     @boundscheck 0 < ij <= rings[end][end] || throw(BoundsError)
     j = 1
     @inbounds while ij > rings[j][end]
--- a/src/abstract_types.jl
+++ b/src/abstract_types.jl
@@ -29,8 +29,8 @@ abstract type AbstractBoundaries{NF} end
 abstract type AbstractColumnVariables{NF} end
 
 # PARAMETERIZATIONS
-abstract type BoundaryLayer end
-abstract type TemperatureRelaxation end
+abstract type BoundaryLayer{NF} end
+abstract type TemperatureRelaxation{NF} end
 
 # INPUT/OUTPUT
 abstract type AbstractFeedback end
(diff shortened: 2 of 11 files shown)


milankl:
This ☝🏼 gets mutability for scalars by doing
```julia
Base.@kwdef struct ColumnVariables
     a::Base.RefValue{Int} = Ref(0)
end
```
Such that accessing becomes
```julia
c = ColumnVariables()
c.a[]
c.a[] = 2
```
which is not really handy

milankl:
Alternatively, the next commit does
```julia
Base.@kwdef mutable struct ColumnVariables{NF<:AbstractFloat} <: AbstractColumnVariables{NF}
    const nlev::Int = 0                     # number of vertical levels
    jring::Int = 0                          # latitude ring the column is on
    const u::Vector{NF} = zeros(NF,nlev)      # zonal velocity
     ...
end
```
So using a `mutable struct` but declaring some fields as `const`, that seems to do the job as well and avoids the ugly `[]` accessing through the pointers and is as fast
```julia
julia> @btime SpeedyWeather.parameterization_tendencies!($d,$m);
  2.248 ms (1 allocation: 896 bytes)
```

milankl:
@maximilian-gelbrecht this PR would add the annotation of `const` inside `mutable struct` which is only available from v1.8, meaning we'd drop compat with earlier Julia versions. Is that an issue from your side?

maximilian-gelbrecht:
> @maximilian-gelbrecht this PR would add the annotation of `const` inside `mutable struct` which is only available from v1.8, meaning we'd drop compat with earlier Julia versions. Is that an issue from your side?

I am fine with that. 1.8 is now released since a while. With a project like this, we can't and shouldn't guarantee backwards compatibility. As long as we wait long enough that everybody had enough time to upgrade (also admins on HPC environments).

milankl:
Sounds like a good philosophy. Especially in Julia I find there isn't a massive lag between a new minor release and packages catching up, with python however, you sometimes need to wait for a year...!

PR #287, opened 2023-04-06
restructure PrognosticVariables

### LINKED ISSUE #255: Renaming in Prognostic/DiagnosticVariables

milankl:
fixes #255 

@maximilian-gelbrecht, currently just a draft as I need to rename a lot of things, but my suggestion is to do

```julia
julia> p = zeros(PrognosticVariables{Float32},31,31,4);

julia> p.
layers     lmax       mmax       n_storage  nlev       surface
```
so `layers` is one field (`Vector{PrognosticLayerTimesteps}`) and `surface` another (was `pres`). This mean if we ever need to add another prognostic variable at the surface (e.g. a slab ocean model) this is more straight forward to implement. It also better matches how `DiagnosticVariables` is defined.

Then inside the `layers`/`surface` we have `timesteps` (was `leapfrog`, just because we may at some point want another time stepping scheme, so I wanted it to be more general)
```julia
julia> p.layers[1].timesteps[1].
div    humid  temp   vor
julia> p.surface.timesteps[1].pres    # pres currently only field here, but e.g. sst could be added
```
`timesteps[1]` is then the past time step and `timesteps[2]` is the current in the leapfrog-logic.

If you agree, I'll try to adapt everything else to this logic.


milankl:
`n_storage` also sounds meaningless. I borrowed this work from the idea that an implementation of a time stepping scheme uses needs to store `n` steps, which doesn't have to be the same as the order or the number of stages. 

I believe we should just rename this to `n_timesteps`, given it's a field of the prognostic variables it should be clear that it doesn't refer to the overall time steps, or just `n_steps` maybe that's even more concise

maximilian-gelbrecht:
I agree with the overall changes. 

I'd also name it `n_steps` or even `n_solversteps`, otherwise I would find it a bit confusing, as it is just about the time steps that the solver needs, not those that are saved (for later use).

Issue #289, opened 2023-04-06
Greenland whole column instability at T255


milankl:
With 
```julia
run_speedy(Float32,PrimitiveDryCore,verbose=false,initial_conditions=StartFromFile(),restart_id=23,trunc=255,n_days=20,output=true,output_dt=1,Grid=OctahedralClenshawGrid,nlev=31)
```
There's an instability after about 12 days that spans the whole column, level 15

![image](https://user-images.githubusercontent.com/25530332/230497918-51a5ad93-1398-4161-8580-520ae522b605.png)

It's barely visible in surface or stratospheric temperature, nor surface pressure. But also mid-tropospheric temperature

![image](https://user-images.githubusercontent.com/25530332/230498640-4cdf4634-8fb3-464f-91bd-8eccffb94e04.png)

and (particularly?) pronounced in vorticity a bit higher (level 10)

![image](https://user-images.githubusercontent.com/25530332/230499285-cea5d84c-8fd5-47db-ad07-69eafb6e55d4.png)

milankl:
Hypothesis: The instability is created because the extra Laplacian in the stratosphere reaches in this model setup to level k=9 but not below. So there's an abrupt change in the horizontal diffusion between k=9,10. Cutting through the instability at about lon=292˚E, sigma coordinates on y but upside down.

![image](https://user-images.githubusercontent.com/25530332/230531671-e63be0e1-b9bf-4d22-93c2-d687dbc9e9bf.png)

milankl:
Another one of those, this time at the coast of Antarctica

![image](https://user-images.githubusercontent.com/25530332/230971026-27b49c9a-aad4-4212-91a3-ed29acddecb4.png)

run with

```julia
run_speedy(Float32,PrimitiveDryCore,verbose=false,initial_conditions=StartFromFile(),restart_id=23,trunc=255,n_days=30,output=true,output_dt=1,output_vars=[:vor],Grid=OctahedralGaussianGrid,nlev=31)
```

after #290 got merged

milankl:
Either disabling the Laplacian with tapering from sigma 0,0.2 or setting its power to 2 (biharmonic) does seem to resolve the issue (level 20)

![image](https://user-images.githubusercontent.com/25530332/231001167-9872d061-968e-47d0-834f-baab9c5cc3e1.png)

run with

```julia
[...]

PR #298, opened 2023-04-10
New stratosphere diffusion defaults

milankl:
Because of #289 the additional diffusion in the stratosphere is changed to a biharmonic diffusion (power=2) and a slightly longer time scale (24hours at T31). Comparison

image

Via 
```julia
julia> p,d,m = initialize_speedy(PrimitiveDryCore,nlev=31,diffusion=HyperDiffusion(power_stratosphere=2,time_scale_stratosphere=24));

julia> semilogy(0:32,1 .- m.horizontal_diffusion.∇²ⁿ_implicit[1],label=("power=2,time_scale=24"))
```

So the new default is red, it was blue, which caused problems at the transition into the stratosphere (defined by tapering_sigma here). No additional diffusion would be green (set tapering to top at the atmosphere, so effectively disable). Judging from #289 blue is unstable at T255 31 levels, but orange, red green are stable.

### DIFF
--- a/src/dynamics/define_diffusion.jl
+++ b/src/dynamics/define_diffusion.jl
@@ -4,8 +4,8 @@ Base.@kwdef struct HyperDiffusion <: DiffusionParameters
     time_scale::Float64 = 2.5                   # Diffusion time scale [hrs] for temp, vor, div
     
     # additional diffusion in stratosphere
-    power_stratosphere::Int = 1                 # additional ∇² for stratosphere
-    time_scale_stratosphere::Float64 = 12       # associated time scale
+    power_stratosphere::Int = 2                 # additional ∇² for stratosphere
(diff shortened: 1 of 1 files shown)


milankl:
This was the transition from below tapering_sigma (orange) into the tapering zone in the stratosphere (blue lines)

image

I guess that transition was just too hard.

milankl:
New default looks like

image

Definitely much smoother, but question then is whether we could just decrease the power from 4 in fractions to 1. But that's probably another project we can pick up at some point.

milankl:
New test incoming: decrease the tropospheric diffusion power in the tapering zone to the stratospheric, but keep the time scale constant (so that higher wavenumbers are always diffused the same way)

image

(black) transition from power=4 to power=1 at the top
(blue dashed) transition from power=4 to power=2 at the top
(orange) default power=4 in the troposphere

PR #300, opened 2023-04-12
Adds literated example in Docs

### LINKED ISSUE #297: Literated example with the animation that is currently in `README`?
Would that be a good idea?

Or the docs can include a simpler/less computationally demanding example?

navidcy:
Resolves #297 

https://speedyweather.github.io/SpeedyWeather.jl/previews/PR300/literated/basic_example/


navidcy:
This is a start. :)

milankl:
Perfect! I'll can add more examples here. Question: Is there a way that the colours in the unicode heatmap also appears in the docs or shall we include the result as images for simplicity?

navidcy:
> Perfect! I'll can add more examples here. Question: Is there a way that the colours in the unicode heatmap also appears in the docs or shall we include the result as images for simplicity?

I’ve never used UnicodePlots so I don’t know. Probably there will be a way.

navidcy:
Well, the "example" I added is a placeholder. Replace it with something more meaningful?

Also, we can use CairoMakie for plotting in the literated examples?

milankl:
With #327 I'm restructuring the model completely to make it more Oceananigans-like an example would then be. Define the grid
```julia
julia> using SpeedyWeather
julia> spectral_grid = SpectralGrid(Barotropic,trunc=85,Grid=OctaHEALPixGrid)
SpectralGrid{Barotropic}(
  Spectral:   T85 LowerTriangularMatrix{Complex{Float32}}, radius = 6.371e6 m
  Grid:       16384-element, 127-ring OctaHEALPixGrid{Float32} (quadratic)
  Resolution: 176km (average)
  Vertical:   1-level NoVerticalCoordinates)
```
define planet, forcing, initial conditions
```julia
julia> planet = SpeedyWeather.Earth(rotation=0)
SpeedyWeather.Earth(
  rotation::Float64 = 0.0
  gravity::Float64 = 9.81
  daily_cycle::Bool = true
  length_of_day::Float64 = 24.0
  seasonal_cycle::Bool = true
  length_of_year::Float64 = 365.25
  equinox::Dates.DateTime = 2000-03-20T00:00:00
  axial_tilt::Float64 = 23.4)

julia> forcing = SpeedyWeather.NoForcing(spectral_grid)
SpeedyWeather.NoForcing{Float32}()

julia> ic = StartWithVorticity(power_law=-3)
StartWithVorticity(-3.0, 1.0e-5)
```
create a model and initialize
```julia
julia> model = BarotropicModel(;spectral_grid,planet,forcing,initial_conditions=ic)
```
and run!

image

navidcy:
#345 transforms code + output that is currently written as text in the docs + img underneath into live code that actually runs and produces the output, e.g., the code+plots here:
https://clima.github.io/OceananigansDocumentation/stable/model_setup/grids/

This PR adds an example script that then, via Literate.jl, converts to a website with text/math/code/figures, like:
https://clima.github.io/OceananigansDocumentation/stable/generated/internal_wave/

The end product is similar. But the latter also gives a `.jl` file in the repository that people could just run as is, e.g.,
```
julia --project example.jl
```

The two methods can complement each other. Not both are necessary but both could be a thing in the docs. Up to you which you think is more useful at the moment and/or which to prioritize.

milankl:
I think we went on to prioritize the current style to document examples as in the code, tempted to close this as I find this currently sufficient. Unless there's something in here that should really be added?

PR #307, opened 2023-04-16
Power spectrum analysis

### LINKED ISSUE #304: Power spectrum
We should include `power_spectrum(::LowerTriangularMatrix{Complex{NF}}) where NF` in `SpeedyTransforms`, following Justin Willmert

image

which could be written as
```julia
julia> function power_spectrum(alms::LowerTriangularMatrix{Complex{NF}}) where NF
           lmax,mmax = size(alms)
           spectrum = zeros(NF,mmax)
           for l in 0:mmax-1
               spectrum[l+1] = alms[l+1,1]^2
               for m in 1:l
                   spectrum[l+1] += 2*abs(alms[l+1,m+1])^2
               end
               spectrum[l+1] /= 2l+1
           end
           return spectrum
       end
```
but should be reordered and probably also contain the last degree for completeness (or `last_degree=false` kwarg)

milankl:
fixes #304

### DIFF
--- a/src/SpeedyTransforms/SpeedyTransforms.jl
+++ b/src/SpeedyTransforms/SpeedyTransforms.jl
@@ -1,47 +1,49 @@
 module SpeedyTransforms
 
-    import Parameters: @unpack
+import Parameters: @unpack
 
-    # NUMERICS
-    import AssociatedLegendrePolynomials as Legendre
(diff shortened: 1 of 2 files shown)


milankl:
Looks like for purple (top) to yellow (surface)

image

PR #308, opened 2023-04-17
HyperDiffusion and VerticalDiffusion

### LINKED ISSUE #306: Diffusion scaling
I believe we need a revision of how the (hyper) diffusion scales with resolution, for example at T85 the model runs stable for ~1 year but I believe that with `HyperDiffusion(time_scale=2.5)` it's at the edge of stability, e.g.

image

So
- reduce the default time scale (make it stronger) for the power=4 diffusion
- maybe introduce a weaker background diffusion of power=1,2
- check how the time scale should change with resolution following that diffusion coefficients ([m^2/s] for harmonic, [m^4/s] for biharmonic) usually reduce ~linearly (?) with resolution, e.g. Jablonowski and Williamson (2006) use

image

but it's 1/2x for T21 to T42 but ~1/10x for T42 to higher with biharmonic!

milankl:
fixes #306


milankl:
Last commit was necessary to avoid blowup at T170 when restarting from stable T85

milankl:
Last commits are an attempt at making a biharmonic diffusion the default, as Jablonowski and Williamson use it too. Make the power=4 idea just isn't suited for higher resolutions. Difference in the default settings look like

image

PR #316, opened 2023-05-09
Keepbits per variable

milankl:
New default is
```julia
default_keepbits() = (u=7,v=7,vor=5,div=5,temp=10,pres=12,humid=7)
```


milankl:
after, before, temperature at top model level 6hours after default model setup initialisation

image

PR #326, opened 2023-05-13
Add compat entry for FLoops

navidcy:
Just wondering: was FLoops added after v0.4? Because as far as I know Julia registry doesn't allow you to tag a release unless all dependencies have a compat entry.


milankl:
Yes, I've only added it a few weeks ago, thanks for spotting this!

milankl (approved):

PR #327, opened 2023-05-15
New SpectralGrid, Model, initialize!, simulation, run! structure

### LINKED ISSUE #320: Away from monolithic `run_speedy()`?
At the moment we have a monolithic interface
```julia
run_speedy(param1,param2,...)
```
which makes it difficult to pass on a struct in which array sizes depend on, e.g. the size of the grid. For example
```julia
Base.@kwdef struct MyParameterization{T<:AbstractFloat} <: ParameterizationClass{T}
    parameter1::Float64 = 1    # user should be able to change those at initialisation 
    parameter2::Float64 = 2

    n::Int = 0                 # size of arrays that may not be known at initialisation
    some_constant_array::Vector{T} = zeros(T,n)
end
```
because `MyParameterization.some_constant_array` is of length `n` which in general is only known *after* run_speedy is called. 

Alternatively, we could go the Oceananigans approach to (1) first define an object `geometry` that contains that information before (2) creating a model which is then (3) run
```julia
geometry = Geometry(OctahedralGaussianGrid,trunc=31,nlev=8,dealiasing=2)
model = PrimitiveDryCoreModel(geometry;param1,param2,param3,...)
run!(model)
```

milankl:
attempt to fix #320 

@glwagner you might like this pull request!


glwagner:
😂 

I'm not 100% sure but the talk here might be relevant for some background:

https://docs.google.com/presentation/d/1eOZm1TOWQFrRSrNJ7KYmsJR_x1yXQCgr62txRAY3JSE/edit?usp=sharing

with some supporting information in

https://github.com/glwagner/ImprovingScientificSoftware2021-HumanCentricAbstractions

navidcy:
I like where this PR is going!

(perhaps the PR title is a bit esoteric and not as descriptive tho...)

glwagner:
Right, as much as it massages my ego...

milankl:
@navidcy do you know why the documentation fails with
```
ERROR: LoadError: UndefVarError: @kwdef not defined
```
I probabaly removed all the `Base.` from `Base.@kwdef` we had before, but only the Documentation complains about this...

navidcy:
> @navidcy do you know why the documentation fails with
> 
> ```
> ERROR: LoadError: UndefVarError: @kwdef not defined
> ```
> 
> I probabaly removed all the `Base.` from `Base.@kwdef` we had before, but only the Documentation complains about this...

Was this the culprit or you still want me to have a look?

milankl:
Yeah, for some reason the documenter doesn't like `@kwdef` but requires `Base.@kwdef`, then I also realised that `@with_kw` has some features that `@kwdef` doesn't have (like you can do `@assert` within) but on the other hand `@with_kw` doesn't allow `const` declarations in mutable structs, so we have a mix of the two now 😆

PR #328, opened 2023-05-15
spectral_truncation/interpolation type stable

### LINKED ISSUE #323: Global shallow-water run
Hi, it is a great model here.  I've already make a default run `run_speedy()`.  Now I need to customize it.  Specifically, I need to:
- modify dynamical core to single-layer shallow-water model,
- change the resolution to a higher one,
- add forcing field (probably in lat/lon gridded nc file), and
- control the dissipation through regular Laplacian diffusion.

Is there any doc or notebook that I can follow?  I guess I don't need to change much because everything was set to the earth's atmosphere (which is great to me).

milankl:
fixes #323

### DIFF
--- a/src/SpeedyTransforms/spectral_truncation.jl
+++ b/src/SpeedyTransforms/spectral_truncation.jl
@@ -92,7 +92,7 @@ function spectral_truncation(   ::Type{NF},                     # number format 
     lmax,mmax = size(alms) .- 1     # 0-based degree l, order m of the spherical harmonics
     
     # interpolate to higher resolution if output larger than input
-    (ltrunc > lmax || mtrunc > mmax) && return spectral_interpolation(alms,ltrunc,mtrunc)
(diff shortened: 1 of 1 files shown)


milankl:
@miniufo can you test whether your issue is resolved using this branch?

miniufo:
How to use this branch?  Should I do
```julia
julia> ] add https://github.com/SpeedyWeather/SpeedyWeather.jl#main
```

once again?

milankl:
```julia
julia> ] add https://github.com/SpeedyWeather/SpeedyWeather.jl#mk/swm
```
`mk/swm` is the name of this branch

milankl:
it's not in main yet as I want to check whether you run into more problems 😆 Thanks for testing our code btw! it's so tricky to keep up to speed with all the different options and making sure it never breaks....

miniufo:
That works fine!  I also tried T511, it is OK so far, as I can see the progress bar going slowly.  It only use a single core of my laptop.  Can I specify more cores for that?

PR #330, opened 2023-05-16
1st order upwind scheme for vertical advection

milankl:
At the moment we have a 2nd order centred scheme for vertical advection, which is known to be very dispersive, e.g. 

image

this arises from the general form

image

for $\xi_{k+1/2} = 1/2*(\xi_k + \xi_{k+1})$, i.e. just average the values above and below, this is a centred scheme:

image

But it becomes upwind for (velocities are positive towards increasing k, i.e. top to bottom)

$$ \xi_{k+1/2} = \begin{cases} \xi_k & \text{for} & M_{k+1/2} > 0 \newline \xi_{k+1} &\text{else.} & \end{cases}$$

thanks to @valeriabarra for the suggestion

### DIFF
--- a/src/dynamics/tendencies_dynamics.jl
+++ b/src/dynamics/tendencies_dynamics.jl
@@ -296,10 +296,20 @@ function _vertical_advection!(  ξ_tend::Grid,           # tendency of quantity 
                                 ξ_below::Grid,          # quantity ξ at k+1
                                 Δσₖ::NF                 # layer thickness on σ levels
                                 ) where {NF<:AbstractFloat,Grid<:AbstractGrid{NF}}
-    Δσₖ2⁻¹ = -1/2Δσₖ                                    # precompute
+    Δσₖ⁻¹ = -1/Δσₖ                                      # precompute
 
     # += as the tendencies already contain the parameterizations
-    @. ξ_tend += Δσₖ2⁻¹ * (σ_tend_below*(ξ_below - ξ) + σ_tend_above*(ξ - ξ_above))
+    for ij in eachgridpoint(ξ_tend,σ_tend_above,σ_tend_below,ξ_above,ξ,ξ_below)
+        # 1st order upwind scheme
+        ξ_face_below = signbit(σ_tend_below[ij]) ? ξ_below[ij] : ξ[ij]  # upwind: if velocity < 0, pick ξ from k+1
+        ξ_face_above = signbit(σ_tend_above[ij]) ? ξ[ij] : ξ_above[ij]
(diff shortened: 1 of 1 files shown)


milankl:
First tests look good (note the reduced vorticity levels already after 10 days)

image

milankl:
The stability criterion for the vertical advection may also give us an idea of how many vertical levels to choose

image

e.g. after 10 days at T63 and 8 levels the max vertical velocity is O(100) note that this is in radius times sigma coordinates because of scaling

```julia
julia> [extrema(d.layers[k].dynamics_variables.σ_tend) for k in 1:8]
8-element Vector{Tuple{Float32, Float32}}:
 (-12.847849, 11.947435)
 (-31.923378, 27.286299)
 (-40.226105, 49.74807)
 (-65.96341, 42.520805)
 (-82.03398, 63.377575)
 (-83.67786, 75.66916)
 (-83.829384, 62.141808)
 (0.0, 0.0)
```

at that resolution the scaled time step is
```julia
julia> m.constants.Δt
0.0001412651f0
```
so for 32 levels CFL is ~0.5 which could become easily critical for larger vertical velocities
```julia
julia> 100*m.constants.Δt/(1/32)
0.45204833149909973
```
instead for 8 levels CFL is ~0.1 which I guess is much more in the safe zone
```julia
julia> 100*m.constants.Δt/(1/8)
0.11301208287477493
```

milankl:
Running at T127 8 levels, the vertical velocities go in the 200s so CFL is again ~0.1!

milankl:
T85 with 12 levels test also looks good, centred top, upwind bottom

milankl:
We can calcualte vertical cfl by
```julia 
amax,amin = extrema(σ̇)
a = max(abs(amax),abs(amin))
cfl = a*model.constants.Δt/Δσₖ
```

shoyer:
> At the moment we have a 2nd order centred scheme for vertical advection, which is known to be very dispersive

Not to say that upwind schemes are a bad idea, but I think this is kind of backwards:

- 1st order upwind is very dispersive
- 2nd order centered is unstable

There are also 2nd order upwind schemes that might be a good idea...

milankl:
By dispersive I mean it gets easily wiggles that one can counteract with vertical diffusion, but 1st order upwind schemes have that diffusion already baked in, but they don't produce the wiggles?

shoyer:
> By dispersive I mean it gets easily wiggles that one can counteract with vertical diffusion, but 1st order upwind schemes have that diffusion already baked in, but they don't produce the wiggles?

I was thinking of "dispersion" as a synonym for "diffusion." But yes, indeed you would potentially need diffusion to stabilize an non-upwind scheme.

milankl:
No no, dispersion as in waves of different wavelengths travelling at different speeds such that a shape which is composed of several waves deforms over time. Not as in the variance decreases over time 😉

PR #338, opened 2023-05-31
SpeedyCondensation implemented

milankl:
@white-alistair I've just overhauled the large scale condensation to fit our new structure, I found some bugs too I believe, e.g. you wrote
```julia
humid_tend[k] += -(humid[k] - humid_threshold) / humid_relax_time_lsc   # Formula 22
temp_tend[k] += -alhc / cₚ * min(humid_tend[k], humid_tend_max*pres)    # Formula 23
```
where the first line would accumulate in humid_tend which may already contain other parameterizations but then use that same value to get the temperature tendency, so I wrote this now as
```julia
humid_tend_k = -(humid[k] - humid_threshold) / time_scale                   # Formula 22
temp_tend[k] += -latent_heat * min(humid_tend_k, humid_tend_max[k]*pres[k])   # Formula 23

...
# only write into humid_tend now to allow humid_tend != 0 before this scheme is called
humid_tend[k] += humid_tend_k

### CODE REVIEW

**Comment by milankl** [comment]
I'm struggeling a bit to initialise humidity, because the Fortran SPEEDY way blows up immediately, I don't understand how this gives reasonable specific humidities of O(1e-3) kg/kg

https://github.com/samhatfield/speedy.f90/blob/0ebfdbfcd62efd46dd1802284908645be4b230be/source/prognostics.f90#L100-L117

If I follow this I end up with O(1e15) which is obviously way too much

**Comment by milankl** [comment]
So I've just turned on humidity, and it gets advected, the large scale precipitation isn't zero and it doesn't immediately blow up but looks funky

image

**Comment by milankl** [comment]
For comparison that's what specific humidity in CAMS/IFS looks like

image

**Comment by milankl** [comment]
Hence I suggest we just start with a coslat^2 lat profile

image

**Comment by milankl** [comment]
Saturation specific humidity seems to be off by ~100x, maybe there's one formula that assumes hPa but we provide Pa?
```julia
julia> simulation.diagnostic_variables.columns[1].temp
8-element Vector{Float32}:
 200.10944
 209.96857
 230.56596
 249.06987
 268.04233
 287.04587
 301.8028
 309.78912

julia> simulation.diagnostic_variables.columns[1].sat_humid
8-element Vector{Float32}:
 2.56574f-7
 3.438067f-7
 2.6195164f-6
 1.2258612f-5
 4.8984508f-5
 0.00014451817
 0.00028417647
 0.00038649476

julia> simulation.diagnostic_variables.columns[1].pres
9-element Vector{Float32}:
   3667.2708
  11911.124
  22187.152
  34931.117
  50454.676
  68294.25
  85883.234
  98965.9
 103820.12
```
this is just calling `get_thermodynamics!` with temp [K] and pres [Pa] but maybe not returning specific humidity [kg/kg] ??

milankl:
Jean-Michel just confirmed: Fortran SPEEDY uses g/kg for specific humidity. Probably the problem here! ☝🏼

milankl:
@white-alistair I forgot to switch on diffusion for humidity, that's what it looked so funky, so for the record and reminder to self: Every variable that's advected gets diffusion. Meaning all but surface pressure. Looks reasonable now: T127 at the surface after a couple of days starting from the initial conditions as above. 

image

Note that we have slightly negative values too, but that's a typical problem with spectral advection, and I wouldn't be worried about this for now. What I find quite cool, this is the top level that starts with zero humidity, but after 2 days the vertical advection behind mountains actually lifts humidity into those layers -- I guess at the least the sign of vertical velocity seems to be correct 😆 

image

milankl:
OMG THERE IS RAIN

image

I'm sure the units are still off, mm/dt is supposed to be mm since last output time step, so mm/6h here. The humidity tendency is $q_t = kg/kg/s$ to get precipitation $P$ is

$$ P = \sum_k \frac{p_s \Delta\sigma_k}{g} q_{t,k}$$

Which should then have units of kg/m^2/s, divided by the density of water 1000kg/m^3 gives m/s. x1000 for mm/s then accumulate over all time steps till output and multiply by the output time step to get mm/dt? Something like this.

milankl:
For comparison, that's total precipitation in IFS

image

milankl:
Multiplication is with time step dt and not output time step dt, now it's

image

so the units are more what they are supposed to be (note that at lower resolution than IFS you'd also expect less intense rainfall). Maybe precipitation is a bit too even, like there don't seem to be many regions where it's like definitely not raining, but that could also be a problem with initial conditions, the time scale for condensation, no strong heating forcing, etc the precipitation does sit on fronts though which is good:

image

image

milankl:
I'll merge this, it's good enough for now and we can tweak things later

PR #343, opened 2023-06-01
barotropic, shallowwater test cases

milankl:
Yay, I've documented two examples on how to run speedyweather, a 2D barotropic turbulence example and the shallow water Galewsky test case.


milankl:
Can one look at the built documentation before it's merged? I had some issues with relativ vs absolute paths in the past!

PR #344, opened 2023-06-01
absolute paths docs img

milankl:
@navidcy I never get relative paths working in the documentation, any insights?


navidcy:
I think because a relative url would be relative compared to where this file lives in the `build` directory of the docs, e.g., at
https://github.com/SpeedyWeather/SpeedyWeather.jl/tree/gh-pages/dev/how_to_run_speedy

Ideally what you'd like is that these plots are generated by the docs so that they are always representative of how the code works at any given time.

navidcy (approved):
lgtm! but a better solution is to have code here to generate the figures (even if the code is hidden).

PR #352, opened 2023-06-08
Fix typos in mk/docs

pitmonticone:


milankl:
Awesome! Some of these typos are so funny, I should really enable some kind of spell checker in VS Code 😆

PR #354, opened 2023-06-09
plot small grids without spaces

milankl:


milankl:
Small grids are just plotted smaller now

image

milankl:
Previously

image

PR #356, opened 2023-06-10
SpectralTransform for square LowerTriMats too

milankl:
So that both work
```julia
alms = randn(LowerTriangularMatrix{ComplexF64},32,32)
gridded(alms)

alms = randn(LowerTriangularMatrix{ComplexF64},33,32)
gridded(alms)
```
previously only the second (which is what's used inside the dynamical core) was working when externally used.


milankl:
@maximilian-gelbrecht this PR will change the `set_vorticity` etc functions slightly to adapt them to the needs of more flexibility with the LowerTriangularMatrix sizes

PR #358, opened 2023-06-16
Support KernelAbstractions 0.9

vchuravy:


vchuravy:
Didn't know that SpeedyWeather had a dependency on KA otherwise I would have opened a PR myself.

note that KA now has functionality to allocate memory, so you might be able to simplify some of your codes.

PR #361, opened 2023-06-20
CI tests on Julia 1.8, 1.9 + test on Julia nightly every week

### LINKED ISSUE #359: CI test on supported Julia versions?
Perhaps we change this

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/bdf02345292895570579d87d5de8d832eedbf7e7/.github/workflows/CI.yml#L17

to match

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/bdf02345292895570579d87d5de8d832eedbf7e7/docs/src/installation.md#L28-L31

milankl:
fixes #359

Is this the correct yaml syntax?

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -14,7 +14,10 @@ jobs:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
         with:
-          version: 1
+          version:
+            - '1.8'
+            - '1.9'
+            - 'nightly'
       - uses: actions/cache@v1
         env:
           cache-name: cache-artifacts
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -10,9 +10,7 @@ on:
 jobs:
   test:
     runs-on: ${{ matrix.os }}
-    continue-on-error: ${{ matrix.experimental }}
     strategy:
-      fail-fast: true
       matrix:
         version:
           - '1.8'
@@ -21,13 +19,8 @@ jobs:
           - ubuntu-latest
         arch:
           - x64
-        experimental: [false]
-        include:
-          - version: 'nightly'
-            experimental: true
-          
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: julia-actions/setup-julia@v1
         with:
           version: ${{ matrix.version }}
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -10,7 +10,9 @@ on:
 jobs:
   test:
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.bleadingedge }}
     strategy:
+      fail-fast: true
       matrix:
         version:
           - '1.8'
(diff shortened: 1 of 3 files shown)


.github/workflows/CI.yml
```diff
@@ -14,7 +14,10 @@ jobs:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
         with:
-          version: 1
+          version:
+            - '1.8'
+            - '1.9'
+            - 'nightly'
```
navidcy, line 20:
```suggestion
            - 1.8
            - 1.9
            - nightly
```

I think without quotes? ci didn't run before... let's try this way

milankl:
Nope, also not that way 😆

milankl:
Yes, I saw this in some packages, what does matrix.os mean?

navidcy:
I think there is a way to *not* mark the failing on nightly CI as failure. But I don't remember now on top of my head how to do it.

white-alistair:
> I think there is a way to _not_ mark the failing on nightly CI as failure. But I don't remember now on top of my head how to do it.

That sounds like the best of both worlds! Anything to avoid the dreaded :x:

navidcy:
hm... I can't figure it out.
Personally, I'm happy with either; nightly or no nightly!
1.8 and 1.9 was what I was thinking with #359

navidcy:
OK, let's take out nightly!

milankl:
nightly fails because of NetCDF.jl 
```julia
ERROR: LoadError: Failed to precompile NetCDF [30363a11-5582-574a-97bb-aa9a979735b9] to "/home/runner/.julia/compiled/v1.10/NetCDF/jl_mVW2pK".
...
ERROR: LoadError: InitError: could not load library "/home/runner/.julia/artifacts/0a7f440ba143b238cc3dd0def2956d34d76755f3/lib/libnetcdf.so"
libmbedtls.so.13: cannot open shared object file: No such file or directory
```

milankl:
Apparently Flux separates out the nightly CI to an independent github action, let's try what that does!

navidcy:
That’s nice. We now need to somehow make the nightly CI not make the tests fail. We want to be award of it but without the dreaded ❌….

milankl:
But that's what we have now, no? 👇🏼  nightly fails because of something in NetCDF.jl so we have a ❌ but it's still ✅ overall!

.github/workflows/nightly-ci.yml
```diff
@@ -0,0 +1,42 @@
...
+    branches:
+      - main
+    tags: '*'
+
+jobs:
+  test:
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
```
milankl, line 17: I've added this, but it doesn't seem to do the job?

navidcy: I'm asking at Slack..

navidcy:
We can do something like

https://github.com/JuliaCI/PkgTemplates.jl/blob/master/.github/workflows/JuliaNightly.yml

which is have the Julia Nightly tests run periodically on main (e.g. every week or so). This way we'll get an idea if something breaks but the Julia nightly won't be part of the CI tests that run for every commit/PR.

navidcy:
How about what I have done with https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/mk/testversions/.github/workflows/julianightly.yml?
I have it run once a week.

milankl:
Awesome! This is great. I'll merge 🎉

Issue #265, opened 2023-03-22
Jablonowski test case


milankl:
In the Jablonowski (#45) test case our day-9 baroclinic wave is currently not strong enough (T255 26 levls Float64 FullGaussianGrid $\alpha=1$ implicit stepping default diffusion)

![image](https://user-images.githubusercontent.com/25530332/227041708-91208df4-d0d3-407d-a02c-a20c778b2b9b.png)

versus

![image](https://user-images.githubusercontent.com/25530332/227042166-9b0a4306-bcdd-4f30-b6fb-9ad5a13fb61f.png)

![image](https://user-images.githubusercontent.com/25530332/227042399-606bcf73-f930-4ce9-8e18-930c349e2359.png)

Maybe an issue in the initial conditions or the implicit stepping dampens too aggressively? Also, I didn't wait for the steady state to equilibrate as Jablonoswki suggest.

milankl:
Same result at lower resolution (T170)

![image](https://user-images.githubusercontent.com/25530332/227054579-5899ff73-11c6-4b15-8222-39859e25a0b8.png)

and apart from some additional waves in pressure also with implicit $\alpha = 0.5$

![image](https://user-images.githubusercontent.com/25530332/227054686-ba226bf9-7ba6-44c7-af0d-da5faa934524.png)

milankl:
@christophernhill @jm-c any idea why that could be? 👆🏼 Compared to Jablonowski, we are about a day behind 🙈 See day 8 below, that one fits quite well and looking at our day 10 is quite similar to Jablonowski's day 9. What controls the error growth time scale in this sytem, proportional to the initial perturbation and the available potential energy in the baroclinic background flow?

![image](https://user-images.githubusercontent.com/25530332/227056479-7c336e47-adb2-4d8b-a165-446c62ee74d5.png)

milankl:
Okay, a stronger perturbation $u_p = 2$ m/s instead of 1m/s and a day later (day 10) yields roughly what we are supposed to see

![image](https://user-images.githubusercontent.com/25530332/227059437-851dce78-b35a-474b-a504-934385b21665.png)

milankl:
With $u_p$ = 3m/s the strength of the baroclinic wave after 9 days is about correct, but it didn't travel far enough: The big vortex sits at about 65˚N, 195˚E for us but more like 60˚N, 215˚E for Jablonowski

[...]

PR #362, opened 2023-07-24
Different vertical advection schemes

simone-silvestri:
The objective of this PR is to implement some stable vertical advection schemes (Upwinding and WENO)
Following http://www.gfdl.noaa.gov/idealized-spectral-models-quickstart, it looks like upwinding with semi-implicit time stepping requires reconstructing advected variables from previous time steps.

An example of 10 days simulation with third order upwinding (left) and second order centered advection (right)

Screenshot 2023-07-21 at 6 07 29 PM


simone-silvestri:
Difference between upwind third order (top, 3-point stencil) and upwind fifth order (bottom, 5-point stencil)

https://github.com/SpeedyWeather/SpeedyWeather.jl/assets/33547697/670fec3e-f5d3-44f8-9191-3884644dc6a2

milankl:
This is awesome!!

The Jablonowski test case (see also #265) can now be run as follows

```julia
using SpeedyWeather

# no mountains, no physics
spectral_grid = SpectralGrid(trunc=31,nlev=8)
orography = NoOrography(spectral_grid);

# I believe I deviated slightly here from the Jablonowski initial temperature profile for the stratosphere, will check
initial_conditions = ZonalWind()
model = PrimitiveDryModel(;spectral_grid,orography,initial_conditions,physics=false);
simulation = initialize!(model);

run!(simulation,n_days=10)
```
which yields
image

milankl:
`ZonalWind` are the Jablonowski initial conditions with perturbation

```julia
julia> initial_conditions = ZonalWind()
ZonalWind <: InitialConditions:
 η₀::Float64 = 0.252
 u₀::Float64 = 35.0
 perturb_lat::Float64 = 40.0
 perturb_lon::Float64 = 20.0
 perturb_uₚ::Float64 = 1.0
 perturb_radius::Float64 = 0.1
 ΔT::Float64 = 0.0
 Tmin::Float64 = 200.0
 pressure_on_orography::Bool = false
```
But I think they use another `ΔT` to have an increasing temperature in the stratosphere

simone-silvestri:
Is there a way to validate the exactness of the solution (i.e., a reference solution?) Or is there a particularly unstable case we can test to see if there is some stability improvement?
Not convecting cases though since, even if more stable than centered, upwind advection will not cut it

simone-silvestri:
Using 
```julia
spectral_grid = SpectralGrid(trunc=170,nlev=40)
```


https://github.com/SpeedyWeather/SpeedyWeather.jl/assets/33547697/99a997ae-6ac6-4060-aa1a-b699529ab639



https://github.com/SpeedyWeather/SpeedyWeather.jl/assets/33547697/1c86252c-9fd9-414d-bdab-f33aa14239b3

src/dynamics/diagnostic_variables.jl
```diff
@@ -59,14 +63,18 @@ function Base.zeros(::Type{GridVariables},SG::SpectralGrid)
     vor_grid            = zeros(Grid{NF},nlat_half)   # vorticity
     div_grid            = zeros(Grid{NF},nlat_half)   # divergence
     temp_grid           = zeros(Grid{NF},nlat_half)   # absolute temperature
+    temp_grid_prev      = zeros(Grid{NF},nlat_half)   # absolute temperature
     temp_virt_grid      = zeros(Grid{NF},nlat_half)   # virtual temperature
     humid_grid          = zeros(Grid{NF},nlat_half)   # specific humidity
+    humid_grid_prev     = zeros(Grid{NF},nlat_half)   # specific humidity
```
milankl, line 69: Currently not used, probably not needed, see below

src/dynamics/diagnostic_variables.jl
```diff
@@ -46,11 +46,15 @@ struct GridVariables{NF<:AbstractFloat,Grid<:AbstractGrid{NF}}
     vor_grid            ::Grid  # vorticity
     div_grid            ::Grid  # divergence
     temp_grid           ::Grid  # absolute temperature [K]
+    temp_grid_prev      ::Grid  # absolute temperature of previous time step [K]
     temp_virt_grid      ::Grid  # virtual tempereature [K]  
     humid_grid          ::Grid  # specific_humidity
+    humid_grid_prev     ::Grid  # specific_humidity of previous time step
```
milankl, line 52: Currently not used, probably not needed, see below

src/dynamics/models.jl
```diff
@@ -139,6 +139,7 @@ Base.@kwdef struct PrimitiveDryModel{NF<:AbstractFloat, D<:AbstractDevice} <: Pr
     boundary_layer_drag::BoundaryLayerDrag{NF} = LinearDrag(spectral_grid)
     temperature_relaxation::TemperatureRelaxation{NF} = HeldSuarez(spectral_grid)
     static_energy_diffusion::VerticalDiffusion{NF} = StaticEnergyDiffusion(spectral_grid)
+    vertical_advection::VerticalAdvection{NF} = UpwindVerticalAdvection(spectral_grid)
```
milankl, line 142: This is now 5th order upwind for the new default, right? If that's a good choice happy with it, I'll let you judge.

src/dynamics/tendencies_dynamics.jl
```diff
@@ -750,9 +673,14 @@ function SpeedyTransforms.gridded!( diagn::DiagnosticVariablesLayer,
     
     (; vor_grid, div_grid, u_grid, v_grid ) = diagn.grid_variables
     (; temp_grid, humid_grid ) = diagn.grid_variables
+    (; temp_grid_prev, u_grid_prev, v_grid_prev) = diagn.grid_variables
     U = diagn.dynamics_variables.a      # reuse work arrays for velocities spectral
     V = diagn.dynamics_variables.b      # U = u*coslat, V=v*coslat
 
+    @. temp_grid_prev = temp_grid
+    @. u_grid_prev = u_grid
+    @. v_grid_prev = v_grid
```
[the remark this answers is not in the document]
simone-silvestri: ok I can take it off and we can test a `PrimitiveWet` model to see how the stability looks like

src/dynamics/models.jl
simone-silvestri: Probably `WENOVerticalAdvection(spectral_grid)` is the right choice. `UpwindVerticalAdvection(spectral_grid, order = 3)` might be too diffusive (we don't want to have it as a default), and `UpwindVerticalAdvection(spectral_grid, order = 5)` might be unstable, so WENO should hit the sweet spot.
On the other hand, `WENOVerticalAdvection` might be more computationally expensive, so we could also leave the centered scheme as default but point out different choices in case of instabilities.

milankl (approved):

src/dynamics/models.jl
milankl: Making `CenteredVerticalAdvection` default.

milankl:
Quick speed comparison

image

milankl (approved):

milankl:
@simone-silvestri is there anything you wanted to add here? Have you checked that it's not allocating?

milankl:
In https://speedyweather.github.io/SpeedyWeather.jl/previews/PR365/primitiveequation/#Vertical-advection (may need some minutes to build) you can see the current state of the documentation. I won't go into details of the reconstruction and what schemes that means, but if you could have a look at that section and feel free to add anything that's important to understand the code! ❤️

simone-silvestri:
> @simone-silvestri is there anything you wanted to add here? Have you checked that it's not allocating?

I think it is ready to go. I will check that it does not allocate but it shouldn't (only tuples)
I ll also take a look at the documentation and fill it accordingly

simone-silvestri:
This seems to be ready, should I merge and look at the docs in your new PR? Should I add some tests before?

PR #368, opened 2023-08-28
Bugfix in vertical advection

simone-silvestri:
I was trying to run a high-resolution simulation (convinced the stability problem was solved) to replace the high-res simulation in the docs but the simulation crashed with upwinding!
I figured that I had changed just a little bit the structure of the code in #362 before merging but just enough to leave a bug not caught by the test.

This PR fixes it
(also produces the high res simulation to put in the README)


milankl:
Awesome, I'm about to review the semi-implicit scheme for the documentation. From the GFDL model it seems that the previous time step is necessary, but I don't fully understand this. Technically every linear term that is solved semi-implicitly needs to be evaluated at the previous time step, but for the adiabatic conversion term for example I'm evaluating that term at the current time step and then move it back to the previous time step with a linear operator in spectral space. Maybe the same is possible here. I'll report back

milankl (approved):

PR #376, opened 2023-09-06
Forcing and drag for the ShallowWaterModel

### LINKED ISSUE #151: Forcing and climatology in the shallow water system
> What is actually a reasonable warm-up / transient integration time for Speedy?

That's at the moment no straight forward to say. Most simulations I've run so far just used some initial conditions for vorticity. In that case you get a reasonably developed turbulence in 10-20 days even at higher resolution. However, that's decaying turbulence due to the lack of forcing obviously. Hence there's no actually interesting invariant measure the model converges to.

With #144 I've added some forcing for the shallow water equations, even including a seasonal cycle that you can switch on with `interface_relaxation = true` (default false) and control with `interface_relax_time` and `interface_relax_amplitude`.

What's happening there is that one pulls the l=1,m=0 zonal mode of the interface displacement / SSH towards something that mimicks an equatorial heating. Other modes are not affected, so you literally just dump energy into that mode. With the `seasonal_cycle = true` (default true) that equatorial heating moves up and down with the seasons, similar to an ITCZ but actually all the way to the tropic of cancer back to the equator and down to the tropic of capricorn and back.

My hope was to get an interesting and stable climate also in the shallow water system. But at the moment it still converges to a steady state (or one that closely follows the seasonal cycle). I probably haven't played around with it enough to get an actually nice setup that produces eddies at all resolutions with a proper energy cascade from some large scale forcing to some small scale dissipation

milankl:
fixes #151 

1. Adds a forcing term of the form
```math
\mathbf{F} = (F_u(\theta),0)
```
which only depends on latitude $\theta$ and mimics a jet stream in mid-latitudes that sits above the shallow water layer.

image

(6h after starting from rest, the perturbations are from orography)

2. Adds a quadratic drag term of the form
```math
\mathbf{D} = -\tfrac{c_D}{H}\vert \mathbf{u} \vert \mathbf{u}
```
with a drag coefficient $c_D$ and the shallow water layer thickness $H$.


milankl:
After 5 days

image

After 10 days

image

After 50 days

image

And vorticity after 50 days

image

milankl:
With $c_D = 10^{-5}$ the wind speeds are with 150m/s regularly somewhat too strong

image

milankl:
For $c_D = 2\cdot 10^{-5}$ this decreases to about 125m/s

image

milankl:
Or similar for a longer forcing time scale (=weaker) forcing, which also makes the northern jet weaker and less stable

image

milankl:
Last commit makes the forcing weaker by 50% looks good (T85)

image

image

milankl:
@ningliu-iga with this PR we can use the ShallowWaterModel for very long integrations, the forcing provides an energy input into the system, the drag removes it. Running it for O(years) seems like we're getting a stable climatology with turblence plus a bunch of Rossy waves that travel along the jet streams on both hemispheres. Shown is day 100 to day 200.
We can probably adjust how turbulent vs how wavy it's supposed to be by adjusting forcing/drag/layer thickness somewhat.

https://github.com/SpeedyWeather/SpeedyWeather.jl/assets/25530332/858e4e83-1a64-4900-81be-909a096d6c59

@maximilian-gelbrecht I think we talked about a climatology in the ShallowWaterModel before, this PR addresses this.

milankl:
Run with

```julia
spectral_grid = SpectralGrid(trunc=85,nlev=1)
quadratic_drag = SpeedyWeather.QuadraticDrag(spectral_grid,c_D=1e-5)
jet_stream_forcing = SpeedyWeather.JetStreamForcing(spectral_grid)
model = ShallowWaterModel(;spectral_grid,drag=quadratic_drag,forcing=jet_stream_forcing,initial_conditions=StartFromRest())
run!(simulation,n_days=100,output=true)
```

PR #384, opened 2023-09-11
Output via NCDatasets.jl, remove NetCDF.jl deps

milankl:
- [x] change from NetCDF.jl to NCDatasets.jl
- [x] also output orography as option
- [x] check whether sync is necessary
- [x] enable compression


milankl:
@natgeo-wong as you suggested we now use NCDatasets.jl instead of NetCDF.jl for output

@ningliu-iga `output = OutputWriter(;spectral_grid,Model,output_vars=[...,:orography])` will now add the orography to the output as used in the dynamical core (exactly for full grids, interpolated for reduced grids, or `as_matrix`)

PR #403, opened 2023-10-19
Soil moisture and vegetation

milankl:
Implements  (from Fortran SPEEDY documentation)

image

soil moisture availability is therefore diagnostic (depends on soil moisture, seasonal, and vegetation cover, constant)

image


milankl:
Surface humidity 6hours after dry start now with vegetation

image

compared to without vegetation

image

milankl:
And with the last commit a test how evaporation is different in NH summer (last pictures were winter)

image

PR #414, opened 2023-11-09
Convection

milankl:
This enables the `diagnose_convection!` function, actually doing `convection!` is still commented.

One issue we had was that Fortran SPEEDY uses humidity in g/kg, we switched to kg/kg but I believe that still has some issues in various places, e.g. I just found (was top, now bottom)

```julia
#    "latent heat of condensation [J/g] for consistency with specific humidity [g/Kg], also called alhc"
#    latent_heat_condensation::Float64 = 2501
    "latent heat of condensation [J/kg] for consistency with specific humidity [kg/kg], also called alhc"
    latent_heat_condensation::Float64 = 2501e3
```

Value seems to be an approximation to (from [wikipedia](https://en.wikipedia.org/wiki/Latent_heat))

image

Cloud top (dark blue means way above the surface) indicates where convection is active

image

which is almost everywhere in the beginning and it also spreads across the domain within days. Not what I was hoping but I bet there's still tons of bugs everywhere. To be fixed!


white-alistair (approved):
LGTM, nothing to add really. I like the new approach and I find the code exceptionally readable :clap: 

Out of curiosity, how would we eventually go about verifying the correctness of this?

milankl:
The last commit adds the calculation of the geopotential $\Phi$ to the physics (was previously just zero). This caused the dry static energy to be off (which is $c_p T + \Phi$) and so also the moist static energy, now with this fix, the convection is activated only in the tropics/sub-tropics reaching somewhat realistically into mid-latitudes I assume

image

milankl:
Starting to diagnose convective precipitation, scaling should still be off (unit is wrong) but pattern-wise doesn't look to unrealistic

image

milankl:
@white-alistair I also realised that the formula we use for saturation humidity is actually the [Tetens equation](https://en.wikipedia.org/wiki/Tetens_equation)

image

milankl:
Oh yeah @white-alistair that looks like some convection to me! T85 over 20 days. Looks good enough for me at this stage (it does qualitatively the right thing, is not absolutely off in terms of magnitude and is seemingly stable)

https://github.com/SpeedyWeather/SpeedyWeather.jl/assets/25530332/a39d1d4a-4539-4144-a83b-b60b3d2b871d

in comparison this is large-scale preicipitation


https://github.com/SpeedyWeather/SpeedyWeather.jl/assets/25530332/8f915a1c-3f1e-4dff-b3e1-99a2ae36aa02

milankl:
For reference this is IMERGE (but note that the unit here is mm/h)

image

white-alistair:
That looks amazing, congrats @milankl!

milankl:
Applied some tweaks, convection rains now in the tropics (better!) but also over a too wide area (bad), and magnitudes are an order of magnitude off (bad). Anyway, it seems to run stable and the rest is tuning we can do when other parameterizations are in too. E.g. Does it rain less when the vertical diffusion doesn't diffuse humidity (what we currently have), how does convection occur when there's no radiation to heat the atmosphere from below? At the moment we're only heating with surface fluxes (temperature relaxation is switched off), so I'm not too worried about things being somewhat unrealistic for now. I feel all of these can have an effect similar to the discrepancy we see.

image

milankl:
Created a comparison for a talk

image

milankl:
@p-hss SpeedyWeather has a not too-unrealistic closed hydrological cycle now!

PR #415, opened 2023-11-16
Fix load_trajectory

maximilian-gelbrecht:
Just fixing a bug due to a change in NCDatasets


milankl (approved):

milankl:
I thought we had a test for this function?

maximilian-gelbrecht:
We have one which tests it for the time variable. This one was also not affected, and the routine never returned an error. 

Before `[:]` returned the dataset as an Array in its regular shape (similar to how the Python netCDF library works (from memory)). Now `[:]` returns it as a Vector so like indexing a regular n-d Array with `[:]` and `Array` is the correct command. The tests just have one test for returning the correct time axis from the dataset. Time is always a vector, so this yielded no error.

Issue #417, opened 2023-11-16
Given user-facing parameters that represent a time period actually units


milankl:
At the moment we often have default parameters with second, minutes, hours, days scattered across the code. This basically requires the user to know whether `output_dt=6` means hours, seconds, days, or whatever. Using `Dates` or `Unitful` might offer neat ways to solve this.

1. `Dates`

Say you have some user-facing type `A` which has a parameter `time_scale`. Then we could do

```julia
import Dates: Second, Minute, Hour, Day

Base.@kwdef struct A
     time_scale::Second = Hour(1)
end
```

So the time scale is by default 1 hour, note that the conversion happens automatically
```julia
julia> A()
A(Second(3600))
```
One can also provide other periods at creation
```julia
julia> A(time_scale=Day(2))
A(Second(172800))
```
Which is exactly what I want. So the user doesn't have to know the unit of `time_scale` nor needs to do a conversion manually (how many seconds are in 15 days???). However this would throw an error
```julia
julia> A(time_scale=1.0)
ERROR: MethodError: Cannot `convert` an object of type Float64 to an object of type Second
```
which we could circumvent with
```julia
function Base.convert(::Type{Second},x::Real)
    xr = round(Int64,x)
    @info "Rounding and converting $x to $xr for integer seconds."
    return convert(Second,xr)
end

function Base.convert(::Type{Second},x::Integer)
    @info "Input $x assumed to have units of seconds. Use Minute(x), Hour(x), Day(x) otherwise."
    return Second(round(Int64,x))
end
```
which yields very neatly a little info that what the assumed unit is when not providing it
```julia
julia> A(time_scale=11.9)
[ Info: Rounding and converting 11.9 to 12 for integer seconds.
[ Info: Input 12 assumed to have units of seconds. Use Minute(x), Hour(x), Day(x) otherwise.
A(Second(12))

julia> A(time_scale=12)
[ Info: Input 12 assumed to have units of seconds. Use Minute(x), Hour(x), Day(x) otherwise.
A(Second(12))
```

One (major?) drawback though is that only integer seconds are supported with `Dates`, sure we could use `Millisecond` etc but I'd really like to have internally everything as SI as possible to prevent a unitwar. @maximilian-gelbrecht @white-alistair what do you think?

maximilian-gelbrecht:
[...]

Issue #416, opened 2023-11-16
Output time axis: option to adhere to output_dt exactly


maximilian-gelbrecht:
Currently, the specified `output_dt` is only approximately kept, as also demonstrated in the documentation on the netCDF output section. 

For long trajectories and small `output_dt` the offset can become quite large, quite quickly. 

One way of solving it would be to interpolate with an external library. But I wonder if we also should offer an alternative. 

What I could implement is the following:

Introduce an optional kwarg to the time stepper (so `Leapfrog`) for `output_dt`. If it is provided, then the scaled(*) `Δt_at_T31` is adjusted to be the closest divisor of `output_dt` of its input argument (or default value if no input for `Δt_at_T31` is provided).  When `output_dt` is divisible by the scaled `Δt_at_T31`, then the time output is kept exactly. 

(*) scaled with trunc, as done currently

What do you think? I definitely need this for myself, but I could also turn it into a PR.

milankl:
I'd be happy for you and go forward implementing these ideas, if I understand an example would be

You want output every 6hrs = 21600s? At your resolution you need a time step of approximately 125s for stability, but that's not a divisor of 21600s so make the time step slightly smaller, 120s so that we would do output every 21600/120 = 180 time steps. 

In that sense we would interpret `Δt_at_T31` as the largest possible time step but the real time step would be somewhat smaller. At the moment our time step is always an integer (even if represented as float) but we can relax that too. Because in the example above if 125s would be turned into 120s that would make the simulation also 4% slower. However, if we said, output every 21600s, ideal time step would be 125s so that's every 172.8 time steps, problem, so let's round this up to 173 time steps of 21600s/173 = 124.855...s the slow down would be insignificant.

I'd actually suggest to add a `adjust_with_output=true/false` to `Leapfrog` for which then the method as you suggest is done. That way it's more verbose that even if you choose a specific time step in `Leapfrog` it may get altered slightly depending on the output time step.

We would then just need to make sure that the model time is counted up correctly

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/b047723d187505438ab981ab67da89d76aeb0333/src/dynamics/time_integration.jl#L293

Because that's the time that's used to write the time axis of the netcdf output here

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/b047723d187505438ab981ab67da89d76aeb0333/src/output/output.jl#L385-L386

[...]

PR #418, opened 2023-11-17
Output_dt in leapfrog time stepping and some adjustments for using Dates 

maximilian-gelbrecht:
This implements #416 and part of #417. 

I made `Δt_sec` a `Second` as well, so it can be only full seconds. This also means that the adjusted `Δt_at_T31` will be a bit smaller and the integration will take a few more steps. If we would want to implement it like Milan wrote in #416 then we would need `Δt_sec` as a Float or Millisecond. I am not opposed to that, but I thought in the initial draft I do everything in SI. 

So with this I already implemented a part of #417 as well, but I did not adjust everything. After this is done, we definitely also need to go over the documentation. I already made some small corrections based on this PR, but no proper rewrite of the respective sections. 

I am on vacation next week, so don't expect any replies from me. Feel free to modify this PR/branch.

### DIFF
--- a/src/SpeedyWeather.jl
+++ b/src/SpeedyWeather.jl
@@ -17,7 +17,7 @@ import Adapt: Adapt, adapt, adapt_structure
 
 # INPUT OUTPUT
 import TOML
-import Dates: Dates, DateTime
+import Dates: Dates, DateTime, Second, Minute, Hour, Day, value
 import Printf: Printf, @sprintf
 import NCDatasets: NCDatasets, NCDataset, defDim, defVar
 import JLD2: jldopen
--- a/src/SpeedyWeather.jl
+++ b/src/SpeedyWeather.jl
@@ -17,7 +17,7 @@ import Adapt: Adapt, adapt, adapt_structure
 
 # INPUT OUTPUT
 import TOML
-import Dates: Dates, DateTime, Second, Minute, Hour, Day, value
+import Dates: Dates, DateTime, Second, Minute, Hour, Day, Week
 import Printf: Printf, @sprintf
 import NCDatasets: NCDatasets, NCDataset, defDim, defVar
 import JLD2: jldopen
--- a/src/SpeedyWeather.jl
+++ b/src/SpeedyWeather.jl
@@ -4,6 +4,7 @@ module SpeedyWeather
 using DocStringExtensions
 
 # NUMERICS
+import Primes
 import Random
 import FastGaussQuadrature
 import LinearAlgebra: LinearAlgebra, Diagonal
--- a/src/SpeedyWeather.jl
+++ b/src/SpeedyWeather.jl
@@ -17,7 +17,7 @@ import Adapt: Adapt, adapt, adapt_structure
 
 # INPUT OUTPUT
 import TOML
-import Dates: Dates, DateTime, Second, Minute, Hour, Day, Week
+import Dates: Dates, DateTime, Millisecond, Second, Minute, Hour, Day
 import Printf: Printf, @sprintf
 import NCDatasets: NCDatasets, NCDataset, defDim, defVar
 import JLD2: jldopen
@@ -27,7 +27,7 @@ import UnicodePlots
 import ProgressMeter
 
 # to avoid a `using Dates` to pass on DateTime arguments
-export DateTime
+export DateTime, Second, Minute, Hour, Day
 
 # EXPORT MONOLITHIC INTERFACE TO SPEEDY
 export  run_speedy,
--- a/src/dynamics/prognostic_variables.jl
+++ b/src/dynamics/prognostic_variables.jl
@@ -28,7 +28,7 @@ end
 $(TYPEDSIGNATURES)
 Initialize the clock with the time step `Δt` in the `time_stepping`."""
 function initialize!(clock::Clock,time_stepping::TimeStepper)
-    clock.n_timesteps = ceil(Int,24*clock.n_days/(value(time_stepping.Δt_sec)/3600))
+    clock.n_timesteps = ceil(Int,3600*24*clock.n_days/time_stepping.Δt_sec.value)
     return clock
 end
 
--- a/src/dynamics/prognostic_variables.jl
+++ b/src/dynamics/prognostic_variables.jl
@@ -28,7 +28,7 @@ end
 $(TYPEDSIGNATURES)
 Initialize the clock with the time step `Δt` in the `time_stepping`."""
 function initialize!(clock::Clock,time_stepping::TimeStepper)
-    clock.n_timesteps = ceil(Int,24*clock.n_days/(value(time_stepping.Δt_sec)/3600))
+    clock.n_timesteps = ceil(Int,24*clock.n_days/(value(Hour(time_stepping.Δt_sec))))
     return clock
 end
 
--- a/src/dynamics/prognostic_variables.jl
+++ b/src/dynamics/prognostic_variables.jl
@@ -28,7 +28,7 @@ end
 $(TYPEDSIGNATURES)
 Initialize the clock with the time step `Δt` in the `time_stepping`."""
 function initialize!(clock::Clock,time_stepping::TimeStepper)
-    clock.n_timesteps = ceil(Int,24*clock.n_days/(value(Hour(time_stepping.Δt_sec))))
+    clock.n_timesteps = ceil(Int,24*clock.n_days/(value(time_stepping.Δt_sec)/3600))
     return clock
 end
 
--- a/src/dynamics/prognostic_variables.jl
+++ b/src/dynamics/prognostic_variables.jl
@@ -28,7 +28,7 @@ end
 $(TYPEDSIGNATURES)
 Initialize the clock with the time step `Δt` in the `time_stepping`."""
 function initialize!(clock::Clock,time_stepping::TimeStepper)
-    clock.n_timesteps = ceil(Int,3600*24*clock.n_days/time_stepping.Δt_sec.value)
+    clock.n_timesteps = ceil(Int,3600*24*clock.n_days/time_stepping.Δt_sec)
     return clock
 end
 
--- a/src/dynamics/prognostic_variables.jl
+++ b/src/dynamics/prognostic_variables.jl
@@ -28,7 +28,7 @@ end
 $(TYPEDSIGNATURES)
 Initialize the clock with the time step `Δt` in the `time_stepping`."""
 function initialize!(clock::Clock,time_stepping::TimeStepper)
-    clock.n_timesteps = ceil(Int,24*clock.n_days/time_stepping.Δt_hrs)
+    clock.n_timesteps = ceil(Int,24*clock.n_days/(value(time_stepping.Δt_sec)/3600))
     return clock
 end
 
--- a/src/dynamics/time_integration.jl
+++ b/src/dynamics/time_integration.jl
@@ -36,8 +36,6 @@ Base.@kwdef mutable struct Leapfrog{NF} <: TimeStepper{NF}
     Δt::NF = Δt_sec/radius  
 end
 
-using Primes
-
 """
 $(TYPEDSIGNATURES)
 Computes the time step in [ms]. `Δt_at_T31` is always scaled with the resolution `trunc` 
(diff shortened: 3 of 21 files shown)


src/SpeedyWeather.jl
```diff
@@ -17,7 +17,7 @@ import Adapt: Adapt, adapt, adapt_structure
 
 # INPUT OUTPUT
 import TOML
-import Dates: Dates, DateTime
+import Dates: Dates, DateTime, Second, Minute, Hour, Day, value
```
milankl, line 20: I would not import `value` into scope simply because it's too common a term and therefore to easy to overwrite and cause unexpected errors, one can do `time_scale.value` to obtain the `Int64` behind a `time_scale::Second` (or any other time period).

src/dynamics/prognostic_variables.jl
```diff
@@ -28,7 +28,7 @@ end
 $(TYPEDSIGNATURES)
 Initialize the clock with the time step `Δt` in the `time_stepping`."""
 function initialize!(clock::Clock,time_stepping::TimeStepper)
-    clock.n_timesteps = ceil(Int,24*clock.n_days/time_stepping.Δt_hrs)
+    clock.n_timesteps = ceil(Int,24*clock.n_days/(value(time_stepping.Δt_sec)/3600))
```
milankl, line 31: This would become `time_stepping.Δt_sec.value` or could even be skipped (see comment in PR)

src/dynamics/time_integration.jl
```diff
@@ -10,30 +10,57 @@ Base.@kwdef struct Leapfrog{NF} <: TimeStepper{NF}
...
     robert_filter::NF = 0.05
 
     "Williams time filter (Amezcua 2011) coefficient for 3rd order acc"
     williams_filter::NF = 0.53
 
-
     # DERIVED FROM OPTIONS    
     "time step Δt [s] at specified resolution"
-    Δt_sec::Int = round(Int,60*Δt_at_T31*(32/(trunc+1)))
+    Δt_sec::Second = get_Δt_sec(Second(Δt_at_T31), trunc, adjust_Δt_with_output, Second(output_dt))
```
milankl, line 34: What I meant with #417 is really just the user facing time-related parameters. One doesn't choose `Δt_sec` directly so this can actually be of type `NF` and therefore doesn't have to be rounded to the second which addresses my point in #416

src/dynamics/time_integration.jl
```diff
@@ -10,30 +10,57 @@ Base.@kwdef struct Leapfrog{NF} <: TimeStepper{NF}
...
 
     "time step Δt [s/m] at specified resolution, scaled by 1/radius"
-    Δt::NF = Δt_sec/radius
-    
-    "convert time step Δt from minutes to hours"
-    Δt_hrs::Float64 = Δt_sec/3600     
+    Δt::NF = value(Δt_sec)/radius
```
milankl, line 37: Especially given that the actual (scaled) time step inside the model isn't an integer anyway!

src/dynamics/time_integration.jl
```diff
@@ -10,30 +10,57 @@ Base.@kwdef struct Leapfrog{NF} <: TimeStepper{NF}
...
+    scaling_factor = (32/(trunc+1))
+    scaled_Δt_at_T31 = value(Δt_at_T31)*scaling_factor
+
+    if adjust_Δt_with_output
+        # by using ceil we will always adjust Δt_at_T31 to be smaller than original one
+        k = ceil(value(output_dt) / scaled_Δt_at_T31)
+    
+        Δt_sec = Second(round(Int, (value(output_dt)/k) * inv(scaling_factor)))
+    else 
+        Δt_sec = Second(round(Int, scaled_Δt_at_T31))
```
milankl, line 58: So I think here I'd just skip the rounding, return a float and let the conversion happen in `Δt_sec::NF = get_Δt_sec(...)`

src/output/feedback.jl
```diff
@@ -80,7 +80,7 @@ function initialize!(feedback::Feedback,clock::Clock,model::ModelSetup)
 
     # hack: redefine element in global constant dt_in_sec
     # used to pass on the time step to ProgressMeter.speedstring
-    DT_IN_SEC[] = model.time_stepping.Δt_sec        
+    DT_IN_SEC[] = value(model.time_stepping.Δt_sec)
```
milankl, line 83: Do `.value` here too

src/physics/large_scale_condensation.jl
```diff
@@ -78,7 +78,7 @@ function large_scale_condensation!(
 
     (;gravity, water_density) = constants
     (;Δt_sec) = time_stepping
-    pₛΔt_gρ = pₛ*Δt_sec/gravity/water_density   # precompute constant
+    pₛΔt_gρ = pₛ*value(Δt_sec)/gravity/water_density   # precompute constant
```
milankl, line 81: `.value` here too, or via implementing *,/ as denoted in comment in PR

test/netcdf_output.jl
```diff
@@ -126,13 +126,23 @@ end
...
+    @test diff(t)[1] == Minute(70)
+
     # this is a nonsense simulation with way too large timesteps, but it's here to test the time axis output
     # for future tests: This simulation blows up because of too large time steps but only a warning is thrown
     # at the moment, no error
     # 1kyrs simulation
     spectral_grid = SpectralGrid()
-    time_stepping = Leapfrog(spectral_grid,Δt_at_T31=60*24*365*10)
-    output = OutputWriter(spectral_grid,PrimitiveDry,path=tmp_output_path,id="long-output-test",output_dt=24*365*10)
+    time_stepping = Leapfrog(spectral_grid,Δt_at_T31=Minute(60*24*365*10))
```
milankl, line 144: I thought this was supposed to make this easier as `Day(365*10)` 😉 ...unfortunately `Year` doesn't work because it's not fixed length.

test/netcdf_output.jl
```diff
@@ -126,13 +126,23 @@ end
...
+    output = OutputWriter(spectral_grid,PrimitiveDry,path=tmp_output_path,id="long-output-test",output_dt=Hour(24*365*10))
```
milankl, line 145: Also `Day(365*10)` here!

milankl:
Sorry I thought the following was a good idea (and therefore referenced it in the comments above) but it turns out it isn't. So forget about it, but for the record:

To avoid `a*time_scale.value` or `a*value(time_scale)` (and similar with `/`) we could also implement for SpeedyWeather the following promotion

```julia
julia> Base.:(*)(x::Second,y::Number) = x.value*y

julia> 5.5*a
ERROR: MethodError: *(::Second, ::Float64) is ambiguous.

Candidates:
  *(x::P, y::Real) where P<:Period
    @ Dates ~/.julia/juliaup/julia-1.9.4+0.x64.apple.darwin14/share/julia/stdlib/v1.9/Dates/src/periods.jl:90
  *(x::Second, y::Number)
    @ Main REPL[12]:1
```
which however hits ambiguity errors which don't seem to be easy to solve and therefore implement safely. I wanted to address this problem:
```julia
julia> a = Second(5)
5 seconds

julia> 5a
25 seconds

julia> 5.0*a
25 seconds

julia> 5.5*a
ERROR: InexactError: Int64(27.5)
```

src/physics/large_scale_condensation.jl
milankl: addressed with [fd2b5e2](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/418/commits/fd2b5e24547033dd18cffaca922ab7b9f559ad6f)

src/SpeedyWeather.jl
milankl: addressed with [fd2b5e2](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/418/commits/fd2b5e24547033dd18cffaca922ab7b9f559ad6f)

milankl:
@maximilian-gelbrecht I realised that this is actually a harder problem that I thought it would be. To illustrate:

Say you want output every 21601s (6 hours + 1 second) then the time step dt times number of time steps till output n needs to be n*dt = 21601. This doesn't work because 21601 is a prime number. So you need (a) either to output less than every 21601s (what we had before) or (b) you need a timestep more precise than seconds or (c) you need to restrict the possible output frequencies to something less precise than seconds.

So with the last commits I've used the chance to: (b) change the time integration time step to milliseconds precision. There's a `Δt_millisec::Millisecond` now that counts the clock up (=integer arithmetic no rounding errors yay!) as well as the corresponding `Δt_sec::NF` and `Δt::NF` which are the(unscaled/radius-scaled) float representations of the time step to be used in the dynamical core/parameterizations.

Some examples
```julia
julia> spectral_grid = SpectralGrid(trunc=42)
julia> time_stepping = Leapfrog(spectral_grid,adjust_with_output=true)
Leapfrog{Float32} <: TimeStepper
├ trunc::Int64 = 42
├ Δt_at_T31::Second = 1800 seconds
├ radius::Float32 = 6.371e6
├ adjust_with_output::Bool = true
├ robert_filter::Float32 = 0.05
├ williams_filter::Float32 = 0.53
├ Δt_millisec::Dates.Millisecond = 1350000 milliseconds
├ Δt_sec::Float32 = 1350.0
└ Δt::Float32 = 0.00021189766
```
which uses the `DEFAULT_OUTPUT_DT` (=21600s/6hrs), now lets be nasty and change to prime number of seconds output
```julia
julia> output = OutputWriter(spectral_grid,PrimitiveWet,output_dt=Second(21601))
julia> model = PrimitiveWetModel(;spectral_grid,time_stepping,output)
[ Info: Adjust with output: Time step shortened to 1080050 milliseconds (-20%)
```

The output dt is in milliseconds 21601000, which is only about 16 time steps à 1350000 ms, given the divisors of 21601000 

```julia
julia> Primes.factor(21601000)
2^3 * 5^3 * 21601
```
i.e. 1,2,4,5,8,10,20,25,40,... we would need to make it 20 time steps each 1080050ms to hit those 21601s exactly even with millisecond time step precision. That's a time step 20% shorter than before hence the `[ Info:` from above. I've also included in the initilisation of the output a check to throw an info as follows (here triggered with `adjust_with_output=false`)

```julia
julia> time_stepping = Leapfrog(spectral_grid,adjust_with_output=false)
julia> output = OutputWriter(spectral_grid,PrimitiveWet,output_dt=Second(21601));
julia> model = PrimitiveWetModel(;spectral_grid,time_stepping,output)
[ Info: 16 steps of Δt = 1339535ms yield output every 21432560ms (=21432.56s), but output_dt = 21601s
```

milankl:
okay next problem 😆

With timesteps in milliseconds but netcdf output in integer seconds the actual timestep written to file is rounded easily leading to an alternative roundup/down in the netcdf time axis (outputting every time step with `OutputWriter(...,output_dt=Second(0))`)
```julia
julia> ds["time"][:]
66-element Vector{DateTime}:
 2000-01-01T00:00:00
 2000-01-01T00:22:20
 2000-01-01T00:44:39
 2000-01-01T01:06:59
 2000-01-01T01:29:18
 ⋮
 
ulia> diff(ds["time"][:])
65-element Vector{Millisecond}:
 1340000 milliseconds
 1339000 milliseconds
 1340000 milliseconds
 1339000 milliseconds
 1340000 milliseconds
 1339000 milliseconds
 1340000 milliseconds
 ⋮
```

I believe we could just change it to hours since X and output in Float64?! Testing ...

milankl:
Okay that looks good now!

```julia
julia> ds["time"][:]
66-element Vector{DateTime}:
 2000-01-02T00:11:09.775
 2000-01-02T00:33:29.310
 2000-01-02T00:55:48.845
 2000-01-02T01:18:08.380
 2000-01-02T01:40:27.915
 ⋮

julia> diff(ds["time"][:])
65-element Vector{Millisecond}:
 1339535 milliseconds
 1339535 milliseconds
 1339535 milliseconds
 1339535 milliseconds
 1339535 milliseconds
 1339535 milliseconds
 1339535 milliseconds
 ⋮
```

src/dynamics/time_integration.jl
milankl: adressed with commits below

src/dynamics/time_integration.jl
milankl: both are NF now

maximilian-gelbrecht:
Sorry, for no responses so far, just got back from vacation yesterday. Got teaching and some other stuff to take care of first, but I can take a detailed look at it on Friday

milankl:
I think it's all good now, just some last tweaks for the documentation (and the Documenter is being naughty) I'll merge beforehand, but if something is not right and you suggest something different we can fix that later. I'm quite confident that you like those changes ;)

milankl:
Many thanks for initiating this in the first place Max!

PR #428, opened 2024-01-17
Use Julia v1.10 for CI and building Docs

navidcy:
This PR run CI on Julia v1.10 and also uses this version to build the documentation.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -15,6 +15,7 @@ jobs:
         version:
           - '1.8'
           - '1.9'
+          - '1.10'
         os:
           - ubuntu-latest
         arch:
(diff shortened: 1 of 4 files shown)


navidcy:
probably we should drop running CI on v1.9? or not yet?

milankl (approved):

milankl:
> probably we should drop running CI on v1.9? or not yet?

I'm fine leaving it in for now. Does Oceananigans always just live on the latest release?

navidcy:
> > probably we should drop running CI on v1.9? or not yet?
> 
> I'm fine leaving it in for now. Does Oceananigans always just live on the latest release?

Out test suit is heavy so we only afford to test on one julia version

PR #430, opened 2024-01-17
Adjust time step with radius

milankl:
Previously the time step was automatically chosen as a function of resolution but not as a function of radius, but larger planets have effectively a coarser grid spacing, meaning they can deal with larger time steps. And vice versa: radius=1e6 currently blows up in the shallow water model because we effectively have a finer resolution by a factor of ~7x, which requires a ~7x smaller time step.

As suggested by @KristianJS


milankl:
Galewsky jet after 4 days on Mars (radius=3.39e6)

image

Galewsky jet after 4 days on a planet 4x Earth's radius

image

However, at the moment larger planets cause stability problems, not sure why:

image

PR #439, opened 2024-02-01
ImmediateCondensation and Clausius-Clapeyron

milankl:
I've implemented a simpler condensation scheme, the immediate condensation from [Frierson 2006](https://doi.org/10.1175/JAS3753.1), in detail

image

which also suggested to use a simpler Clausius-Clapeyron than the current Tetens equation. Technically one could go back to Tetens, but the derivative wrt to temperature is a bit more complicated so I thought why not keep it simple. The immediate condensation from Frierson has the advantage that it comes with basically no parameter, but I've introduced a maximum heating for latent heat release which kept causing stability issues otherwise.


milankl:
@p-hss Large scale condensation is a bit patchier now, but that's probably a good thing. Frontal precipitation isn't as smooth anymore.

image

p-hss:
yepp, a bit more intermittency in the spatial patterns should make it more realistic, I think.

PR #441, opened 2024-02-03
Calculate bulk Richardson number

milankl:


milankl:
Surface bulk Richardson numbers $Ri_a$ after 10 days of model simulation

image

tropics are generally unstable $Ri_a > 10$ but mit-latitudes are more stable, sounds about right!

PR #450, opened 2024-02-22
AquaPlanetMask

milankl:
@natgeo-wong requested an `AquaPlanetMask`, here it is 🎉


navidcy:
Shall we include something in the docs mentioning this? Perhaps a minimal example?

milankl:
> Shall we include something in the docs mentioning this? Perhaps a minimal example?

With the last commit I've added the following to the `AbstractLandSeaMask` docstring
```julia
"""
Abstract type for land-sea masks. Custom land-sea masks have to be defined as

    CustomMask{NF<:AbstractFloat,Grid<:AbstractGrid{NF}} <: AbstractLandSeaMask{NF,Grid}

and need to have at least a field called `land_sea_mask::Grid` that uses a `Grid` as defined
by the spectral grid object, so of correct size and with the numberformat NF.
Then the initialize function has to be extended for that new mask

    initialize!(mask::CustomMask,model::PrimitiveEquation)

which generally is used to tweak the mask.land_sea_mask grid as you like, using
(preferrably read-only) anything from `model`.

The land-sea mask grid is expected to have values between [0,1] as we use a fractional mask,
allowing for grid points being, e.g. quarter land and three quarters sea for 0.25
with 0 (=sea) and 1 (=land). The surface fluxes will weight proportionally the fluxes e.g.
from sea and land surface temperatures. Note however, that the land-sea mask can declare
grid points being (at least partially) ocean even though the sea surface temperatures
aren't defined (=NaN) in that grid point. In that case, not flux is applied."""
abstract type AbstractLandSeaMask{NF,Grid} end
```

which is generally information I want to provide for every extandable type. And then we should have a little section in the documentation showcasing how to define a new, in this case, land-sea mask. It's just a lot of work =)

PR #453, opened 2024-02-22
Run LowerTriangularMatrix and output examples while building the docs

navidcy:
This PR changes some of the example so that they run live when docs are being built. This ensures that the output produced is up to date with the latest version of the package.


milankl (approved):

milankl:
Awesome, yeah I never seem to have gotten around this for the lower triangular matrices... Thanks Navid.

PR #455, opened 2024-02-22
Callbacks

milankl:
As suggested by @sandreza this implements a first version of a possible callback interface!

E.g.

```julia
julia> spectral_grid = SpectralGrid(trunc=31,nlev=8)
julia> callback = GlobalSurfaceTemperatureCallback(spectral_grid)
julia> model = PrimitiveDryModel(;spectral_grid,callbacks=[callback])
julia> run!(simulation,period=Day(10))
```
now `model.callbacks[1]` contains a vector of global mean surface temperature on every time step
```julia
julia> model.callbacks[1].temp
481-element Vector{Float32}:
 285.2934
 285.27222
 285.16898
   ⋮
 280.54077
```


navidcy:
Could we add the callback _after_ the model is created?

I'm just asking because I was thinking that this:

```Julia
using SpeedyWeather
output = OutputWriter(spectral_grid, ShallowWater)
model = ShallowWaterModel(;spectral_grid, output=output)
```

is a bit weird, in the sense that we need to construct the output and provide it with arg `ShallowWater` before we construct the model. And if there was a callback functionality one could write something like:

```Julia
using SpeedyWeather
model = ShallowWaterModel(;spectral_grid)
add_callback!(model, OutputWriter(spectral_grid))
```

Anyway, this is not necessarily for this PR.

maximilian-gelbrecht:
The `OutputWriter` can impact the time stepping though, if that option is turned on (which it is per default now). This was done so that one can have a regular e.g. hourly output without additional interpolation options. 
So allowing the `OutputWriter` to be specified later would be a bit messy. 

Callbacks are great though! And allowing those to be specified at a later point would be fine as well.

sandreza:
Thanks for adding this @milankl! I'm looking forward to testing this with multiple threads

milankl:
@navidcy `model.callbacks` is a vector so I've added some methods to make the following possible

```julia
julia> model = PrimitiveWetModel(;spectral_grid)

julia> append!(model.callbacks, GlobalSurfaceTemperatureCallback())
1-element Vector{Main.SpeedyWeather.AbstractCallback}:
 GlobalSurfaceTemperatureCallback{Int64} <: AbstractCallback
├ timestep_counter::Int64 = 0
└── arrays: temp
```
so using `append!` one can just add additional callbacks after model construction. It starts off with an empty vector and one can add as many as one likes!

milankl:
@navidcy I agree with your confusion about before/after with the output writer, I'll revise that at some point next week. For the callbacks what I find more important though, at the moment we have the following order of things on a given time step

- start with zero tendencies

then 
- compute parameterizations, accumulate tendencies
- compute dynamics, accumulate tendencies
- time stepping
- set tendencies to zero
- callbacks
- repeat

The nice thing about this is that the callbacks could be used to add something to the tendencies and it would be accumulated and not overwritten on the next time step. So that would be a situation where a callback could influence the simulation. However, you could not analyse the tendencies after a given time step. I'm wondering whether generally we want callbacks to be purely diagnostic and only use extensible forcing/drag terms to influence the simulation. An alternative I can see would be

- set tendencies to zero
- compute parameterizations, accumulate tendencies
- compute dynamics, accumulate tendencies
- time stepping
- callbacks
- repeat

which would allow one to analyse the tendencies with a callback but writing into the tendencies would have no effect as they'd be zeroed anyway.

milankl:
I'm tempted to define callbacks as purely diagnostic and let forcing/drag/parameterizations be the extensible components that can and should influence the simulation. Maybe @maximilian-gelbrecht @p-hss you have also opinions about that?

milankl:
@sandreza for projects we discussed, a purely diagnostic callback would be enough, right?

maximilian-gelbrecht:
> I'm tempted to define callbacks as purely diagnostic and let forcing/drag/parameterizations be the extensible components that can and should influence the simulation. Maybe @maximilian-gelbrecht @p-hss you have also opinions about that?

I thinks it's totally fine to do it like this and have purely diagnostic callbacks. For any possible future extension e.g. by integrating ML components we'll have to have separate discussions anyway and callbacks might not be the best way to do that anyway, as they'll likely need to interact more directly with the simulation. So yes, I agree it's better to have the parametrizations etc as extensible components.

milankl:
Yep, I think this is the right strategy
 
>     * set tendencies to zero
> 
>     * compute parameterizations, accumulate tendencies
> 
>     * compute dynamics, accumulate tendencies
> 
>     * time stepping
> 
>     * callbacks
> 
>     * repeat

with the limitation that a callback cannot influence the tendencies. It can, however, influence the simulation: A callback could be written that slowly lowers orography over Antarctica, representing a melting ice sheet, simply by changing `orography.orography` and `orography.geopot_surf` on calls.

milankl:
Last commit was necessary because callbacks should be a field of
```julia
    callbacks::Vector{AbstractCallback} = AbstractCallback[]
```
but with `model = PrimitiveWetModel(;spectral_grid,callbacks=[NoCallback()])` that threw an error, which however got resolved when specifying the parameters like `PrimitiveWetModel{Float32,CPUDevice}` which means that somehow the model constructor wasn't able to infer `NF, D` when passing on a `Vector{NoCallback}` that had to be converted (and isn't a subtype of) to `Vector{AbstractCallback}`. I don't fully understand what's been going on though.

Issue #443, opened 2024-02-14
[JOSS review] Nat's Review of SpeedyWeather.jl


natgeo-wong:
Review of SpeedyWeather.jl by Nat (https://github.com/openjournals/joss-reviews/issues/6323#issuecomment-1927459691)

See the pdf for detailed comments.
[reviewv1.pdf](https://github.com/SpeedyWeather/SpeedyWeather.jl/files/14288668/reviewv1.pdf)

Current outstanding issues on the checklist:
- [x] **Functionality:** Have the functional claims of the software been confirmed?
     - [x] We need a more detailed explanation of the different modules / `AbstractTypes` available for the different models
     - [x] which of these`AbstractType` modules are relevant for each of the different types of models (e.g. `ShallowWaterModel`, `PrimitiveDryModel`) available in SpeedyWeather.jl
- [x] **Functionality documentation:** Is the core functionality of the software documented to a satisfactory level (e.g., API method documentation)?
    - [x] A reorganization of the documentation into a more hierarchical form 
- [x] **Community guidelines:** Are there clear guidelines for third parties wishing to 1) Contribute to the software 2) Report issues or problems with the software 3) Seek support

natgeo-wong:
https://github.com/openjournals/joss-reviews/issues/6323

milankl:
Nat also suggests looking into [Paulius and Garner, 2006](https://doi.org/10.1175/JAS3705.1) for a simple radiation scheme

milankl:
> Nat also suggests looking into [Paulius and Garner, 2006](https://doi.org/10.1175/JAS3705.1) for a simple radiation scheme

implemented with #444

maximilian-gelbrecht:
@milankl Let me know if you want me to do something

milankl:
I can create a type tree for the documentation and outline more clearly what new types within that tree need to fulfill in order to be used properly, e.g. what would a `MyConvection <: AbstractConvection` need to have/do in order to serve as an alternative to our current `SimplifiedBettsMiller` convection?

However, @maximilian-gelbrecht what do you think our community guidelines are or should be? I think "open an issue" is the answer to all of them, but maybe we can just make that clearer in the readme or documentation?

maximilian-gelbrecht:
I do think this is also a problem many Julia library have, so much so that I think that Julia should have some kind of built-in system that handles that. But that's probably besides the point here. Personally, I'd put it in the doc string of each abstract type. 

[...]

PR #459, opened 2024-02-24
Parametric models, non-parametric abstract types, remove DynamicsConstants, Dict of callbacks

milankl:
Monster pull request that

- removes the parameters `{NF}` or `{NF,Grid{NF}}` from abstract parameterization types and other model components as we never dispatch on them anyway. This allows new parameterizations/model components to be defined as parametric types that may be anything
- introduces an `AbstractModelComponent` which streamlines for example the `show` function
- moves most `export` statements just above where those concrete types to be interfaced with the user are defined. We had a long, unmaintained list in `src/SpeedyWeather.jl` that's difficult to cross-reference with what's actually defined in the code and where
- removes the `DynamicsConstants` struct, which was just a leftover bin for some constants/precomputed arrays that I first didn't know where else to put. Now we just use `model.planet` and `model.atmosphere` more widely directly inside the model (we had a user-facing vs dynamical core-facing philosophy that I didn't like) and I've added `model.geopotential`, `model.coriolis`, `model.adiabatic_conversion`, that each have a constructor and an `initialize!` method
- starts to define our concrete model types as BarotropicModel{A,B,C,D,...} like Oceananigans does it to remove all abstract types and make them parametric instead.

@maximilian-gelbrecht FYI, happy to hear any thoughts you have on these changes ☝🏼


src/models/barotropic.jl
```diff
@@ -0,0 +1,79 @@
...
+    AT<:AbstractAtmosphere,
+    CO<:AbstractCoriolis,
+    FR<:AbstractForcing,
+    DR<:AbstractDrag,
+    IC<:InitialConditions,
+    TS<:AbstractTimeStepper,
+    ST<:SpectralTransform{NF},
+    HD<:AbstractHorizontalDiffusion,
+    GE<:AbstractGeometry,
+    DS<:AbstractDevice,
```
maximilian-gelbrecht, line 20: I'd say it's best to be consistent and always have the `AbstractDevice` as the second parameter. Will make it easier to write rountines that dispatch on it when we tackle the GPU models

maximilian-gelbrecht:
So, I haven't had a super deep look into it yet, but so far I really like those changes, especially moving to fully parametric model types! 

With regards to the modularisation and #443, I'd say this is also a good opportunity to document what each concrete type is expected to implement: Generally we expect each concrete type to have a `initialize!(module::Module, model::ModelSetup)` and then we should document what other method each concrete module type is expected to have.

src/models/barotropic.jl
milankl: Yes, that's good call. But also something we can change anytime: At the moment we do not dispatch over the parameters of model types, so changing their order will not cause any issues. The change here is first intended to remove abstract types from models so that the compiler can directly reason about all components of the model

milankl:
> With regards to the modularisation and #443, I'd say this is also a good opportunity to document what each concrete type is expected to implement: Generally we expect each concrete type to have a `initialize!(module::Module, model::ModelSetup)` and then we should document what other method each concrete module type is expected to have.

Yes, although in the new language that would be `initialize!(::AbstractModelComponent,::ModelSetup)`. In #455 I've just added that documentation for callbacks and it's really super similar among most components. It's just work and I didn't want to document it before I wasn't sure that's how we want to do things

maximilian-gelbrecht:
General Julia wisdom is that mutable structs are slow. Are you sure we need all of those to be mutable and this doesn't come at a performance penalty instead of working with `remake` or `reconstruct` to redo those structs if needed? I've only ever did some super basic speed comparisons for this though, and never in particularly complex code.

milankl:
Just checked that with the shallow water model turning every component I could mutable vs immutable, it doesn't make a difference. I don't think we'd be in a situation of creating heap allocations with mutable structs anyway as don't pack them into vectors and just read from them or change value in arrays in these mutable structs. I agree it's something we should keep in mind, but I'll probably make them for now as mutable/immutable as it makes sense from a user interface perspective?

I just like the idea that I could run a simulation for N days, define a callback that alters or switches on/off parameterizations on the fly by resetting components or values within them.

maximilian-gelbrecht:
> I just like the idea that I could run a simulation for N days, define a callback that alters or switches on/off parameterizations on the fly by resetting components or values within them.

Oh yes, that is great. I was just a bit worried about the performance, but if that's not an issue, then it's great to have. 

By the way, should we add a speed test to the CI to catch those things during development? I just looked through a few CI reports and the elepsed time for the runs is very consistent. One could do it as an extra CI step, that's not part of the regular tests and really only executed by GitHubActions where the results will be fairly consistent.

milankl:
> By the way, should we add a speed test to the CI to catch those things during development? I just looked through a few CI reports and the elepsed time for the runs is very consistent. One could do it as an extra CI step, that's not part of the regular tests and really only executed by GitHubActions where the results will be fairly consistent.

I'd love that. I just opened a new issue on that #464.

On this note here, I've done a quick comparison between main and this branch

| Simulation | Main: Speed [SYPD] | This PR: Speedy [SYPD] |
| - | - | - |
| T31L1 ShallowWater | 14k | 14k |
| T31L8 PrimitiveDry |  ~520 | ~720 |
| T31L8 PrimtiveWet |  ~440 | ~550 |
| Fortran SPEEDY | 240 | |

just by making the model parametric.

PR #476, opened 2024-03-06
Remove testing for 1.8 and 1.9 temporarily

milankl:
Because it keeps stalling


milankl:
Yeah, something is really fishy here. The tests now are just queued and don't even start.

navidcy:
There is a chance the repo hit its limit and jobs might start tomorrow? I noticed that some jobs weren’t being automatically cancelled despite new commits were pushed in the PR. But also these are hypotheses…

PR #483, opened 2024-03-11
Resume CI on Julia v1.9

navidcy:


milankl:
Are you testing whether this works again? 😉

milankl:
I mean technically on every release the JuliaRegistrator also tests for older versions, so I don't actually mind testing only 1.10 and then for every release we get a flag whether it also works for 1.8/1.9 and worst case we revert back some Julia features that were only released in 1.10, but honestly there weren't many anyway?

navidcy:
I don't think the registration runs the full test-suite for previous Julia versions...

milankl:
This test is really killing my nerves
```julia
Test Summary:        | Pass  Fail  Total  Time
D, ζ -> u, v -> D, ζ | 2243     1   2244  1.8s
  NF = Float32       | 1122         1122  0.3s
  NF = Float64       | 1121     1   1122  1.4s
```

it keeps failing every now and then, and I don't know what test tolerances for such a round transform are appropriate so I keep changing them.

milankl:
Same problem as before, CI on 1.8/1.9 keeps stalling after the interpolation tests. We could try and comment out the `set_vars.jl` testsets, maybe they cause some issues

```
include("interpolation.jl")
include("set_vars.jl")           # could be commented out?
```

milankl:
I don't get it, now it hangs somewhere else.

@giordano have you heard of issues where CI on 1.10 passes but on 1.8 and 1.9 just hangs somewhere halfway through the tests?

giordano:
Ugh, no, I'm not aware of this problem, sounds like it's something very specific to your application.  If anything, it's positive the newer version is working correctly (although it'd be nice to track down the culprit) 😅

milankl:
As just pointed out by @longemen3000 this issue could be related to the creation of files (our .nc output files that are created in some tests). We could make them smaller and see what happens!

milankl:
With the Makie extension #511 we should probably drop Julia v1.8. I haven't added anything extension related to the tests but the docs depend on it -- so they will fail as extensions got introduced with Julia v1.9

maximilian-gelbrecht:
The stalling CI here is the same problem as #508, and I had it in #503 as well. 

It's stalling on initialising `PrimitiveDryModel` and/or `PrimitiveWetModel`. I made some initial trial and error and couldn't find a specific culprit yet. It seems that Julia 1.9 struggles to dertermine so many parameters (maybe).

milankl:
superseded by #519

PR #492, opened 2024-03-14
Field trees

milankl:
One can now do

```julia
julia> tree(simulation)
Simulation{PrimitiveWetModel}
├┐prognostic_variables
│├ trunc
│├ nlat_half
│├ nlev
│├ n_steps
│├┐layers
││└┐timesteps
││ ├ trunc
││ ├ vor
││ ├ div
││ ├ temp
││ └ humid
│├┐surface
││└┐timesteps
││ ├ trunc
││ └ pres
│├┐ocean
││├ nlat_half
││├ time
││├ sea_surface_temperature
││└ sea_ice_concentration
│├┐land
││├ nlat_half
││├ time
││├ land_surface_temperature
...
```

to get a full field tree (a tree of all fields and their fields as long as they are defined in SpeedyWeather) printed. It's convenient to visualise nested structs. Also added to the docs to have that automatically documented. 

@natgeo-wong this is a first step towards "a more detailed explanation of modules" by which I think you mean the components of a SpeedyWeather simulation (avoiding the term "module" because of its use in Julia)

### DIFF
--- a/docs/src/structure.md
+++ b/docs/src/structure.md
@@ -11,6 +11,7 @@ extended Julia's `show` function to give you an overview of its contents,
 e.g. a `clock::Clock` is printed as
 
 ```@example structure
+using SpeedyWeather
 clock = Clock()
 ```
 
--- a/docs/src/structure.md
+++ b/docs/src/structure.md
@@ -1,7 +1,8 @@
 # Tree structure
 
 At the top of SpeedyWeather's type tree sits the `Simulation`, containing
-variables and model, which in itself contains model components and so on.
+variables and model, which in itself contains model components with their
+own fields and so on.
 (Note that we are talking about the structure of structs within structs
 not the type hierarchy as defined by subtyping abstract types.)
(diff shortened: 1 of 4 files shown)


natgeo-wong:
<3

docs/src/structure.md
```diff
@@ -0,0 +1,101 @@
...
+## Simulation
+
+When creating a Simulation, it's fields are
+```@example structure
+model = BarotropicModel()
+simulation = initialize!(model)
+```
+the `prognostic_variables`, the `diagnostic_variables` and the `model` (that we just
+initialized.) We could now do `tree(simulation)` but that gets very lengthy and
+so will will split things into `tree(simulation.prognostic_variables)`,
```
navidcy, line 37:
```suggestion
so will split things into `tree(simulation.prognostic_variables)`,
```

navidcy (approved):

docs/src/structure.md
navidcy: I took the liberty to commit :)

Issue #400, opened 2023-10-09
Conserved quantity verifying


mini-DONG:
Hi, thanks a lot for your help at #391 . Now, I have run a shallow water model like:
``` julia
spectral_grid = SpectralGrid(trunc=511,nlev=1)
initial_conditions = StartFromRest()
output = OutputWriter(spectral_grid,ShallowWater,
                      output_vars=[:u,:v,:pres,:vor,:div])
model = ShallowWaterModel(;spectral_grid,initial_conditions,output=output)
simulation = initialize!(model)
```
with initial conditions:
```julia
simulation.prognostic_variables.layers[1].timesteps[1].vor .= vor/spectral_grid.radius
simulation.prognostic_variables.layers[1].timesteps[1].div .= div/spectral_grid.radius
simulation.prognostic_variables.surface.timesteps[1].pres .= etas
```
I have run 1 year, and the first output (i.e. initial conditions) and last output is like:
![1696822295135](https://github.com/SpeedyWeather/SpeedyWeather.jl/assets/145758947/65f80458-1c9e-41fa-9c9a-4a3cce0e1897)
Like #323, I also want to verify the conserved quantity. I calculate the total thickness with python codes:
``` python
h = ds.pres + 8500
cos_phi =  np.cos(np.radians(ds.lat))
H = (h*cos_phi).sum(['lat','lon'])
```
The plot of H:
![1696822472399](https://github.com/SpeedyWeather/SpeedyWeather.jl/assets/145758947/0dfb978a-138b-4664-b560-94b8970f6c70)
H is growing fast, which doesn't seem reasonable. Or maybe I should use other way to calculate?

milankl:
If I read your y axis correctly then you have
```julia
julia> h1 = 6200+6.3875e9
6.3875062e9

julia> h0 = 5500+6.3875e9
6.3875055e9

julia> h1/h0
1.0000001095889468
```
It looks like it's growing "fast" but actually it's conserved to 1e-7 which I find very reasonable given there are still several sources of error

- It's single precision so more than 1e-8 you cannot expect anyway
- By default you are using the OctahedralGaussianGrid, but the output will be on the corresponding full grid, that's the FullGaussianGrid. The default interpolation is not conservative.
- you `sum` the array, but on the FullGaussianGrid even with coslat scaling the grid cells are not of even size because of the Gaussian latitudes. You use a high resolution T511, but this error may still be not negligible.

[...]

PR #494, opened 2024-03-19
Export `coriolis`, unscale diagnostics at finish, add! callbacks directly to model too

milankl:
Changes arose from #400

### DIFF
--- a/docs/src/callbacks.md
+++ b/docs/src/callbacks.md
@@ -184,13 +184,17 @@ add!(callbacks, :key1 => NoCallback(), :key2 => NoCallback())   # keys provided
 ```
 Meaning that callbacks can be added before and after model construction
 
-
 ```@example callbacks
 spectral_grid = SpectralGrid()
 callbacks = CallbackDict(:callback_added_before => NoCallback())
 model = PrimitiveWetModel(; spectral_grid, callbacks)
 add!(model.callbacks, :callback_added_afterwards => NoCallback())
+add!(model, :callback_added_afterwards2 => NoCallback())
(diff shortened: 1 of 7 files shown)


milankl:
@miniufo changes now in #main

miniufo:
I should update my version.  Also thinking other variables such as $\cos\phi$ and $\sin\phi$.  I need these too when calculating the absolute angular momentum.  Is there any easy way to do this like the exported `coriolis` function?

milankl:
They are all in the `RingGrids` module, which is also documented here https://speedyweather.github.io/SpeedyWeather.jl/dev/ringgrids/ in short, you do `RingGrids.get_*(::AbstractGrid)` with `*` one of the following
```julia
julia> grid = rand(OctahedralGaussianGrid,4);

julia> RingGrids.get_

get_colat               get_colatlons           get_lat
get_latd                get_latdlonds           get_latlons
get_lon                 get_lond                get_nlat
get_nlat_half           get_nlon                get_nlon_max
get_nlon_per_ring       get_nlons               get_npoints
get_quadrature_weights  get_resolution          get_solid_angles
get_truncation
julia> RingGrids.get_lat(grid)
8-element Vector{Float64}:
  1.2880392632011
  0.921759746334117
  0.5533407873047431
  0.18447924790276526
 -0.18447924790276526
 -0.5533407873047431
 -0.921759746334117
 -1.2880392632011
```
with `lat` in radians, `colat` in radians, `d` means degree, `s` means for every grid point and not just on the zonal rings or meridians.

PR #497, opened 2024-03-20
Document typical online analysis steps

milankl:
Motivated from #400 

@miniufo I'll merge this if it passes and then you can create a pull request extending the remaining section if you like! (which would be awesome!!)


miniufo:
Great, we could just focus on `analysis.md`.

PR #506, opened 2024-03-29
Broadcasting for RingGrids

milankl:
Previously we had broadcasting operations with RingGrids escaping the grid and returning just the underlying `Vector`
```julia
julia> G = ones(OctahedralGaussianGrid,1);

julia> G+G
40-element Vector{Float64}:
 2.0
 2.0
 ⋮
 2.0
 2.0
```
which was really annoying. While we don't use these operations inside the dynamical core (all in-place) in the documentation for example we always had allocate first and then in-place
```julia
A = zero(B)
@. A = B + 1
```
because otherwise the `B+1` operation would just return a vector again and so you lost the information of the grid.

Now this PR implements broadcasting for all `AbstractGrid`, meaning that `A+B`, `A-B`, `A.*B`, `A./B`, `A+1`, `A-1`, `2A`, `A*2.0`, `A/2` etc are all possible and a grid is returned again. This also works with type promotion, i.e. `FullGaussianGrid{Float32}` + `FullGaussianGrid{Float64}` will return that grid with eltype Float64 and same for the operations with scalars - like it also works with `Vector{T}`.

@miniufo In practice, this means that while previously we had to do
```julia
    h = zero(ζ)
    q = zero(ζ)
    
    H = model.atmosphere.layer_thickness
    Hb = model.orography.orography
    f = coriolis(ζ)        # create f on that grid
    
    @. h = η + H - Hb   # thickness
    @. q = (ζ + f) / h  # Potential vorticity
```
to calculate the potential vorticity, now you can just do
```julia
    H = model.atmosphere.layer_thickness
    Hb = model.orography.orography
    h = @. η + H - Hb   # thickness

    f = coriolis(ζ)     # create f on that grid
    q = @. (ζ + f) / h  # Potential vorticity
```

### DIFF
--- a/docs/src/analysis.md
+++ b/docs/src/analysis.md
@@ -338,19 +338,14 @@ We define a function ``total_enstrophy`` for this
 
 ```@example analysis
 function total_enstrophy(ζ, η, model)
-    h = zero(ζ)
-    q = zero(ζ)
-    Q = zero(ζ)
-    
-    # constants used by Speedy model
+    # constants from model
     H = model.atmosphere.layer_thickness
     Hb = model.orography.orography
-
-    f = coriolis(u)        # create f on that grid
+    f = coriolis(ζ)     # create f on the grid
     
-    @. h = η + H - Hb   # thickness
-    @. q = (ζ + f) / h  # Potential vorticity
-    @. Q = q^2 / 2      # Potential enstrophy
+    h = @. η + H - Hb   # thickness
+    q = @. (ζ + f) / h  # Potential vorticity
+    Q = @. q^2 / 2      # Potential enstrophy
 
     # transform to spectral, take l=m=0 mode at [1] and normalize for mean
     return Q_mean = real(spectral(Q)[1]) / model.spectral_transform.norm_sphere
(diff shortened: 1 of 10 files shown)


milankl:
Oh neat, even broadcasting between grids and vectors works!
```julia
julia> H
3-element, 1-ring HEALPixGrid{Float16}:
 1.0
 1.0
 1.0

julia> H + [1,2,3]
3-element, 1-ring HEALPixGrid{Float16}:
 2.0
 3.0
 4.0

julia> H + [1,2.0,3]
3-element, 1-ring HEALPixGrid{Float64}:
 2.0
 3.0
 4.0
```

maximilian-gelbrecht (approved):
Great! 

Yeah, this should also immediately work on GPUs as well, as far as I can see it. 

I encountered the problem with inferring the non-parametric type before as well. It's odd that there's no direct way to do this. I think it's fine to explicitly define it. This seems like the most stable variant. If you'd rather avoid to explicit definition one could also do something like  
```Julia 
L = FullClenshawGrid(randn(NF, 96*47), 24)
typeof(L).name.wrapper # returns: FullClenshawGrid
```
But I am not sure, if this is considered part of Julia's stable API.

Issue #482, opened 2024-03-11
Replace `PythonPlot` with a Julia-based plotting library in Docs


navidcy:
E.g. CairoMakie

milankl:
Yes, absolutely. My problem is just that I'm still more used to matplotlib + cartopy than to Makie, but I should change that. If you want to add `GeoMakie, GLMakie` to `docs/Project.toml` go ahead, then we can replace the examples one by one.

milankl:
All Makie now

PR #511, opened 2024-04-03
Add Makie extension

milankl:
@simondanisch you're probably happy to see this!
@navidcy getting there #482 

```julia
julia> vor
18688-element, 128-ring OctahedralGaussianGrid{Float32}:
  3.0710835f-5
  3.1388656f-5
  ⋮
 -8.172375f-6
 -7.466016f-6

julia> heatmap(vor, title="Surface vorticity [1/s]")
```

yields

image


navidcy:
I can help out!

milankl:
> I can help out!

Yeah, I don't think I've set up the extension correctly. I thought

- put Makie (not CairoMakie, GLMakie) into the extension and define its `heatmap` function therein for `::AbstractGrid`
- but do I then need to also `export heatmap` in SpeedyWeather?
- it currently doesn't load the new `heatmap` method for `::AbstractGrid` when `using CairoMakie` (which also imports Makie, so I assumed that would trigger the extension)
- I then add CairoMakie to the `Project.toml` of the docs (because that's the backend I want to use there)

milankl:
@navidcy Okay, I got it to work

```julia
julia> using SpeedyWeather

julia> heatmap
ERROR: UndefVarError: `heatmap` not defined

julia> using CairoMakie
Precompiling SpeedyWeatherMakieExt
  1 dependency successfully precompiled in 12 seconds. 360 already precompiled.

julia> methods(heatmap)
# 4 methods for generic function "heatmap" from MakieCore:
 [1] heatmap()
     @ ~/.julia/packages/MakieCore/UAwps/src/recipes.jl:172
 [2] heatmap(grid::SpeedyWeather.RingGrids.AbstractFullGrid; title)
     @ SpeedyWeatherMakieExt ~/.julia/packages/SpeedyWeather/kXPNB/ext/SpeedyWeatherMakieExt.jl:14
 [3] heatmap(grid::SpeedyWeather.RingGrids.AbstractGrid; title)
     @ SpeedyWeatherMakieExt ~/.julia/packages/SpeedyWeather/kXPNB/ext/SpeedyWeatherMakieExt.jl:6
 [4] heatmap(args...; kw...)
     @ ~/.julia/packages/MakieCore/UAwps/src/recipes.jl:175

julia> heatmap(rand(FullGaussianGrid,4))
```
pops up a CairoMakie window

image


Lessons learned

- You can have an extension that depends on `Makie` which is also imported when `using CairoMakie` (which imports Makie). That way we define the function once but the user can still have any backend
- It's tricky to test this locally as `using .SpeedyWeather` is in the SpeedyWeather environment, which isn't supposed to have a hard dependency on Makie (=it's not there) so ever `using CairoMakie` will try and add this dependency to the environment (which is exactly not what we want
- So pushing to branch and then installing it in another environment, e.g. `(@v1.10) pkg> add https://github.com/SpeedyWeather/SpeedyWeather.jl#mk/makie` is the way to go as it'll detect the extension correctly without requiring Makie, but then CairoMakie is installed in that environment and `using CairoMakie` will then also precompile and import the extension

@SimonDanisch I've struggled with this ☝🏼 as I haven't found any info. The [pkg docs](https://pkgdocs.julialang.org/v1.10/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)) only describe extensions from the perspective if Makie would want to define some of its plotting functions for external types, not the probably more common use case (?) that a package owns a type and wants to define some plotting function (owned by Makie) for it. If you want me to write this up as a quick guide in the makie docs let me know where and I'm happy to add it.

docs/src/installation.md
```diff
@@ -20,9 +20,22 @@ julia> Pkg.add(url="https://github.com/SpeedyWeather/SpeedyWeather.jl", rev="mai
 ```
 In a similar manner, other branches can be also installed. You can also type, equivalently,
 ```julia
-julia> ] add https://github.com/SpeedyWeather/SpeedyWeather.jl#main
+julia>] add https://github.com/SpeedyWeather/SpeedyWeather.jl#main
```
navidcy, line 23: why this space was deleted?
it's really minor, but at least on my repls there is an empty space between `>` and whatever I want to start typing

milankl: You are right! I just felt wrong to type " " when I don't type it in the REPL but didn't think about it too much. Will correct it now!

milankl: space is back in with [d2f7133](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/514/commits/d2f713370ff34715264ef24a8176cb1caa53c28d)

navidcy: this is really my OCD at its best...

SimonDanisch:
@milankl would be amazing to have this in the docs. 
I think @jkrumbiegel thought about having a section in the docs for this, maybe he has some thoughts.
Otherwise I'd say make a new page for package extensions in `how-to`:
https://docs.makie.org/stable/how-to/
https://github.com/MakieOrg/Makie.jl/tree/master/docs/how-to

PR #520, opened 2024-04-12
AbstractGridArray N-dimensional and GPU ready

milankl:
Accompanying PR to #503 but for RingGrids

TODO

- [x] AbstractGridArray indexing
- [x] Broadcasting
- [x] GPU broadcasting (using JLArrays)
- [x] `rings::Vector{UnitRange{Int}}` lives in every AbstractGridArray instance
- [x] ~~make rings same arraytype as data?~~
- [x] AbstractGridArray constructors
- [x] Retrieve coordinates
- [x] extract 2D grids with `:` 
- [x] non-allocating loops over all grids, rings 
- [x] ~~Define everything in terms of lat in radians not colat in radians~~
- [x] N-dimensional `FullGaussianArray`, `FullClenshawArray`, `FullHEALPixArray`, `FullOctaHEALPixArray`
- [x] N-dimensional `OctahedralGaussianArray`, `OctehedralClenshawArray`, `HEALPixArray`, `OctaHEALPixArray`
- [x] Generalise reduced grids more
- [x] ~~Document how new grids can be defined~~ (every grid has it's own file now so it's somewhat clear but writing this up into the documentation can be postponed)
- [x] ~~3D interpolation?~~
- [x] Tests
- [x] Documentation

This works already

```julia
julia> G = rand(FullGaussianGrid{Float16},1,2)
8×2 FullGaussianArray{Float16, 2, Matrix{Float16}}:
 0.06494  0.6514
 0.9834   0.147
 0.0845   0.536
 0.1787   0.85
 0.0635   0.2065
 0.1011   0.4233
 0.9565   0.7886
 0.673    0.991

julia> G[1] = 1
1

julia> G[2,1] = 0
0

julia> G
8×2 FullGaussianArray{Float16, 2, Matrix{Float16}}:
 1.0     0.6514
 0.0     0.147
 0.0845  0.536
 0.1787  0.85
 0.0635  0.2065
 0.1011  0.4233
 0.9565  0.7886
 0.673   0.991

julia> G[:,1]
8-element, 2-ring FullGaussianGrid{Float16}:
 1.0
 0.0
 0.0845
 0.1787
 0.0635
 0.1011
 0.9565
 0.673
```

and of course the unicodeplots 😉 

image


milankl:
With 
```julia
@inline eachgrid(grid::AbstractGridArray) = CartesianIndices(size(grid)[2:end])
```

looping over all grid points can be done as follows

```julia
function _scale_lat!(grid::AbstractGridArray{T}, v::AbstractVector) where T
    @boundscheck get_nlat(grid) == length(v) || throw(BoundsError)
    
    for k in eachgrid(grid)
        for (j, ring) in enumerate(eachring(grid))
            vj = convert(T, v[j])
            for ij in ring
                grid[ij, k] *= vj
            end
        end
    end

    return grid
end
```
the ring indices are precomputed in `grid.rings`, and `eachgrid` is non-allocating, so the whole thing is too
```julia
julia> @btime RingGrids._scale_lat!($grid, $coslat);
  535.558 μs (0 allocations: 0 bytes)
```

src/RingGrids/general.jl
```diff
@@ -0,0 +1,324 @@
...
+
+## ITERATORS
+"""
+$(TYPEDSIGNATURES)
+CartesianIndices for the 2nd to last dimension of an AbstractGridArray,
+to be used like
+
+for k in eachgrid(grid)
+    for ring in eachring(grid)
+        for ij in ring
```
maximilian-gelbrecht, line 230: I like that! Reads very intuitively.

src/RingGrids/general.jl
```diff
@@ -0,0 +1,324 @@
...
+    each_index_in_ring!(rings, Grid, nlat_half)
+    return rings
+end
+
+"""
+$(TYPEDSIGNATURES)
+Same as `eachring(grid)` but performs a bounds check to assess that all grids
+in `grids` are of same size."""
+function eachring(grid1::Grid, grids::Grid...) where {Grid<:AbstractGridArray}
+    @inline
```
maximilian-gelbrecht, line 256: The `@inline` in an individual line is fine?

src/RingGrids/general.jl
```diff
@@ -0,0 +1,324 @@
...
+function whichring(ij::Integer, rings::Vector{UnitRange{Int}})
+    @boundscheck 0 < ij <= rings[end][end] || throw(BoundsError)
+    j = 1
+    @inbounds while ij > rings[j][end]
+        j += 1
+    end
+    return j
+end
+
+## BROADCASTING
```
maximilian-gelbrecht, line 301: Is there currently a problem with the broadcasting? I think in general, it should be much simpler than for `LowerTriangularArray`, as we really only need to specialise the `similar` function.

src/RingGrids/general.jl
```diff
@@ -0,0 +1,324 @@
...
+
+# # important to remove Grid{T} parameter T (eltype/number format) here to broadcast
+# # automatically across the same grid type but with different T
+# # e.g. FullGaussianGrid{Float32} and FullGaussianGrid{Float64}
+# Base.BroadcastStyle(::Type{Grid}) where {Grid<:AbstractGridArray} =
+#     AbstractGridArrayStyle{nonparametric_type(Grid)}()
+
+# # allocation for broadcasting, create a new Grid with undef of type/number format T
+# function Base.similar(bc::Broadcasted{AbstractGridArrayStyle{Grid}}, ::Type{T}) where {Grid, T}
+#     Grid(Vector{T}(undef,length(bc)))
```
maximilian-gelbrecht, line 317: Maybe it's already enough to change this line here to include the `ArrayType`. Maybe this needs an additional parameter in the broadcast style to distinguish between CPU and GPU style.

src/RingGrids/full_grids.jl
```diff
@@ -77,128 +52,31 @@ function each_index_in_ring!(   rings::Vector{<:UnitRange{<:Integer}},
...
-        "H$nlat_half ($(4nlat_half)x$(2nlat_half-1)) FullHEALPixGrid{$T}.")
+## COORDINATES
+function get_lond(Grid::Type{<:AbstractFullGridArray}, nlat_half::Integer)
+    lon = get_lon(Grid, nlat_half)
+    lon .*= 360/2π      # convert to lond in-place
+    return lon          # = lond
 end
 
-nonparametric_type(::Type{<:FullHEALPixGrid}) = FullHEALPixGrid
+function get_colatlons(Grid::Type{<:AbstractFullGridArray}, nlat_half::Integer)
```
maximilian-gelbrecht, line 62: I think some of these functions for `full_grids.jl` deserve a docstring, at least those that can be used for all grid types.

src/RingGrids/grids/full_clenshaw.jl
```diff
@@ -0,0 +1,38 @@
+struct FullClenshawArray{T, N, ArrayType <: AbstractArray{T, N}} <: AbstractFullGridArray{T, N, ArrayType}
+    data::ArrayType                 # data array, ring by ring, north to south
+    nlat_half::Int                  # number of latitudes on one hemisphere
+    rings::Vector{UnitRange{Int}}   # TODO make same array type as data?
```
maximilian-gelbrecht, line 4: Indices should always stay on the CPU as far as I know, so that should be fine (but might warrant a small test). (Same comment applies to all other grid types)

src/RingGrids/grids/full_clenshaw.jl
```diff
@@ -0,0 +1,38 @@
+struct FullClenshawArray{T, N, ArrayType <: AbstractArray{T, N}} <: AbstractFullGridArray{T, N, ArrayType}
```
maximilian-gelbrecht, line 1: Add docstring for the N-dim types as well? (Same comment applies to all other grid types)

src/RingGrids/grids/new_grids.jl
```diff
@@ -0,0 +1,42 @@
...
+
+Spectral truncation
+    truncation_order    # linear, quadratic, cubic = 1, 2, 3 for grid 
+    get_truncation      # spectral truncation given a grid resolution
+    get_resolution      # grid resolution given a spectral truncation
+
+Quadrature weights and solid angles
+    get_quadrature_weights  # = sinθ Δθ for grid points on ring j for meridional integration
+    get_solid_angle         # = sinθ Δθ Δϕ, solid angle of grid points on ring j
+"""
```
maximilian-gelbrecht, line 42: The abstract type that is annotated with the docstring is missing?

src/RingGrids/grids/octahedral_gaussian.jl
```diff
@@ -0,0 +1,118 @@
...
+An Octahedral Gaussian grid that uses `nlat` Gaussian latitudes, but a decreasing number of longitude
+points per latitude ring towards the poles. Starting with 20 equi-spaced longitude points (starting at 0˚E)
+on the rings around the poles, each latitude ring towards the equator has consecuitively 4 more points,
+one for each face of the octahedron. E.g. 20, 24, 28, 32, ...nlon-4, nlon, nlon, nlon-4, ..., 32, 28, 24, 20.
+The maximum number of longitue points is `nlon`. The values of all grid points are stored in a vector
+field `v` that unravels the data 0 to 360˚, then ring by ring, which are sorted north to south."""
+struct OctahedralGaussianGrid{T} <: AbstractOctahedralGrid{T}
+    data::Vector{T}    # data vector, ring by ring, north to south
+    nlat_half::Int  # number of latitudes on one hemisphere
+
```
maximilian-gelbrecht, line 13: This doesn't have the `rings` object yet, or is it not needed because the indices are easier to compute for the Octahedral grids?

src/RingGrids/interpolation.jl
```diff
@@ -213,9 +213,9 @@ function interpolate(   A::AbstractGrid{NF},        # field to interpolate
                         I::AbstractInterpolator     # indices in I are assumed to be calculated already!
                         ) where NF                  # use number format from input data also for output
 
-    (; npoints ) = I.locator             # number of points to interpolate onto
-    Aout = Vector{NF}(undef, npoints)        # preallocate: onto θs, λs interpolated values of A
-    interpolate!(Aout, A, I)                  # perform interpolation, store in As
+    (; npoints ) = I.locator                # number of points to interpolate onto
+    Aout = Vector{NF}(undef, npoints)       # preallocate: onto θs, λs interpolated values of A
```
maximilian-gelbrecht, line 217: The interpolation is still WIP, right? 

Because here and in the following it would need a more flexible constructor, a `similar` or a `ArrayType{...}(undef, ...)` or something similar

src/RingGrids/utility_functions.jl
```diff
@@ -1,43 +1,32 @@
-"""
-    true/false = isincreasing(v::Vector)
-
+"""$(TYPEDSIGNATURES)
 Check whether elements of a vector `v` are strictly increasing."""
-function isincreasing(x::Vector)
+function isincreasing(x::AbstractVector)
     is_increasing = true
     for i in 2:length(x)
         is_increasing &= x[i-1] < x[i] ? true : false
```
maximilian-gelbrecht, line 6: This will not work on GPU 

Could do a 

```Julia
is_increasing = all(diff(x) .> 0) 
``` 
instead

src/RingGrids/utility_functions.jl
```diff
@@ -1,43 +1,32 @@
...
-"""
-    true/false = isdecreasing(v::Vector)
-
+"""$(TYPEDSIGNATURES)
 Check whether elements of a vector `v` are strictly decreasing."""
-function isdecreasing(x::Vector)
+function isdecreasing(x::AbstractVector)
     is_decreasing = true
     for i in 2:length(x)
         is_decreasing &= x[i-1] > x[i] ? true : false
```
maximilian-gelbrecht, line 16: Same as above

src/RingGrids/grids/octahedral_clenshaw.jl
```diff
@@ -107,12 +135,47 @@ function Matrix!(   MGs::Tuple{AbstractMatrix{T}, OctahedralClenshawGrid}...;
             r += (sr-1)*nside                       # shift row into matrix quadrant
             c += (sc-1)*nside                       # shift column into matrix quadrant
 
-            for (Mi, Gi) in MGs                      # for every (matrix, grid) tuple
-                Mi[r, c] = convert(T, Gi[ij])         # convert data and copy over
+            for (Mi, Gi) in MGs                     # for every (matrix, grid) tuple
+                Mi[r, c] = convert(T, Gi[ij])       # convert data and copy over
```
maximilian-gelbrecht, line 186: This wouldn't work on GPU, and we'd have to think about another version. We'd have to check also other conversion routine. However, initially it's probably fine to define only those that matter for the dynamical core. If it is e.g. only used for the output, then it's fine to have it CPU only, as the output always has to be on CPU anyway.

src/RingGrids/grids/octahealpix.jl
```diff
@@ -80,47 +84,7 @@ function each_index_in_ring!(   rings::Vector{<:UnitRange{<:Integer}},
...
-    end
-    return colat
-end
-
-full_grid(::Type{<:OctaHEALPixGrid}) = FullOctaHEALPixGrid    # the full grid with same latitudes
-
-matrix_size(grid::OctaHEALPixGrid) = (2grid.nlat_half, 2grid.nlat_half)
-matrix_size(::Type{OctaHEALPixGrid}, nlat_half::Integer) = (2nlat_half, 2nlat_half)
+## CONVERSION
 Base.Matrix(G::OctaHEALPixGrid{T}; kwargs...) where T = Matrix!(zeros(T, matrix_size(G)...), G; kwargs...)
```
maximilian-gelbrecht, line 88: Needs to be adjusted: `zeros` is `Array`

milankl:
Broadcasting works now too

```julia
julia> G = rand(OctaHEALPixGrid{Float16},1,4)
4×4, 1-ring OctaHEALPixArray{Float16, 2, Matrix{Float16}}:
 0.063   0.6636   0.4683  0.188
 0.1118  0.04248  0.3584  0.9688
 0.2769  0.8354   0.3257  0.10547
 0.988   0.6245   0.545   0.5293

julia> H = rand(OctaHEALPixGrid{Float32},1,4)
4×4, 1-ring OctaHEALPixArray{Float32, 2, Matrix{Float32}}:
 0.495338  0.510038   0.579926  0.504566
 0.550702  0.761179   0.196999  0.333867
 0.328203  0.0446362  0.817386  0.791339
 0.303796  0.884892   0.414292  0.308456

julia> G + H
4×4, 1-ring OctaHEALPixArray{Float32, 2, Matrix{Float32}}:
 0.558326  1.17361   1.04819   0.692554
 0.662518  0.80366   0.555398  1.30262
 0.605059  0.880085  1.14307   0.896808
 1.29159   1.5094    0.959214  0.837753
```

src/RingGrids/general.jl
```diff
@@ -0,0 +1,324 @@
...
+get_npoints(G::Type{<:AbstractGridArray}, nlat_half::Integer, k::Integer...) = prod(k) * get_npoints2D(G, nlat_half)
+get_npoints2D(grid::Grid) where {Grid<:AbstractGridArray} = get_npoints2D(Grid, grid.nlat_half)
+
+# size of the matrix of the horizontal grid if representable as such (not all grids)
+matrix_size(grid::Grid) where {Grid<:AbstractGridArray} = matrix_size(Grid, grid.nlat_half)
+
+## INDEXING
+@inline Base.getindex(G::AbstractGridArray, ijk::Integer...) = getindex(G.data,ijk...)
+@inline Base.getindex(G::AbstractGridArray, r::AbstractRange, k::Integer...) = getindex(G.data, r, k...)
+@inline Base.getindex(G::AbstractGridArray, ij::Integer, k::AbstractRange) = getindex(G.data, ij, k)
```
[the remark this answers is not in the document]
milankl: Yes, but not needed see below.

src/RingGrids/general.jl
```diff
@@ -0,0 +1,324 @@
...
+# if no nlat_half provided calculate it
+function (::Type{Grid})(data::AbstractArray) where {Grid<:AbstractGridArray}
+    npoints2D = size(data, 1)
+    nlat_half = get_nlat_half(Grid, npoints2D)
+    return Grid(data, nlat_half)
+end
+
+for f in (:zeros, :ones, :rand, :randn)
+    @eval begin
+        # general version with ArrayType(zeros(...)) conversion
```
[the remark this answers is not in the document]
milankl: Only when allocating the prognostic and diagnostic variables, yes! Good call, but also sounds like something that we can swap out later if we need to?

src/RingGrids/general.jl
milankl: Yes, I remember I put it there deliberately for some performance reasons, but it doesn't seem to be needed anymore. Technically it's a callsite `@inline` which is also possible since some Julia version. I removed it now, the code runs at 2ns with or without

src/RingGrids/general.jl
milankl: Nah, when you reviewed this I just hadn't worked on the broadcasting yet, seems to work fine. I had the feeling that the `if isstructurepreserving(bc) || fzeropreserving(bc)` wasn't needed for RingGrids, so I skipped that. If you think it is, please elaborate! 💚

src/RingGrids/general.jl
milankl: I skipped the GPU broadcasting for now as I'm not sure it's actually needed? We won't be interactively allocating arrays there through broadcasting anyway maybe?

src/RingGrids/full_grids.jl
milankl: Yes, absolutely, docstrings will be added

src/RingGrids/grids/full_clenshaw.jl
milankl: Okay sounds good, worst case we can still change that later without many problems I believe as we mostly use `eachring(grid) = grid.rings` so there's room to do something else instead.

src/RingGrids/grids/new_grids.jl
milankl: Yeah, I just copied that over for now, it needs some reworking!

src/RingGrids/grids/octahedral_gaussian.jl
milankl: Now it does ;) hadn't tackled that yet when you reviewed

src/RingGrids/utility_functions.jl
milankl: Yeah, it's not even used at the moment. Don't remember where it was necessary, but that piece of code is gone.

src/RingGrids/grids/octahedral_clenshaw.jl
milankl: Yes the `Matrix!` conversion is some optional output that I used at some point to create these area-preserving projections

matrix_output

So I'd just not touch it for now, it's not high priority for GPU stuff.

src/RingGrids/general.jl
```diff
@@ -0,0 +1,324 @@
...
+@inline Base.getindex(G::AbstractGridArray, I::CartesianIndex) = getindex(G, Tuple(I)...)
+
+@inline function Base.getindex(
+    G::GridArray,
+    col::Colon,
+    k...,
+) where {GridArray<:AbstractGridArray{T, N, ArrayType}} where {T, N, ArrayType}
+    GridArray_ = nonparametric_type(GridArray)
+    return GridArray_{T, 1}(getindex(G.data, col, k...), G.nlat_half)
+end
```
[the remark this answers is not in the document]
maximilian-gelbrecht: Great!

src/RingGrids/general.jl
maximilian-gelbrecht: mh..., maybe not, but we should give it a proper test if typical broadcasting operations work on GPU (or with JLArrays as well) 

So something like 
```Julia 
A .+= b 
```

src/RingGrids/grids/octahedral_clenshaw.jl
maximilian-gelbrecht: Okay, agreed!

src/RingGrids/interpolation.jl
milankl: Yes, interpolation is currently only used for output to get from reduced grids to full grids or for the interactive visualisation with `plot`. The only place where interpolation takes place in the dynamical core is the particle advection and that's passive and switched off by default. So I'm tending towards postponing a GPU-ready version of the interpolation for when we actually need it.

maximilian-gelbrecht: Ok, agreed!

maximilian-gelbrecht:
I see tests is now checked, but I'd say we should add some tests with `JLArrays` to make sure everything works with other array types and GPU arrays as well.

milankl:
Currently the type of the data overwrites the parameters of a grid, i.e.
```julia
julia> OctaHEALPixArray(jl(ones(4,2)))
4×2, 1-ring OctaHEALPixArray{Float64, 2, JLArray{Float64, 2}}:
 1.0  1.0
 1.0  1.0
 1.0  1.0
 1.0  1.0

julia> OctaHEALPixArray{Float32, 2, Matrix{Float32}}(jl(ones(4,2)))
4×2, 1-ring OctaHEALPixArray{Float64, 2, JLArray{Float64, 2}}:
 1.0  1.0
 1.0  1.0
 1.0  1.0
 1.0  1.0
```
wheras julia arrays would include a conversion,
```julia
julia> Matrix{Float32}(zeros(Float64,2,2))
2×2 Matrix{Float32}:
 0.0  0.0
 0.0  0.0
```
should we do the same @maximilian-gelbrecht ?

milankl:
I keep having problems with this bit
```julia
function Base.similar(bc::Broadcasted{AbstractGridArrayStyle{N, Grid}}, ::Type{T}) where {N, Grid, T}
    # this escapes for Array and JLArray
    # if isstructurepreserving(bc) || fzeropreserving(bc) 
    #     return Grid(Array{T}(undef, size(bc)...))
    # end
    # return similar(convert(Broadcasted{DefaultArrayStyle{ndims(bc)}}, bc), T)

    # this doesn't escape for Array but for JLArray
    return Grid(Array{T}(undef, size(bc)...))
end
```

if I use the commented version (as adapted from #503) then both Array and JLArray escape (= grid + grid -> array not grid), the currently uncommented version at least works for Array
```julia
julia> G = rand(OctaHEALPixGrid{Float16}, 1, 2)
4×2, 1-ring OctaHEALPixArray{Float16, 2, Matrix{Float16}}:
 0.3022  0.1201
 0.7876  0.5757
 0.6616  0.4717
 0.439   0.1323

julia> G + G   # doesn't escape, good!
4×2, 1-ring OctaHEALPixArray{Float16, 2, Matrix{Float16}}:
 0.6045  0.2402
 1.575   1.151
 1.323   0.9434
 0.878   0.2646

julia> JL = adapt(JLArray, G)
4×2, 1-ring OctaHEALPixArray{Float16, 2, JLArray{Float16, 2}}:
 0.3022  0.1201
 0.7876  0.5757
 0.6616  0.4717
 0.439   0.1323

julia> JL + JL     # escapes, baaad!
4×2 JLArray{Float16, 2}:
 0.6045  0.2402
 1.575   1.151
 1.323   0.9434
 0.878   0.2646
```

maximilian-gelbrecht:
I'll have a look at the broadcast (and the rest of the updated PR) tomorrow. The `isstructurepreserving(bc) || fzeropreserving(bc)` is something that `Base.Broadcast` specifically defined for structured matrices like the LinearAlgebra.jl definition of lower triangular matrices. So, here we definitely don't need it.

maximilian-gelbrecht:
Ok, I already did the broadcast fix now. I didn't test it properly though. And there's one detail I'll comment on the code. For the rest I'll do a review tomorrow from the train and then I am on vacation.

src/RingGrids/general.jl
```diff
@@ -369,32 +369,29 @@ Base.BroadcastStyle(::Type{Grid}) where {Grid<:AbstractGridArray{T, N, ArrayType
...
+
+# ::Val{0} for broadcasting with 0-dimensional, ::Val{1} for broadcasting with vectors, etc
+AbstractGPUGridArrayStyle{N, ArrayType, Grid}(::Val{M}) where {N, ArrayType, Grid, M} = AbstractGPUGridArrayStyle{N, ArrayType, Grid}()
+
+function GPUArrays.backend(::Type{Grid}) where {Grid <: AbstractGridArray{T, N, ArrayType}} where {T, N, ArrayType <: GPUArrays.AbstractGPUArray}
     return GPUArrays.backend(ArrayType)
 end
 
+function Base.similar(bc::Broadcasted{AbstractGPUGridArrayStyle{N, ArrayType, Grid}}, ::Type{T}) where {N, ArrayType, Grid, T}
+    return Grid(T.(ArrayType(undef, size(bc)...)))
```
maximilian-gelbrecht, line 392: The `T.(ArrayType(undef, size(bc)...))` is not ideal of course. It's the same problem again with determining the non-parametric type. We can keep it if we don't care about the minor performance, or we define the `nonparametric_type` function for `CuArray` and `JLArray` as well (the latter only in the tests, to not have a dependency).

milankl: Ah yes because `ArrayType{T}` doesn't work because `ArrayType` is already parametric? I see the problem.

milankl:
> Ok, I already did the broadcast fix now. I didn't test it properly though. But your code from above works and `JL + JL` returns a `RingGrid`. For the rest I'll do a review tomorrow from the train and then I am on vacation.

Awesome!!! Thanks so much!! I see you have "simply" defined a new broadcasting style for GPUs, which makes sense, but I don't understand why that's not necessary for LowerTriangularArrays??

This is perfect timing then, because once we're happy with these pull requests I can
- restructure the variables
- accordingly restructure the dynamical core
- but keep most of the scalar indexing as is for now

and once you're back we can then discuss how to make the kernels then ready for GPU.

maximilian-gelbrecht:
With LowerTriangularArrays we have `+` and `*` explicitly defined and not via broadcast and all the other operation that we test are of the type `LT .= A .+ B` or something like that, so I guess the `similar` is never hit there. Maybe we could still add it, but we tested all situations that appear in the dynamical core and things like `+`,... etc work as well as mentioned, so we also don't really seem to need it.

src/RingGrids/general.jl
maximilian-gelbrecht: Yep

milankl: `nonparametric_type` is really something that Julia Base should have automatically for every parametric type.

src/RingGrids/RingGrids.jl
```diff
@@ -9,50 +9,78 @@ import Statistics: mean
 import FastGaussQuadrature
 import LinearAlgebra
 
+# GPU
+import Adapt
+import KernelAbstractions
```
maximilian-gelbrecht, line 14: KernelAbstractions is not used anywhere in RingGrids, isn't it?

src/RingGrids/general.jl
```diff
@@ -0,0 +1,398 @@
...
+horizontal_grid_type(grid::AbstractGridArray) = horizontal_grid_type(typeof(grid))
+
+## SIZE
+Base.length(G::AbstractGridArray) = length(G.data)  # total number of grid points
+Base.size(G::AbstractGridArray) = size(G.data)      # size of underlying data (horizontal is unravelled)
+
+"""$(TYPEDSIGNATURES) Size of underlying data array plus precomputed ring indices."""
+Base.sizeof(G::AbstractGridArray) = sizeof(G.data) + sizeof(G.rings)
+
+"""$(TYPEDSIGNATURES) Resolution paraemeters `nlat_half` of a `grid`.
```
maximilian-gelbrecht, line 43: typo

src/RingGrids/general.jl
```diff
@@ -0,0 +1,398 @@
...
+    return nonparametric_type(Grid)(similar_data, nlat_half)
+end
+
+# general version with ArrayType{T, N}(undef, ...) generator
+function (::Type{Grid})(
+    ::UndefInitializer,
+    nlat_half::Integer,
+    k::Integer...,
+) where {Grid<:AbstractGridArray{T, N, ArrayType}} where {T, N, ArrayType}
+    return Grid(ArrayType{T, N}(undef, get_npoints2D(Grid, nlat_half), k...), nlat_half)
```
maximilian-gelbrecht, line 196: I am not sure this works. 

`ArrayType` is already something like `Array{T,1}`, so you can't do `Array{T,1}{T,N}`. 
Is this function tested with JLArray?

Maybe we should really just define 

```Julia
nonparametric_type(::Array{T,N}) where {T,N} = Array
nonparametric_type(::CuArray{T,N}) where {T,N} = CuArray

# and in the tests, but not the model:
import SpeedyWeather.nonparametric_type
SpeedyWeather.nonparametric_type(::JLArray{T,N}) where {T,N} = JLArray
```
This could then also be used by the broadcast

src/RingGrids/quadrature_weights.jl
```diff
@@ -3,19 +3,26 @@
...
-    [nlon_healpix(nlat_half, j) for j in 1:nlat_half].*(2/npoints_healpix(nlat_half))
-octahealpix_weights(nlat_half::Integer) =
-    [nlon_octahealpix(nlat_half, j) for j in 1:nlat_half].*(2/npoints_octahealpix(nlat_half))
+function equal_area_weights(Grid::Type{<:AbstractGridArray}, nlat_half::Integer)
+    weights = zeros(nlat_half)
+    for j in 1:nlat_half
+        nlon = get_nlon_per_ring(Grid, nlat_half, j)
+        weights[j] = 2nlon/get_npoints2D(Grid, nlat_half)
+    end
+    return weights
```
maximilian-gelbrecht, line 18: I assume the quadrature weights need to go on the GPU as well when the Grid is on the GPU. But I would say that we can also adjust that when we tackle the transform submodule

src/RingGrids/RingGrids.jl
milankl: No, right, it was also added in `LowerTriangularMatrices` and I copied that over, will take it out.

src/RingGrids/general.jl
milankl: You don't know parämeters? They're like parameters but crääzy!

src/RingGrids/quadrature_weights.jl
milankl: Yes, they're always precomputed in stored in `SpectralTransform`, (and there's more vectors in there that need to be array-agnostic) so I'd suggest we don't change this here but in `SpeedyTransforms`

src/RingGrids/general.jl
maximilian-gelbrecht: Yes, `CuArray` is exported by CUDA.jl 

`JLArray` really has zero productive use. An extension would work as well, but they are really only used for tests, so both is fine. 

Yep, the parameters aren't necessary.

src/RingGrids/general.jl
milankl: Added with [fd6d9d8](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/520/commits/fd6d9d8fdf7065577be505ee729d1b3a0dc24093)

src/RingGrids/general.jl
milankl: `nonparametric_type` was added for `Array` and `CuArray` and as extension also for `JLArray`

milankl:
I hit a few scalar indexing errors that I didn't know how to fix generally via broadcasting so I've just defined those methods to propagate to `grid.data`
- `Base.fill!(G::AbstractGridArray, x) = fill!(G.data, x)`
- `Base.:(==)(G1::AbstractGridArray, G2::AbstractGridArray) = grids_match(G1, G2) && G1.data == G2.data`
- `Base.all(G::AbstractGridArray) = all(G.data)`

now one can do `grid == grid` or `grid .== grid` even with JLArray without scalar indexing. I don't know what's required to solve these more generally without a new method per function but that's all we need right now for testing

PR #531, opened 2024-05-03
Add explicit arithmetics again to LowerTriangularArrays

maximilian-gelbrecht:
See discussion in #503


test/lower_triangular_matrix.jl
```diff
@@ -597,7 +597,7 @@ end
                 L2 = deepcopy(L1) 
 
                 L2 ./= 5
-                @test L1/5 == L2
+                @test L1/5 ≈ L2
```
milankl, line 600: Given this, I'm wondering whether we should actually do `L/s = L*inv(s)` because the latter introduces two rounding errors. The latter is what an algorithm should implement for speed, but maybe we should move the method for exactly this reason here.

src/LowerTriangularMatrices/lower_triangular_matrix.jl
```diff
@@ -414,8 +414,41 @@ Base.similar(L::LowerTriangularArray{T,N,ArrayType}, ::Type{T}) where {T, N, Arr
 Base.similar(L::LowerTriangularArray{T}) where T = similar(L, T)
  
 # ARITHMETIC
+# While these would work just via broadcast (and the abstractarray interface), these explicit definitions are significantly faster
+# only mul/div with scalar and addition/subtraction, others are handled via the custom broadcast
+function Base.:(*)(L::LowerTriangularArray{T}, s::Number) where T
+    sT = convert(T, s)
```
milankl, line 420: I don't think that any of the `convert` are necessary because they should be handled upstream in `L.data * s`

src/LowerTriangularMatrices/lower_triangular_matrix.jl
```diff
@@ -414,8 +414,41 @@ Base.similar(L::LowerTriangularArray{T,N,ArrayType}, ::Type{T}) where {T, N, Arr
...
+    LowerTriangularArray(L.data .* sT, L.m, L.n)
+end
+
+Base.:(*)(s::Number, L::LowerTriangularArray) = L*s         # commutative
+Base.:(/)(L::LowerTriangularArray, s::Number) = L*inv(s)
+
+function Base.:(+)(
+    L1::LowerTriangularArray{T, N, ArrayType},
+    L2::LowerTriangularArray{S, N, ArrayType},
+) where {T, S, N, ArrayType}
```
milankl, line 430: As far as I see this this wouldn't cover the situation of `@. A = B + C` which is the non-allocating version of broadcasting that we do use inside the dynamical core, `A = B + C` isn't because of the allocation, so it's speed is much less relevant.

maximilian-gelbrecht: Good point, I can do a small benchmark test for that on Monday

maximilian-gelbrecht:
```Julia 
NF=Float64
idims=(5,)
L = randn(LowerTriangularArray{NF}, 10, 10, idims...)
L2 = randn(LowerTriangularArray{NF}, 10, 10, idims...)
A = zeros(LowerTriangularArray{NF}, 10, 10, idims...)

@benchmark L + L2
BenchmarkTools.Trial: 10000 samples with 272 evaluations.
 Range (min … max):  291.673 ns … 60.475 μs  ┊ GC (min … max):  0.00% … 98.71%
 Time  (median):     428.287 ns              ┊ GC (median):     0.00%
 Time  (mean ± σ):     1.107 μs ±  5.384 μs  ┊ GC (mean ± σ):  57.42% ± 11.60%

  █▃                                                           ▁
  ██▅▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▁▃▁▁▁▁▁▃▁▄▁▁▁▄▄▃▃▃▄▄▆▆ █
  292 ns        Histogram: log(frequency) by time      43.4 μs <

 Memory estimate: 2.34 KiB, allocs estimate: 2.

@benchmark L .+ L2
 Range (min … max):  2.332 μs …  1.535 ms  ┊ GC (min … max):  0.00% … 99.61%
 Time  (median):     2.515 μs              ┊ GC (median):     0.00%
 Time  (mean ± σ):   3.349 μs ± 30.675 μs  ┊ GC (mean ± σ):  20.21% ±  2.23%

  ▅▄▇██▇▄▃▂▁▁  ▁   ▁  ▁▁▂▁▂▂▂▂▁                              ▂
  ███████████▇▇██▇▆█▇██████████▇▅▅▄▅▃▅▃▅▄▅▄▅▄▆▆▅▃▄▅▅▆▅▆▄▂▄▃▄ █
  2.33 μs      Histogram: log(frequency) by time     4.94 μs <

 Memory estimate: 2.47 KiB, allocs estimate: 4.

@benchmark @. A = L + L2
 Range (min … max):  2.191 μs …  14.428 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     2.211 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.315 μs ± 464.178 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▄▃   ▂ ▂ ▁ ▁      ▁       ▂                                ▁
  ████▇████▅█▄█▅█▄▇▆▄█▄█▆▁▇▄▁█▆▃▁▁▃▁▁▃▃▁▁▁▁▁▁▄▄▁▁▃▁▆▆▅▅▆▄▁▅▅▆ █
  2.19 μs      Histogram: log(frequency) by time      4.38 μs <

 Memory estimate: 160 bytes, allocs estimate: 3.

@benchmark  LowerTriangularArray{NF, 3, Array{NF}}(undef, 10, 10, idims...)
 Range (min … max):  2.634 μs …  1.396 ms  ┊ GC (min … max):  0.00% … 99.29%
 Time  (median):     2.787 μs              ┊ GC (median):     0.00%
 Time  (mean ± σ):   3.673 μs ± 29.901 μs  ┊ GC (mean ± σ):  19.70% ±  2.43%

  ▆██▅▂▂▂▂   ▁  ▁  ▁                                         ▂
  █████████████▇█▇███▇▇▄▄▄▅▆▇▇▇▇▇▇▇▇▆▃▅▅▃▄▅▄▃▅▃▃▃▄▃▁▁▄▄▅▆▆▆▆ █
  2.63 μs      Histogram: log(frequency) by time     6.78 μs <

 Memory estimate: 3.34 KiB, allocs estimate: 25.

@benchmark Array{NF,3}(undef, 10, 10, idims...)
BenchmarkTools.Trial: 10000 samples with 199 evaluations.
 Range (min … max):  410.935 ns … 71.406 μs  ┊ GC (min … max):  0.00% … 98.80%
 Time  (median):     594.294 ns              ┊ GC (median):     0.00%
 Time  (mean ± σ):     1.524 μs ±  6.784 μs  ┊ GC (mean ± σ):  59.18% ± 13.03%

  █▁                                                           ▁
  ███▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▃▁▃▁▁▄▃▃▃▄▄▅▅▄▅▆▅▅▅▆ █
  411 ns        Histogram: log(frequency) by time      50.6 μs <

 Memory estimate: 4.06 KiB, allocs estimate: 1.

maximilian-gelbrecht: Of course we could just define a `add!(C, A, B)` function. Maybe that's just the easiest thing to do. 

One of the reasons why the broadcast is slower is that the undef initialiser is slower than for regular arrays. I can't immediately see why that's the case, because the `nonzeros` function that's called by it only takes 1.5 ns, so it's shouldn't matter.

milankl: > Of course we could just define a add!(C, A, B) function. Maybe that's just the easiest thing to do.

I wouldn't be opposed to that, but if it's just the undef initializer that might be easy to solve? I'll have a look! At the moment it looks like almost all time is spent in the undef initializer?

maximilian-gelbrecht: Initially I thought that, but I just tested it and in the `@. C = A + B` case, the undef initializer is not called. So, I assume it is the `copyto!(::LTA, ::Broadcasted)` function. Due to the way broadcast works, there are restrictions on it. The `Broadcasted` object needs to be called either with a linear index (over all dimensions/axes), or with an `CartesianIndex`. Currently we do the latter and also have to count up SH numbers for that. That might be just be enough to make it slower. On GPU it is even slower than that because we have to use the `k2ij` function. 

Again, we could get rid of this problem by not making LTA a subtype of `AbstractArray{T,N}`, but of `AbstractArray{T,N-1}` and pre-computing the `k2ij`. But to be honest, everything works the way we done it now, and for the few cases we use that broadcast in the dynamical core, we can also explicitly define those faster functions.

maximilian-gelbrecht: Maybe one GPU just pre-computing the `k2ij` in the current version (`<:AbstractArray{T,N}`) is already enough to make it equal or close in speed.

maximilian-gelbrecht: But again, previously we agreed that we don't care so much about the broadcasting speed. If we just use the explicit functions in the dynamical core, we would be fine.

milankl: How big are the `k2ij`? I thought naively it would need to be a `Int -> (Int, Int)` pair, but you said it was a `BitVector`?

maximilian-gelbrecht: I mean, ultimately, we can do both and make both work, I think. It's just additional work to now change everything to `AbstractArray{T,N-1}` that I am not sure we absolutely need to do.

milankl: `int -> (int, int)` looks like a killer no? At T31 and single precision we'd have

- 1 layer: 560*4 = 2.2KB data but 560*3*8 = 13.4KB for `k2ij` when using `Int64`, so 7x larger = +700%
- 8 layers: +75% larger still
- 8 layers but int32: +37.5% larger
- 8 layers but int16: +18.75% larger
- 2x8 layers and int16: still +9% larger

maximilian-gelbrecht: Ultimately, we just have to decide: 
Do we want to use the broadcast in the dynamical core: 
* Yes: Then we should either accerelate it, or change LTA to `AbstractArray{T,N-1}` 
* No: It's fine like it is, and we use those explicit arithmetics (and should add the `add!(C,A,B)` as well)

milankl: I agree, I think I already had compiled a list of the broadcastable operations that we'd want to define manually (and doing that might be in general smart so that we can change internals without changing the user-facing functionality)

milankl:
List of in-place operations we would want to define manually

- `A += B` here
https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/6860c05ce1fe3c15d1e1e506f2af28ec574c5746/src/dynamics/tendencies.jl#L243
and here
https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/6860c05ce1fe3c15d1e1e506f2af28ec574c5746/src/dynamics/tendencies.jl#L703

- `A = -A - B` here
https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/6860c05ce1fe3c15d1e1e506f2af28ec574c5746/src/dynamics/tendencies.jl#L279

- `A += c*B` here
https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/6860c05ce1fe3c15d1e1e506f2af28ec574c5746/src/dynamics/tendencies.jl#L733
and here
https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/6860c05ce1fe3c15d1e1e506f2af28ec574c5746/src/dynamics/geopotential.jl#L125

- `A .*= c` here (and other places in same file)
https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/6860c05ce1fe3c15d1e1e506f2af28ec574c5746/src/SpeedyTransforms/spectral_gradients.jl#L123

maximilian-gelbrecht:
I added the `add!` 

But, to be honest, the more I think about it, the less I like the non-performant broadcast, just because they are so many special cases that we'd have to cover to really have good performance. 

I don't have much time this week (public holiday in Germany and so on), but I consider working on it in some way next week. 

There are a few possibilities: 
* The kernel one: Do we actually want to do use a broadcast here in any way? Maybe we just put those operations in a KernelAbstractions kernel anyway, then we don't need to think about broadcasting at all  
* The pragmatic one: Just use broadcasts with `.data` in the dynamical core
* The fancy pragmatic one: Write a macro that does the thing above automatically (`@L. $L+=$A` -> `@. L.data += A.data`)
* Make the broadcast more performant: This would be ideal of course. I was thinking about if there's a way to write a more performant `copyto` function using linear indexing, but I don't really know yet if that's possible. 
* Making it `AbstractArray{T,N-1}` (and keep both indexing styles), solves the broadcasting issue, but makes all kind of operations with regular matrices/arrays quite messy, would need a lot of extra definitions/functions there. However, those are not used in the dynamical core, this is more an interfacing issue

milankl:
> Sorry for repeating this discussion again, but to be honest, the more I think about it, the less I like the non-performant broadcast, just because they are so many special cases that we'd have to cover to really have good performance everywhere.

I agree with you! But I'm actually in favour of this one

> Make it <:AbstractArray{T,N-1} (but keep both indexing styles), solves the broadcasting performance issue, but makes all kind of operations with regular N-dim matrices/arrays quite messy, would need a lot of extra definitions/functions there. However, those are not used in the dynamical core, this is more an interfacing issue. Probably it would be fine to drop all kind of arithmetics with regular N-dim arrays and just provide a constructor for conversion.

because then at least from an internal perspective it's all consistent and we're not abusing any subtyping (like a vector that's a matrix as we currently do). We'd then only need to define

- [l,m] indexing as additional method for convience
- and maybe some matrix-like printing for the matrix case to see the triangular structure, this might come in handy for testing too!

maximilian-gelbrecht:
Yeah, I think that's good as well. 

I have some time tomorrow and can start working on a draft. The basic change only takes little work, it's more about adjusting tests, some interfacing etc. 

But would you agree that we don't support for any kind of operations with regular `AbstractArray{T,N}` then, and just provide constructors/conversion `Array`/`full``... for it?

milankl:
> But would you agree that we don't support for any kind of operations with regular `AbstractArray{T,N}` then, and just provide constructors/conversion `Array`/`full``... for it?

Yeah I think that's fine. Operations between LTAs should be possible, but I don't see much of a new for `LTA + Array` or `LTA * Vector` for example.

Do you want to merge this pull request first though? Looks like it wouldn't hurt to have them?

maximilian-gelbrecht:
Yeah, we can merge this. The next PR will partially revert it again, but maybe I need more time than I think right now. So merging it now makes some sense and won't hurt.

PR #538, opened 2024-05-15
Scale orography and smooth fraction of resolved wavenumbers

milankl:
Default is no smoothing but with smoothing it would be T31, 0.05 => smoothing of highest 2 wavenumbers, T170 smoothing of highest 9 wavenumbers, etc.


milankl:
For T170 smoothes top to what's seen below

image

and for T31

image

so mostly takes out negative values and grid-scale crumbles

PR #543, opened 2024-05-18
Change LowerTriangularArrays to be subtype of AbstractArray{T,N-1}

maximilian-gelbrecht:
- [x] Basic tests for LTA passing
- [x] show LowerTriangularMatrix as Matrix but higher dims as Vector 
- [x] Check other occurrences of LTA in the rest of the code 
- [x] Clean up and benchmark arithmetics
- [x] Update plot! 
- [x] Update Doc
- [x] Update Broadcast test


milankl:
This is awesome, thanks Maximilian. I'll let you continute work on this, but feel free to let me know if I can contribute

maximilian-gelbrecht:
I'll look at it again later this week, and will look if I find any possible solutions

milankl:
Maybe some inlining missing? I can have a check too

maximilian-gelbrecht:
Had a look again, and I solved it. 

The problem was slow indexing with `CartesianIndex` 

So before we only had 
```Julia 
Base.@propagate_inbounds Base.getindex(L::LowerTriangularArray{T,N}, I::CartesianIndex{M}) where {T,N,M} = getindex(L, Tuple(I)...)
``` 

Now, I we also have: 
```Julia 
Base.@propagate_inbounds Base.getindex(L::LowerTriangularArray{T,N}, I::CartesianIndex{N}) where {T,N} = getindex(L.data, I)
``` 

This apparently made all the difference because internally broadcasting uses cartesian indexing.

maximilian-gelbrecht:
```Julia  
@benchmark L .+ L2
 Range (min … max):  2.318 μs …  1.983 ms  ┊ GC (min … max):  0.00% … 98.92%
 Time  (median):     4.224 μs              ┊ GC (median):     0.00%
 Time  (mean ± σ):   8.019 μs ± 47.512 μs  ┊ GC (mean ± σ):  36.22% ±  6.99%

  ▅▆▅▅▄▅▇█▇▅▃▂▂▂▂▁▁▁                  ▁▂▂▂▂▂▁                ▂
  ██████████████████████▇█▇▆▇▆▇▆▆▆▅▆▇████████████▇▅▆▇▆▆▅▅▅▅▅ █
  2.32 μs      Histogram: log(frequency) by time     17.5 μs <

 Memory estimate: 25.14 KiB, allocs estimate: 5.

julia> @benchmark L.data .+ L2.data
 Range (min … max):  2.236 μs …  1.869 ms  ┊ GC (min … max):  0.00% … 99.13%
 Time  (median):     4.154 μs              ┊ GC (median):     0.00%
 Time  (mean ± σ):   7.459 μs ± 47.129 μs  ┊ GC (mean ± σ):  38.76% ±  6.92%

  ▅▆▆▆▅▆▇██▆▄▃▃▂▂▂▂▁▁▁▁                                      ▂
  ████████████████████████▇▇█▇▆▇▇▆▆▅▆▆▇▇█▇▇█▇▇▇▆▇▆▆▆▆▄▄▄▅▃▃▅ █
  2.24 μs      Histogram: log(frequency) by time     17.4 μs <

 Memory estimate: 25.11 KiB, allocs estimate: 6.
```

maximilian-gelbrecht:
What's left to do basically is to adjust the usage of AssociatedLegendrePolynomials

milankl:
Last commit shows a `LowerTriangularMatrix` as
```julia
julia> L = rand(LowerTriangularMatrix,5,5)
15-element, 5x5 LowerTriangularMatrix{Float64}
 0.787422   0.0       0.0        0.0       0.0
 0.260061   0.819204  0.0        0.0       0.0
 0.973352   0.620726  0.161257   0.0       0.0
 0.0461552  0.685936  0.0117785  0.382289  0.0
 0.13645    0.501547  0.419401   0.309778  0.688196

julia> Float32.(L)
15-element, 5x5 LowerTriangularMatrix{Float32}
 0.787422   0.0       0.0        0.0       0.0
 0.260061   0.819204  0.0        0.0       0.0
 0.973352   0.620726  0.161257   0.0       0.0
 0.0461552  0.685936  0.0117785  0.382289  0.0
 0.13645    0.501547  0.419401   0.309778  0.688196

julia> L = rand(LowerTriangularMatrix,15,15)
120-element, 15x15 LowerTriangularMatrix{Float64}
 0.261788   0.0        0.0        0.0        …  0.0       0.0        0.0       0.0
 0.495482   0.117453   0.0        0.0           0.0       0.0        0.0       0.0
 0.0925019  0.796409   0.937247   0.0           0.0       0.0        0.0       0.0
 0.816407   0.409503   0.779616   0.182904      0.0       0.0        0.0       0.0
 0.915145   0.823384   0.270083   0.565954      0.0       0.0        0.0       0.0
 0.737812   0.795098   0.194549   0.698349   …  0.0       0.0        0.0       0.0
 0.275376   0.939357   0.535364   0.515381      0.0       0.0        0.0       0.0
 0.910015   0.22325    0.972193   0.0490044     0.0       0.0        0.0       0.0
 0.823201   0.128216   0.0735159  0.122388      0.0       0.0        0.0       0.0
 0.169003   0.862217   0.621427   0.695971      0.0       0.0        0.0       0.0
 0.659415   0.0279775  0.245747   0.247737   …  0.0       0.0        0.0       0.0
 0.596095   0.988664   0.441109   0.990378      0.177168  0.0        0.0       0.0
 0.28178    0.189885   0.609562   0.415566      0.155296  0.989642   0.0       0.0
 0.710654   0.138662   0.849563   0.444248      0.467292  0.0712397  0.722201  0.0
 0.104208   0.775013   0.634025   0.199434      0.955224  0.219269   0.102539  0.0512979
```

milankl:
But then for higher dimensions the matrices just flattened out into columns
```julia
julia> L = rand(LowerTriangularArray,5,5,5)
15×5 LowerTriangularArray{Float64, 2, Matrix{Float64}}:
 0.90549    0.179909   0.722785   0.656362   0.215722
 0.0663613  0.209663   0.393662   0.29946    0.142641
 0.31763    0.121395   0.941059   0.14699    0.738631
 0.132192   0.160128   0.157869   0.751886   0.0505202
 0.61786    0.7052     0.841462   0.915375   0.0175474
 0.802477   0.113889   0.344423   0.261357   0.484409
 0.0662565  0.933641   0.271906   0.0445899  0.085901
 0.0866564  0.0510311  0.115502   0.477471   0.195907
 0.995015   0.367898   0.14462    0.5566     0.626406
 0.613344   0.327654   0.292307   0.598366   0.416606
 0.863694   0.273074   0.0687791  0.901928   0.422603
 0.77978    0.471441   0.133331   0.0685466  0.529635
 0.617265   0.944824   0.989218   0.985348   0.371635
 0.40759    0.570559   0.0131059  0.229718   0.532602
 0.998965   0.150414   0.0708596  0.853031   0.308048
```

maximilian-gelbrecht:
The `CartesienIndex{0}` is a bit tricky to handle in an elegant way for us because we have the different indexing styles that are dispatched by checking the number of input arguments so `getindex(L::LTA{T,N}, I::Varargs{Any,N})` etc. 

I'll add this in a bit hacky way by explicitly defining a separate `getindex(L::LTA{T,1}, i::Integer, C::CartesianIndex{0})`. That should be enough.

milankl:
Yes agree. Leading integer argument then trailing cartesianindex is all we need! Thanks Max!!

maximilian-gelbrecht:
There's currently an error in the transforms here, that is a bit hard for me to see where it comes from. 

The "Transform: Orography (exact grids)" test fails, but only for the ClenshawGrids and only in the last row of the spectral coefficients.

maximilian-gelbrecht:
Made a brief check again. I don't really understand why this error is happening. 

@milankl do you have any idea? I didn't really change anything in SpectralTransforms that should be able to cause this, and that it is only appearing for ClenshawGrids and not GaussianGrid is a bit weird. 

```Julia 
using SpeedyWeather
trunc = 31
NF = Float32 
Grid = FullClenshawGrid

dealiasing = Grid in (FullGaussianGrid, OctahedralGaussianGrid) ? 2 : 3

SG = SpectralGrid(NF; trunc, Grid, dealiasing)
S = SpectralTransform(SG, recompute_legendre=false)
O = EarthOrography(SG, smoothing=true)
E = Earth(SG)
initialize!(O, E, S)

oro_grid = O.orography
oro_spec = spectral(oro_grid, S)

oro_grid1 = gridded(oro_spec, S)
oro_spec1 = spectral(oro_grid1, S)
oro_grid2 = gridded(oro_spec1, S)
oro_spec2 = spectral(oro_grid2, S)
```

```Julia 
 oro_spec1 - oro_spec2
560-element, 33x32 LowerTriangularMatrix{ComplexF32}
 -0.00012207+0.0im          0.0+0.0im                 0.0+0.0im         …          0.0+0.0im                 0.0+0.0im
 -3.05176f-5+0.0im  -1.52588f-5-1.52588f-5im          0.0+0.0im                    0.0+0.0im                 0.0+0.0im
  6.10352f-5+0.0im  -2.38419f-7+1.52588f-5im  -3.05176f-5+0.0im                    0.0+0.0im                 0.0+0.0im
 -3.05176f-5+0.0im   7.62939f-6+3.05176f-5im  -1.52588f-5-7.62939f-6im             0.0+0.0im                 0.0+0.0im
 -3.05176f-5+0.0im          0.0-1.52588f-5im          0.0-4.76837f-6im             0.0+0.0im                 0.0+0.0im
         0.0+0.0im  -1.90735f-6+2.28882f-5im  -1.14441f-5+0.0im         …          0.0+0.0im                 0.0+0.0im
 -3.05176f-5+0.0im   1.90735f-6-3.05176f-5im   7.62939f-6+3.8147f-6im              0.0+0.0im                 0.0+0.0im
            ⋮                                                           ⋱             ⋮              
  3.05176f-5+0.0im  -2.86102f-6-1.33514f-5im   1.90735f-6+3.8147f-6im              0.0+0.0im                 0.0+0.0im
  1.52588f-5+0.0im  -3.21865f-6+0.0im          9.53674f-6+2.38419f-6im             0.0+0.0im                 0.0+0.0im
 -9.53674f-6+0.0im   3.33786f-6-1.57356f-5im    3.8147f-6-1.90735f-6im             0.0+0.0im                 0.0+0.0im
  8.58307f-6+0.0im  -1.90735f-6-1.52588f-5im  -1.43051f-5-1.90735f-6im             0.0+0.0im                 0.0+0.0im
  3.05176f-5+0.0im  -1.90735f-6-3.8147f-6im    1.33514f-5-4.76837f-7im  …  -6.25849f-7+5.96046f-7im          0.0+0.0im
 -2.86102f-6+0.0im  -6.91414f-6-1.33514f-5im  -1.71661f-5-3.8147f-6im      -2.38419f-6-9.53674f-7im  -9.53674f-7+9.53674f-7im
    0.238285+0.0im    -0.194557-0.108487im      0.0348457+0.129144im               0.0+1.90735f-6im   9.53674f-7+0.0im
```

milankl:
Back in Oxford, I can have a look today. For now: I find it suspicious that `9.53674f-7im` shows up 4 times in the highest wavenumbers, maybe there's some indexing going wrong?

src/SpeedyTransforms/spectral_transform.jl
```diff
@@ -331,7 +330,7 @@ function gridded!(  map::AbstractGrid{NF},                      # gridded output
 
         # Recalculate or use precomputed Legendre polynomials Λ
         recompute_legendre && Legendre.unsafe_legendre!(Λw, Λ, lmax, mmax, Float64(cos_colat[j_north]))
-        Λj = recompute_legendre ? Λ : Λs[j_north]
+        Λj = recompute_legendre ? LowerTriangularMatrix(Λ) : Λs[j_north]
```
milankl, line 333: I believe this allocates for a T31 transform 24 T31 matrices -- not ideal!

src/SpeedyTransforms/spectral_transform.jl
```diff
@@ -452,7 +448,7 @@ function spectral!( alms::LowerTriangularMatrix{Complex{NF}},   # output: spectr
         # LEGENDRE TRANSFORM in meridional direction
         # Recalculate or use precomputed Legendre polynomials Λ
         recompute_legendre && Legendre.unsafe_legendre!(Λw, Λ, lmax, mmax, Float64(cos_colat[j_north]))
-        Λj = recompute_legendre ? Λ : Λs[j_north]
+        Λj = recompute_legendre ? LowerTriangularMatrix(Λ) : Λs[j_north]
```
milankl, line 451: same here

src/SpeedyTransforms/spectral_transform.jl
```diff
@@ -513,12 +509,28 @@ function gridded(   alms::AbstractMatrix{T};            # spectral coefficients
...
+    return gridded(alms, S; kwargs...)
+end
+
 """
 $(TYPEDSIGNATURES)
 Spectral transform (spectral to grid space) from spherical coefficients `alms` to a newly allocated gridded
 field `map` with precalculated properties based on the SpectralTransform struct `S`. `alms` is converted to
 a `LowerTriangularMatrix` to execute the in-place `gridded!`."""
-function gridded(   alms::AbstractMatrix,       # spectral coefficients
+function gridded(   alms::Union{AbstractMatrix, LowerTriangularMatrix},       # spectral coefficients
```
milankl, line 533: I think I want to disable the use of `gridded(M::Matrix)` and `spectral(M::Matrix)`, people should be explicit that their data represents spectral/gridded data and do `gridded(LowerTriangularMatrix(M))` or `spectral(FullGaussianGrid(M))` manually...

test/lower_triangular_matrix.jl
```diff
@@ -597,50 +604,42 @@ end
                 L2 = deepcopy(L1) 
 
                 L2 ./= 5
-                @test L1/5 == L2
+                @test (L1.data ./ 5) == L2.data
```
milankl, line 607: This shouldn't be necessary?

test/lower_triangular_matrix.jl
```diff
@@ -597,50 +604,42 @@ end
...
 
                 L1 = adapt(ArrayType, randn(LowerTriangularArray{NF}, 10, 10, idims...))
                 L2 = deepcopy(L1)
 
                 L2 .^= 2
-                @test L1.^2 == L2
+                @test L1.data.^2 == L2.data
```
milankl, line 613: Same here and below

src/SpeedyTransforms/spectral_transform.jl
maximilian-gelbrecht: Yeah, absolutely, not ideal in case `recompute_legendre == true`. 

I didn't go very deep into AssociatedLegendrePolynomials.jl, but it does repeatably use `ndims` of the array it's given, even the actual compute routines. So, we can use the `LowerTriangularMatrix` there completely easily. 

I didn't test what happens if you redefine `ndims`, but I expect this to be cause some troubles somewhere else. 

The first quite "hacky" solution that comes to my mind to solve it would be to define another type that is just used for the transforms: 

```Julia
struct LegendePolArray{T,N,ArrayType} <: AbstractArray{T,N}
   data::LowerTriangularArray{T,N-1,ArrayType}
end 

getindex(L:: LegendePolArray, I...) = getindex(L.data, I...)
setindex!(L:: LegendePolArray, x, I...) = setindex!(L.data, x, I...)
```
Broadcast and some other operations wouldn't work for it, but as far as I can see it, everything in `SpectralTransforms` would. 

But yeah, that is a bit hacky as a solution to that problem.

milankl:
@maximilian-gelbrecht I'm making progress on the rewrite of the dynamical core in #525 too to inform some of the decisions here, I think there's really only two things I've been adding/tweaking

1. Vertical iterator `eachmatrix` and size match checker `lowertriangular_match`

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/b8737f7cf95507a0fdab517570bfeea2c6abc9e9/src/LowerTriangularMatrices/lower_triangular_matrix.jl#L282-L285

Something like `eachmatrix` should automatically check compatible sizes, like `eachindex(a,b)` does too, avoids a lot of manual `@boundscheck size(a) == size(b)`` stuff which needed tweaking anyway because (5,5) and (5,5,1) should match (pres and velocity for example). I've implemented something similar for `eachgrid` for the ring grids.

I think we should rename `lowertriangular_match` in `ismatching` because we need conceptually the same thing for the ring grids and the spectral transforms... `ismatching` would then just take many combinations of lower triangular matrices, or ring grids with transforms and tell you whether they "match". This function then gets a `horizontal_only` keyword so that `ismatching(A, B, horizontal_only=true)` would return true for something like `(5, 5)` and `(5, 5, 2)` (e.g. the former is constant across the vertical)

2. `matrix_size` with OneBased, ZeroBased

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/b8737f7cf95507a0fdab517570bfeea2c6abc9e9/src/LowerTriangularMatrices/lower_triangular_matrix.jl#L48-L59

The spectral gradients and other parts of the dynamical core are somewhat inconsistent on whether 1-based or 0-based indexing is used (and there are places where it makes sense to mix), to make this clearer I'm suggesting these abstract types and then

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/b8737f7cf95507a0fdab517570bfeea2c6abc9e9/src/LowerTriangularMatrices/lower_triangular_matrix.jl#L68-L70

so that one can neatly write

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/b8737f7cf95507a0fdab517570bfeea2c6abc9e9/src/SpeedyTransforms/spectral_gradients.jl#L64-L65

maximilian-gelbrecht:
I did try to do it like that initially but this doesn't work with the dispatching rules apparently. You'll immediately get warnings that the function definitions are being overwritten

src/SpeedyTransforms/spectral_transform.jl
maximilian-gelbrecht: I implemented it like this now

maximilian-gelbrecht:
@milankl I tried to look into the Crenshaw grid error again, but couldn't fix it. 

I can't see where I changed something that would even affect it to be honest. Could you do a full code review and look if you  can find a part of code change that could affect it?

milankl:
> I did try to do it like that initially but this doesn't work with the dispatching rules apparently. You'll immediately get warnings that the function definitions are being overwritten

Right, because multiple dispatch doesn't apply to keyword arguments, what about
```julia
julia> struct S end

julia> Base.size(s::S, as::Type{<:Matrix}) = "matrix"

julia> Base.size(s::S, as::Type{<:Vector}) = "vector"

julia> Base.size(s::S; as=Vector) = size(s, as)

julia> size(S())
"vector"

julia> size(S(), as=Matrix)
"matrix"

julia> size(S(), as=Vector)
"vector"
```
then? make `as` the second argument and define a method that turns keyword `as` into the 2nd argument?

milankl:
The transform tests that are currently failing do have some differences in the orography field to begin with, left this branch (not so smooth), right is main (smoother). That could already lead to higher errors, maybe in relation to the higher resolution in these tests with the clenshaw grids because they use. I'm investigating!

image

milankl:
@maximilian-gelbrecht okay I got it, there was a `size(geopot_surf, 1)` (=560) call in the smoothing of the orography (which is used as default field to test the transforms that is now corrected to `size(geopot_surf, 1, as=Matrix)` (=33) the former effectively disabled the spectral smoothing as it smoothed relative to wavenumber 560 not 33 ...

milankl:
Also you think when we allow getindex with `[:, 1]` we should also allow for `[1,:]` ? Only the former currentlty works, the latter fails
```julia
julia> L = ones(LowerTriangularMatrix, 5, 5)
15-element, 5x5 LowerTriangularMatrix{Float64}
 1.0  0.0  0.0  0.0  0.0
 1.0  1.0  0.0  0.0  0.0
 1.0  1.0  1.0  0.0  0.0
 1.0  1.0  1.0  1.0  0.0
 1.0  1.0  1.0  1.0  1.0

julia> L[:, 1]
5-element Vector{Float64}:
 1.0
 1.0
 1.0
 1.0
 1.0

julia> L[1, :]
ERROR: MethodError: no method matching isless(::Int64, ::Colon)

Closest candidates are:
  isless(::Missing, ::Any)
   @ Base missing.jl:87
  isless(::Any, ::Missing)
   @ Base missing.jl:88
  isless(::Real, ::Union{StatsBase.PValue, StatsBase.TestStat})
   @ StatsBase ~/.julia/packages/StatsBase/ebrT3/src/statmodels.jl:91
  ...

Stacktrace:
 [1] <(x::Int64, y::Function)
   @ Base ./operators.jl:352
 [2] getindex(::LowerTriangularMatrix{Float64}, ::Int64, ::Function)
   @ Main.SpeedyWeather.LowerTriangularMatrices ~/git/SpeedyWeather.jl/src/LowerTriangularMatrices/lower_triangular_matrix.jl:185
 [3] top-level scope
   @ REPL[72]:1
```

milankl:
Maybe we should fix this one too?
```julia
julia> size(L, as=Matrix)
(5, 5)

julia> L[:, 5]
5-element Vector{Float64}:
 0.0
 0.0
 0.0
 0.0
 1.0

julia> L[:, end]
5-element Vector{Float64}:
 1.0
 1.0
 1.0
 1.0
 1.0
```

maximilian-gelbrecht:
> @maximilian-gelbrecht okay I got it, there was a `size(geopot_surf, 1)` (=560) call in the smoothing of the orography (which is used as default field to test the transforms that is now corrected to `size(geopot_surf, 1, as=Matrix)` (=33) the former effectively disabled the spectral smoothing as it smoothed relative to wavenumber 560 not 33 ...

Oh great, I tried to look for every occasion of `size(::LTA)` but this one I overlooked.

maximilian-gelbrecht:
> Maybe we should fix this one too?
> 
> ```julia
> julia> size(L, as=Matrix)
> (5, 5)
> 
> julia> L[:, 5]
> 5-element Vector{Float64}:
>  0.0
>  0.0
>  0.0
>  0.0
>  1.0
> 
> julia> L[:, end]
> 5-element Vector{Float64}:
>  1.0
>  1.0
>  1.0
>  1.0
>  1.0
> ```

Because of the way `end` works, this might be hard to fix. But I'll have a look into it.

maximilian-gelbrecht:
I could take over here again and finish the PR until Monday or Tuesday if you want (and there are no unforeseen other troubles)

milankl:
> I could take over here again and finish the PR until Monday or Tuesday if you want (and there are no unforeseen other troubles)

I'll try to get this merged so that I can continue working on #525 and hopefully merge this before JuliaCon! I feel the remaining points here can be in their own PR as it's mostly adding user-facing functionality

PR #554, opened 2024-06-12
Fix docs badges

navidcy:


milankl:
But latest release is stable and main is dev?

milankl (changes requested):
Before was correct?

navidcy:
Damn! I got confused!!

PR #555, opened 2024-06-13
All parameterizations at prev timestep, flux limiters removed

milankl:
(hopefully) fixes #553

PR #556, opened 2024-06-13
Correct vertical advection of humidity

simone-silvestri:
I don't know why, at the moment, humidity is reconstructed differently from temperature and velocities when it comes to vertical advection.

This should not be the case (i.e., temperature and humidity should follow the same transport equation).
I suspect this was just a mishap given that humidity is separated by the other variables because it is active only in a wet core. 
This PR corrects this "problem". 
I will try some tests to make sure everything is correct before merging

### DIFF
--- a/src/dynamics/vertical_advection.jl
+++ b/src/dynamics/vertical_advection.jl
@@ -67,8 +67,8 @@ function vertical_advection!(   layer::DiagnosticVariablesLayer,
 
     if wet_core
         ξ_tend = getproperty(layer.tendencies, :humid_tend_grid)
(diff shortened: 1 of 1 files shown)


milankl:
Thanks Simone, this is only relevant for WENO advection though right? Because all others use the current time step anyway? Asking because with #555 I've moved the evaluating of *all* parameterizations to the previous instead of the current timestep which seems to help massively with numerical stability. It also largely makes all flux limiters we previously had redundant. Just checking in relation to the vertical advection whether it would also be important to evaluate more at the previous rather than the current time step?

PR #559, opened 2024-07-06
Create CHANGELOG.md

milankl:
Using https://github.com/dangoslen/changelog-enforcer


milankl:
@maximilian-gelbrecht I copied over this workflow setup from how Makie does it. I thought it might be a good practice to write a few words summary after every PR into a changelog file? Or do you think that would be covered by the "closed pull requests" summary for releases?

maximilian-gelbrecht:
I like it. Good to get an overview of recent changes

milankl:
Not sure whether the workflow should already trigger here?

milankl:
> Not sure whether the workflow should already trigger here?

Haha, I forgot the `.yml` ending, now it works.

Issue #558, opened 2024-07-05
Additional LowerTriangularArray functionality


milankl:
Spillover functionality from #543 that we could consider implementing or, alternatively, this is an issue to document the stuff that doesn't work but that user might expect to work

1. [Indexing with trailing colon `:` ](https://github.com/SpeedyWeather/SpeedyWeather.jl/issues/543#issuecomment-2209436927_)

```julia
julia> L[1, :]
ERROR: MethodError: no method matching isless(::Int64, ::Colon)
```

2. [Indexing with `end`](https://github.com/SpeedyWeather/SpeedyWeather.jl/issues/543#issuecomment-2209438311_)

```julia
julia> L[:, end]
3-element Vector{Float64}:
 1.0
 1.0
 1.0
```
should be `0.0, 0.0, ..., 1.0` (= zeros for the upper triangle)

3. Broadcasting of vector with LowerTriangularMatrix

```julia
k = 1:32
A = randn(LowerTriangularMatrix{Complex{Float32}}, 32, 32)
A .*= k.^-2
```

doesn't work because `A` and `k` are technically both vectors of different sizes (560 for `A` vs 32 for `k`)

milankl:
4. `LowerTriangularMatrix` as `ones`, `zeros`, `rand`, `randn` generator for any size

Currently this fails
```julia
julia> L = zeros(LowerTriangularMatrix{ComplexF32}, 5, 5, 1);
ERROR: MethodError: no method matching Vector{ComplexF32}(::Matrix{ComplexF32})
```
Although `zeros(LowerTriangularMatrix{ComplexF32}, 5, 5)` works fine. One needs to use `Array` for any higher dimensions, which I find confusing.

milankl:
5. `.=` between Array and LowerTriangularArray

```julia
julia> L = zeros(LowerTriangularArray{ComplexF32}, 5, 5)
15-element, 5x5 LowerTriangularMatrix{ComplexF32}
 0.0+0.0im  0.0+0.0im  0.0+0.0im  0.0+0.0im  0.0+0.0im
 0.0+0.0im  0.0+0.0im  0.0+0.0im  0.0+0.0im  0.0+0.0im
 0.0+0.0im  0.0+0.0im  0.0+0.0im  0.0+0.0im  0.0+0.0im
 0.0+0.0im  0.0+0.0im  0.0+0.0im  0.0+0.0im  0.0+0.0im
[...]

PR #560, opened 2024-07-08
LTA: Copyto! function for Vectors

maximilian-gelbrecht:
Adresses #558 point 5


maximilian-gelbrecht:
Oops, I think I hadn't had my main updated? I am a bit confused why this includes the other commits. I am redoing this PR

PR #563, opened 2024-07-12
Set! function for PrognosticVariables and ICs for new structure

maximilian-gelbrecht:
This is currently pointing towards merging into `mk/newvariables` which the branch is also based on

Currently the `set_vordiv!` test fails, I have to look into that and also add a few more test for setting it functions to really cover all cases

(I am off work Monday/Tuesday, don’t expect a quick reply form me)


maximilian-gelbrecht:
@milankl As I am struggling a bit to get the vor/div part correct, can you just confirm that: 

`u::Grid`/`v::Grid` -> `scale_coslat⁻¹` -> `transform` -> `curl`/`divergence` is the right order/procedure? 

And that this should equal `UV_from_vordiv!` -> `transform` -> `scale_coslat⁻¹` -> `u::Grid`/`v::Grid` ?

milankl:
Yes, see the `transform!` method at the end of `src/dynamics/tendencies.jl`. Thanks Max from Belgium, went cycling for 3 days on a Brompton 😅

maximilian-gelbrecht:
Nice, I've also just been on a bike trip from Thüringer Wald back to Berlin. 

I also thought that this is the right way to do it with routines, but I am confused why this test: 

```Julia 
using SpeedyWeather

N_lev = 8 
N_trunc = 31
spectral_grid = SpectralGrid(trunc=N_trunc, nlev=N_lev)
model = PrimitiveWetModel(; spectral_grid)   # construct model

A1 = rand(spectral_grid.Grid{Float32}, spectral_grid.nlat_half, N_lev)   
A2 = rand(spectral_grid.Grid{Float32}, spectral_grid.nlat_half, N_lev)   

RingGrids.scale_coslat⁻¹!(A1)
RingGrids.scale_coslat⁻¹!(A2)

A1_spec = transform(A1, model.spectral_transform)
A2_spec = transform(A2, model.spectral_transform)

vor = similar(A1_spec)
div = similar(A1_spec)

curl!(vor, A1_spec, A2_spec, model.spectral_transform)
divergence!(div, A1_spec, A2_spec, model.spectral_transform)

u2_spec = similar(A1_spec)
v2_spec = similar(A1_spec)

SpeedyWeather.SpeedyTransforms.UV_from_vordiv!(u2_spec, v2_spec, vor, div, model.spectral_transform)

u2 = transform(u2_spec, model.spectral_transform)
v2 = transform(v2_spec, model.spectral_transform)

RingGrids.scale_coslat⁻¹!(u2)
RingGrids.scale_coslat⁻¹!(v2)

u2 ≈ A1 # false 
v2 ≈ A2 # false
```
yields wrong results. Not sure what I missed?

milankl:
The error tolerance should be set higher. The transforms introduce an error that is usually a multiple of eps. Maybe even sqrt(eps). How did I write the respective test in #525? Error tolerances are something that I've been constantly battling as I never know where the bounds should be and whether there is a strong bound in the first place. How large are the errors currently?

maximilian-gelbrecht:
It's not a matter of missing tolerances. The fields are really completely different. 

The test suite includes a test `"D, ζ ::LTA -> u, v -> D, ζ ::LTA"` (using the `transform!(::DiagnosticVariables)` so it's slightly different, but overall the approach is similar. But I still feel that I might be overlooking something.

maximilian-gelbrecht:
Mh... I had another look at it just now, and it's not clear to me. I made sure that I compute the vorticity and divergence correctly and verified that with the other routine (see updated code about). But I can't recover the same `u_grid`,`v_grid` via `UV_from_vordiv!`. 

For the `set!` function itself this doesn't really matter, I could add the possibility to set vorticity and divergence from u, v without it, but it's still weird that I can't get this test to pass / recover the same u, v field via  `UV_from_vordiv!`

milankl:
Thanks for checking, can you set the respective tests to broken for now? But I can also have a look tomorrow!

maximilian-gelbrecht:
Outside of the issue regarding `UV_from_vordiv!` I see this as more or less ready. I started converting some of the initial conditions as well. I don't think it's really that practical for all initial conditions though. 

Maybe one could also more generally allow a function `(m::ModelSetup) -> Grid / LTA` to be handed over to `set!`. This would allow the direct use of e.g. the geometry struct and the transforms etc. Maybe the geometry alone would already be sufficient as well.

milankl:
> Maybe one could also more generally allow a function `(m::ModelSetup) -> Grid / LTA` to be handed over to `set!`. This would allow the direct use of e.g. the geometry struct and the transforms etc. Maybe the geometry alone would already be sufficient as well.

With `set!(sim::AbstractSimulation, ...)` the `sim.model` would be available no?

maximilian-gelbrecht:
Yes and no. 

Inside of the `set!` functions here as well. But as suggested by you, I also implemented the possibility to just hand over a function `(lat, lon, sigma) -> val` to set the variable. And I think in practice, it's more handy if this function can access all of this information easily. 

Take for example the pressure initial condition: 
https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/4fd8c045fbc493f94b70f121cabf8f8fc3c15f66/src/dynamics/initial_conditions.jl#L453-487 

It's a simple function of the orography but implementing it as function `(lat, lon, sigma) -> val` is a bit wonky. Just going `(g::Geometry) -> AbstractGrid / AbstractLTA` would make it easier to program a bit mire complex initial conditions I think.

maximilian-gelbrecht:
Had a another small look at this and updated the code example above with the third way of computing it by manually doing ∇⁻² -> ∇, and I got yet another third completely different field in the end. Not sure, I made a mistake somewhere. 

We should definitely add a unit test and some documentation on this once we resolve it, as it's really unclear to me what's wrong here.

maximilian-gelbrecht:
Except for the aforementioned failing test this is ready, and also adjusts the initial conditions to the new variable structure

milankl:
Cool I'll have a look!

milankl (approved):

milankl:
Sorry to hear, hope you're better soon! I think I need to merge this as the CI isn't running anymore? And then fix any fails there?

maximilian-gelbrecht:
Yes, feel free to merge it, if it's better to work on it in the big PR. 

There are still the broken tests for checking that $$u$$ and $$v$$ are set correctly with `UV_from_vordiv!`. See also the script I posted above. Probably I did something wrong there, but I am not quite sure what, because the steps should be correct, and I double checked it.

PR #569, opened 2024-08-04
Fixed floating point exception in adiabatic conversion `initialize!`

seleneonowe:
## Description
I was running some simulations where I saw the NaR warning during runtime. To debug exactly where this first appears, I switched on trapping for floating point exceptions.

In this calculation in `adiabatic_conversion.jl`:
```julia
    σ_lnp_B .= 1 .- σ_levels_half[1:end-1]./σ_levels_thick .*
                    log.(σ_levels_half[2:end]./σ_levels_half[1:end-1])
    σ_lnp_B[1] = σ_levels_half[1] <= 0 ? log(2) : σ_lnp_B[1]    # set α₁ = log(2), eq. 3.19
```
if `σ_levels_half[1]` is less than or equal to zero, then we would have a division by zero, or invalid arguments to `log`.

This doesn't normally cause any bugs, because this simply returns `NaN` in the first element and in any of those cases the next line immediately throws away the `NaN` and swaps it for `log(2)`.

However, if we are trapping all floating point exceptions for debug purposes, the first expression would always raise e.g. `FE_DIVBYZERO` where we divide by zero, and `FE_INVALID` where we take the logarithm of a negative number; even though any problems caused by these exceptions are already being treated in the next line.

This means floating point exceptions caused by genuine bugs in model initialization that happen after `AdiabaticConversion` is initialized will be masked by this benign error. It is better if these lines are reordered so that no invalid operations are intentionally performed.

## To reproduce

In the REPL, run the following :

```julia
using SpeedyWeather
```

```julia
begin
    if Sys.ARCH == :x86_64
        const FE_INVALID  = 0x1
        const FE_DIVBYZERO  = 0x4
        const FE_OVERFLOW   = 0x8
        const FE_UNDERFLOW  = 0x10
        const FE_INEXACT  = 0x20
    elseif Sys.ARCH == :aarch64
        const FE_INVALID  = 0x1
        const FE_DIVBYZERO  = 0x2
        const FE_OVERFLOW   = 0x4
        const FE_UNDERFLOW  = 0x8
        const FE_INEXACT  = 0x10
     else
        # let me know if you use some other architecture.
        error("Unsupported architecture: $(Sys.ARCH)")
end
    
    fpexceptions() = ccall(:fegetexcept, Cint, ())

    function setfpexceptions(f, modes...)
        mode = foldl(|, modes)
        prev = ccall(:feenableexcept, Cint, (Cint,), mode)
        try
            f()
        finally
            ccall(:fedisableexcept, Cint, (Cint,), mode & ~prev)
        end
    end
end
```

```julia
spectral_grid = SpectralGrid(trunc=41, nlev=8)
```

```julia
model = PrimitiveWetModel(; spectral_grid)
```

```julia
setfpexceptions(FE_DIVBYZERO, FE_INVALID) do
end
```
This throws an error with a stacktrace pointing to the changed lines.
Expected behaviour would be `FE_DIVBYZERO` and `FE_INVALID` never being triggered intentionally where possible (otherwise it becomes very hard to debug occurrences of this happening unintentionally).

## Fix

Reorder the calculation so that rather than replacing the `NaN` after it is generated, it is never generated in the first place.


milankl:
Hi @seleneonowe, sorry for late response, I've been travelling/vacationing mostly in August, back at the desk now. Thanks for your pull request, this is very much appreciated. Indeed, this code was initially written to have a NaN intermediately which is then supposed to be replaced with a finite value. Very happy to change that to what you suggest!

src/dynamics/adiabatic_conversion.jl
```diff
@@ -26,9 +26,13 @@ function initialize!(
     σ_lnp_A[1] = 0  # the corresponding sum is 1:k-1 so 0 to replace log(0) from above
     
     # precompute the αₖ = 1 - p_k-1/2/Δpₖ*log(p_k+1/2/p_k-1/2) term in σ coordinates
-    σ_lnp_B .= 1 .- σ_levels_half[1:end-1]./σ_levels_thick .*
-                    log.(σ_levels_half[2:end]./σ_levels_half[1:end-1])
-    σ_lnp_B[1] = σ_levels_half[1] <= 0 ? log(2) : σ_lnp_B[1]    # set α₁ = log(2), eq. 3.19
+    if σ_levels_half[1] <= 0
+        σ_lnp_B[1] = log(2)   # set α₁ = log(2), eq. 3.19
+    else
+        σ_lnp_B[1] *= 1 - σ_levels_half[1] / σ_levels_thick[1] * log(σ_levels_half[2] / σ_levels_half[1])
```
milankl, line 32: What's the `*=` for? I believe this array is intialized with zeros, so it should be set with `=`?

milankl:
This should be addressed in a follow-up to #1137 when also the calculation of the adiabatic term in sigma-pressure coordinates is overhauled

Issue #570, opened 2024-08-14
RingGrids: N-dim Matrix/Array constructors 


maximilian-gelbrecht:
While revising the `OutputWriter` (#571), I noticed some inconsistencies with how the `RingGrids` can be initialised from data. 

So in 2D everything is as it is before: 
```Julia 
map = randn(Float32, 8, 4)
FullGaussianGrid(map) # -> 32-element, 4-ring FullGaussianGrid{Float64}
```

But when we add more dimensions to the data, the following happens: 
```Julia 
map = rand(Float32, 8, 4, 2) # in our logic in the other parts of the code, these are 2 (8x4) maps 
FullGaussianGrid(map) # -> 8×4×2, 2-ring FullGaussianArray{Float64, 3, Array{Float64, 3}}:
```
So this returns a 4D `8×4×2, 2-ring FullGaussianArray{Float64, 3, Array{Float64, 3}}:`, even though I would expect it to just return a `32x2 - 4-ring FullGaussianGrid{Float64}`. At least in my opinion, this would be consistent with how we handle the multidimensional data in the other functions. 

I don't directly see another way to generate a `RingGrid` from a full array currently. 

Also, the documentation should probably be extended to include a section for N-dim RingGrids.

milankl:
Thinking about this, we should make the second version (horizontal dimensions flattened) the default for consistency with other grids. But yes, it would be great to have an easy constructor for the full grids that also allows for a non-flattened input.

I see your desire to have a new function `FullGrid_from_Array` but that's not really clear whether it's a flattened or matrix representation for the horizontal dimensions no? For me this sounds like an option rather than a new function, you think we could implement this using a keyword argument, something like

```julia
FullGaussianGrid(map, horizontal_as=Matrix)
```

Or maybe `horizontal_input_as=Matrix` vs `Vector`? similar to how we use `as=Matrix` for `size(::LowerTriangularMatrices)`? Maybe one would understand this keyword more as affecting the full Gaussian grid rather than the interpretation of the input...

maximilian-gelbrecht:
Yeah, that would work as well. I can change my PR like that this afternoon

Issue #522, opened 2024-04-16
Modularise netCDF output


milankl:
Motivated by discussion in #507 we need to refactor our `OutputWriter`. Something like this should be possible

```julia
# initialise output writer generally but not variable specific
output = OutputWriter(spectral_grid, schedule = Schedule(every = Hour(6)), output_Grid = ...)

# add some variables that should be output with some variable-specific options
add!(output, StreamfunctionOutput(spectral_grid), keepbits=15, compression_level = 2)

# also adds default variables to output, u, v, vor, ...
model = PrimitiveWetModel(;spectral_grid, output)

# remove or add other variables to output
remove!(output, DivergenceOutput)
add!(output, ConvectivePrecipitationOutput(spectral_grid))
```

so that we have for every variable that should be outputted something like

```julia
struct TemperatureOutput{NF} <: AbstractOutputVariable
    NF::DataType = Float32
    dimensions = [true, true, true, true]    # x, y, z, t
    long_name::String = "Temperature"
    units::String = "degC"
    ...
end

output!(::OutputWriter, ::StreamfunctionOutput, ::PrognosticVariables, ::DiagnosticVariables, ::ModelSetup)
    # define here how to take variables from progn, diagn and put them into
    # the nc file in output
    # some variables need unscaling, temperature needs -273.15 for K->˚C
    # pressure needs a exp, precipitation needs unit adjustment etc...
end
```

which would allow users to define their own output variables with any possible processing

milankl:
@minifuo with #525 merged this is now also documented in the `dev` docs, should be built any moment

PR #571, opened 2024-08-14
Adjust OutputWriter to new structure

maximilian-gelbrecht:
While our HPC is still offline / making problems, I had some time revise the `OutputWrite` according to the new structure. This isn't the complete overhaul suggested in #522, it just revises the current code. I might make an attempt at #522 later if the problems with our HPC continue to persist. 

This doesn't fully work yet because of #570 

But as soon as this issue is resolved, this could be merged into the PR #525


milankl:
That's great. I have started to make the output writer modular but I forgot to commit the changes before leaving for Scotland. Those changes should address everything for #522 but I'll be able to push these in 10 days... sorry!

milankl:
@maximilian-gelbrecht you think this is superceded by #573 ?

maximilian-gelbrecht:
Yes, not needed anymore

PR #572, opened 2024-08-14
RingGrids: Fix RingGrids Array conversion

### LINKED ISSUE #570: RingGrids: N-dim Matrix/Array constructors

maximilian-gelbrecht:
Fixes #570

### DIFF
--- a/src/RingGrids/full_grids.jl
+++ b/src/RingGrids/full_grids.jl
@@ -21,18 +21,17 @@ matrix_size(Grid::Type{<:AbstractFullGridArray}, nlat_half::Integer) =
 # convert an AbstractMatrix to the full grids, and vice versa
 """
 ($TYPEDSIGNATURES)
-Initialize an instance of the grid from an Array. If `input_as==Vector` this leading dimension is 
-interpreted as a flat vector of all horizontal entries in one layer. If `input_as==Matrx` the first
-two leading dimensions are interpreted as longitute and latitude. This is only possible for full 
-grids that are a subtype of `AbstractFullGridArray`.
+Initialize an instance of the grid from an Array. For keyword argument `input_as=Vector` (default)
+the leading dimension is interpreted as a flat vector of all horizontal entries in one layer.
+For `input_as==Matrx` the first two leading dimensions are interpreted as longitute and latitude.
+This is only possible for full grids that are a subtype of `AbstractFullGridArray`.
 """
 (Grid::Type{<:AbstractFullGridArray})(M::AbstractArray; input_as=Vector) = Grid(M, input_as)
 
 function (Grid::Type{<:AbstractFullGridArray})(M::AbstractArray, input_as::Type{Matrix})
-    M_flat = reshape(M, :, size(M)[3:end]...) # identical to vec(M) for M <: AbstractMatrix
-    npoints2D = size(M_flat, 1)
-    nlat_half = get_nlat_half(Grid, npoints2D)  # get nlat_half of Grid
-    Grid(M_flat, nlat_half)
+    # flatten the two horizontal dimensions into one, identical to vec(M) for M <: AbstractMatrix
+    M_flat = reshape(M, :, size(M)[3:end]...)
+    Grid(M_flat)
 end 
 
 Base.Array(grid::AbstractFullGridArray) = Array(reshape(grid.data, :, get_nlat(grid), size(grid.data)[2:end]...))
--- a/src/RingGrids/full_grids.jl
+++ b/src/RingGrids/full_grids.jl
@@ -19,7 +19,12 @@ matrix_size(Grid::Type{<:AbstractFullGridArray}, nlat_half::Integer) =
 
 ## CONVERSION
 # convert an AbstractMatrix to the full grids, and vice versa
-(Grid::Type{<:AbstractFullGrid})(M::AbstractMatrix) = Grid(vec(M))
+function (Grid::Type{<:AbstractFullGrid})(M::AbstractArray) 
+    M_flat = reshape(M, :, size(M)[3:end]...) # identical to vec(M) for M <: AbstractMatrix
+    npoints2D = size(M_flat, 1)
+    nlat_half = get_nlat_half(Grid, npoints2D)  # get nlat_half of Grid
+    Grid(M_flat, nlat_half)
+end 
 Base.Array(grid::AbstractFullGridArray) = Array(reshape(grid.data, :, get_nlat(grid), size(grid.data)[2:end]...))
 Base.Matrix(grid::AbstractFullGridArray) = Array(grid)
 
--- a/src/RingGrids/full_grids.jl
+++ b/src/RingGrids/full_grids.jl
@@ -19,15 +19,28 @@ matrix_size(Grid::Type{<:AbstractFullGridArray}, nlat_half::Integer) =
 
 ## CONVERSION
 # convert an AbstractMatrix to the full grids, and vice versa
-function (Grid::Type{<:AbstractFullGrid})(M::AbstractArray) 
+"""
+($TYPEDSIGNATURES)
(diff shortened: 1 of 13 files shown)


milankl:
Don't wait for any reviews from my side in the next 10 days 😉 Just go ahead 🏴󠁧󠁢󠁳󠁣󠁴󠁿🚵‍♂️

maximilian-gelbrecht:
I had a brief look at it again before merging, and I noticed that this might cause problems in other parts even though the tests are passing. 

With the N-dim arrays we can't overload the constructors with both, from flat arrays / rings: 

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/b5e8eb298eb0dda9b006c74cd219117b123868a0/src/RingGrids/general.jl#L120

and the constructor from Matrix / lon-lat arrays for Full Grids: 

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/b5e8eb298eb0dda9b006c74cd219117b123868a0/src/RingGrids/full_grids.jl#L22

at the same time. 

I tried some changes but so far, but my quick attempt wasn't not really working. As this also affects a lot of the code (e.g. initializiations of all kinds of components of the model) I'll wait for you Milan to do the final changes.

src/RingGrids/full_grids.jl
```diff
@@ -19,10 +19,28 @@ matrix_size(Grid::Type{<:AbstractFullGridArray}, nlat_half::Integer) =
...
+    nlat_half = get_nlat_half(Grid, npoints2D)  # get nlat_half of Grid
+    Grid(M_flat, nlat_half)
+end 
+
 Base.Array(grid::AbstractFullGridArray) = Array(reshape(grid.data, :, get_nlat(grid), size(grid.data)[2:end]...))
 Base.Matrix(grid::AbstractFullGridArray) = Array(grid)
 
+function full_grid_from_matrix(Grid::Type{<:AbstractFullGridArray{T,N}}, M::AbstractArray{S,N2}) where {T,S,N,N2}
+   
+end
```
milankl, line 43: hu?

src/physics/land.jl
```diff
@@ -27,7 +27,7 @@ Base.@kwdef struct SeasonalLandTemperature{NF, Grid<:AbstractGrid{NF}} <: Abstra
     varname::String = "lst"
 
     "Grid the land surface temperature file comes on"
-    file_Grid::Type{<:AbstractGrid} = FullGaussianGrid
+    file_Grid::Type{<:AbstractGrid} = FullGaussianGrid{NF}
```
milankl, line 30: Why was that necessary? The idea here is to just store the information of what grid the gridded netcdf file comes on -- technically one could infer this from the latitudes, but I thought that was tedious

src/RingGrids/full_grids.jl
maximilian-gelbrecht: ups, forgot to delete that

src/physics/land.jl
maximilian-gelbrecht: Yes you are right, I'll remove it again. 
Initially I had a version of this PR that was trying to do this solely by multiple dispatch. But this wasn't working nicely

milankl: There is a related issue that NCDatasets returns `Union{Missing, Float32}` as eltype maybe when reading the netcdf file this was dealt with by doing `FullGaussianGrid{NF}(...) ` so fixing that type on the side of the grid instead of doing something like `FullGaussianGrid(Array{NF}(ds[var]))` which maybe we should do generally?

maximilian-gelbrecht: Yes, I am aware that NCDatasets returns `Union{Missing, Float32}` but where do we read like that? I've only seen something like `file_Grid(ncfile["orog"][:, :])` in the code, which seems to work without the `NF` (why actually?).

milankl: I think it works because I didn't specify `_FillValue` for those netcdf files in which case the returned eltype is Float32 and not a Union. In any case we maybe should do `Grid(Array(ds["varname"].var), input_as=Matrix)` which always ignores masked values, doesn't require the dimensions and obtains the eltype from the netcdf automatically?

maximilian-gelbrecht: Does this always work? I just tried it with some netCDF file I had lying around. If the netCDF variables is `Union{Missing, Float64}`, 
* `Array(ds["varname"])` also results in an `Array{Union{Missing, Float64}, 3}` 
* and `Array(ds["varname"].var)` in `Array{Float32, 3}`
I find this a bit confusing, but if you're sure it always works, it's probably safer to change it to ``Array(ds["varname"].var)`

milankl: Did some speed comparison https://github.com/Alexander-Barth/NCDatasets.jl/issues/227#issuecomment-2321498232 I believe in our case we should do one of these

1. No missing values (i.e. ignore them), dimensions known, number format as in NetCDF
```julia
ds["vor"].var[:, :, :, :]
```

2. No missing values (i.e. ignore them), number format as in NetCDF but dimensions unknown

```julia
Array(ds["vor"].var)
```

3. like 2 but convert to another number format on the fly

```julia
Array(ds["vor"].var)
```

milankl (commented):
Only adjusted the `.var` notation to netCDF files to not end up with a `Union{Missing, Float32}` eltype

src/physics/land.jl
milankl: Implemented 1. with [e81208c](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/572/commits/e81208c06be4db11e7e15f26cac61c2a8c738492)

src/RingGrids/full_grids.jl
```diff
@@ -19,12 +19,25 @@ matrix_size(Grid::Type{<:AbstractFullGridArray}, nlat_half::Integer) =
...
+For `input_as==Matrx` the first two leading dimensions are interpreted as longitute and latitude.
+This is only possible for full grids that are a subtype of `AbstractFullGridArray`.
+"""
+(Grid::Type{<:AbstractFullGridArray})(M::AbstractArray; input_as=Vector) = Grid(M, input_as)
+
+function (Grid::Type{<:AbstractFullGridArray})(M::AbstractArray, input_as::Type{Matrix})
+    # flatten the two horizontal dimensions into one, identical to vec(M) for M <: AbstractMatrix
+    M_flat = reshape(M, :, size(M)[3:end]...)
+    Grid(M_flat)
+end
```
milankl, line 35: Oh also shortened this, as the two additional lines you wrote is exactly what happens if you wrap an array into a grid: get the number of horizontal points `get_npoints2D` then determine what  resolution parameter `nlat_half` that is

milankl:
This is ready to be merged @maximilian-gelbrecht ?

maximilian-gelbrecht:
Yes

PR #573, opened 2024-08-27
NetCDF output modularised

milankl:
@maximilian-gelbrecht @miniufo Okay, first commit to solve #522 which is needed for #525. The idea is that "NetCDFOutput" is a model component that takes all sorts of 2D/3D fields and potentially interpolates them onto _one_ other grid of a possibly different resolution or a possibly different number format. I found the contrain that _one_ output grid + resolution is considered only helpful as otherwise every variable would need its own interpolator (which I find too complicated for now and especially for the modularisation as this would require every custom output variable to deal with interpolator tweaks). Meaning now, all variables in the netcdf file share x,y,z,t but can drop any of them (e.g. outputting orography wouldn't use z). Constructed via

```julia
julia> output = NetCDFOutput(spectral_grid)
NetCDFOutput{FullGaussianGrid{Float32}}
├ status: inactive/uninitialized
├ interpolator: Main.SpeedyWeather.RingGrids.AnvilInterpolator{Float32, OctahedralGaussianGrid}
├ path: output.nc
├ frequency: 21600 seconds
└┐ variables:
 ├ v: meridional wind [m/s]
 ├ u: zonal wind [m/s]
 ├ vor: relative vorticity [s^-1]
```

(default for `Barotropic`) or using other default variables with a 2nd argument of the model type, i.e. `output = NetCDFOutput(spectral_grid, PrimitiveWet)`. The actual variables are a dictionary

```julia
julia> output.variables
Dict{Symbol, Main.SpeedyWeather.AbstractOutputVariable} with 6 entries:
  :v     => MeridionalVelocityOutput("v", "m/s", "meridional wind", (true, true, true, true), NaN, 3, true, 7)
  :humid => HumidityOutput("humid", "kg/kg", "specific humidity", (true, true, true, true), NaN, 3, true, 7)
  :temp  => TemperatureOutput("temp", "degC", "temperature", (true, true, true, true), NaN, 3, true, 10)
  :u     => ZonalVelocityOutput("u", "m/s", "zonal wind", (true, true, true, true), NaN, 3, true, 7)
  :pres  => SurfacePressureOutput("pres", "hPa", "surface pressure", (true, true, false, true), NaN, 3, true, 12)
  :vor   => VorticityOutput("vor", "s^-1", "relative vorticity", (true, true, true, true), NaN, 3, true, 5)
```

So that we can just add to the output an additional variable like so

```julia
julia> output = NetCDFOutput(spectral_grid, PrimitiveWet)

julia> add!(output, SpeedyWeather.OrographyOutput())
NetCDFOutput{FullGaussianGrid{Float32}}
├ status: inactive/uninitialized
├ interpolator: Main.SpeedyWeather.RingGrids.AnvilInterpolator{Float32, OctahedralGaussianGrid}
├ path: output.nc
├ frequency: 21600 seconds
└┐ variables:
 ├ orography: orography [m]
 ├ v: meridional wind [m/s]
 ├ humid: specific humidity [kg/kg]
 ├ temp: temperature [degC]
 ├ u: zonal wind [m/s]
 ├ pres: surface pressure [hPa]
 ├ vor: relative vorticity [s^-1]
```

output variables are defined as (possibly externally)

```julia
@kwdef mutable struct VorticityOutput <: AbstractOutputVariable
    name::String = "vor"
    unit::String = "s^-1"
    long_name::String = "relative vorticity"
    dims_xyzt::NTuple{4, Bool} = (true, true, true, true)
    missing_value::Float64 = NaN
    compression_level::Int = 3
    shuffle::Bool = true
    keepbits::Int = 5
end
```

and require a method for the `output!` function because we can generally not guess which variable people want to grab from where and with whatever post-processing (e.g. scaling, Kelvin->degC, ...)

```julia
function output!(
    output::NetCDFOutput,
    variable::VorticityOutput,
    progn::PrognosticVariables,
    diagn::DiagnosticVariables,
    model::AbstractModel,
)
    vor = output.grid3D             # use output grid, vorticity is 3D
    (; vor_grid) = diagn.grid       # unpack vorticity from gridded diagnostic variables
    RingGrids.interpolate!(vor, vor_grid, output.interpolator)

    unscale!(vor, diagn.scale[])    # was vor*radius, back to vor
    round!(vor, variable.keepbits)  # bitrounding for compression
    i = output.output_counter       # write into timestep i
    output.netcdf_file[variable.name][:, :, :, i] = vor
    return nothing
end
```

all of these look very similar and people could easily add their own tweaks to this.


maximilian-gelbrecht:
Looks good! I had an initial look at the code and I don't have any direct comments or questions about it.

PR #577, opened 2024-09-25
Benchmarks for individual dynamics functions

maximilian-gelbrecht:
I added the individual dynamics functions as another benchmark suite. I had to change a few things around because they didn’t fit in the existing `BenchmarkSuite` scheme. 

I use `BenchmarkTools` which isn’t in the main environment, so I added a new local environment. This isn’t working ideally because the `Project.toml` doesn’t know that Speedy was added with `dev ..`. Maybe we should add the Manifest.toml as well here, or you know, just be aware of it when executing the benchmarks.

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -1,6 +1,6 @@
 # Benchmarks
 
-created for SpeedyWeather.jl v0.9.0 on Tue, 02 Apr 2024 14:05:51. 
+created for SpeedyWeather.jl v0.10.0 on Wed, 25 Sep 2024 18:36:59. 
 
 All simulations have been benchmarked over several seconds (wallclock time) without output. Benchmarking excludes initialization and is started just before the main time loop and finishes right after. The benchmarking results here are not very robust, timings that change with +-50% are not uncommon. Proper benchmarking for performance optimization uses the minimum or median of many executions, while we run a simulation for several time steps which effectively represents the mean, susceptible to outliers that slow down the simulation. However, this is what a user will experience in most situations anyway and the following therefore presents a rough idea of how fast a SpeedyWeather simulation will run, and how much memory it requires.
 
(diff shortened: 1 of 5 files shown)


milankl:
> I use `BenchmarkTools` which isn’t in the main environment, so I added a new local environment. This isn’t working ideally because the `Project.toml` doesn’t know that Speedy was added with `dev ..`. Maybe we should add the Manifest.toml as well here, or you know, just be aware of it when executing the benchmarks.

In practice this just means someone has to have BenchmarkTools in their environment and it's not automatically installed no? Because I wouldn't find that thaaat problematic given how mature and widely used BenchmarkTools is

benchmark/README.md
```diff
@@ -1,6 +1,6 @@
 # Benchmarks
 
-created for SpeedyWeather.jl v0.9.0 on Tue, 02 Apr 2024 14:05:51. 
+created for SpeedyWeather.jl v0.10.0 on Wed, 25 Sep 2024 18:36:59.
```
milankl, line 3: Should we already change the version to 0.11 here?

benchmark/benchmark_suite.jl
```diff
@@ -60,4 +63,196 @@ function run_benchmark_suite!(suite::BenchmarkSuite)
...
+    allocs::Vector{Vector{Int}} = [fill(0, length(function_names)) for i=1:nruns]
+end 
+
+default_function_names() = ["pressure_gradient_flux!", "linear_virtual_temperature!", "temperature_anomaly!", "geopotential!", "vertical_integration!", "surface_pressure_tendency!", "vertical_velocity!", "linear_pressure_gradient!", "vertical_advection!","vordiv_tendencies!", "temperature_tendency!", "humidity_tendency!", "bernoulli_potential!"]
+
+function add_results!(suite::BenchmarkSuiteDynamics, trial::BenchmarkTools.Trial, i_run::Integer, i_func::Integer)
+
+    t = median(trial)
+
+    suite.time_median[i_run][i_func] = t.time
```
milankl, line 138: Is median what's typically printed with `@btime`? I thought that was lowest?

benchmark/README.md
maximilian-gelbrecht: Yeah, makes sense. It's a major change, and some of it is also user facing (the output)

benchmark/benchmark_suite.jl
maximilian-gelbrecht: Yes, you are right `@btime` prints the minimum. I didn't know that before. `@benchmark` prints range, median and mean. Any way, not that it matters a lot, but I'd still prefer median.

milankl: I believe in benchmarking the minimum is often used because it is more robust because that's the most uninterrupted execution of this code

PR #579, opened 2024-09-25
Add tracer advection

### LINKED ISSUE #564: arbitrary tracer support
Hello! I've just recently started playing around with speedyweather and I love it, definitely so far accomplishing its goals of being easy to use compared to other models I have experience with.

I'm interested in using it to model atmospheres of other solar system bodies, which ties nicely into your to-do list item of wanting an "exoplanet mode" and #462.

For other worlds, different physical processes dominate and so differences in parameterisation schemes are essential. That's fine, as those are generally going to be fine-tuned by and/or written entirely by the user.

Speedy currently has support for a single tracer in the PrimitiveWetModel, water vapour by default - it isn't currently documented, but its properties can be changed by the user. This was discussed in #460 - also mentioned there was that the computational expense of adding new tracers making you hesitant to add support for new ones (in the context of #460, the new ones being water condensates). We can see from the benchmarks the performance hit you get just going from PrimitiveDry to PrimitiveWet.

# Tracers in other atmospheres 

I'd have to argue from the point of view of generalising to other atmospheres, that support for users adding their own tracers is essential - they should of course be warned about the computational hit of new prognostics, but if this is to be a sandbox for playing with a wide variety of atmospheres rather than just Earth's, then there's really no getting around this. You can just about get away with one tracer on Earth as a first approximation, but on other worlds, that's not necessarily the case.

## Mars

At the absolute bare minimum for simulating the Martian atmosphere, I would say you need two tracers: one to keep track of water ice, and the other dust. You'd have to then crudely diagnose the number of particles from only the mass, or vice versa. A slower but more accurate model might have four tracers: mass and number mixing ratios (mmr and nmr) for both, and then parameterisations could assume some properties of the distribution of particle sizes given both of these numbers. 

You can't readily neglect either ice or dust, nor can you diagnose one from the other. 

NASA's legacy model used two additional tracers (water vapour mmr and dust core mmr) for a total of six.

Summary: definitely can't have less than 2, 4 would be less crude, 6 is great.

## Venus

Has complicated chemistry and I'm no Venusian expert. Both water and sulfur based stuff is important and you can't diagnose one from the other, so that's already multiple desirable tracers.

I've seen simplifying assumptions that certain tracers only exist at fixed altitudes and then advecting them in 2d rather than 3d.

I'm not sure what a reasonable number of tracers is or a perfect number, but I suspect one is far too much of an oversimplification.

## Titan

Titan has some of the most complex atmospheric chemistry in the solar system. The most important tracer is methane, which can also exist in several phases just like water on earth, so wanting three tracers for it is a very reasonable request, although much like water, you could massively simplify and treat the condensates as diagnostics of the vapour.

You'd also want ethane as a tracer, though, it's also very important.

There is also a haze that is typically treated as a single homogeneous thing rather than a soup of many organic compounds.

You could maybe get away with just two, and diagnose everything from those with some other assumptions about droplet formation etc. I've seen the haze often being treated as a constant or even globally uniform field.

Ideally though the choice of tracer number would be user-specified not baked into the model, since "how few tracers can we get away with while still reasonably approximating the Titanaen atmosphere?" is an interesting research question that's arguably still an open problem; and speedy would be a great tool for exploring that.

## Jupiter and Saturn 

From what I know about them, personally I'd want to advect two tracers if I had to pick a minimum number: one to represent ammonia and the other methane.

## Uranus and Neptune

These you could get away with only one tracer I think? Just methane would probably get you pretty far in simulating these planets.

## Triton and Pluto

It's not wildly unreasonable you'd want more than one. There's nitrogen ice but you could diagnose that by altitude maybe. There's methane and carbon monoxide you might want to advect.

# Summary

Even just looking at the atmospheres we know a lot about, without going to exoplanet research, support for user-defined tracers is important if we want speedy to be generalised to other atmospheres. I personally feel it's the main thing it's missing - people studying a new atmosphere might well expect they have to write an entirely new paramterisation scheme, but not that they'd have to also know how to modify the dynamical equations to account for new tracers.

## Implementation 

Not coming from a dynamics background myself I'm not sure exactly how implementation of new tracers should be done. Correct me if this is nonsense but I think "all" that needs doing is:

- add a new tracer transport equation to the primitive equations, with the time derivative of the tracer being the negative wind dot grad tracer (plus some source and sink terms from parameterisations).

- solver would have to be modified to loop over all tracers

- tracers could be passive (in which case that's it) or active (which I'm using to mean the effect of the tracer on density is not negligible) in the latter case the virtual temperature would need to be modified to be the `temperature * (1+ sum_over_active_tracers((gas_constant_ratio_for_tracer_i - 1)*qi))`.

## User interface

Very much up for discussion, but I was envisioning something like having a `PrimitiveExoticModel` (unsure of naming) in addition to the zero and one tracer `Dry` and `Wet` models.

This could be created with a 1d array of `
AbstractTracer` objects.

Each abstract tracer object could be required to have a name (for reference in setting initial conditions or writing output or something), and be either passive or active, and if active then it would also need a gas constant.

The ratio of tracer gas constants to the bulk atmospheric gas constant (dry air on Earth) could then be computed and stored internally since those are the important numbers in the dynamics (I think).

## Motivation

This does seem to square with the goals of speedy, since a desire for an exoplanet mode is mentioned on the to-do list and it definitely leans into the "play atmospheric modelling like it's LEGO" desire for flexibility. Unfortunately approximating the atmosphere with a single tracer only sounds reasonable for a subset of known planets, and I think avoiding being too Earth-centric is a noble goal.

This would also assist with the statement about extending speedyweather with new parameterisations: the "we're happy to provide an interface for a complicated cloud scheme you want to attach to speedy, even if it'll live in its own repository rather than here because it's too slow and complex" statement in speedyweather's docs does rather imply that it'd be nice to just say extend AbstractCloudScheme, as opposed to the cloud parameterisation developer having to do some complicated implementation of new prognostics to the dynamical core themselves (a much higher barrier of entry since the people who study one of physics or dynamics tend to be much less familiar with the other).

As an aside, many of these worlds are smaller than Earth, so the same angular grid spacing gives you a higher effective horizontal resolution. You can make up a lot of lost performance from new tracers by the performance gain of simulating a smaller world's atmosphere. Of course the opposite is true for very large worlds unfortunately, but I'm quite interested in the smaller worlds myself!

I am willing to write code to help with this where necessary, though I suspect I'd be more useful helping write some parameterisation implementations than updating the dynamics.

milankl:
fixes #564 by

- [x] define `Tracer` (name, ~id~, active)
- [x] add `model.tracers` for a list of tracers
- [x] based on `length(model.tracers)` add additional progn/diagn variables
- [x] advect tracer variables
- [x] time stepping for tracers
- [x] `set!` for tracers
- [x] 3D advection
- [x] output for tracers (preliminary)
- [x] tests
- [x] documentation


milankl:
new syntax is

```julia
add!(model, Tracer(:ch4))      # before initialize! or
add!(simulation, Tracer(:ch4)) # after
```

tracers are now just identified with their `Symbol` (not an id anymore) as their just live in a dictionary in the prognostic/diagnostic variables

milankl:
Tracers can be added (`add!(simulation, Tracer(:newtracer))` (and technically also be deleted) anytime, but with `model.tracer[:co2].active = false` also deactivated (=frozen in time). Implemented the `set!` function for tracers too,
so
```julia
julia> simulation.prognostic_variables.tracers
Dict{Symbol, Tuple{LowerTriangularArray{ComplexF32, 2, Matrix{ComplexF32}}, LowerTriangularArray{ComplexF32, 2, Matrix{ComplexF32}}}} with 2 entries:
  :ch4 => ([0.0+0.0im; 0.0+0.0im; … ; 0.0+0.0im; 0.0+0.0im;;], [0.0+0.0im; 0.0+0.0im; … ; 0.0+0.0im; 0.0+0.0im;;])
  :co2 => ([0.0+0.0im; 0.0+0.0im; … ; 0.0+0.0im; 0.0+0.0im;;], [0.0+0.0im; 0.0+0.0im; … ; 0.0+0.0im; 0.0+0.0im;;])
```

can be `set!` via

```julia
julia> set!(simulation, ch4=1)

julia> simulation.prognostic_variables.tracers
Dict{Symbol, Tuple{LowerTriangularArray{ComplexF32, 2, Matrix{ComplexF32}}, LowerTriangularArray{ComplexF32, 2, Matrix{ComplexF32}}}} with 2 entries:
  :ch4 => ([3.54491+0.0im; 0.0+0.0im; … ; 0.0+0.0im; 0.0+0.0im;;], [0.0+0.0im; 0.0+0.0im; … ; 0.0+0.0im; 0.0+0.0im;;])
  :co2 => ([0.0+0.0im; 0.0+0.0im; … ; 0.0+0.0im; 0.0+0.0im;;], [0.0+0.0im; 0.0+0.0im; … ; 0.0+0.0im; 0.0+0.0im;;])

julia> set!(simulation, co2=2)

julia> simulation.prognostic_variables.tracers
Dict{Symbol, Tuple{LowerTriangularArray{ComplexF32, 2, Matrix{ComplexF32}}, LowerTriangularArray{ComplexF32, 2, Matrix{ComplexF32}}}} with 2 entries:
  :ch4 => ([3.54491+0.0im; 0.0+0.0im; … ; 0.0+0.0im; 0.0+0.0im;;], [0.0+0.0im; 0.0+0.0im; … ; 0.0+0.0im; 0.0+0.0im;;])
  :co2 => ([7.08982+0.0im; 0.0+0.0im; … ; 0.0+0.0im; 0.0+0.0im;;], [0.0+0.0im; 0.0+0.0im; … ; 0.0+0.0im; 0.0+0.0im;;])
``` 

@maximilian-gelbrecht which I implemented simply by checking whether additional `kwargs` are in the keys for `progn.tracers`.  This happens if not

```julia
julia> set!(simulation, co3=2)
ERROR: UndefVarError: `co3` not defined
```

Technically we could do the same to avoid all the predefined `vor = nothing, ...` with `isnothing(vor)` checks. One could simply do `:vor in keys(kwargs) && set!(..., kwargs[:vor], ...)`

milankl:
This seems to work!

https://github.com/user-attachments/assets/fde81e79-ae9d-42b0-a2c1-d4bbb2752452

done via

```julia
spectral_grid = SpectralGrid(trunc=63, nlayers=1)
model = ShallowWaterModel(spectral_grid)

# add tracer
add!(model, Tracer(:co2))
add!(model, SpeedyWeather.TracerOutput())

# set tracer and run a simulation
set!(simulation, co2 = (λ, φ, σ) -> 0 < λ < 10 ? 1 : 0)

# reset tracer but continue simulation
set!(simulation, co2 = (λ, φ, σ) -> exp(-(λ-180)^2/10^2))

**Comment by milankl** [comment]
And T170 resolution and 20days long

https://github.com/user-attachments/assets/2d493ba5-29bd-42b2-8264-517878c6e65e

**Comment by maximilian-gelbrecht** [comment]
Great! I can do a review tomorrow or Friday.

**Comment by milankl** [comment]
I also think this is a good test case for the semi-Lagrangian advection I've been dreaming about! Because that should allow us to get to maybe 3-4k SYPD at T31 by allowing for larger time steps and cutting down on the number of transforms, especially for all the tracers. Every tracer adds 4 transforms (1x spec-> grid, 1x grid tendencies -> spec, 2x grid -> spec for the fluxes with u,v). So we could

1. start using semi-Lagrange for tracers only, compare speed, compare to Eulerian advection in terms of diffusion/accuracy/conservation
2. use semi-Lagrange for humidity (active tracer but maybe just straight forward otherwise), virtual temperature?
3. use semi-Lagrange for temperature (even more active tracer)
4. use for momentum too

The reason I say "dreaming" is that the semi-implicit semi-Lagrangian scheme is at the heart of what makes ECMWF's model so fast, allowing really big time steps and making the model very stable.

File: src/dynamics/tracers.jl [hunk 1]
```diff
@@ -0,0 +1,52 @@
...
+Tracer(name::Symbol; kwargs...) = Tracer(; name, kwargs...)
+
+const TRACER_DICT = Dict{Symbol, AbstractTracer}
+
+function add!(dict::TRACER_DICT, tracers::Tracer...)
+    for tracer in tracers
+        dict[tracer.name] = tracer
+    end
+    return dict
+end
```
maximilian-gelbrecht, line 18: I am a bit confused, where is this method of `add!` with the `TRACER_DICT` used? I find it.

src/output/netcdf_output.jl
```diff
@@ -1095,6 +1095,39 @@ function output!(
     return nothing
 end
 
+"""Defines netCDF output for a specific variables, see `VorticityOutput` for details.
```
maximilian-gelbrecht, line 1098: adjust docstring

maximilian-gelbrecht (commented):
This looks really great! 

I can't fully check that the advection is done correctly, just from looking at the code, but everything looks to be well implemented and makes sense to me. 

I mainly have comments about the documentation.

docs/src/tracers.md
```diff
@@ -0,0 +1,189 @@
...
+If the tracer ``q`` does not impact the flow it is considered _passive_.
+Humidity, for example, is an _active_ tracer as it changes the [Geopotential](@ref)
+(and therefore the pressure gradient force) through the [Virtual temperature](@ref).
+A tracer is conserved in the absence of sources or sinks (the zero on the right-hand side above).
+Aerosols from wildfires might be considered to be a passive tracer, but the
+source term should increase the aerosol concentration whereever and whenever there is 
+a wildfire. And also a sink term should be added representing aerosols being washed out
+in rainfall, or deposited on the ground. However, aersols should be considered
+_active_ and not _passive_ if they influence the radiation and hence the temperature
+which couples the tracer equation above two-way with the other equations.
```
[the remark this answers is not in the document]
milankl: Added a few sentences on this!

docs/src/tracers.md
```diff
@@ -0,0 +1,189 @@
...
+You can also add a tracer to a `simulation`, i.e. after the model is initialized.
+
+```@example tracers
+simulation = initialize!(model)
+add!(simulation, Tracer(:xyz))
+```
+
+which will add the tracer to `model.tracers` as above but also add
+it to the prognostic and diagnostic variables (otherwise done at `initialize!`).
+What you should not do is add a tracer to the `model` _after_ it has been initialized.
```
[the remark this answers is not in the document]
milankl: Yes, technically we could only allow/document how to add a tracer to a simulation. However, conceptually I find it important to be able to add a tracer to the model too as this is exactly what happens if one chooses "PrimitiveWetModel" over "PrimitiveDryModel" -- hence the existence of a tracer should be defined in the model itself.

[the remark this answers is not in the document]
milankl: Added some sentences to clarify that

src/dynamics/tracers.jl
milankl: just in the line below, added some comments to clarify

src/output/netcdf_output.jl
milankl: will do in another pull request, I need to overhaul the netcdf output anyway

PR #585, opened 2024-10-10
Stop supporting Julia 1.9

milankl:
Especially given that 1.10 is now LTS. I'd rather do this early than running into problems supporting older Julia versions. I've updated CI, docs, and removed some definitions for 1.9 and earlier


maximilian-gelbrecht (approved):
Yes I fully agree. Supporting a LTS version and the current version is enough. Let's not cause ourselves extra work by supporting old versions.

Issue #584, opened 2024-10-10
Set! discontinuous across prime meridian


milankl:
```julia
orography = NoOrography(spectral_grid)
model = PrimitiveWetModel(;spectral_grid, orography)
H, λ₀, φ₀, σ = 4000, 0, 50, 15         # height, lon, lat position, and width
set!(model, orography=(λ,φ) -> H*exp((-(λ-λ₀)^2 - (φ-φ₀)^2)/2σ^2), add=true)
heatmap(model.orography.orography)
```

has a discontinuity at 0˚E

![image](https://github.com/user-attachments/assets/bc40e71d-59e9-4dbe-a11e-8f3faddbd5d8)

@maximilian-gelbrecht you think we can fix that within the `set!` function or does one need to account for that in the `(λ,φ) ->` function definition?

maximilian-gelbrecht:
Huh, I'll look into it. There has to be a possibility to do that neatly within `set!`

maximilian-gelbrecht:
Okay, after thinking about this a bit more. This is too poorly defined mathematicall to do it within `set!`. You'd have to provide more additional information of how you want to have the continuity to be enforced and this just becomes messy. 

I'd just add a "ghost point" to the function definition in the individual cases 

```Julia 
using SpeedyWeather 
spectral_grid = SpectralGrid()
orography = NoOrography(spectral_grid)
model = PrimitiveWetModel(;spectral_grid, orography)
H, λ₀, φ₀, σ = 4000, 0, 50, 15         # height, lon, lat position, and width
set!(model, orography=(λ,φ) -> H*(exp((-(λ-λ₀)^2 - (φ-φ₀)^2)/2σ^2) + exp((-(λ-(λ₀+360))^2 - (φ-φ₀)^2)/2σ^2)), add=true)
plot(model.orography.orography)
```

`

maximilian-gelbrecht:
It's not easy to generalise this to arbitrary functions, so I'd prefer to keep this out of `set!` I think. 

One thing we could think to implement is a function that shifts a RingGrid by a certain longitude. I am even sure if this is easy to do with all reduced grids, but it would be helpful in cases in which you - in some way - want the prime meridian to be somewhere else to solve a discontinuity.

milankl:
[...]

PR #588, opened 2024-10-16
Add Haversine and general `spherical_distance` and `AbstractSphericalDistance` functionality

treigerm:
This adds a simple haversine distance function to compute distance on the sphere; partly addressing #584.

TODOs:
- [x] Add tests (would need some guidance where to best add the tests)
- [x] Check documentation is added correctly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Haversine formula and AbstractSphericalDistance [#588](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/588)
 - Array-agnostic SpectralTransform [#583](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/583)
 - Move CUDA dependency into extension [#586](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/586)
 - Stop supporting Julia v1.9 [#585](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/585)
--- a/src/RingGrids/geodesics.jl
+++ b/src/RingGrids/geodesics.jl
@@ -40,7 +40,7 @@ function Haversine(             # functor for Haversine struct
 end
 
 # allow for any <:AbstractSphericalDistance also non-tupled arguments lon1, lat1, lon2, lat2
-(F::AbstractSphericalDistance)(lon1, lat1, lon2, lat2; kwargs...) = F((lon1, lat1), (lon2, lat2); kwargs...) 
+(F::Type{<:AbstractSphericalDistance})(lon1, lat1, lon2, lat2; kwargs...) = F((lon1, lat1), (lon2, lat2); kwargs...) 
 
 """$(TYPEDSIGNATURES)
 Spherical distance, or great-circle distance, between two points `lonlat1` and `lonlat2`
--- a/src/RingGrids/geodesics.jl
+++ b/src/RingGrids/geodesics.jl
@@ -42,6 +42,8 @@ end
 # allow for any <:AbstractSphericalDistance also non-tupled arguments lon1, lat1, lon2, lat2
 (F::Type{<:AbstractSphericalDistance})(lon1, lat1, lon2, lat2; kwargs...) = F((lon1, lat1), (lon2, lat2); kwargs...) 
 
+export spherical_distance
+
 """$(TYPEDSIGNATURES)
 Spherical distance, or great-circle distance, between two points `lonlat1` and `lonlat2`
 using the `Formula` (default `Haversine`)."""
--- a/src/RingGrids/geodesics.jl
+++ b/src/RingGrids/geodesics.jl
@@ -1,0 +1,17 @@
+const EARTH_RADIUS = 6371e3  # Earth mean radius in meters.
+
+"""$(TYPEDSIGNATURES) 
+Haversine formula calculating the great-circle distance between a pair of 
+longitude-latitude points. Assumes the Earth is a perfect sphere with radius 
+6371 km."""
+function haversine((lon1, lat1), (lon2, lat2))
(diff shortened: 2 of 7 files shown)


milankl:
Yeha, welcome welcome! 🥳 

Tests go in `test/` we are currently testing the grids in `grids.jl` but this function is somewhat different I guess. So I would just create its own file `test/geodesics.jl` and test in there! While `RingGrids` is its own module it's still being tested as part of SpeedyWeather. We eventually want to move RingGrids, LowerTriangularMatrices and SpeedyTransforms into their own packages but want them to mature a bit more first.

src/RingGrids/geodesics.jl
```diff
@@ -0,0 +1,54 @@
...
+Super type of formulas to calculate the spherical distance or great-circle distance.
+To define a `NewFormula`, define `struct NewFormula <: AbstractSphericalDistance end`
+and the actual calculation as a functor
+
+    function NewFormula(lonlat1::Tuple, lonlat2::Tuple; radius=DEFAULT_RADIUS, kwargs...)
+
+assuming inputs in degrees and returning the distance in meters (or radians for radius=1).
+Then use the general interface `spherical_distance(NewFormula, args...; kwargs...)`"""
+abstract type AbstractSphericalDistance end
+struct Haversine <: AbstractSphericalDistance end
```
milankl, line 15: introduced an `AbstractSphericalDistance` so that any formula is a subtype of it which allows us to handle a "class" of types abstractly together. See how I've defined both `lon1, lat1, lon2, lat2` as well as tupled arguments `(lon1, lat1), (lon2, lat2)` for all new formulas collectively. The actual calculation is then implemented as a function of the same name, here `Haversine` which in Julia is called a functor. It's essentially an object that also has a function defined. The object here is "empty" (there are no fields in `struct Haversine ... end`)

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Haversine formula and AbstractSphericalDistance [#588](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/588)
```
milankl, line 5: Oh and we've set this repo up so that every pull request is required to change our changelog, so that we have a little log and what's been introduced/changed in what version

src/RingGrids/geodesics.jl
```diff
@@ -0,0 +1,54 @@
...
+    φ1 = deg2rad(lat1)
+    φ2 = deg2rad(lat2)
+
+    Δφ = deg2rad(lat2 - lat1)
+    Δλ = deg2rad(lon2 - lon1)
+
+    # Haversine formula, see https://en.wikipedia.org/wiki/Haversine_formula
+    a = sin(Δφ / 2)^2 + cos(φ1) * cos(φ2) * sin(Δλ / 2)^2
+    c = 2 * atan(sqrt(a), sqrt(1 - a))
+    return c*radius             # in meters
```
milankl, line 39: There is one more issue here: `radius::Float64` by default so it's always float64 that is returned, but it would be nice to have the function being type stable, in a sense of that the promotion of the input types in the tuples determines the output type. We could either do `c*convert(typeof(c), radius)` but then a user defined radius would be inferior to the latlons or we simply define the `const DEFAULT_RADIUS = 6371000` as integer which is probably my preferred option!

treigerm:
Thanks for all the helpful comments @milankl . That is very insightful for learning Julia best-practices. I have now added some tests. 

Two more points that need to be clarified:
1. Do we want to export the `spherical_distance` function directly from SpeedyWeather? Right now it needs to be called with `RingGrids.spherical_distance`.
2. Should we explicitly enforce the constraint that `longitude \in [0, 360]` and `latitude \in [-90, 90]`?

milankl:
> That is very insightful for learning Julia best-practices.

It's more like SpeedyWeather's modularity standards to be honest. Your version was perfectly fine Julia code, I just wanted to pepper it a little bit for future uses in e.g. the ParticleAdvection which needs to compute a spherical distance which is currently just done in cartesian coordinates. So I'm aiming for (one day, not now) something like

```julia
particle_advection = ParticleAdvection(spectral_grid, spherical_distance=Haversine)
```

or `spherical_distance=CartesianDistance` so that internally one or the other formula is picked. Or one can define manually a new `MySphericalDistance <: AbstractSphericalDistance` and then just write the formula as `MySphericalDistance(lon1, lat1, lon2, lat2, ...) = ...` and would be automatically propagated everywhere in the model 🎉 

>     1. Do we want to export the `spherical_distance` function directly from SpeedyWeather? Right now it needs to be called with `RingGrids.spherical_distance`.

Yes, we can add `export` it to SpeedyWeather and then also reexport it from there! So that `using SpeedyWeather` would bring that function directly into scope, I'll add that!

>     2. Should we explicitly enforce the constraint that `longitude \in [0, 360]` and `latitude \in [-90, 90]`?

hmm, through the `deg2rad` both -180 to 180˚E and 0 to 360˚ should work no? Maybe add a test on that? Then for `latitude` I think it's fine, the other common coordinate is colatitudes and we clearly say "lat" not "colat". But on that note, maybe we should write `latd` and `lond` whereever we mean degrees, that's at least the convention we have in other parts of RingGrids

treigerm:
> hmm, through the deg2rad both -180 to 180˚E and 0 to 360˚ should work no? 

Thanks for adding the test, and yes it should work by default. But the function also currently doesn't complain if you give values outside these ranges so `spherical_distance((140, 30), (240, -40)) == spherical_distance((500, 30), (600, -40))`. I guess that's not really an issue because it gives you reasonable values outside these bounds.

milankl:
Typical problem again that the docs don't deploy because you pull requested from a fork, I'll merge if the docs don't look like what they supposed to look like then we patch, but they build so don't fail at least!

PR #589, opened 2024-10-17
SpeedyTransforms: Differentiable transforms via custom Enzyme rule

maximilian-gelbrecht:
We want to make the spherical harmonics transforms differentiable. 

The complete model is supposed to rely on Enzyme for that, Enzyme can deal with many of our functions already by itself in principle. However, the FFTs are not computed by Julia code, but with FFTW and cuFFT. Transforms are also the most performance critical part of the model, so optimising the gradients of the transforms also makes sense. 

## Enzyme 

* The Legendre transformation `_legendre!` Enzyme can handle by itself
* We do Fourier transforms consecutively on each latitude ring in `_fourier!`, the forward version uses `rfft` plans, the inverse `brfft` plans. `_fourier!` isn't normalised, the normalisation is done in `_legendre!`. 
* The FFT needs a rule here, the adjoint of a Fourier transform is the unnormalised inverse Fourier transform (so `brfft`). For real-valued FFT you have an additional scale in the adjoint reflecting the different coefficients in the formula (there's a two in front of the cos/sine terms)  
* Rules are tested with the test tools from Enzyme and FD comparison for the whole transform

## ChainRules/Zygote/…

For uses outside of our model (e.g. Spherical FNOs) it would be great to also have this differentiable by other ADs. These ADs usually don’t support mutation, so that I would define rules for the non-mutating `transform(grid, S)`. I’ll do that a bit later. This might also need some changes to the code. 

## This PR 

* This PR implements the Enzyme Rule for `_fourier!` and tests it.  
* As far as I can see it, this rule should also work with the GPU version as soon as we have it with just a tiny adjustment. 
* We could pre-compute the scaling. It’s a relatively small Int valued matrix. That would make the adjoint marginally faster. 
* I also introduce a new extension for FiniteDifferences.jl that makes our data structure compatible with the library 
* I set up a new environment for the tests. 
* Enzyme Custom rules go into another extension
* <s>Currently I just add Enzyme, EnzymeRules and EnzymeTestUtils to the main env to make testing a bit easier for me while developing this PR, that will be changed before merging this
    * Do we want to do the differentiability test always in the same files as the other tests, or put them in a separate folder?
    * Enzyme (and most of this code) could go in an extension. Is this what we want long-term? Probably yes, but I don’t have a good impression yet how many custom rules we have to define. Hopefully not many though. </s>
    * 
## Testing 

* I do a manual test comparing with FiniteDifferences of the full spherical harmonics transforms, a test that $$\frac{d}{dx}\mathcal{S}^{-1}(S(x)) = 1$$  and tests with EnzymeTestUtils (which also do a comparison to FD)
* <s>I'd love to test the rules directly with EnzymeTestUtils as well, but there are several problems currently
    * There's a problem with the FFT plans having an uninitailzied fields. There's a quite hacky way around this though
    * There's a problem with the function having complex valued outputs 
    * The tests are stuck, I waited 30 min without a result. 
    * I have to look into this a bit further</s>
* <s>FiniteDifferences tests can take a while. Going forward do we want to put all gradient correctness checks in the regular CI? The four quite simple tests that are currently in this PR already take 5 mins. </s>

## Complex Numbers 

Differentiating complex numbers is a bit of a topic for itself. Enzyme also has a quite long explanation of their approach in their FAQs (https://enzymead.github.io/Enzyme.jl/stable/faq/#Complex-numbers). I think for us, we can treat the real and imaginary part basically as separate numbers, as they would be for real SPH (for which coefficients with m<0 correspond to our imaginary part (sort of)). I am not quite sure yet if this works well out of the box with Enzyme or if we have to be more carefully somewhere. It's something we should think about.


maximilian-gelbrecht:
It turns out our data structures are not compatible with FiniteDifferences.jl. FiniteDifferences.jl needs to convert each data structure to a vector and then back. I added an extension for FiniteDifferences.jl to provide those extension. I'll do that for RingGrids and LowerTriangularMatrices. Maybe we have to add further ones later.

maximilian-gelbrecht:
Ok, JVPs with FiniteDifferences are now possible

maximilian-gelbrecht:
If tested manually and not via EnzymeTestTools, the gradient through Enzyme is very wrong unfortunately. I hope to track down the mistake next week.

maximilian-gelbrecht:
The forward and backward transforms are differentiable via Enzyme! 
The tests comparing the results to finite differences pass on my laptop, hopefully they will in the CI as well 

@milankl This might be a good point for you to already have a brief look at it. Especially at what I wrote about how we want to organise this code and testing going forward

maximilian-gelbrecht:
There seem to be some issue running those tests in the CI that I don't have on my computer. I'll investigate.

milankl:
With the last commit: You think we can have all Enzyme functionality as an extension? Happy to try, but also loading/precompiling it is not as expensive as CUDA I believe?

maximilian-gelbrecht:
I think so. Because the only we'd need Enzyme for directly in the package is custom rules. And I don't think we need that many. If we are lucky, those two might even be the only ones (because they are calling the only non-Julia code). 

Enzyme.jl installs the C++ Enzyme library. So I thought, maybe good to move it to an extension when it's not needed for every user.

ext/SpeedyWeatherEnzymeExt.jl
```diff
@@ -0,0 +1,115 @@
...
+    else
+        tape = nothing
+    end
+
+    return AugmentedReturn(nothing, nothing, tape) # because the function actually returns nothing
+
+end 
+
+function reverse(config::EnzymeRules.RevConfigWidth{1}, func::Const{typeof(_fourier!)}, ::Type{<:Const}, tape,
+    f_north::Duplicated, f_south::Duplicated, grids::Duplicated{<:AbstractGridArray}, S::Union{Const, MixedDuplicated})
```
milankl, line 58: So `::Any` becomes `::Duplicated` and `::T` becomes `::Duplicated{<:T}`? And that's how to specify the method of a function?

ext/SpeedyWeatherEnzymeExt.jl
```diff
@@ -0,0 +1,115 @@
...
+    grids::Duplicated{<:AbstractGridArray}, f_north::Duplicated, f_south::Duplicated, S::Union{Const, MixedDuplicated})
+
+    # adjoint/vjp of FFT has a different scaling, compute it, apply it later to f_north, f_south
+    scale = adjoint_scale(S.val)
+    
+    # TODO: retrieve from tape here if overwritten? 
+
+    # compute the actual vjp
+    dfnorthval = zero(f_north.val)
+    dfsouthval = zero(f_south.val)
```
milankl, line 101: Because these allocate: Is that because we can't reuse the forward memory for the gradient?

ext/SpeedyWeatherEnzymeExt.jl
```diff
@@ -0,0 +1,115 @@
...
+
+    _fourier!(dfnorthval, dfsouthval, grids.dval, S.val) # inverse FFT (w/o normalization)
+
+    f_north.dval .+= scale .* dfnorthval
+    f_south.dval .+= scale .* dfsouthval 
+
+    # no derivative wrt the grids that were input because they are overwritten
+    make_zero!(grids.dval)
```
milankl, line 109: How is this different from `fill!`?

test/runtests.jl
```diff
@@ -2,6 +2,7 @@ using SpeedyWeather
 using Test
 
 # GENERAL
+include("test_transforms_ad_rules.jl") # TODO: put the test somewhere else (and next to the transforms)
```
milankl, line 5: Don't actually mind in which order the tests are executed but yeah feel free to move it after the transforms tests

test/test_transforms_ad_rules.jl
```diff
@@ -0,0 +1,101 @@
...
+            # forwards 
+            S = SpectralTransform(spectral_grid)
+            dS = deepcopy(S)
+            grid = rand(spectral_grid.Grid{spectral_grid.NF}, spectral_grid.nlat_half, spectral_grid.nlayers)
+            dgrid = zero(grid)
+            specs = zeros(LowerTriangularArray{Complex{spectral_grid.NF}}, spectral_grid.trunc+2, spectral_grid.trunc+1, spectral_grid.nlayers)
+            
+            # seed
+            dspecs = zero(specs)
+            fill!(dspecs, 1+1im)
```
milankl, line 66: What does "seed" mean in this context and why is it 1+i?

test/test_transforms_ad_rules.jl
```diff
@@ -0,0 +1,101 @@
...
+
+            autodiff(Reverse, transform!, Const, Duplicated(specs, dspecs), Duplicated(grid, dgrid), Duplicated(S, dS))
```
milankl, line 68: Très cool!!

test/test_transforms_ad_rules.jl
```diff
@@ -0,0 +1,101 @@
...
+
+            autodiff(Reverse, transform!, Const, Duplicated(grid, dgrid), Duplicated(specs, dspecs), Duplicated(S, dS))
+
+            # new seed 
+            dgrid2 = similar(grid)
+            fill!(dgrid2, 1)
+
+            fd_jvp = FiniteDifferences.j′vp(central_fdm(5,1), x -> transform(x, S), dgrid2, specs)
+
+            @test isapprox(dspecs, fd_jvp[1])
```
milankl, line 93: We could also test that the roundtrip transform is identity no?

ext/SpeedyWeatherEnzymeExt.jl
maximilian-gelbrecht: I didn't find this documented anywhere, but this is how I think it works.

ext/SpeedyWeatherEnzymeExt.jl
maximilian-gelbrecht: It's used across Enzyme rule definitions. I think it's more general than `fill!`, but for array-type data it does actually call `fill!`. So here, `make_zero!` should also call the `fill!` we defined for our data types.

test/test_transforms_ad_rules.jl
maximilian-gelbrecht: What roundtrip exactly? I am not sure I understand. You could differentiate through $\mathcal{S}^{-1}(\mathcal{S}(x))$, but that's just identity, so the derivative wrt the input would just vanish. Or is this exactly what you were thinking of?

test/test_transforms_ad_rules.jl
maximilian-gelbrecht: We do a reverse mode differentiation here. So you propagate the value of the output in reverse direction to determine the sensitivity of output wrt its inputs. The seed is an example output so to say. If you set it to one, you do actually compute directly the gradient of the function. 

As I wrote above in the big post, differentiating complex numbers is a bit of special topics in itself. In our case we treat the imaginary and real part kind of like separate numbers, so setting both to one made sense to me intuitively. But differentiating complex numbers is something we should still think about a bit.

test/test_transforms_ad_rules.jl
milankl: Vanish=1 no? It's d(S^-1(S(x)))/dx = dx/dx = 1 I thought? And we could test this starting in either space?

maximilian-gelbrecht: Ah yes sure, you are right.

maximilian-gelbrecht:
The CI is passing for Julia 1.10.

There are some problems with Enzyme for Julia 1.11, but they are working on it.

test/test_transforms_ad_rules.jl
maximilian-gelbrecht: I added this as tests. 

Starting in spectral space it works very well
Starting in grid space, the result is sort of close to 1, but not quite. Currently I am not sure what goes wrong there.

milankl: I am not surprised. This depends on what data you start with in grid space, which can hold more information than spectral space, so the first roundtrip is usually lossy. But subsequent ones should be fine. I would start in spectral space, transform to grid space then start the test grid->spectral->grid there

test/test_transforms_ad_rules.jl
milankl: Yes please! Just checking, because I think of 
```julia
julia> z = exp(im*π/4)
0.7071067811865476 + 0.7071067811865475im

julia> abs(z)
1.0
```
more as complex "1" than `1 + 1im` because it's actually a unit vector

milankl:
@vchuravy we (well, Max!) are making some progress to differentiate through SpeedyWeather!

milankl:
@swilliamson7 this might be also relevant for you, let us know if you have any more insights / general wisdom!!

swilliamson7:
Looking through the thread, but rather exciting for SpeedyWeather.jl! In addition to Valentin, maybe @wsmoses would be interested in seeing this?

test/test_transforms_ad_rules.jl
maximilian-gelbrecht: We already do that in the tests. But in principal you are right, there's something wrong with the setup for this test. Because the finite difference differentiation yields the same gradient as Enzyme. So it's not an issue of the custom rule.

maximilian-gelbrecht:
I had a very quick check if some non-sensical differentiation through a timestep of a barotropic model is possible. 

I get an error from Enzyme and the compiler from the `_divergence!` function. That's already a bit what I suspected. After the transform itself, the spatial gradient functions are the other challenging bit. I'll look into differentiating them again, when we also make steps towards GPU-ifying those bits. That should be next on the agenda after the transforms.

test/test_transforms_ad_rules.jl
maximilian-gelbrecht: The test passes now, I changed the function to test the identity to not directly overwrite the input data, but instead save the output in a pre-allocated array. This wasn't just an issue though with Enzyme, FiniteDifferences.jl showed the same error. Now it works.

ext/SpeedyWeatherEnzymeExt.jl
maximilian-gelbrecht: I have no idea to be honest. I didn't do it initially because I am a bit careful with mutating the arguments within things the AD might still need. But I just tested to use `f_north.val` as a buffer, and this seems to work. At least the tests pass. But I don't know if this is generally something we can do. I didn't commit this yet to the PR.

maximilian-gelbrecht:
So, all my direct tests pass. This is moving towards being ready. 

Left to do: 

- [ ] Figure out why EnzymeTestUtils is stuck 
- [ ] Do we want to keep all of these tests in the CI? Right now for one grid type it's 8min. If we add more differentiability tests later, it might get too much. Enzyme is fast, but Finite Differences are slow. That's why it takes so long 
- [ ] Clear up some smaller open questions/To-do along the code

milankl:
> Do we want to keep all of these tests in the CI? Right now for one grid type it's 8min. If we add more differentiability tests later, it might get too much. Enzyme is fast, but Finite Differences are slow. That's why it takes so long

Hmmm it would be good if the tests aren't adding massive time for all other development but they are obviously crucial for differentiability testing. Not sure how we can achieve that. Could we only keep the test without FiniteDifferences? Use a lower resolution transform? You're in a better decision than me to decide.

maximilian-gelbrecht:
Ah yes, of course, you are right. So far I was even using 8 layers in the test, which is just a bit stupid.

maximilian-gelbrecht:
Now it's much faster, I think we can keep them like that in the CI

maximilian-gelbrecht:
Okay, I feel a bit stupid, haha. EnzymeTestUtils is also not stuck anymore. Before I was letting it run longer than 2 hours, now it finishes in less in 1 second for a lower dimensional problem. 

In order the tests of EnzymeTestUtils to pass I had to increase the tolerance a bit from the default (to `1e-4`) and increase the accuracy of the FD. For my QG3.jl I had to do the same, so that's not really a surprise to me.

maximilian-gelbrecht:
The tests pass on my laptop, but not in the CI. I'll investigate this a bit later.

maximilian-gelbrecht:
All tests pass also in the 1.10 CI for the `FullGaussianGrid`.  

For the `OctahedralGaussianGrid`, it's a bit weird. 

* The FD comparisons of the full transforms that I wrote, pass 
* EnzymeTestUtils passes most tests, especially the relevant ones (derivative of the function wrt to its relevant input)
* But for some reason the test of the derivative wrt to the pre-allocated `f_north`,`f_south` don't pass. We set those manually to zero, as is recommended by the Enzyme docs because `f_north` is overwritten by `_fourier!`, so the derivative wrt to that input is zero. But FiniteDifferences.jl computes a few non-zero entries in these derivative matrices. This doesn't happen for the `FullGaussianGrid`. It's a bit odd, but we know that those derivatives should be zero, so this is likely an issue with FiniteDifferences and how it's called, and not an issue with our custom rule or Enzyme.

maximilian-gelbrecht:
With the lastest update to Enzyme the tests now pass on 1.11 as well.

vchuravy (commented):
Looks fine to me.

@michel2323 was looking at custom rules for FFT IIRC

PR #591, opened 2024-10-18
Rossby-Haurwitz wave initial conditions +docs

milankl:
Tobias Ehlert @rivwgk had implemented the Rossby-Haurwitz wave as initial conditions for SpeedyWeather, see [Williamson et al. 1992](https://www.sciencedirect.com/science/article/pii/S0021999105800166), i.e. initial conditions for vorticity are (eq. 145)

image

With our `set!` function this is now some easy three lines of code that take care of the transform between grid and spectral space for you

```julia
using SpeedyWeather
spectral_grid = SpectralGrid(trunc=127, nlayers=1)
model = BarotropicModel(;spectral_grid)

# Rossby-Haurwitz wave initial conditions for vorticity
R, K, ω = 4, 7.848e-6, 7.848e-6
ζ(λ, θ, σ) = 2ω*sind(θ) - K*sind(θ)*cosd(θ)^R*(R^2 + 3R + 2)*cosd(R*λ)
set!(simulation, vor=ζ)
```

So only differences to the mathematical version from above is that we use degrees (`sind`, `cosd`) as coordinates, and `σ` is our vertical coordinate that needs to be specified for compatibility with our 3D primitive equations models (yes, it's exactly the same syntax in `PrimitiveDryModel` and `PrimitiveWetModel`!). Looks like

image

@rivwgk can you provide a reference for the cut-off amplitude? It's not discussed in Williamson et al., 1992. You call it $$\sigma$$ but we use that for the vertical coordinate, so it's `c` here. Anyway, you can do that as easy as

```julia
# filter low values below cutoff amplitude c
vor = simulation.prognostic_variables.vor[1]  # 1 = first leapfrog timestep
low_values = abs.(vor) .< c
vor[low_values] .= 0
```

no need for the loop you wrote, very similar to numpy if you ask me ;)

This pull request

- introduces and exports `RossbyHaurwitzWave <: AbstractInitialConditions` to have these initial conditions predefined
- adds a section on "Initial conditions" to the documentation outlining how to set the initial conditions using the Rossby-Haurwitz wave as an example

@clessig you wanted to be kept updated ☝🏼

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Rossby-Haurwitz wave initial conditions [#591](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/591)
 - Array-agnostic SpectralTransform [#583](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/583)
 - Move CUDA dependency into extension [#586](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/586)
 - Stop supporting Julia v1.9 [#585](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/585)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Array-agnostic SpectralTransform [#583](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/583)
 - Move CUDA dependency into extension [#586](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/586)
 - Stop supporting Julia v1.9 [#585](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/585)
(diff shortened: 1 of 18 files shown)


rivwgk:
Sadly, I don't have a reference for the cut-off amplitude and don't exactly remember why I added it. I think it might be that I added it because at first I had the Rossby-Haurwitz wave given by three spherical harmonics coefficients for $Y_1^0$, $Y_5^4$ and $Y_5^{-4}$ and wanted to cut other numerical noise off?

milankl:
> Sadly, I don't have a reference for the cut-off amplitude and don't exactly remember why I added it. I think it might be that I added it because at first I had the Rossby-Haurwitz wave given by three spherical harmonics coefficients for Y 1 0 , Y 5 4 and Y 5 − 4 and wanted to cut other numerical noise off?

Well it acts now as an example even if it's just you who came up with it ;) I hope that's okay. If you want to make any changes feel free to create a pull request, otherwise you find the Rossby-Haurwitz wave test case now in the documentation:

https://speedyweather.github.io/SpeedyWeather.jl/dev/initial_conditions/#Rossby-Haurwitz-wave-in-a-BarotropicModel

PR #592, opened 2024-10-21
SpectralAR1Process and AbstractRandomProcess

milankl:
Implements two random processes `<:AbstractRandomProcess` to generate random patterns (2D only for now) with temporal and spatial autocorrelation following idea from the Stochastically Perturbed Parameterization Schemes (SPPT) developed at ECMWF, see [here](https://www.ecmwf.int/sites/default/files/elibrary/2016/16733-model-uncertainty-representations-ifs.pdf) for example:

image

@samhatfield the mentioned length scales are 500, 1000, and 2000km. Do you know which wavenumbers these are? I picked ``l = 24, 16, 8`` for now.  It seems you do this [here](https://github.com/samhatfield/speedy.f90/blob/0ebfdbfcd62efd46dd1802284908645be4b230be/source/sppt.f90#L74-L84) in speedy.f90

@minqi6

### DIFF
--- a/src/SpeedyWeather.jl
+++ b/src/SpeedyWeather.jl
@@ -115,6 +115,7 @@ include("dynamics/scaling.jl")
 include("dynamics/tendencies.jl")
 include("dynamics/hole_filling.jl")
 include("dynamics/particle_advection.jl")
+include("dynamics/random_process.jl")
 
 # PARAMETERIZATIONS
 include("physics/albedo.jl")
--- a/src/dynamics/diagnostic_variables.jl
+++ b/src/dynamics/diagnostic_variables.jl
@@ -119,7 +119,9 @@ $TYPEDFIELDS."""
     v_grid          ::GridVariable3D = zeros(GridVariable3D, nlat_half, nlayers)
     "Logarithm of surface pressure [Pa]"
     pres_grid       ::GridVariable2D = zeros(GridVariable2D, nlat_half)
-    
+    "Random pattern controlled by random process [1]"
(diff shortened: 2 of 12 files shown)


milankl:
Superposition of these 3 wavenumbers looks like

image

samhatfield:
Oh wow, that code takes me back. It looks like I only implemented a single length scale of [500 km](https://github.com/samhatfield/speedy.f90/blob/0ebfdbfcd62efd46dd1802284908645be4b230be/source/sppt.f90#L35), right?

milankl:
> It looks like I only implemented a single length scale of [500 km](https://github.com/samhatfield/speedy.f90/blob/0ebfdbfcd62efd46dd1802284908645be4b230be/source/sppt.f90#L35), right?


yes, but that's fine, the multiscale is in the end just an addition. The latest commit now implements that pretty similar to yours although there's an additional 1/2 factor in the `exp` that's not in the memorandum. I skipped that, not sure why that's needed. Also it seems that ECMWF uses a different normalisation, is it possible that your `m = n = 0` zero mode is *the* mean of the field? Because we use a normalisation where there's an additional factor of $$2\sqrt(\pi) = 3.5449...$$ that I have to multiply otherwise the standard deviation of the grid field is off by a factor of about that!

Looks now like

image

with length scale of 500km being translated to wavenumber 12 (I wanted to have it independent of the earth's radius) and a standard deviation of 1/3 before clamping which yields a distribution like

image

and the clamping to $$[-1, 1]$$ is hardly doing much.

PR #593, opened 2024-10-21
Model constructors with positional SpectralGrid argument

milankl:
At the moment one *has to* write
```julia
model = PrimitiveWetModel(; spectral_grid, kwargs...)
```
to construct any model. To lower the confusion for beginners ("what's that `;` for?") also allow for the spectral grid to be passed on as first positional argument, i.e.

```julia
model = PrimitiveWetModel(spectral_grid, kwargs...)
```

which is also consistent with how we create all model components.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


maximilian-gelbrecht (approved):
Yeah, I am fine with that, as it's a required argument, this could avoid confusion.

PR #596, opened 2024-10-25
NetCDF output: Precipitation accumulated and rate

milankl:
NetCDF output has now by default accumulated precipitation (large-scale condensation and convection) so that the fields in `diagnostic_variables.physics.precip_large_scale` and `_convection` are not affected by the NetCDF output, previously the output writer was setting those accumulators back to zero which interferes with `RainGauge` in RainMaker.jl for example. Now the model records accumulated precipitation and everything else is done in output. Now `LargeScalePrecipitationRateOutput` and `ConvectivePrecipitationRateOutput` are implemented as a post-processing step, called during the `finalize!` of `model.output`. Also renames `finish!` generally to `finalize!` for consistency with `initialize!`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Output both accumulated and precipitation rate as netCDF [#596](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/596)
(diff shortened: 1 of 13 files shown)


milankl:
@minqi6 with this pull request the model will by default output accumulated precipitation and precipitation rate which might come handy for you.

image

PR #601, opened 2024-10-30
Array-agnostic horizontal diffusion, +modularised

### LINKED ISSUE #547: Spectral filtering instead of hyperdiffusion
We currently do the solid lines for horizontal diffusion of vorticity and divergence which change shape with higher resolution (hence the time scale is adapted to compensate that somehow)

image

I believe a more scale-selective diffusion just via a spectral filter might avoid overly diffusive simulations while efficiently taking out power at high wavenumbers, like the dashed lines above, which were created with

$$ K(l) = \frac{1}{2} (1 - \text{erf}(s(l - T - k)))$$

with parameters $k = 0$ (the shift, for k=-5 the 5th highest wavenumbers is dampened like the highest wavenumber with $k=0$) and the stretch $s = 0.12$ which controls the scale-selectiveness, ($s=0.7$ is more like a biharmonic Laplacian, etc.)

Coded like

```julia
using SpecialFunctions
T = 31
l = 0:T
s = 0.05
k = 0
1/2 * (erfc(s*(l - T - k)))
```
(`erfc = 1 - erf`)

milankl:
fixes #547 

- implements `SpectralFilter` as an alternative to `HyperDiffusion` but leaves that one as default
- makes `HyperDiffusion somewhat weaker for all variables except divergence
- corrects the gravity wave example in the docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+## v0.12.0
+
 - OctaminimalGaussianArray/Grid to start with 4 points around the poles [#595](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/595)
 - Output both accumulated and precipitation rate as netCDF [#596](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/596)
 - Random processes for random pattern generation [#592](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/592)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- SpectralFiltering for horizontal diffusion [#601](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/601)
+
 ## v0.12.0
 
 - OctaminimalGaussianArray/Grid to start with 4 points around the poles [#595](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/595)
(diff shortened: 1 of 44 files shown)


milankl:
Experimenting with new defaults, the new explicit/implicit terms from spectral filtering based on the `tanh` function (could also use `erf` but that requires a SpecialFunctions.jl dependency. `SpectralFilter` is now defined as

```julia
time_scale = Second(diffusion.time_scale).value/radius * (32/(trunc+1))^resolution_scaling
expl[l+1, k] =  -(1 + tanh(scale*(l - trunc - shift)))^power / time_scale
impl[l+1, k] = 1/(1-2Δt*expl[l+1, k])      
```

with maybe default values being

```julia
    "[OPTION] shift diffusion to higher (positive shift) or lower (neg) wavenumbers, relative to trunc"
    shift::NF = 0

    "[OPTION] Scale-selectiveness, steepness of the sigmoid, higher is more selective"
    scale::NF = 0.05
    
    "[OPTION] diffusion time scale"
    time_scale::Second = Hour(4)

    "[OPTION] resolution scaling to shorten time_scale with trunc"
    resolution_scaling::NF = 1

    "[OPTION] power of the error function"
    power::NF = 4
```

This yields a weaker but for higher resolutions more scale-selective diffusion

image

milankl:
The Galwesky test case is slightly more wavy with SpectralFilter

image

compared to HyperDiffusion

image


but the enstrophy spectrum looks much better

image

I believe the underlying reason is that the enstrophy cascades are different to the divergence cascades so that actually the divergence needs a higher diffusion, because here the divergence spectrum actually looks like

image

with this being the divergence in the Galeswky test case

image

Issue #575, opened 2024-09-12
Simplified spectral transform towards GPU version


milankl:
```julia
using SpeedyWeather

function SpeedyTransforms.transform_simple!(  # SPECTRAL TO GRID
    grids::FullGaussianArray,                 # gridded output
    specs::LowerTriangularArray,              # spectral coefficients input
    S::SpectralTransform{NF};                 # precomputed transform
) where NF                                    # number format NF

    (; nlat, nlon_max, nlat_half) = S
    (; Λs, brfft_plans ) = S

    @boundscheck ismatching(S, grids) || throw(DimensionMismatch(S, grids))
    @boundscheck ismatching(S, specs) || throw(DimensionMismatch(S, grids))

    lmax = specs.m - 1            # 0-based maximum degree l of spherical harmonics
    mmax = specs.n - 1            # 0-based maximum order m of spherical harmonics

    # preallocate work arrays
    nfreq_max = nlon_max÷2 + 1
    gn = zeros(Complex{NF}, nfreq_max)      # phase factors for northern latitudes
    gs = zeros(Complex{NF}, nfreq_max)      # phase factors for southern latitudes

    rings = eachring(grids)                 # precomputed ring indices

    for k in eachgrid(grids)                # loop over all grids (e.g. vertical dimension)
        for j_north in 1:nlat_half              # symmetry: loop over northern latitudes only
            j_south = nlat - j_north + 1        # southern latitude index
            Λj = Λs[j_north]                    # Use precomputed Legendre polynomials Λ

            # INVERSE LEGENDRE TRANSFORM by looping over wavenumbers l, m
            lm = 1                              # flattened index for non-zero l, m indices
            for m in 1:mmax+1                   # loop over orders, Σ_{m=0}^{mmax}, but 1-based index
                acc_odd  = zero(Complex{NF})    # accumulator for isodd(l+m)
                acc_even = zero(Complex{NF})    # accumulator for iseven(l+m)

                for l in m:lmax+1               # loop over degrees, Σ_{l=m}^{lmax} but 1-based index
                    # anti-symmetry: sign change of odd harmonics on southern hemisphere
                    if iseven(l+m)
                        acc_even += specs[lm, k] * Λj[lm]
                    else
                        acc_odd += specs[lm, k] * Λj[lm]
                    end

                    lm += 1                     # count up flattened index for next harmonic
                end

                gn[m] += (acc_even + acc_odd)   # accumulators for northern
[...]

PR #602, opened 2024-10-31
Implementation of CUDA-ised transform

jackleland:
With apologies to @milankl for the delay on getting this PR made! 

This is a working-progress implementation of the transform on CUDA, currently with a functioning forward fourier transform component utilising CUFFT. A couple of things to note:

1. I had to refactor the plan creation as CUFFT doesn't seem to play nicely while using views of the CuArray, so it instead has to slice them. I thought initially that the slicing would also work for the CPU arrays but that seems to break the use of views in the actual `fourier_batched!` and `fourier_serial`. There might be a way around this but I have not been able to find one!
2. The actual running of the FFT within `fourier_batched` was done, again, with slices i.e.
``` julia
view(f_north, 1:nfreq, 1:nlayers, j) .= rfft_plan * grids.data[ilons, :]
```
which is slow and allocatey but again I could not find a way around. This doesn't appear to be a limitation for the serial implementation however...

Speed-wise we get significant speedup for larger grids `(trunc=127, nlayers=64)` compared to the CPU, but is 10x slower than the CPU at default grid size. 

Feedback much appreciated as this is my first Julia PR. I'll be back on this next week to have a go at the Legendre.

Addresses #575


src/SpeedyTransforms/SpeedyTransforms.jl
```diff
@@ -9,11 +9,16 @@ import FFTW
...
 import Primes
+import Adapt: adapt
 
 # SPEEDYWEATHER MODULES
 using ..LowerTriangularMatrices
 using ..RingGrids
 
+# import SpeedyWeatherCUDAExt
+using SpeedyWeather
+Base.get_extension(SpeedyWeather, :SpeedyWeatherCUDAExt)
```
maximilian-gelbrecht, line 20: Is this really needed? 

As far as I can see it you only extend functions that are already there in the main module, so it doesn't seem necessary to me.

src/SpeedyTransforms/spectral_transform.jl
```diff
@@ -140,7 +139,8 @@ function SpectralTransform(
     end
     
     # SCRATCH MEMORY FOR FOURIER NOT YET LEGENDRE TRANSFORMED AND VICE VERSA
-    scratch_memory_north = zeros(Complex{NF}, nfreq_max, nlayers, nlat_half)
+    # north is converted to array type now for help with making the FFT plans
+    scratch_memory_north = ArrayType(zeros(Complex{NF}, nfreq_max, nlayers, nlat_half))
     scratch_memory_south = zeros(Complex{NF}, nfreq_max, nlayers, nlat_half)
```
maximilian-gelbrecht, line 144: This one also needs to be `ArrayType`

maximilian-gelbrecht, line 144: This one also needs to be `ArrayType` , same for the `scratch_memory_column_north` below

maximilian-gelbrecht (commented):
Nice progress! 

We should now really add GPU CI, then you can add CI tests for those things as well.

ext/SpeedyWeatherCUDAExt/fourier.jl
```diff
@@ -0,0 +1,119 @@
...
+        j_south = nlat - j_north + 1        # corresponding southern latitude index
+        nlon = nlons[j]                     # number of longitudes on this ring
+        nfreq = nlon÷2 + 1                  # linear max Fourier frequency wrt to nlon
+        not_equator = j_north != j_south    # is the latitude ring not on equator?
+
+        rfft_plan = rfft_plans[j]           # FFT planned wrt nlon on ring
+        ilons = rings[j_north]              # in-ring indices northern ring
+        
+        # FOURIER TRANSFORM in zonal direction, northern latitude
+        view(f_north, 1:nfreq, 1:nlayers, j) .= rfft_plan * grids.data[ilons, :]
```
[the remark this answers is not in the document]
milankl: @jackleland because at the moment it sounds we would just need an `rfft!`/`brfft!` method that calls this line, either written with `LinearAlgebra.mul!` for CPUs or with something else for GPUs, so the `_fourier_batched!` function is actually the same and we wouldn't need any code duplication here, just a new method for `rfft!`, `brfft!`

src/SpeedyTransforms/SpeedyTransforms.jl
milankl, line 20: Ouh why is that needed? Wouldn't that load CUDA everytime SpeedyTransforms is loaded, i.e. everytime SpeedyWeather is loaded?

src/SpeedyTransforms/spectral_transform.jl
```diff
@@ -27,12 +26,12 @@ struct SpectralTransform{
     mmax::Int                       # Maximum order m=[0, l] of spherical harmonics
     nfreq_max::Int                  # Maximum (at Equator) number of Fourier frequencies (real FFT)
     LegendreShortcut::Type{<:AbstractLegendreShortcut} # Legendre shortcut for truncation of m loop
-    mmax_truncation::VectorIntType  # Maximum order m to retain per latitude ring
+    mmax_truncation::Vector{Int}   # Maximum order m to retain per latitude ring
```
milankl, line 29: @maximilian-gelbrecht vectors of integers that don't actually contain any data, in this case just defines the end of a range for a loop, aren't supposed to live on the GPU, right?

ext/SpeedyWeatherCUDAExt/fourier.jl
```diff
@@ -0,0 +1,119 @@
...
+
+    return rfft_plans, brfft_plans, rfft_plans_1D, brfft_plans_1D
+end
+
+
+function SpeedyTransforms._fourier_batched!(                     # GRID TO SPECTRAL
+    f_north::CuArray{<:Complex, 3},             # Fourier-transformed output
+    f_south::CuArray{<:Complex, 3},             # and for southern latitudes
+    grids::AbstractGridArray{NF, N, <:CuArray}, # gridded input
+    S::SpectralTransform,                       # precomputed transform
```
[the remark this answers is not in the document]
maximilian-gelbrecht: With how Jack wrote it as an extension just doing the dispatch on the array type seems the natural fit. Nonetheless we might need to do a dispatch on the type/device within `SpectralTransform` as well in other parts of the code. 

The Fourier transform are also completely device specific. If at any later point we want to do be compatible with AMD, Metal, etc, we have to use their FFT library. So there's no point in doing it even more general here.

src/SpeedyTransforms/spectral_transform.jl
maximilian-gelbrecht: Indices don't have to be stored on the GPU.

src/SpeedyTransforms/SpeedyTransforms.jl
jackleland: Ahh good catch, left that in accidentally

src/SpeedyTransforms/spectral_transform.jl
jackleland: Everything gets converted to CuArrays at the end of the constructor, only the north scratch memory is needed for the fft planning so I didn't explicitly convert anything else but I can do happily.

ext/SpeedyWeatherCUDAExt/fourier.jl
```diff
@@ -0,0 +1,119 @@
...
+        real_matrix_input = fake_grid_data.data[rings[j], :]
+        complex_matrix_input = scratch_memory_north[1:nlon÷2 + 1, :, j]
+        real_vector_input = fake_grid_data.data[rings[j], 1]
+        complex_vector_input = scratch_memory_north[1:nlon÷2 + 1, 1, j]
+
+        rfft_plans[j] = FFT_package.plan_rfft(real_matrix_input, 1)
+        brfft_plans[j] = FFT_package.plan_brfft(complex_matrix_input, nlon, 1)
+        rfft_plans_1D[j] = FFT_package.plan_rfft(real_vector_input, 1)
+        brfft_plans_1D[j] = FFT_package.plan_brfft(complex_vector_input, nlon, 1)
+    end
```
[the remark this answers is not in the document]
jackleland: I thought this but when I tested it seemed to break the current implementation of `fourier_batched!` as FFTW apparently allows planning on views and restricts those plans to only working on views. I will test again on Monday as in retrospect that does seem odd behaviour.

src/SpeedyTransforms/SpeedyTransforms.jl
jackleland: Accidental commit, will roll back.

src/SpeedyTransforms/spectral_transform.jl
maximilian-gelbrecht: Ah right, it gets automatically converted by the constructor, right? I missed that.

maximilian-gelbrecht:
@milankl After my vacation, in early December, I could look into setting up CI for GPU. What do you think?

maximilian-gelbrecht:
I'll try to take care of it early in December when I'm back. 

@jackleland When you are writing unit tests, just commit them to a separate file in the test folder, and don't include those into the standard test suite in the `runtests.jl` file. 

In principle we are likely able to set up a separate CI pipeline on Builtkite that does the GPU only tests, so a reduced version of our tests just for the GPU functionality.

jackleland:
@maximilian-gelbrecht Possibly not the place to discuss this, but how does GPU CI work? Do you pay for public cloud hosting or is there some free service available?

ext/SpeedyWeatherCUDAExt/fourier.jl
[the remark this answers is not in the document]
jackleland: Turns out I just needed to apply the plan on the same stride array in `fourier_batched!` rather than on a subarray, but this means allocating in the transform and slows things down a bit. I think it's probably worth the small amount of duplicated code for a the speedup and lower memory use?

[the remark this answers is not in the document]
milankl: yes

ext/SpeedyWeatherCUDAExt/legendre.jl
```diff
@@ -0,0 +1,127 @@
+# convert i, j indices of a matrix (here 0-based l,m though...) to a single 1-based running index
+import LowerTriangularMatrices: ij2k
+
+# range of the running indices lm in a l-column (degrees of spherical harmonics)
+# given the column index m (order of harmonics) 
+get_lm_range(m, lmax) = ij2k(m, m, lmax):ij2k(lmax, m, lmax)
```
milankl, line 6: This function assumes 1-based m, lmax inputs. I uses the `ij2k` function from LowerTriangularMatrices to transform between i, j indices of a matrix (here l, m) to the running index `k` as it's called in that module (here lm).

ext/SpeedyWeatherCUDAExt/legendre.jl
```diff
@@ -0,0 +1,127 @@
...
+    lon_offsets,
+    kjm_indices
+)
+    tid = (CUDA.blockIdx().x - 1) * CUDA.blockDim().x + CUDA.threadIdx().x
+
+    if tid <= length(kjm_indices)
+        k, j, m  = kjm_indices[tid]
+
+        # are m, lmax 0-based here or 1-based? 
+        lm_range = get_lm_range(m, lmax)    # assumes 1-based
```
milankl, line 24: this needs to be adapted (+comment maybe!) for 0-based m, lmax

ext/SpeedyWeatherCUDAExt/legendre.jl
```diff
@@ -0,0 +1,127 @@
...
+    nlayers = axes(specs, 2)            # get number of layers of specs for fewer layers than precomputed in S
+
+    @boundscheck SpeedyTransforms.ismatching(S, specs) || throw(DimensionMismatch(S, specs))
+    @boundscheck size(g_north) == size(g_south) == (S.nfreq_max, S.nlayers, nlat_half) || throw(DimensionMismatch(S, specs))
+
+    g_north .= 0
+    g_south .= 0
+
+    # INVERSE LEGENDRE TRANSFORM by looping over wavenumbers l, m and layer k
+    kernel = CUDA.@cuda launch=false phase_factor_kernel!(
```
milankl, line 95: renamed this to `kernel`

ext/SpeedyWeatherCUDAExt/legendre.jl
```diff
@@ -0,0 +1,127 @@
+# convert i, j indices of a matrix (here 0-based l,m though...) to a single 1-based running index
+import SpeedyWeather.LowerTriangularMatrices: ij2k
+
+# range of the running indices lm in a l-column (degrees of spherical harmonics)
+# given the column index m (order of harmonics) 
+get_lm_range(m, lmax) = ij2k(m, m, lmax):ij2k(lmax, m, lmax)
+
+# (inverse) legendre transform kernel, called from _legendre!
+function phase_factor_kernel!(
```
milankl, line 9: this change hasn't been pushed yet?

jackleland: Not sure what change you're referring to?

milankl: Where it is actually called from?

jackleland: It's invoked by the `@cuda` macro, so lines 95 (and 109)

jackleland: The name is bad, I'll change it

milankl:
We talked about load balancing last week, which isn't trivial with the lower triangular matrices we have when mapping rows or columns to threads. However, one can two columns/rows, one long, one short together similar to how conceptually the triangle number can be explicitly computed, so instead of mapping row/column m=1,...,n to thread 1,...,n one could do
```julia
julia> m = 15
15

julia> for (r, i) in enumerate(0:m÷2)
           if r == 1 && isodd(m)
               println("thread $r: column $m")
           elseif r == 1
               println("thread $r: (idle)")
           else
               println("thread $r: column $i, $(m-i+iseven(m))")
           end
       end
thread 1: column 15
thread 2: column 1, 14
thread 3: column 2, 13
thread 4: column 3, 12
thread 5: column 4, 11
thread 6: column 5, 10
thread 7: column 6, 9
thread 8: column 7, 8

julia> m = 16
16

julia> for (r, i) in enumerate(0:m÷2)
           if r == 1 && isodd(m)
               println("thread $r: column $m")
           elseif r == 1
               println("thread $r: (idle)")
           else
               println("thread $r: column $i, $(m-i+iseven(m))")
           end
       end
thread 1: (idle)
thread 2: column 1, 16
thread 3: column 2, 15
thread 4: column 3, 14
thread 5: column 4, 13
thread 6: column 5, 12
thread 7: column 6, 11
thread 8: column 7, 10
thread 9: column 8, 9
```

(two examples one with even m one with odd m). Don't know whether this is actually helpful in practice might might be worth thinking about it

simone-silvestri:
excited to see SpeedyWeather running on GPUs! 

If it might be of any help, `nsys` and `ncu` are perfect tools to visualize the bottlenecks.
The usage is quite simple 
```julia
nsys profile --trace=cuda julia --project mytest.jl
```
and it is really great to see what the code does

jackleland:
> I'll look into CI.
> 
> For GPU CI, we'd just run those tests that absolutely need to run on GPU, on GPU. So, we need to define another test set. I'd suggest to just put all GPU tests in a new `test/gpu` folder.
> 
> @jackleland Do you already have some unit tests that you use locally for this?

Have been doing a lot of manual testing with notebooks, formalising into unit tests now

maximilian-gelbrecht:
GPU CI is on the way. We have access now, and I'll do a PR to set up the pipelines within the next days.

maximilian-gelbrecht:
The error in the regular CI is a bit odd, if you don't find the cause of it, I can look into it early next week.

jackleland:
> The error in the regular CI is a bit odd, it doesn't seem to be related to the GPU versions of the transforms at all. If you don't find the cause of it, I can look into it early next week.

It was being caused by a fix I found last year for  a problem I was having with `mapreducedim!` on LTA (Milan mentioned it in #558 I think), but seemingly it is causing ambiguity problems now. Not sure what changed between then and now that would cause that but I've removed the fix as I'm no longer using `mapreducedim!`, so no more error.

jackleland:
We now have a fully working round-trip transform on GPUs, with the caveat that I'm not sure it works on other grids just yet. I still need to:
- [ ] Add `transform(grid)` etc convenience functions for GPU
- [ ] Make better scaling benchmark plots
- [ ] Refactor redundant fft calling code
- [ ] Test with other grids
- [ ] Rewrite kernels in KA form (not sure this is necessary just yet?)
Before merging across, also happy to hear from others obviously. 

@milankl For the forward legendre kernel I ended up using an atomic add to deal with the race condition on reducing into the `specs` output, it seems to work pretty well but I had to allow some allocations in the kernel to get the @atomic call to work (it doesn't seem to like calculating indices or writing into views) but any suggestions on better ways of approaching this would be welcomed.

maximilian-gelbrecht:
Ah I see. Good to know. I'll give fixing this issue an attempt next week anyway. Seems important for other parts of the model as well.

milankl:
Wooohoo tests are passing! Let me know if you're ready for a review from my side!!

jackleland:
## Benchmark update:

Reran the benchmarks on the tranform components up to `T511, nlayers=64`, and wanted to get some feedback on the results. 

Speedup:
image

Actual times:
image

with the important bit in `benchmarks/CUDA/README.md`: 

### CPU() | Float32 | T511 L64 | OctahedralGaussianGrid | 768 Rings

| - | - | - | - |
| forward_legendre | 1.909 s| 0 bytes| 0 |
| inverse_legendre | 1.841 s| 0 bytes| 0 |
| forward_fourier | 237.968 ms| 96.00 KiB| 1536 |
| inverse_fourier | 237.545 ms| 108.00 KiB| 1536 |

### GPU() | Float32 | T511 L64 | OctahedralGaussianGrid | 768 Rings

| - | - | - | - |
| forward_legendre | 137.463 ms| 3.34 KiB| 117 |
| inverse_legendre | 145.007 ms| 4.00 KiB| 130 |
| forward_fourier | 15.543 ms| 2.25 MiB| 79874 |
| inverse_fourier | 28.117 ms| 4.55 MiB| 172034 |

Initially I ran it improperly (without `CUDA.sync`) and it seemed like the forward legendre, which uses an `@atomic` call, was 10-100x slower than the inverse, but after running it properly it seems they're roughly equal. Milan and I were concerned that the `@atomic` call would be a major bottleneck but _I no longer think that's the case_. 

@milankl I do think that the loop over `l` in the legendre kernels is probably the major bottleneck for larger values of `trunc`, so your [previous comment](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/602#issuecomment-2514372880) is probably a good shout as a next step. Another possibility is that we could unroll the loop over `l` into the thread index (i.e. make each kernel a function of `k, j, m, l`), which leaves us with another race condition on the cumulative sum for both the forward and inverse legendre transforms. We could therefore do either an atomic add for the cumulative sum or utilise shared memory and then reduce down at the end (more on that in a sec) but I feel we could easily overoptimise here...

---
# `@atomic` vs. shared memory scan

The current implementation of the forward legendre kernel utilises an `@atomic` call to avoid a race condition on the write into the output specs. A toy example to illustrate:

``` julia
# in the kernel

k, j, m = get_kjm_from_threadIdx(i)

# read from input arrays 
even = offset * input_north[m, k, j]
odd = offset * input_south[m, k, j]

# get values of l to loop over
lm_range = get_lm_range(m)

# loop over l and atomic add into output spectrum
for l in lm_range
    @atomic output[l, k] += legendre_polys[l, j] * even
    @atomic output[l+1, k] += legendre_polys[l+1, j] * odd
end
```

I've simplified but you get the rough idea: read a separate value from input for each `k, j, m` and then write to output mapped from only `m` and `k`, so the writes for all the different `j` values intersect, hence the need for an atomic. Milan has suggested that we could avoid the need for an `@atomic` call by utilising shared memory for each of the different j values and then combining into output, probably by the same logic used in a scan algorithm. The questions needing answers are therefore:
1) would this be possible?
2) would it provide noticeable speedup compared to the atomic implementation above?

I have a feeling that the answer to (2) is no, as there isn't much difference in the benchmarking between the forward and inverse legendre kernels where the forward has the `@atomic` call and the inverse doesn't (just a straight embarassingly parallel write), *but* this may change if we were to utilise more `@atomic` calls when unrolling the loop over `l` within each kernel.

milankl:
Also it seems that the bottleneck is the Legendre transform (forward/inverse) so we really should focus our efforts there. If I see this correctly, at T511L64 at the moment the Legendre transform is ~10x the cost of the Fourier transform, so if we get that faster by 10x we'd get effectively 5x on top, so 50x in total!?

ext/SpeedyWeatherCUDAExt/legendre.jl
```diff
@@ -0,0 +1,404 @@
+# convert i, j indices of a matrix (here 0-based l,m though...) to a single 1-based running index
+import SpeedyWeather.LowerTriangularMatrices: ij2k
+
+# range of the running indices lm in a l-column (degrees of spherical harmonics)
+# given the column index m (order of harmonics) 
+get_lm_range(m, lmax) = ij2k(2*m - 1, m, lmax):ij2k(lmax+m, m, lmax)
+get_2lm_range(m, lmax) = 2*ij2k(2*m - 1, m, lmax)-1:2*ij2k(lmax+m, m, lmax)
```
maximilian-gelbrecht, line 7: Those utility functions are quite useful outside of the GPU legendre transform as well (e.g. in some of the regular unit tests as well). We could also just move them into the regular LowerTriangularMatrices submodule.

src/SpeedyTransforms/spectral_transform.jl
```diff
@@ -143,38 +146,48 @@ function SpectralTransform(
...
-        rfft_plans_1D[j] = FFT_package.plan_rfft(real_vector_input, 1)
-        brfft_plans_1D[j] = FFT_package.plan_brfft(complex_vector_input, nlon, 1)
-    end
+    # PLAN THE FFTs
+    plan_FFTs!(
+        rfft_plans, brfft_plans, rfft_plans_1D, brfft_plans_1D,
+        fake_grid_data, scratch_memory_north, rings, nlons
+    )
     
+    # PRECOMPUTE KJM INDICES FOR LEGENDRE TRANSFORM (0-based)
```
maximilian-gelbrecht, line 174: Can we slightly more explicit in the documentation/comments what variables like `kjm` stands for? I was confused at first, but it's 
layer - ring - m (SPH number), right?

maximilian-gelbrecht (commented):
Very nice! 

Could you write a bit more on the general strategy that's applied for the Legendre transform? Especially as we go forward and adapt more functions. E.g. choosing block size, threads etc. Triangular matrices are of course a bit awkward to deal with. So right now it's just all columns in parallel and it doesn't matter that they are not equally long?

src/SpeedyTransforms/spectral_transform.jl
milankl: that's usually the indexing I've been using, `i` is the in-ring index (i.e. longitude), `j` is the ring index (latitude), `ij` is the running index for all grid points across rings, `l` is the spherical harmonics degree (1st dimension, row), `m` is the order (2nd dimension, column), `k` is for both grid and spherical harmonics any additional dimensions, i.e. vertical layer mostly.

milankl: Oh and `lm` is consistently a running index across all spherical harmonics ...

ext/SpeedyWeatherCUDAExt/fourier.jl
```diff
@@ -0,0 +1,132 @@
...
+    j::Int,
+    nfreq::Int,
+    ilons::UnitRange{Int};
+    not_equator::Bool = true
+) where {NF<:AbstractFloat, N}
+    rfft_plan = S.rfft_plans[j]     # FFT planned wrt nlon on ring
+    nlayers = size(grids, 2)        # number of vertical layers
+
+    if not_equator
+        view(f_out, 1:nfreq, 1:nlayers, j) .= rfft_plan * grids.data[ilons, :]
```
milankl, line 56: There seem's to be an in-place version implemented here

https://github.com/JuliaGPU/CUDA.jl/blob/b51c6228621cc37bc3fb40d6b3467487c8e02063/lib/cufft/fft.jl#L338

benchmark/CUDA/plot_benchmarks.jl
```diff
@@ -11,7 +11,7 @@ using CairoMakie
 include("benchmark_suite.jl")
 
 # Define the range of array sizes (N) to benchmark
-trunc_list = [15, 31, 63, 127, 255, 511]
+trunc_list = [15, 31, 63, 127, 255, 511, 1024]
```
milankl, line 14: Make it 1023 for consistency ;)

jackleland: Oops, yep

jackleland:
@milankl this is now good to merge

jackleland:
# Benchmark pngs to reference from the readme
benchmark_speedup
benchmark_times

milankl:
> @milankl this is now good to merge

PRESS THE BUTTON 💥  (when tests pass...)

PR #604, opened 2024-11-06
Rossby-Haurwitz wave initial conditions

rivwgk:
This pull request adds the height (eqs. 146-149 from Williamson et al. 1992) to the Rossby-Haurwitz wave initial condition.

There is a small difference that this implementation is not including $h_0$ in the height as otherwise this somehow produces NaNs, still the computed solution is comparable to results using other methods.

### DIFF
--- a/src/dynamics/initial_conditions.jl
+++ b/src/dynamics/initial_conditions.jl
@@ -302,6 +302,10 @@ Parameters are $(TYPEDFIELDS)"""
     m::Int = 4
     ω::Float64 = 7.848e-6
     K::Float64 = 7.848e-6
+    Ω::Float64 = 7.292e-5
+    a::Float64 = 6.37122e6
+    g::Float64 = 9.80616
+    h0::Float64 = 8e3
     c::Float64 = 1e-10
 end
 
@@ -313,20 +317,31 @@ function initialize!(
     initial_conditions::RossbyHaurwitzWave,
     model::AbstractModel,
 )
-    (; m, ω, K, c) = initial_conditions
+    (; m, ω, K, Ω, a, g, h0, c) = initial_conditions
     (; geometry) = model
 
     # Rossby-Haurwitz wave defined through vorticity ζ as a function of
     # longitude λ, latitude θ (in degrees), sigma level σ (vertically constant though)
     # see Williamson et al. 1992, J Computational Physics, eq 145
     ζ(λ, θ, σ) = 2ω*sind(θ) - K*sind(θ)*cosd(θ)^m*(m^2 + 3m + 2)*cosd(m*λ)
+    # see Williamson et al. 1992, J Computational Physics, eq 147 - 149, 146
+    A(λ, θ) = ω/2 * (2Ω+ω)*cosd(θ)^2 + K^2/4*cosd(θ)^(2m)*((m+1)*cosd(θ)^2 + (2m^2-m-2) - 2m^2/(cosd(θ)^2))
+    B(λ, θ) = (2(Ω+ω)*K)/((m+1)*(m+2))*cosd(θ)^m*( (m^2+2m+2) - (m+1)^2*cosd(θ)^2 )
+    C(λ, θ) = K^2/4*cosd(θ)^(2m)*((m+1) * cosd(θ)^2 - (m + 2))
+    #η(λ, θ) = h0 + a^2/g*(A(λ,θ) + B(λ,θ)*cosd(m*λ) + C(λ,θ)*cosd(2m*λ))
+    η(λ, θ) = a^2/g*(A(λ,θ) + B(λ,θ)*cosd(m*λ) + C(λ,θ)*cosd(2m*λ))
+
     set!(progn, geometry, vor = ζ)
+    set!(progn, geometry, pres = η)
     set!(progn, geometry, div = 0)  # technically not needed, but set to zero for completeness
 
     # filter low values below cutoff amplitude c
     vor = progn.vor[1]  # 1 = first leapfrog timestep
     low_values = abs.(vor) .< c
     vor[low_values] .= 0
+    pres = progn.pres[1]
+    low_value = abs.(pres) .< c
+    pres[low_value] .= 0
 
     return nothing
 end


src/dynamics/initial_conditions.jl
```diff
@@ -302,6 +302,10 @@ Parameters are $(TYPEDFIELDS)"""
     m::Int = 4
     ω::Float64 = 7.848e-6
     K::Float64 = 7.848e-6
+    Ω::Float64 = 7.292e-5
```
milankl, line 305: you find this in `model.planet.rotation`, this way it's consistent with the rotation rate as defined in the planet

src/dynamics/initial_conditions.jl
```diff
@@ -302,6 +302,10 @@ Parameters are $(TYPEDFIELDS)"""
...
+    a::Float64 = 6.37122e6
```
milankl, line 306: you find this in `spectral_grid.radius`, it may sound more intuitive to have `radius` in `planet`, but actually radius is also fundamental to choosing the resolution, it determines a lot in other components so it lives in `spectral_grid`.

src/dynamics/initial_conditions.jl
```diff
@@ -302,6 +302,10 @@ Parameters are $(TYPEDFIELDS)"""
...
+    g::Float64 = 9.80616
```
milankl, line 307: in `model.planet.gravity`

src/dynamics/initial_conditions.jl
```diff
@@ -302,6 +302,10 @@ Parameters are $(TYPEDFIELDS)"""
...
+    h0::Float64 = 8e3
```
milankl, line 308: there is technically `model.atmosphere.layer_thickness` which is used for the shallow water model, but as you correctly found out, the $\eta$ / `pres` in the model is actually the anomaly (like sea surface height / interface displacement), so you shouldn't add it anyway.

src/dynamics/initial_conditions.jl
```diff
@@ -313,20 +317,31 @@ function initialize!(
     initial_conditions::RossbyHaurwitzWave,
     model::AbstractModel,
```
milankl, line 318: You have different options here: Either you constraint this method here to `model::ShallowWater` as you're now defining something that only really makes sense in the shallow water model. Imagine now someone wants to use the Rossby-Haurwitz initial conditions in the primitive equation model, there `pres` has a different meaning (logarithm of surface pressure)

src/dynamics/initial_conditions.jl
```diff
@@ -313,20 +317,31 @@ function initialize!(
...
     ζ(λ, θ, σ) = 2ω*sind(θ) - K*sind(θ)*cosd(θ)^m*(m^2 + 3m + 2)*cosd(m*λ)
+    # see Williamson et al. 1992, J Computational Physics, eq 147 - 149, 146
+    A(λ, θ) = ω/2 * (2Ω+ω)*cosd(θ)^2 + K^2/4*cosd(θ)^(2m)*((m+1)*cosd(θ)^2 + (2m^2-m-2) - 2m^2/(cosd(θ)^2))
+    B(λ, θ) = (2(Ω+ω)*K)/((m+1)*(m+2))*cosd(θ)^m*( (m^2+2m+2) - (m+1)^2*cosd(θ)^2 )
+    C(λ, θ) = K^2/4*cosd(θ)^(2m)*((m+1) * cosd(θ)^2 - (m + 2))
+    #η(λ, θ) = h0 + a^2/g*(A(λ,θ) + B(λ,θ)*cosd(m*λ) + C(λ,θ)*cosd(2m*λ))
+    η(λ, θ) = a^2/g*(A(λ,θ) + B(λ,θ)*cosd(m*λ) + C(λ,θ)*cosd(2m*λ))
+
     set!(progn, geometry, vor = ζ)
+    set!(progn, geometry, pres = η)
```
milankl, line 335: alternatively (and I think that's the easier solution) you can put a `model isa ShallowWater &&` in front of this line, so only in the case of the model being a shallow water model the `pres` is actually set -- otherwise it just remains zero.

milankl:
Fantastic Tobias! I've put some comments in, let me know if you'd like to address them or if you want me to do it? There's also
- every pull request requires a change to our `CHANGELOG.md` so I'd add there under `# Unreleased` a line like `- Rossby-Haurwitz wave with initial conditions for interface displacement for shallow water models [#604](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/604)` -- that's why CI is currently failing.

rivwgk:
I addressed your comments in the recent commit. I will try to add a bit to the documentation.

src/dynamics/initial_conditions.jl
milankl: But then went for the conditional `set!` of eta here, I like!

milankl:
> I addressed your comments in the recent commit. I will try to add a bit to the documentation.

Awesome, thanks a lot, this looks great, let me know if you need any help!!

rivwgk:
now it should be ready (though, documentation is not finished yet) :) there was a part that I missed which was also shallow water exclusive

milankl (approved):

milankl:
> now it should be ready (though, documentation is not finished yet)

Thanks Tobias, feel free to push anytime!

rivwgk:
You're welcome. I just pushed a draft for the documentation. Though, I am not entirely sure about my changes for the last section.

milankl:
Fantastic I'll have a look and will make changes where needed!

src/dynamics/initial_conditions.jl
```diff
@@ -315,18 +315,34 @@ function initialize!(
 )
     (; m, ω, K, c) = initial_conditions
     (; geometry) = model
+    Ω = model.planet.rotation
+    a = model.spectral_grid.radius
```
milankl, line 319: you think we could rename the radius from `a` to `R`? That way it would be consistent with other parts of the model/documentation!

docs/src/initial_conditions.md
```diff
@@ -123,16 +123,62 @@ scalars, like `set!(simulation, div=0)` (which is always the case in the
...
+h(θ, λ) = h_0 + a^2 / g ( A(θ) + B(θ) \cos(Rλ) + C(θ) \cos(2Rλ),
+
+A(θ) = \frac{ω(2Ω + ω)}{2}\cos(θ)^2 + \frac{1}{4}K^2\cos(θ)^{2R}\left((R+1)\cos(θ)^2 + (2R^2 - R - 2) - \frac{2R^2}{\cos(θ)^2}\right),
+
+B(θ) = \frac{2(Ω + ω)K}{(R+1)(R+2)} \cos(θ)^R\left((R^2 + 2R + 2) - (R+1)^2\cos(θ)^2\right),
+
+C(θ) = \frac{1}{4}K^2 \cos(θ)^{2R}\left((R+1)\cos(θ)^2 - (R + 2)\right).
+```
+
+Where ``a`` is the circumference of the planet on which we consider the
```
milankl, line 141: It's the radius, no?

src/dynamics/initial_conditions.jl
milankl: I think it's now consistently called `R` for radius but `m` for order / the zonal wavenumber, could you check?

docs/src/initial_conditions.md
rivwgk: Yes, it's the radius.

rivwgk:
There were two places in the code, where the radius was still called `a`, which got renamed in the recent commit.

milankl (approved):

PR #609, opened 2024-11-19
large-scale condensation at <100% 

minqi6:
The problem is that large-scale condensation at <100% currently gives negative precipitation as the `ImplicitCondensation` isn't correctly generalised for other thresholds than 1=100%

from #614 

```julia
using SpeedyWeather
large_scale_condensation = ImplicitCondensation(spectral_grid, relative_humidity_threshold = 0.8)
convection = SimplifiedBettsMiller(spectral_grid)
model = PrimitiveWetModel(; spectral_grid, large_scale_condensation, convection)
simulation = initialize!(model, time=DateTime(2007, 1, 1))
run!(simulation, period=Day(20), output=true)
```

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - SpectralFilter for horizontal diffusion [#601](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/601)
 - GeoMakie weak dependency, globe function for 3D data visualisation [#600](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/600)
 - Zonal mean for AbstractGridArray [#603](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/603)
+- bugfix: large-scale condensation also at <100% [#609](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/609)
 
 ## v0.12.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,15 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


milankl:
The tests are running now, would also want to write some additional ones simply testing what you tried with `0.8, 0.9, 1.0` and checking that a 5-day simulation is stable?

milankl:
1) We have a [changelog](https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/main/CHANGELOG.md) that needs to be updated with every pull request, you can write something like `- bugfix: large-scale condensation also at <100% [#609](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/609)` this just helps us to keep track of changes that were made in different versions of the model. [Past versions](https://github.com/SpeedyWeather/SpeedyWeather.jl/releases/tag/v0.12.0) have this too but it sometimes also good to have one file with a list of all changes listed by version

2) The tests we'd need would look something like

```julia
@testset "Large-scale condensation" begin
    @testset for r in (0.8, 0.9, 1.0)
        large_scale_condensation = ImplicitCondensation(spectral_grid, relative_humidity_threshold=r)
        model = PrimitiveWetModel(spectral_grid; large_scale_condensation)
        run!(simulation, period=Day(5)
        
        precip = simulation.diagnostic_variables.physics.precip_large_scale

        for ij in eachindex(precip)
            @test precip[ij] >= 0   # precipitation should always be non-negative
        end
    end
end
```
I would create a new file in `test/large_scale_condensation.jl` which then needs to be included in `test/runtests.jl`

PR #613, opened 2024-11-25
Slab ocean and net surface and top-of-atmosphere fluxes

milankl:
After 20 days at T31 the outgoing/reflected shortwave radiation (transparent clouds, just albedo reflection and cos zenith based diurnal cycle)

image

surface fluxes positive down, including transparent shortwave and surface temperature fluxes 

image

SlabOcean (no slab land) that heats up (unrealistically much though...)

image

Similarly that's the outgoing longwave radiation OLR

image


milankl:
Merging this as the introduced new components aren't default and further changes should be made with #624 merged too. Both PRs seem to work fine from a software engineering perspective and from a modelling perspective we need to make sure that these don't cause runaway temperatures

PR #623, opened 2024-12-01
Change timestep to 40min at T31

milankl:
This is what Fortran SPEEDY uses but I had shortened this to 30min a while back to be on the safer side. Just testing whether we could make the time step longer again to gain some additional +33% speedup


milankl:
@maximilian-gelbrecht asking for your opinion here, this would speedup the default model, is the same time step that Fortran SPEEDY uses that I only had reduced at some point to avoid instabilities. But it doesn't seem that there's any need for a 30min time step anymore and we could increase that to 40min (at T31). However, it's tricky to test this for higher resolutions in all cases with combinations of parameterizations etc. I'm tempted to make this the new default without knowing exactly what will happen at T127 for example. What do you think?

maximilian-gelbrecht:
Where's the time step scaling actually coming from that we use? 

One could also be more conservative for higher resolutions that we don't test thoroughly.

milankl:
> Where's the time step scaling actually coming from that we use?

It's linear, which is just based on CFL condition $$U\Delta t/\Delta x$$. To have a constant CFL condition, if you half the grid space you half the time step, assuming the characteristic speed $$U$$ doesn't change, if $$U$$ increases with resolution you technically would need to decrease $$\Delta t$$. That's definitely something you have to be careful with in the ocean between eddy/non-eddy resolving grids but not that much of a problem in the atmosphere as strong wind speeds in the free troposphere are usually quite laminar.

> One could also be more conservative for higher resolutions that we don't test thoroughly.

Yeah, but it's not clear what scaling law should apply other than linear?

maximilian-gelbrecht:
Ok, I see. 

From my point of view we can change it to 30min. 

If we want to be extra thorough, we could also make a few more high resolutions runs part of the extended test set that we have defined now.

milankl:
> If we want to be extra thorough, we could also make a few more high resolutions runs part of the extended test set that we have defined now.

That's a good call. I've just tagged the current main as v0.13 and then we can merge this branch here towards v0.14 which gives us more time to test that the longer time step doesn't cause any problems

milankl:
I want to make the new radiation and slab ocean default anyway which I need to test more

PR #625, opened 2024-12-02
[SpeedyTransforms] Restrict finite difference tests to one grid only 

maximilian-gelbrecht:
The AD finite differences test take quite a while in the CI, this does those just for one grid type. For the other grid the transform identity is still tested.


milankl:
reminder to squash merge 😉 Otherwise the number of commits in this repo is increasing too quickly which increases the repo size ... but tbh I've never figured out why the repo size is 200MB?

image

maximilian-gelbrecht:
We could remove all the old branches and run the `git gc`

milankl:
yes, feel free to delete your old branches, I don't want delete yours though!

PR #628, opened 2024-12-03
Introduce separate extended tests in CI 

maximilian-gelbrecht:
This introduces a flag `FLAG_EXTENDED_TESTS` in the tests that mark tests that are not run in the regular CI, but only in the weekly CI. 

The first tests that went into this category are the more high-level differentiability tests. The low-level `test_reverse` from `EnzymeTestUtils` are staying in the main tests.

### DIFF
--- a/.github/workflows/extended_tests.yml
+++ b/.github/workflows/extended_tests.yml
@@ -1,0 +1,29 @@
+name: Extended Test
+# Weekly extended tests
+on:
+  schedule:
+    - cron: '0 2 * * 0'  # Daily at 2 AM UTC every Sunday
+jobs:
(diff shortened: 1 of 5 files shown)


milankl (commented):
Happy with this but it would always run extended tests on nightly not on the currently supported version like 1.10 or 1.11 atm, is that what we would want? Alternatively, we could create `.github/workflows/extended_tests.yml` but run with current minor release independent of the nightly workflow?

maximilian-gelbrecht:
Ok, that should be ready to go now. I'll merge it later today if you don't object

milankl (approved):

Issue #598, opened 2024-10-29
Differentiability of Spatial Gradients 


maximilian-gelbrecht:
After/with #589, I continued my hunt for differentiability with Enzyme a bit. With the transforms (more or less) differentiable, I found the first error from Enzyme at the spatial gradients, more precisely `_divergence!`. 

As it is, Enzyme seems to have a problem with the way we write these methods with function kernels/closures that are conditional. Because as the following shows, Enzyme only yields an error when we use a kernel with a conditional closure as we commonly do in our code for these functions: 

```Julia 
import Pkg
Pkg.activate(".")

using SpeedyWeather
using Enzyme

grid_types = [FullGaussianGrid] #, OctahedralGaussianGrid] # one full and one reduced grid, both Gaussian to have exact transforms 
grid_dealiasing = [2] #, 3]

grid_type = grid_types[1]
i_grid = 1 

spectral_grid = SpectralGrid(Grid=grid_type, dealiasing=grid_dealiasing[i_grid])

# forwards 
S = SpectralTransform(spectral_grid)
dS = deepcopy(S)

u_grid = rand(spectral_grid.Grid{spectral_grid.NF}, spectral_grid.nlat_half, spectral_grid.nlayers)
v_grid = rand(spectral_grid.Grid{spectral_grid.NF}, spectral_grid.nlat_half, spectral_grid.nlayers)

u = transform(u_grid, S)
v = transform(v_grid, S)
du = zero(u)
dv = zero(v)

div = zero(u)
ddiv = zero(u)
fill!(ddiv, 1+1im)

func(o, a, b, c) = a - b + c

# no error 
autodiff(Reverse, SpeedyWeather.SpeedyTransforms._divergence!, Const, Const(func), Duplicated(div, ddiv), Duplicated(u, du), Duplicated(v, dv), Duplicated(S, dS)) # this doesn't give an error directly

mul = 2
kernel2(o, a, b, c) = mul .* (a-b+c)  # just some arbitrary closure over a function

# no error 
autodiff(Reverse, SpeedyWeather.SpeedyTransforms._divergence!, Const, Const(kernel2), Duplicated(div, ddiv), Duplicated(u, du), Duplicated(v, dv), Duplicated(S, dS)) # this also doesn't give an error directly

add = false
kernel(o, a, b, c) = add ? o-(a-b+c) : -(a-b+c) # this is (similar to) what we actually use in divergence!

# directly an error
[...]

PR #638, opened 2024-12-13
Spectral Gradients Gradients

maximilian-gelbrecht:
This is WIP to add tests to check the correctness of the gradients of the spectral gradients. 

While some of the underlying implementation will possibly change (and GPU-ifiyed). The tests of the gradients of `curl!`, `divergence!` will stay the same. 

Part of the extended test set. 

Currently `curl!` is passing and `divergence!` failing.


maximilian-gelbrecht:
Making some progress. So far, so good. 

I am slightly confused by what I thought would be a wrong sign based on the [formula](https://speedyweather.github.io/SpeedyWeather.jl/dev/spectral_transform/#Derivatives-in-spherical-coordinates)  in the curl/div gradient, but Enzyme and FD both agree on it, so I must have a mistake in my thinking there.

milankl:
Something I screwed up in the docs?

maximilian-gelbrecht:
No, code and docs both say the same thing how it's computed. But the gradient of curl/div wrt u/v returns a "-im". But maybe I am also making in a mistake with how I call Enzyme and FD / the seed.

maximilian-gelbrecht:
All Gradients of Spectral Gradients are working now.

I tried to set up a MWE why the closure style of kernel functions won't work with Enzyme, but the problem doesn't appear when setting up a MWE without Speedy. Maybe it's some weird interaction with `LowerTriangularArray`. But as the anonymous function kernel work, and FD verifies the correctness, I am not too worried about this.

maximilian-gelbrecht:
FYI: With this PR, differentiating through a single time step of a barotropic model with Enzyme at the very least doesn't return an error. After the holidays, I'll investigate if the gradient is correct as well and what else already works.

milankl:
So cool 😍

maximilian-gelbrecht:
After the discussion in #598, I think we can merge this now. Anonymous function are the recommended way for this anyway as it seems. 

In the next PR I'll look into the whole barotropic model and probably also structure the whole differentiability tests a little bit better.

milankl:
Feel free to replace all kernels with anonymous functions btw! But I think we only use them in `spectral_gradients.jl`?

maximilian-gelbrecht:
Yeah good call, there were some for `horizontal_advection!` and `set!`. 

Once all tests pass now, I'll merge it.

PR #639, opened 2024-12-14
ConvectiveHeating implemented

### LINKED ISSUE #567: Implement `Heating` as model component
For example to implement a tropical temperature tendency lie [Lee and Kim, 2003](https://doi.org/10.1175/1520-0469(2003)060%3C1490:TDRBSA%3E2.0.CO;2)

Suggested to have a `AbstractHeating`, then `ConvectiveHeating <: AbstractHeating` probably living in `ConvectiveHeating.jl` repository.

image

Suggested by @KristianJS

milankl:
closes #567


milankl:
@KristianJS to be used like

```julia
convection = ConvectiveHeating(spectral_grid, p₀ = 525) # specify non-defaults as keyword args here     
model = PrimitiveDryModel(spectral_grid; convection)    # pass on to model constructor
run!(simulation)
```

`ConvectiveHeating` is implemented as a parameterization (not a forcing) given it's formulated in pressure coordinates and especially because it's presumably not to be used in combination with other convection schemes. So passing `convection` here on will use that instead of the more complicated default Betts-Miller scheme. Not tested yet, would you?

PR #645, opened 2024-12-17
GPU CI Pipeline 

maximilian-gelbrecht:
This merges into `jl/fourier_cuda` 

This is currently a "draft", but marked as a regular PR to hopefully trigger the pipeline.

PR #646, opened 2024-12-18
Dummy Buildkite Pipeline

maximilian-gelbrecht:
See #645, to not get errors while our proper pipeline and GPU tests are not in `main` yet.


maximilian-gelbrecht:
yes

PR #647, opened 2025-01-03
Grid coordinates defined in degrees not colatitudes and radians

milankl:
- [x] `get_lond`, `get_latd`, `get_londlatds` as the lowest-level functions, others call those.
- [x] Other functionality adpated to the always lon, lat not lat, lon logic
- [x] Also assume lon, lat (not lat, lon) for interpolation

### DIFF
--- a/src/dynamics/geometry.jl
+++ b/src/dynamics/geometry.jl
@@ -23,9 +23,6 @@ $(TYPEDFIELDS)
     "maximum number of longitudes (at/around Equator)"
     nlon_max::Int = get_nlon_max(Grid, nlat_half)
 
-    "=nlon_max, same (used for compatibility), TODO: still needed?"
-    nlon::Int = nlon_max
-
     "number of latitude rings"
     nlat::Int = spectral_grid.nlat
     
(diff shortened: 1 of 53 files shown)


milankl:
@maximilian-gelbrecht don't remember where we discussed this, but this also aims to make lon, lat; not lat, lon the default order for interpolation which you correctly pointed out was inconsistent!

src/dynamics/geometry.jl
```diff
@@ -7,22 +7,22 @@ Construct Geometry struct containing parameters and arrays describing an iso-lat
 and the vertical levels. Pass on `SpectralGrid` to calculate the following fields
 $(TYPEDFIELDS)
 """
-@kwdef struct Geometry{NF<:AbstractFloat, Grid<:AbstractGrid} <: AbstractGeometry
+@kwdef struct Geometry{
+    NF,
+    Grid,
+    VectorType,
```
maximilian-gelbrecht, line 13: I am not even sure we need everything from the Geometry on GPU for the GPU model. 

In principle, indices stay on the CPU, even when it's a GPU model. And indices are sometimes derived from the geometry, aren't they? 

Of course, it doesn't hurt to write the struct like that, and in the end we can still just always initialise with `Array` if that's what we want to use.

maximilian-gelbrecht (commented):
Nice! I like this consistency. What made you do the change now? 

I think we had this discussion in something RingGrids related before.

src/dynamics/geometry.jl
milankl: Yeah, for `Geometry` I don't think it's necessary to move things to the GPU, but actually the `coslat` vectors would be good. For the interpolation though, we'll need that as it's actually "data".

PR #650, opened 2025-01-07
`set!` for time step

milankl:
To be used like

```julia
spectral_grid = SpectralGrid(trunc=63)
L = Leapfrog(spectral_grid)
set!(L, Minute(10))
```

Also disables the adjustment of the timestep towards the desired output frequency and recalculates all dt-based fields in `Leapfrog`.

### DIFF
--- a/.github/workflows/extended_tests.yml
+++ b/.github/workflows/extended_tests.yml
@@ -9,6 +9,7 @@ jobs:
     strategy:
       matrix:
         version:
+          - '1.10'
           - '1.11'
         os:
           - ubuntu-latest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - set! for time step [#650](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/650)
+- Bug: ZonalWind had sqrt of negative for uncommon resolutions [#649](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/649)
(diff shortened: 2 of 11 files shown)


glwagner:
This will probably be useful for running simulations in conditions that differ from the ones typical to Earth today, especially idealized simulations.

maximilian-gelbrecht (approved):
Nice! 

Maybe just make it clear in the docs (string) that this is meant as the actual time step and not the scaled time set that we use a keyword to configure the model. Could be a bit confusing for users otherwise.

milankl:
I've decided to allow for positional argument and keyword argument, i.e.
```julia
set!(L, Δt)
set!(L, Δt = Minute(10))
```
but feel that Δt_at_T31 should be just set when creating a `Leapfrog` in the first place?

glwagner:
Out of curiosity, what is the relevance of "time step at T31" if you are not using the T31 grid?

milankl:
> Out of curiosity, what is the relevance of "time step at T31" if you are not using the T31 grid?

It just scales linearly. If you have 40min at T31, then it's 20min at T63, 10min at T127, 5min ... etc. So normally a user wouldn't need to worry about the time step, but if they want the timestep to be, say, 10% shorter, then choose 36min at T31 and whatever resolution you run at is then automatically scaled to. We just wanted to hide the issue of figuring out what a stable time step size is given some resolution as best as possible

glwagner:
Okay, so if you are using T127 and want a shorter time-step, then a user should figure out what the T31 time-step is and shorten that by an appropriate amount to obtain the desired scaling of the T127 time step?

Is the time-step determined by a CFL limitation?

PR #652, opened 2025-01-07
Group initialize!, timestep!, finalize! of main time loop

milankl:
@simone-silvestri this is to bundle functionality into

```julia
initialize!(simulation)
timstep!(simulation)
finalize!(simulation)
```

### DIFF
--- a/src/dynamics/time_integration.jl
+++ b/src/dynamics/time_integration.jl
@@ -370,7 +370,7 @@ Initializes a `simulation`. Scales the variables, initializes
 the output, stores initial conditions, initializes the progress meter feedback,
 callbacks and performs the first two initial time steps to spin up the
 leapfrogging scheme."""
-function initialize!(simulaton::AbstractSimulation)
+function initialize!(simulation::AbstractSimulation)
     progn = simulation.prognostic_variables         # unpack stuff
     diagn = simulation.diagnostic_variables
     (; clock) = progn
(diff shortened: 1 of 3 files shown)


maximilian-gelbrecht:
This is great! I was about the suggest something like this because it makes working on the differentiability of the time stepping also much easier.

src/dynamics/time_integration.jl
```diff
@@ -365,25 +365,23 @@ function timestep!(
...
-    diagn::DiagnosticVariables,         # all pre-allocated diagnostic variables
-    model::AbstractModel,               # all model components
-)          
-    
+"""$(TYPEDSIGNATURES)
+Initializes a `simulation`. Scales the variables, initializes
+the output, stores initial conditions, initializes the progress meter feedback,
+callbacks and performs the first two initial time steps to spin up the
+leapfrogging scheme."""
+function initialize!(simulation::AbstractSimulation)
```
simone-silvestri, line 373: So the first time steps are included in the `initialize!(simulation)` function, which means that when we start the time stepping with `timestep!(simulation)` the simulation will be at a `time = 2Δt`?

Could this be made optional so that we can start timestepping at `time = 0`? 
I wonder if we could also embed those `first_timesteps!` in the `timestep!` function, something like
```
function timestep!(simulation)
   if iter == 1
       first_timestep!()
   elseif iter == 2 
       second_timestep!()
   else
       timestep!()
   end
end
```
This would make it easier to couple. However, if it is not possible we can work around it by resetting the clock.

milankl: at `time = Δt` (the first steps are smaller) but yes. I see your point that this would be more consistent. The first time step is then actually two, but that should be fine, conceptually they count as one, also given they step forward dt only. I'll make another PR to change that.

PR #654, opened 2025-01-09
bug: south pole at -90 for vertices

milankl:
@asinghvi17 found this issue that the southernmost faces aren't properly drawn. Problem is just a missing sign

image


milankl:
Fixed: 

image

PR #656, opened 2025-01-09
Differentiability tests for time stepping 

maximilian-gelbrecht:
Work in progress

While doing so I restructered all the differentiability tests und put them in a separate folder.


maximilian-gelbrecht:
I am struggling a bit with correctness checks with FiniteDifferences.jl at the moment, but I also made some attempts at differentiation through the other models. ShallowWater works, I am working on the PrimitiveModel 

@milankl  
One thing I caught: Enzyme really doesn't like type instabilities. We had a number of time lines like 
```Julia 
some_variable = flag ? model.parameter : 0 
``` 
Enzyme always caused errors for this. I adjusted all of these that I found, but we have to avoid them in the future as well.

maximilian-gelbrecht:
There seems to be some problem with the way we do broadcasting for the RingGrids. 

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/78258930405ab13ae9c77db065c4aaf873fa4b4f/src/dynamics/tendencies.jl#L855-L860

Caused segmentation faults while differentiating via Enzyme. 

Changing those at least temporaliy to 

```Julia 
@. temp_grid_prev.data = temp_grid.data
```

worked, and the seg faults didn't appear again. 

With this change I was also able to differentiate through a single time step of the `PrimitiveWetModel` as well. So, while there's still some work to do, that's already great news 🎉

maximilian-gelbrecht:
To-Do list 

- [ ] Do proper correctness tests with FiniteDifferences.jl
- [ ] What's wrong RingGrids 
- [ ] Think about proper test cases of the differentiability (might not be part of this PR)

ext/SpeedyWeatherEnzymeExt.jl
```diff
@@ -17,23 +19,23 @@ function adjoint_scale(S::SpectralTransform)
...
 end 
 
 # Computes the scale for the adjoint/pullback of a real discrete fourier transform.
 function rfft_adjoint_scale(n_freq::Int, n_real::Int)
     if iseven(n_real)
-        return [1; [2 for i=2:(n_freq-1)]; 1]
+        return [1 < i < n_freq ? 2 : 1 for i=1:n_freq]
     else 
-        return [1; [2 for i=2:n_freq]]
+        return [1 < i ? 2 : 1 for i=1:n_freq]
```
maximilian-gelbrecht, line 38: The old way of computing the adjoint scale caused an error with Enzyme and I don't know why. This new one is faster any way though.

ext/SpeedyWeatherEnzymeExt.jl
```diff
@@ -17,23 +19,23 @@ function adjoint_scale(S::SpectralTransform)
...
+    scale = zeros(Int, maximum(nfreqs), 1, nlat_half) # the scratch memory is (Freq x lvl x lat), so we insert 
+                                                      # an additional dimension here for easier matrix multiply
 
     for i=1:nlat_half
-        scale[1:nfreqs[i],i] = rfft_adjoint_scale(nfreqs[i], nlons[i])
+        scale[1:nfreqs[i],1,i] = rfft_adjoint_scale(nfreqs[i], nlons[i])
     end 
 
-    # TODO: transfer array to GPU in case we are on GPU
-    return reshape(scale, maximum(nfreqs), 1, nlat_half) # the scratch memory is (Freq x lvl x lat), so we insert
```
maximilian-gelbrecht, line 27: The reshape caused an error with Enzyme (segmentation fault). I am not sure why, but the new way of just directly creating the array is more efficient any way.

src/dynamics/clock.jl
```diff
@@ -38,6 +38,18 @@ function Base.show(io::IO, C::Clock)
...
+function Base.copy!(clock::Clock, clock_old::Clock)
+    clock.time = clock_old.time 
+    clock.start = clock_old.start 
+    clock.period = clock_old.period 
+    clock.timestep_counter = clock_old.timestep_counter
+    clock.n_timesteps = clock_old.n_timesteps
+    clock.Δt = clock_old.Δt 
+    
+    return nothing 
+end
```
milankl, line 51: Looks good, but why is that needed? I'm suspecting to restart a simulation? Cause in that case we'd need to check whether all of them are supposed to just be copied over or whether some of them should be set differently afterwards

ext/SpeedyWeatherEnzymeExt.jl
milankl: I think the former allocates twice (throwing away the inner vector) the latter only allocates once

ext/SpeedyWeatherFiniteDifferencesExt.jl
```diff
@@ -27,4 +27,79 @@ function FiniteDifferences.to_vec(x::LTA) where LTA <: LowerTriangularArray
...
+
+included_fields(::Type{<:GridVariables}) = (:vor_grid, :div_grid, :temp_grid, :temp_virt_grid, :humid_grid, :u_grid, :v_grid, :pres_grid, :tracers_grid, :random_pattern, :temp_grid_prev, :humid_grid_prev, :u_grid_prev, :v_grid_prev, :pres_grid_prev, :tracers_grid_prev)
+excluded_fields_pre(::Type{<:GridVariables}) = (:nlat_half, :nlayers)
+excluded_fields_post(::Type{<:GridVariables}) = ()
+
+included_fields(::Type{<:DynamicsVariables}) = (:a, :b, :a_grid, :b_grid, :a_2D, :b_2D, :a_2D_grid, :b_2D_grid, :uv∇lnp, :uv∇lnp_sum_above, :div_sum_above, :temp_virt, :geopot, :σ_tend, :∇lnp_x, :∇lnp_y, :u_mean_grid, :v_mean_grid, :div_mean_grid, :div_mean)
+excluded_fields_pre(::Type{<:DynamicsVariables}) = (:trunc, :nlat_half, :nlayers)
+excluded_fields_post(::Type{<:DynamicsVariables}) = ()
+
+included_fields(::Type{<:PhysicsVariables}) = (:precip_large_scale, :precip_convection, :precip_rate_large_scale, :precip_rate_convection, :cloud_top, :soil_moisture_availability, :sensible_heat_flux, :evaporative_flux, :surface_shortwave_up, :surface_shortwave_down, :surface_longwave_up, :surface_longwave_down, :outgoing_shortwave_radiation, :outgoing_longwave_radiation, :cos_zenith)
```
milankl, line 97: these all have to be hardcoded? Because I'm frequently changing these fields. But note that everything in `diagnostic_variables.physics` is actually only stored from the parameterizations for output. None of these values feed back into the model. Well here in this version they do but with #667 merged I moved the prescribed heat fluxes into the prognostic variables as they do influence the simulation.

test/differentiability/test_utils.jl
```diff
@@ -0,0 +1,38 @@
...
+    end 
+
+    nvars_grid = 6 
+    land_ocean_array = zeros(GridVariable2D, nlat_half, nvars_grid)
+    land_ocean_array[:, 1] = prog.ocean.sea_surface_temperature
+    land_ocean_array[:, 2] = prog.ocean.sea_ice_concentration
+    land_ocean_array[:, 3] = prog.land.land_surface_temperature
+    land_ocean_array[:, 4] = prog.land.snow_depth
+    land_ocean_array[:, 5] = prog.land.soil_moisture_layer1
+    land_ocean_array[:, 6] = prog.land.soil_moisture_layer2
```
milankl, line 29: land temperature and soil moisture will become 3D grids with some flexible number of vertical layers, just as a heads up that this will fail at some point in the future.

milankl:
Sorry I changed a bunch of stuff in `src/physics/surface_fluxes.jl` causing a conflict but your idea of writing `zero(NF)` still applies!

src/dynamics/clock.jl
maximilian-gelbrecht: As you see in the code of the test, we really need to copy the structs quite often. I also sometimes use out-of-place forms of the functions, so my intend was to really have them identical if I copy them. Intuitively that's also what I would expect from a `copy!` function. 
I am happy to revert the change though, I also realise now that it's a bit more directly related to the integration. But maybe we want to have it like this? Because before a new integration is started `set_period` and `initialise!(::Clock)` is called again anyway. So this doesn't break anything as far as I can see it.

ext/SpeedyWeatherFiniteDifferencesExt.jl
maximilian-gelbrecht: Yeah, I totally see that this is maintenance heavy code to do it like this. It was a quick thing to code it like that. 

Let's see how could we do this different. I think this could probably also be just based on the types of the fields. E.g. exclude `Int`, include all `AbstractArray` and container formats.

test/differentiability/test_utils.jl
maximilian-gelbrecht: I think I will remove this function anyway. I had this still in from an earlier version, in which I hadn't coded the `to_vec` yet. No it's just used for exactly one test and for this test I could also use the `to_vec` as well.

src/dynamics/clock.jl
milankl: No I agree that's what I expect a `copy!` function to do, but you know that `deepcopy` is always defined for all types?
```julia
julia> struct S
           a
       end

julia> s = S(1)
S(1)

julia> deepcopy(s)
S(1)
```

ext/SpeedyWeatherFiniteDifferencesExt.jl
maximilian-gelbrecht: I committed a much less hardcoded version. The types are still hardcoded though.

test/differentiability/test_utils.jl
maximilian-gelbrecht: I removed this function now

src/dynamics/clock.jl
maximilian-gelbrecht: yeah,`deepcopy` is very generic and works for all types, `copy` has to be specifically defined.

maximilian-gelbrecht:
Mh, merging #667 did break the differentiability again, even after fixing the type instabilities. I'll look into it. 

For the future, I already included a test in the regular CI as well to catch this early.

maximilian-gelbrecht:
After minor fixes, differentiating through the `timestep!` does work again. The problem is actually the seemingly simple `copy!(::PrognosticVariables, ::PrognosticVariables)`, which might be related to the problem described above about RingGrids. It's hard to condense this down to a MWE, but I have to try when I find the time.

milankl:
Some renamings relevant for `copy!(::PrognosticVariables, ::PrognosticVariables)` coming with #671 so that we have the following names

```julia
julia> simulation.prognostic_variables
├ vor:   T31, 8-layer, 2-steps LowerTriangularArray{Float32}
├ div:   T31, 8-layer, 2-steps LowerTriangularArray{Float32}
├ temp:  T31, 8-layer, 2-steps LowerTriangularArray{Float32}
├ humid: T31, 8-layer, 2-steps LowerTriangularArray{Float32}
├ pres:  T31, 1-layer, 2-steps LowerTriangularArray{Float32}
├ random_pattern: T31, 1-layer LowerTriangularArray{Float32}
├┐ocean: PrognosticVariablesOcean{Float32}
│├ sea_surface_temperature:  48-ring OctahedralGaussianGrid{Float32}
│└ sea_ice_concentration:    48-ring OctahedralGaussianGrid{Float32}
├┐land:  PrognosticVariablesLand{Float32}
│├ soil_temperature: 3-layer, 48-ring OctahedralGaussianGrid{Float32}
│├ soil_moisture:    3-layer, 48-ring OctahedralGaussianGrid{Float32}
│└ snow_depth: 48-ring OctahedralGaussianGrid{Float32}
├ tracers: 0, Symbol[]
├ scale: 1.0
└ clock: 2000-01-21T00:00:00
```

this is necessary as we now have n soil layers meaning `land_surface_temperature` (uppermost layer) is actually a soil temperature. And `soil_moisture_layer1` and `_layer2` are also a 3D grid variables now condensing those 2 variables together

maximilian-gelbrecht:
I found the issue regarding RingGrids 🎉

We were apperently missing definitions of `Base.dataids` in our `RingGrids` that help unalias mutable variables. The doctoring of that function do actually recommend that array wrapper struct do implement it, but it's not part of the public API and otherwise not well documented.

maximilian-gelbrecht:
Some problems with Enzyme in Julia 1.11, a minimal example without SpeedyWeather posted there https://github.com/EnzymeAD/Enzyme.jl/issues/2295

maximilian-gelbrecht:
I was able to fix the correctness tests for the high-level `timestepping!` tests now quite quickly. So it seems we can in fact differentiate through all models correctly! 

There's still something wrong for the more detailed tests of the primitive equation model, but I assume that's just user error, as the more high-level tests are correct. 

There's not so much remaining to do here. Enzyme in Julia 1.11 is a bit of an issue, but also partially outside of our control. 

@milankl I am off work for a week, I think this is already a good time for a proper code review in the meanwhile

docs/Project.toml
```diff
@@ -1,6 +1,7 @@
 [deps]
 CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
```
maximilian-gelbrecht, line 4: Adding Enzyme to the docs and actually running the example code, makes the whole CI process for the docs much longer. We may also just hardcode the Enzyme example in the docs.

maximilian-gelbrecht:
As an outlook: I did already try to differentiate through multiple time steps with `run!` as well. And this does seem to work (correctness not separately checked), but of course for any actually longer runs, naively differentiating through everything becomes unfeasible due to the memory usage. This is when one needs to think about checkpointing e.g. with Checkpointing.jl.

docs/Project.toml
milankl: Happy with that for now! We can later check how to automate tests if that's currently unfeasible

maximilian-gelbrecht:
I actually skip the test in the regular CI for Julia 1.11. I think that's fine for now. I'd rather merge this sooner than later, and continue work on differentiability in further PRs.

src/physics/land/land.jl
```diff
@@ -15,14 +15,14 @@ export LandModel
...
-function LandModel(SG::SpectralGrid)
+#=function LandModel(SG::SpectralGrid)
     return LandModel(
         SeasonalLandTemperature(SG),
         SeasonalSoilMoisture(SG),
         VegetationClimatology(SG),
         NoRivers(SG),
     )
-end
+end =#
```
maximilian-gelbrecht, line 25: I removed/commented off this constructor. The warning with the double definition of the constructor stops the pre-compilation which in turn messes with the extensions. Without the extensions working properly the differentiability tests also don't work

src/physics/surface_fluxes/heat.jl
```diff
@@ -49,10 +49,10 @@ SurfaceOceanHeatFlux(SG::SpectralGrid; kwargs...) = SurfaceOceanHeatFlux{SG.NF}(
 initialize!(::SurfaceOceanHeatFlux, ::PrimitiveEquation) = nothing
 
 function surface_heat_flux!(
-    column::ColumnVariables,
+    column::ColumnVariables{NF},
```
milankl, line 52: I'm wondering whether he should generally use `eltype(column)` instead of the `where NF` syntax?

docs/src/differentiability.md
```diff
@@ -0,0 +1,56 @@
...
+```julia
+dprogn = one(progn) # shadow for the progn values 
+ddiagn = make_zero(diagn) # shadow for the diagn values 
+dmodel = make_zero(model) # here, we'll accumulate all parameter derivatives 
+```
+
+Then, we can already do the differentiation with Enzyme
+
+```julia
+autodiff(Reverse, SpeedyWeather.timestep!, Const, Duplicated(progn, dprogn), Duplicated(diagn, ddiagn), Const(dt), Duplicated(model, dmodel))
```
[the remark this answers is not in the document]
maximilian-gelbrecht: Yes, I tried and the documentation job timed out.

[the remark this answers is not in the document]
maximilian-gelbrecht: Which is a bit weird though, because running just that example and the `autodiff` command doesn't take THAT long.

src/physics/surface_fluxes/heat.jl
maximilian-gelbrecht: Currently we don't have `eltype(column)` explicitly defined, so it yields `Any`

src/dynamics/prognostic_variables.jl
```diff
@@ -223,12 +234,78 @@ function Base.copy!(progn_new::PrognosticVariables, progn_old::PrognosticVariabl
...
     end
 
-    progn_new.clock.time = progn_old.clock.time
+    progn_new.random_pattern .= progn_old.random_pattern.data
+
+    copy!(progn_new.clock, progn_old.clock)
     progn_new.scale[] = progn_old.scale[]
 
-    return progn_new
+    return nothing
```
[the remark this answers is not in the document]
maximilian-gelbrecht: Yep, you are right.

PR #657, opened 2025-01-10
NetCDF output for tracers, precipitation rates at initialize! again

milankl:


milankl:
Awesome, this works

image

via

```julia
add!(model, SpeedyWeather.TracerOutput(name="co2"))
add!(model, SpeedyWeather.TracerOutput(name="co2", long_name="Carbon dioxide", unit="ppm"))
set!(simulation, co2=(λ, φ, σ) -> sind(λ))
```

PR #658, opened 2025-01-10
bump compat to GPUArrays v11, JLArrays v0.1, KernelAbstractions 0.9

milankl:


milankl:
@jackleland @maximilian-gelbrecht I believe we can now upgrade GPUArrays, JLArrays, and KernelAbstractions compat.

One would (fingers crossed) roll forward these two changes again

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/f90b9d8c793a762c08eafa4ee527803737d2ceef/src/RingGrids/general.jl#L485-L489

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/719dce952a26bf7713f8c9a2332313e52cbb512b/src/LowerTriangularMatrices/lower_triangular_array.jl#L633-L637

that #636 rolled back

maximilian-gelbrecht:
Tests are passing here. I'd like to briefly test on GPU before merging as well.

maximilian-gelbrecht:
Ah, the tests are passing here, but I think the compat entries are actually wrong, we do loose compatibility with GPUArrays v10. It'll be only compatible with v11.

milankl:
riiiight, is there a reason why we wouldn't upgrade then? E.g. I guess this means that older CUDA jl versions don't support v11, does that restrict us in any other way?

maximilian-gelbrecht:
I mean it's only relevant if someone wants to use Speedy with some package that isn't updated yet to be compatible with GPUArrays v11. Not sure if there are many around. This is only relevant for packages that do something with GPUs. Even if that's the case, they'd still just get a slightly older version of Speedy.

milankl:
Oceananigans still uses GPUArrays v10, so at the moment that would disallow running both simultaneously (coupling), but I believe by the time we do GPU-GPU coupling they will have upgraded already (@simone-silvestri do you plan to change compat to GPUArrays v11?). And it's better to go to the new version anyway. Green light from my side!

simone-silvestri:
There has been a feature removal in `GPUArrays@11.1.0` that breaks a bunch of code for GPU oceananigans see this [PR](https://github.com/CliMA/Oceananigans.jl/pull/4037). So for the moment until the issue has been resolved, we have pinned a previous version of GPU arrays. Hopefully, if it is solved soon, we can move to the new version.

maximilian-gelbrecht:
We are in no hurry with this. I think we can wait until Oceananigans is compatible as well.

milankl:
Because even if we run CPU-GPU with Oceananigans this will create a compat conflict? Yes makes sense

milankl:
@simone-silvestri I'd love to update to GPUArrays v11, but I believe that would currently cause a conflict with Oceananigans?

giordano:
> but I believe that would currently cause a conflict with Oceananigans?

https://github.com/CliMA/Oceananigans.jl/pull/4081

PR #668, opened 2025-01-19
[LowerTriangularMatrices] Sum for LTA

maximilian-gelbrecht:
As discussed in #602 and #558 

This adds `sum` for `LowerTriangularArrays`. It just forwards it to `sum(L.data)`. This means it follows the flat indexing dimension logic. I think this is the most consistent with our treatment of `LowerTriangularArrays` and how we also describe it in the docs. It also is by far the easies version to implement for `sum`. 

I didn't do a version for `cat` yet, as it is actually quite hard to do one that is reasonable performant and general at the same time. E.g. `cat([L.data for L in Ls]...)` would be quite poor due to the splatting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fixed a bug, so that sum now works for LowerTraingularArrays [#668](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/668)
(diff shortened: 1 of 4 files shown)


milankl (approved):

milankl:
Agree that we should continue to use matrix indexing in only as an add-on and for construction with `rand`, `zeros`, etc.

PR #669, opened 2025-01-21
Precipitation rate for coupling

milankl:


milankl:
The last commit was necessary as convection is allowed to escape early, so the rate should always be set back to zero, otherwise the same rate is used for all grid points until convection occurs again where it's then overwritten. 

Was

image


is now

image

PR #671, opened 2025-01-29
Land model modularised

milankl:
- `model.land` has now subcomponents for temperature, soil moisture, vegetation and rivers
- `LandModel` and `DryLandModel` defined whereas the latter is for the primitive dry model with only temperature and no soil moisture, vegetation or rivers
- `SpectralGrid` has now a central `nlayers_soil` to allocate 3D grid variables for the land. Default 2 layers.
- NetCDF output also on that `nlayers_soil` dimension, 3D land variables require a `is_land::Bool = true` field so that that dimension is used and not the vertical sigma coordinate
- `add!(model, SpeedyWeather.LandOutput()...)` can be used to output soil temperature, moisture and availability index at every time step
- `model.land.temperature.monthly_temperature` is now a 4D grid variable, same for `land.soil_moisture.monthly_soil_moisture`


milankl:
@maximilian-gelbrecht I'm also finally structuring the `/test` folder into more subfolders with this PR

PR #672, opened 2025-01-31
MITgcm's 2-layer land bucket model

milankl:
Implementing a 2-layer land bucket model written by Jean-Michel many years ago for MITgcm (see [docs here](https://mitgcm.readthedocs.io/en/latest/phys_pkgs/land.html))

- store and release heat and moisture on two layers, 
- diffuse heat between these layers such that roughly the top layer reacts to the daily cycle but the bottom layer to the seasonal cycle only
- diffuse moisture between layers on a 2-day time scale
- diagnoses heat capacity from soil moisture
- overflow excess soil moisture into nothing or rivers that could provide an instantaneous fresh water flux to the ocean
- uses prescribed high and low vegetation (e.g. constant or seasonal cycle) to determine a soil moisture availability to evaporation

image


milankl:
Soil temperature in the primitive dry model goes nuts after ~10 days

image

not sure what temperature one would expect with a `DryLandModel` but it's suspicious that the hot spots are concentrated along the coast, maybe there are some fluxes that aren't weighted land/sea accordingly

milankl:
I accidentally merged the commits here directly into main 🙈 closing this and fixing bugs in another pull request!

Issue #456, opened 2024-02-23
Another simple radiation scheme


milankl:
Andrew Williams @AndrewILWilliams suggested

> Another simple radiation scheme which might be worth trying, and which more faithfully mimics realistic radiative cooling profiles is that of Jeevanjee and Romps (2018), and recently used in Sec 2.3 of Jake Seeley's paper (https://www.jacobtseeley.com/files/seeley2023.pdf) and Nadir's 2022 paper (https://nadirjeevanjee.com/papers/19iris.pdf -- Eq. 2).

> The basic insight is that, if F is the net upwards radiative flux, the flux divergence with respect to temperature doesn't vary that much. That is, \partial_{T} F is approximately a fixed function of temperature (see Eq. 1 of the Seeley paper). This means one can get radiative heating rates by calculating "\partial_{T} F" (which is just a fixed function of your local gridbox temperature) and then transforming to pressure coordinates by multiplying by dT/dP (or whatever vertical coordinate SpeedyWeather uses). Then you can get your K/s heating rates.

milankl:
@AndrewILWilliams maybe you can help me understand that, it says

![image](https://github.com/SpeedyWeather/SpeedyWeather.jl/assets/25530332/fe2ecf12-591c-4c6f-beba-19852f7ff888)

With $T_t = 200K$ and $\alpha = 0.025 W/m^2/K$, I'd discretize this with $F_k$ being the upward flux between layer $k$ and $k+1$ so that we'd have

$$F_k = F_{k+1} + (T_k - T_{k+1}) \alpha (T_t - T_{k+1})$$

For $N=3$ vertical levels with $T_1 = 280 K, T_2 = 290 K, T_3 = 300 K$ and sea/land surface temperature $T_4 = 300K$ we'd have

$$ F_N  = 0$$

because there's no temperature difference between the surface and the lowermost layer but

$$F_2 = F_N + 25 W/m^2K = 25 W/m^2K$$

so the lowermost layer doesn't receive any net radiative flux, but loses it to the layer above. Then

$$F_1 = F_2 + 22.5 W/m^2K = 47.5 W/m^2 K$$

[...]

PR #677, opened 2025-02-14
Modular land model + diagnostic albedo, independent for ocean/land

milankl:
How to run a rocky planet simulation

```julia
land = LandModel(spectral_grid, temperature=LandBucketTemperature, soil_moisture=LandBucketMoisture)
land_sea_mask = RockyPlanetMask(spectral_grid)
model = PrimitiveWetModel(spectral_grid; land_sea_mask, land)
```

Aquaplanet simulation with slab ocean

```julia
ocean = SlabOcean(spectral_grid)
land_sea_mask = AquaPlanetMask(spectral_grid)
model = PrimitiveWetModel(spectral_grid; land_sea_mask, ocean)
```

Earth then just by using the default land-sea mask

```julia
land_sea_mask = LandSeaMask(spectral_grid)
model = PrimitiveWetModel(spectral_grid; land_sea_mask, land, ocean)
```


milankl:
Looks much better now! Rocky planet simulation: Soil moisture is higher were it rained, capped at 1 due to runoff, soil moisture availability is consequently higher, but the evaporative flux from land is only concentrated in some areas...

image

milankl:
Okay the evaporative flux in that experiment is just focused on the day side, after 40.5 days

image

milankl:
Removed the `max( ,0)` formulation of Fortran SPEEDY's surface fluxes to allow for surface condensation, now

image

daytime evaporation, nighttime condensation, looks reasonable to me!

milankl:
Top layer of soil moisture nicely reacting to rainfall and evaporation over 30 days

https://github.com/user-attachments/assets/bad14e9c-f26e-49de-bd5a-0489ed5fb11a

milankl:
Surface sensible heat flux, daily cycle over land but not in mountaineous regions, strongly dominated by wind over ocean

https://github.com/user-attachments/assets/efa68b78-b333-415a-8e9c-b33b82cb2533

milankl:
@maximilian-gelbrecht okay first version of the land model seems to be up and running! need to add tests though and update the documentation ... 🥳

pierluigividale:
This is great. What can I do to help test?

pierluigividale:
> @pierluigividale looks like the land model is up and running! 🥳 At least it's not too unreasonable, to get the surface climatology better I'm thinking of
> 
> * diagnosing a sea ice concentration based on SST, e.g. -10˚C -> 1, then linearly to 0˚C -> 0 sea ice
> * assuming snow cover for land surface temperatures below zero
> * diagnosing albedo from this (any references?)

The formula for sea ice seems a bit odd. I do believe that 100% sea ice requires indeed at least -10C, but I am not sure where I would set the 0% sea ice, probably still at negative temperatures? I have an old book at the office, with simpler models, I can look it up.
For land, many simple LSMs set precipitation to snow when the near-surface temperature is <=0. This is one of those annoying ON/OFF things, and I would rather set it with a bit of stochasticity, but for now OK
For albedo, the simplest is to take a fixed value for ice and one for snow, from any LSM, e.g. JULES or CLM. Then people change the albedo of snow based on age, but that is overkill. I can give you values tomorrow.

pierluigividale:
> Removed the `max( ,0)` formulation of Fortran SPEEDY's surface fluxes to allow for surface condensation, now
> 
> image
> 
> daytime evaporation, nighttime condensation, looks reasonable to me!

This is indeed starting to look like a habitable planet.

pierluigividale:
> Surface evaporative flux (night time condensation over land!)
> 
>  sef.mp4

So, this is u'q'?
Do you have the precipitation animation that goes with that soil moisture evolution? It looks very realistic, just wanted to check. I better start doing my own, so I can answer all my own questions.

pierluigividale:
> Top layer soil temperature still gets quite hot (100-120K) which might be due to albedo=0.4 everywhere for land now, the not T^4 longwave radiation formulation within the Jeevanjee radiation, or something else, but it's stable so doesn't look like a bug to me. Also there's something wrong with the land-sea mask here, compared to soil moisture!
> 
>  soil_temperature_layer1.mp4

OK, albedo of 0.4 for land is rather high, it is more like the albedo of bare rocks. If you were to set it to a more realistic value for normal land, your temperatures would be even hotter. But what worries me more is hearing that LW up is not epsilon sigma T_s^4? What is this Jevanjee radiation?

milankl:
Some instructions, you can output all variables with
```julia
add!(model, SpeedyWeather.AllOutputVariables()...)
```
before `run!(simulation)`. You can also only output variable groups
```julia
PrecipitationOutput()...
BoundaryOutput()...
RadiationOutput()...,
RandomPatternOutput(),
SurfaceFluxesOutput()...,
LandOutput()...
SeaSurfaceTemperatureOutput()
```
(add the moment you have to unpack the tuples with `...` (note to self: this should probably just work either way)). You can't add a variable twice, it's an entry in a dictionary, so it'll just be overwritten by the second.

milankl:
> The formula for sea ice seems a bit odd. I do believe that 100% sea ice requires indeed at least -10C, but I am not sure where I would set the 0% sea ice, probably still at negative temperatures? I have an old book at the office, with simpler models, I can look it up.

@glwagner do you have a reference for a simple thermodynamic sea ice model?

> For land, many simple LSMs set precipitation to snow when the near-surface temperature is <=0. This is one of those annoying ON/OFF things, and I would rather set it with a bit of stochasticity, but for now OK

Would that be prognostic or diagnostic? Because I guess you don't want to melt all the snow immediately if it doesn't snow on one timestep? I'm wondering whether there is a simple diagnostic formulation that just increases albedo on a given time step under certain conditions.

> So, this is u'q'?
Do you have the precipitation animation that goes with that soil moisture evolution? It looks very realistic, just wanted to check. I better start doing my own, so I can answer all my own questions.

Nah it's this one https://speedyweather.github.io/SpeedyWeather.jl/dev/surface_fluxes/#Surface-evaporation but without the `max(x, 0)`, I removed that!

For land it's here

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/3e623cdea29d960c8bdc1f29d66ac10aabc6bea4/src/physics/surface_fluxes/moisture.jl#L122-L123

But I just realised that without the `max`, the multiplication with `α` (the soil moisture availability) doesn't really make that much sense? For ocean it's here

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/3e623cdea29d960c8bdc1f29d66ac10aabc6bea4/src/physics/surface_fluxes/moisture.jl#L74-L75

pierluigividale:
Hi

Ok, let me ask the FAMOUS people downstairs, I do see the point about staying away from prognostics.

Thanks for the other pointers.

Cheers,
PL

pierluigividale:
Talked to Robin, who is the main developer of FAMOUS. He said this about sea ice and its albedo:
1) doing an instantaneous calculation for sea ice, that is, using an air temperature to say ice on/off works, but is a bad idea
2) anything involving sea ice fraction requires indeed a prognostic equation. In terms of formulations, he would have to search, but his search is as good as ours. My search through textbooks and via Google keeps coming up with prognostic equations for ice volume, to then be converted to extent, nothing as simple as a formula for area.
3) attempting to include sea ice only from the point of view of a) albedo and b) blocking air-sea fluxes gets into all sorts of other complications, so it may not be worth the effort.

pierluigividale:
> > The formula for sea ice seems a bit odd. I do believe that 100% sea ice requires indeed at least -10C, but I am not sure where I would set the 0% sea ice, probably still at negative temperatures? I have an old book at the office, with simpler models, I can look it up.
> 
> @glwagner do you have a reference for a simple thermodynamic sea ice model?
> 
> > For land, many simple LSMs set precipitation to snow when the near-surface temperature is <=0. This is one of those annoying ON/OFF things, and I would rather set it with a bit of stochasticity, but for now OK
> 
> Would that be prognostic or diagnostic? Because I guess you don't want to melt all the snow immediately if it doesn't snow on one timestep? I'm wondering whether there is a simple diagnostic formulation that just increases albedo on a given time step under certain conditions.
> 
> > So, this is u'q'?
> > Do you have the precipitation animation that goes with that soil moisture evolution? It looks very realistic, just wanted to check. I better start doing my own, so I can answer all my own questions.
> 
> Nah it's this one https://speedyweather.github.io/SpeedyWeather.jl/dev/surface_fluxes/#Surface-evaporation but without the `max(x, 0)`, I removed that!
> 
> For land it's here
> 
> https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/3e623cdea29d960c8bdc1f29d66ac10aabc6bea4/src/physics/surface_fluxes/moisture.jl#L122-L123
> 
> But I just realised that without the `max`, the multiplication with `α` (the soil moisture availability) doesn't really make that much sense? For ocean it's here
> 
> https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/3e623cdea29d960c8bdc1f29d66ac10aabc6bea4/src/physics/surface_fluxes/moisture.jl#L74-L75

Yes, for land you do need to limit LH by multiplying with alpha. For ocean, there is no such constraint.

milankl:
>     3. attempting to include sea ice only from the point of view of a) albedo and b) blocking air-sea fluxes gets into all sorts of other complications, so it may not be worth the effort.

Thanks for investigating this. But there must be a way to have a prognostic sea ice concentration forced by sea surface temperatures no? (allowing for supercooled water, but if we reduce the air-sea fluxes whenever that happens then it wouldn't be significantly supercooled anyway?) And then diagnostically one reduces air-sea fluxes and increases albedo. I mean one can think of a freeze rate $f$ [1/K/s] and a melt rate $m$ [1/K/s] (maybe even the same?) with a threshold $t$ (I guess -2˚C for sea water) such that

```math
\frac{dc}{dt} = -f*\min(SST - t, 0) - m*\max(SST - t, 0) 
```

with a constraint that $0 <= c <= 1$ always. So we would have freezing (increasing ice concentration) whenever the temperatures are below freezing, proportional to how cold it is, and melting for temperatures above freezing again proportional to how warm it is. One is therefore left with two parameters, and one could assume $f = m$ too. I get it, it's not the most sophisticated model, but already much better than having no sea ice?

milankl:
> Yes, for land you do need to limit LH by multiplying with alpha. For ocean, there is no such constraint.

Great then I leave this as is. Feel free to open a pull request if there's any proposed changes to the current formulation. The vegetation is just a climatology of low/high vegetation, using the formula from [here](https://speedyweather.github.io/SpeedyWeather.jl/stable/surface_fluxes/#Surface-evaporation)

milankl:
> OK, albedo of 0.4 for land is rather high, it is more like the albedo of bare rocks. If you were to set it to a more realistic value for normal land, your temperatures would be even hotter. But what worries me more is hearing that LW up is not epsilon sigma T_s^4? What is this Jevanjee radiation?

Yes, I should work on the radiation scheme before investigating this further. Jeevanjee was implemented here in this pull request #502 following this discussion #456. We use this as it calculates radiative fluxes based on temperature directly without solving a radiative transfer equation. As I was working on the modularisation with optical depth I found that useful as it bypasses that. It's not a good radiation code but it's not awful either. After land, radiation is up next to be overhauled!

pierluigividale:
Yes, I think that this is crude, albeit better than nothing at all.

I was just going through the CMSS lectures again, and I suddenly remembered that, in fact, this is not that different to how partial cloud fraction is imposed (albeit non-linearly), which is much better than the on/off approach that was used before that simple approach. See a screenshot below.

PL

￼

pierluigividale:
OK, I had a brief look at it, and I found the one equation rather confusing, but the discretisation a bit less so.

I have the impression that this is designed to restore towards a reference profile, and that the bottom boundary condition does not matter? Also suggested by this statement: " It is not meant as a full radiative transfer emulator — i.e., no need to worry about the numerical value of net flux into/out of the column, or to tie that boundary condition to the underlying land/ocean model. All that matters is the vertical derivative, “

The next step would be to include that bottom boundary condition, and to tie it to the surface energy balance, but can such a model support that?

PL

milankl:
> The next step would be to include that bottom boundary condition, and to tie it to the surface energy balance, but can such a model support that?

Yes, I've added these lines

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/cbafe4b98a8dc6e4c631b315718ecccbd20188b0/src/physics/longwave_radiation.jl#L125-L133

as if the surface was another layer. This also sets then the (net) surface longwave upward flux (downward flux is zero) to force slab ocean / land surface model.

pierluigividale:
OK, that would enable the surface to cool by LW, and it should help reduce the warm bias at the surface.
I have been looking at that original paper, but the equations there are quite standard radiative flux divergences, while these are expressed as a d/dtau of temperatures, which looks like a sort of vertical flux divergence, I just cannot figure out the transformation from radiances to temperatures.

pierluigividale:
I do need to try and understand this Jeevanjee and Romps, 2018 model, which would also answer the question of whether or not we can impose climate change via changes in optical depth.

milankl:
So with the latest commits I've modularised the whole thing and added some tests. We now have

- `LandModel`, with temperature, soil moisture, vegetation and rivers
- `DryLandModel` with temperature only

Both land models can be used with `PrimitiveWetModel` and `PrimitiveDryModel`. The latter will automatically skip the functionality around soil moisture that's not needed with a dry atmosphere. Both land models have

- `land.geometry` that defines the vertical `layer_thickness`
- `land.thermodynamics` that defines constants like heat capacity and conductivity for land

For temperature we now have

```julia
julia> subtypes(SpeedyWeather.AbstractLandTemperature)
3-element Vector{Any}:
 ConstantLandTemperature
 LandBucketTemperature
 SeasonalLandTemperature
```

for soil moisture we have

```julia
julia> subtypes(SpeedyWeather.AbstractSoilMoisture)
3-element Vector{Any}:
 LandBucketMoisture
 NoSoilMoisture
 SeasonalSoilMoisture
```

for vegetation

```julia
julia> subtypes(SpeedyWeather.AbstractVegetation)
2-element Vector{Any}:
 NoVegetation
 VegetationClimatology
```

and for rivers only

```julia
julia> subtypes(SpeedyWeather.AbstractRivers)
1-element Vector{Any}:
 NoRivers
```

So a land model for example can be constructed as

```julia
thermodynamics = LandThermodynamics(spectral_grid, heat_conductivity=0.42)
temperature = LandBucketTemperature(spectral_grid)
soil_moisture = SeasonalSoilMoisture(spectral_grid)
vegetation = NoVegetation(spectral_grid)
land = LandModel(spectral_grid; thermodynamics, temperature, soil_moisture, vegetation)
```

printed to the terminal as

```julia
julia> model.land
LandModel <: SpeedyWeather.AbstractWetLand
├ spectral_grid: SpectralGrid
├ geometry: LandGeometry{Float32, Vector{Float32}}
├ thermodynamics: LandThermodynamics{Float32}
├ temperature: LandBucketTemperature{Float32}
├ soil_moisture: SeasonalSoilMoisture{Float32, OctahedralGaussianArray{Float32, 3, Array{Float32, 3}}}
├ vegetation: NoVegetation
└ rivers: NoRivers
```

any non-specified component (here rivers) will be taken as default. Such a land model can then be passed onto the model constructor

```julia
model = PrimitiveWetModel(spectral_grid; land)
```

So we are essentially applying the same logic of how to construct an atmospheric model to the land surface model but the land surface model is then inside the atmospheric model. The whole logic can be further extended like Russian dolls.

pierluigividale:
Yes, you added an equation that looks like one more layer in that Seeley and Wordsworth, 2023, eq (1): dF/dT = α*(T_t - T)
which looks like a flux divergence, and the idea is to treat the surface as a sort of sponge layer, similar to an air layer, which is actually quite compatible with the philosophy of the vegetation layer in ECLand and JULES.
What I was saying is that I do not yet understand how equation (1) relates to more traditional models. It feels like we can change optical depth by playing with α, but how, and by how much, would require that I understand that parametrisation.

pierluigividale:
> So with the latest commits I've modularised the whole thing and added some tests. We now have
> 
> * `LandModel`, with temperature, soil moisture, vegetation and rivers
> * `DryLandModel` with temperature only
> 
> Both land models can be used with `PrimitiveWetModel` and `PrimitiveDryModel`. The latter will automatically skip the functionality around soil moisture that's not needed with a dry atmosphere. Both land models have
> 
> * `land.geometry` that defines the vertical `layer_thickness`
> * `land.thermodynamics` that defines constants like heat capacity and conductivity for land
> 
> For temperature we now have
> 
> ```julia
> julia> subtypes(SpeedyWeather.AbstractLandTemperature)
> 3-element Vector{Any}:
>  ConstantLandTemperature
>  LandBucketTemperature
>  SeasonalLandTemperature
> ```
> 
> for soil moisture we have
> 
> ```julia
> julia> subtypes(SpeedyWeather.AbstractSoilMoisture)
> 3-element Vector{Any}:
>  LandBucketMoisture
>  NoSoilMoisture
>  SeasonalSoilMoisture
> ```
> 
> for vegetation
> 
> ```julia
> julia> subtypes(SpeedyWeather.AbstractVegetation)
> 2-element Vector{Any}:
>  NoVegetation
>  VegetationClimatology
> ```
> 
> and for rivers only
> 
> ```julia
> julia> subtypes(SpeedyWeather.AbstractRivers)
> 1-element Vector{Any}:
>  NoRivers
> ```
> 
> So a land model for example can be constructed as
> 
> ```julia
> thermodynamics = LandThermodynamics(spectral_grid, heat_conductivity=0.42)
> temperature = LandBucketTemperature(spectral_grid)
> soil_moisture = SeasonalSoilMoisture(spectral_grid)
> vegetation = NoVegetation(spectral_grid)
> land = LandModel(spectral_grid; thermodynamics, temperature, soil_moisture, vegetation)
> ```
> 
> printed to the terminal as
> 
> ```julia
> julia> model.land
> LandModel <: SpeedyWeather.AbstractWetLand
> ├ spectral_grid: SpectralGrid
> ├ geometry: LandGeometry{Float32, Vector{Float32}}
> ├ thermodynamics: LandThermodynamics{Float32}
> ├ temperature: LandBucketTemperature{Float32}
> ├ soil_moisture: SeasonalSoilMoisture{Float32, OctahedralGaussianArray{Float32, 3, Array{Float32, 3}}}
> ├ vegetation: NoVegetation
> └ rivers: NoRivers
> ```
> 
> any non-specified component (here rivers) will be taken as default. Such a land model can then be passed onto the model constructor
> 
> ```julia
> model = PrimitiveWetModel(spectral_grid; land)
> ```
> 
> So we are essentially applying the same logic of how to construct an atmospheric model to the land surface model but the land surface model is then inside the atmospheric model. The whole logic can be further extended like Russian dolls.

OK, this is a great construct and I cannot overstate how excellent this is for teaching. I am bogged down with too many things, and have yet to play, but I know what to do now.

milankl:
Soil moisture is definitely responding correctly to precipitation, it's high where there was a good amount of rain in the week prior. But something is off there with the land-sea mask in soil moisture

image

milankl:
I did an analysis of why the land/ocean gets so warm. Bottom line: the [Jeevanjee radiation](https://speedyweather.github.io/SpeedyWeather.jl/dev/radiation/#Jeevanjee-radiation), see also #502, we use currently is not radiating enough surface longwave upward for high temperatures. Because of no clouds we're in a warmer climate anyway but then for those higher surface temperatures the Jeevanjee radiation does not provide enough of a negative feedback cooling the surface down -- only when the surface temperatures goes unrealistically high. I did that assessment here. Using surface (sea/soil) temperatures from a simulation and the air temperatures in the lowermost layer above I've calculated the upward and downward radiative fluxes following Jeevanjee and following Stefan-Boltzmann $\sigma T^4$. Jeevanjee has a higher upward flux (bright pink) and a lower (i.e. negative) downward flux (not so bright pink) such that the net (orange) is up to 2x cooling less than the corresponding net $\sigma T^4$ flux (black). 

image

I split the Jeevanjee flux $F$ as (with $T_s$ the respective surface temperature, ocean/soil, and $T_a$ the temperature of the atmospheric layer above)

```math
\begin{align}
F &= (T_s - T_a)~\alpha~(T_s - T_t) = F^{\uparrow} + F_{\downarrow} \\
F^{\uparrow} &= T_s~\alpha~(T_s - T_t) \\
F_{\downarrow} &= -T_a~\alpha~(T_s - T_t) \\
\end{align}
```

@jtseeley you may find that interesting ☝🏼 ?

milankl:
@pierluigividale this means to me that I would like to merge this pull request but not make the land model with `LandBucketTemperature`, `LandBucketMoisture` the default yet as I don't think there's anything wrong with the changes of the pull request but with other model components that we can fix in other pull requests. There is currently simply no masking for the output data, but that's also more a question of the land-sea mask we use (how many lakes/islands etc) then a question for this pull request.

pierluigividale:
> I did an analysis of why the land/ocean gets so warm. Bottom line: the [Jeevanjee radiation](https://speedyweather.github.io/SpeedyWeather.jl/dev/radiation/#Jeevanjee-radiation), see also #502, we use currently is not radiating enough surface longwave upward for high temperatures. Because of no clouds we're in a warmer climate anyway but then for those higher surface temperatures the Jeevanjee radiation does not provide enough of a negative feedback cooling the surface down -- only when the surface temperatures goes unrealistically high. I did that assessment here. Using surface (sea/soil) temperatures from a simulation and the air temperatures in the lowermost layer above I've calculated the upward and downward radiative fluxes following Jeevanjee and following Stefan-Boltzmann σ T 4 . Jeevanjee has a higher upward flux (bright pink) and a lower (i.e. negative) downward flux (not so bright pink) such that the net (orange) is up to 2x cooling less than the corresponding net σ T 4 flux (black).
> 
> image
> 
> I split the Jeevanjee flux F as (with T s the respective surface temperature, ocean/soil, and T a the temperature of the atmospheric layer above)
> 
> F	=(
> T
> s
> −
> T
> a
> ) α (
> T
> s
> −
> T
> t
> )=
> F
> ↑
> +
> F
> ↓
>  
> F
> ↑
> =
> T
> s
>  α (
> T
> s
> −
> T
> t
> )	 
> F
> ↓
> =−
> T
> a
>  α (
> T
> s
> −
> T
> t
> )	  
> @jtseeley you may find that interesting ☝🏼 ?

Hmmmm. I have been thinking a lot about this alpha factor, and how it relates to emissivity. But it seems from your current analysis that the far more fundamental issue is that a proper S-B law would insert a T^4 dependence, while this Jeevanjee is just linear with temperature. One possible hack would be to use a different alpha for the surface than for the atmosphere, which would correspond to a higher emissivity for the surface. But the linear dependence continues to be an issue.

milankl:
@maximilian-gelbrecht the last commit also allows one to drop the `geometry` argument from `set!` for 2D grids as the only coordinates required can be recomputed with `londs, latds = RingGrids.get_londlatds(grid)` which of course shouldn't be done if `geometry` is available. Unfortunately not possible for 3D grids as we'd need to assume something
for the vertical. In contrast lon, lat is actually encoded implicitly already in the grid type.

```julia
function set!(
    var::AbstractGridArray{T, 1},
    f::Function,
    geometry::Union{Nothing, Geometry}=nothing,
    S::Union{Nothing, SpectralTransform}=nothing;
    add::Bool=false,
) where T
    
    if isnothing(geometry)      # recompute the coordinates
        londs, latds = RingGrids.get_londlatds(var)
    else
        (; londs, latds) = geometry
    end
    
    kernel = add ? (a,b) -> a+b : (a,b) -> b
    for ij in eachgridpoint(var)
        var[ij] = kernel(var[ij], f(londs[ij], latds[ij]))
    end
    return var
end
```

milankl:
@pierluigividale I merged this as the next bits of work should be on the radiation and cloud scheme. And wrote a little documentation for it, feel free to check it out

https://speedyweather.github.io/SpeedyWeather.jl/dev/land/

pierluigividale:
Hi 
Back from my mega-trip

I did re-install the latest from the Julia repository, and I re-ran a number of experiments while I was on the plane.
I have not tried the land yet, but it is unclear to me what is available via the install above, and how much I need to do by hand?

About the radiation and cloud scheme, do you want to have a chat? I am around for a couple of weeks now, maybe more, courtesy of DOGE…

Cheers,|
PL

milankl:
The pull request here is merged, so when you install the main branch you get all of the functionality as outlined in the documentation https://speedyweather.github.io/SpeedyWeather.jl/dev/land/. But none of that is the default yet. I want to have more time to play with it before making it the default. For example I do a stress test with north-south flipped land-sea mask at ~100km resolution like so

```julia
spectral_grid = SpectralGrid(trunc=127, nlayers=12, nlayers_soil=2)

# define land model
temperature = LandBucketTemperature(spectral_grid)
soil_moisture = LandBucketMoisture(spectral_grid)
land = LandModel(spectral_grid; temperature, soil_moisture)

# define an albedo
albedo = Albedo(
    ocean = GlobalConstantAlbedo(spectral_grid, albedo=0.1),
    land = ManualAlbedo(spectral_grid))

set!(albedo.land, (λ, φ) -> 0.3 + 0.3*abs(φ)/90)

# define the ocean
ocean = SlabOcean(spectral_grid)

# construct the model
model = PrimitiveWetModel(spectral_grid; land, albedo, ocean)
add!(model, SpeedyWeather.AllOutputVariables()...)

# after initialize! before run! reverse land-sea mask in latitudes
reverse!(model.land_sea_mask.mask, dims=:lat)
# plot(model.land_sea_mask.mask) # in case you want to check that this is right

# run 10 day spinup and 10 days to store data
```

this video shows lower-most layer temperature. It nicely shows that orography (not lat-reversed!) reduces temperatures with elevation (you can see the andes, himalayas, etc where they are supposed to be). But that the land-sea mask, which defines where the land bucket model is active vs the slab ocean model largely defines the areas of pronounced daily cycle through lower heat capacity. So yes, this simulation has an ocean at the top of Mt Everest so obv not realistic but nicely shows what's due to the land surface model!

https://github.com/user-attachments/assets/6ec06be3-c66f-49b9-8d15-56c722be3cf9

pierluigividale:
Hi
I did the manual installation, because the land model was still not available, and after that it all works.

I tried a few definitions of the albedo, and what I see makes sense. 
I am still blown away by how one constructs models, and how they self-declare each time. I have been at it for quite a while now, and the novelty is still revolutionary to me.

Let me try your stress test below.

Cheers,
PL

milankl:
This is surface layer soil moisture in that same experiment

https://github.com/user-attachments/assets/4f1041f5-1083-4016-a3cb-b447f01fac1c

milankl:
And the surface humidity fluxes with evaporation but night-time condensation over land

https://github.com/user-attachments/assets/cb9b1344-0802-479d-a595-6f1726274007

pierluigividale:
That is a lot of fun.
Blimey, I am running at T127 truncation with 12 layers, on an M2 chip from 2023, and it runs at 12 SYPD. 

How do you do those animations? I was setting up CairoML with record, but I cannot figure out the shape of the arrays while running.
Do you always write the data first, then do the animations?

Best,
PL

milankl:
> That is a lot of fun. Blimey, I am running at T127 truncation with 12 layers, on an M2 chip from 2023, and it runs at 12 SYPD. 

Yeah, on my M1 from 2020 I get similar speeds! I really need to find some time in the coming months to work on the GPU version, that should give similarish speeds at 4x the resolution and more vertical layers, fingers crossed! 

> How do you do those animations? I was setting up CairoML with record, but I cannot figure out the shape of the arrays while running. Do you always write the data first, then do the animations? 

At the moment my workflow is save to netcdf with `output=true` then use [ncvis](https://github.com/SEATStandards/ncvis) to open the netcdf file and click `Export...` and `Export multiple frames` and then ffmpeg them to an mp4. I do want to streamline this with GeoMakie, GLMakie etc so opening a netcdf file and storing it as a video becomes a one-click adventure but haven't found the time yet.

pierluigividale:
Do you recommend that I clone that site and do a manual installation, or should I use the condo install, which points to something called ncvis-climate?

PL

milankl:
> Do you recommend that I clone that site and do a manual installation, or should I use the condo install, which points to something called ncvis-climate? PL

I have installed ncvis through cloning and via the `build.sh` yes. After doing a `conda install -c conda-forge wxwidgets libnetcdf` to have the required libraries but I didn't bother creating a separate environment.

pierluigividale:
Yes, I also never bother with separate envs, which means that I often end up in tears of blood, but multiple envs end up unmanageable for me.

Cheers,
PL

pierluigividale:
OK, I have it all working.

Lots of things look good.
LST still above 100C, which may also explain why it does not rain on land?

LW up looks ok, but there is no LW down at all.

Atmospheric temperature looks absolutely fine.

A few puzzles about radiation and clouds indeed.

PL

milankl:
> LW up looks ok, but there is no LW down at all.

LW up is for the Jeevanjee radiation actually LW net. That's why the other one is zero.

maximilian-gelbrecht:
I just noticed that the wet land model tests that the PR introduces are now our longest lasting unit tests at 2mins. Would it be okay to reduce the number of layers and/or resolution for these tests, so that we reduce the CI time a bit?

pierluigividale:
> > LW up looks ok, but there is no LW down at all.
> 
> LW up is for the Jeevanjee radiation actually LW net. That's why the other one is zero.

OK, good, that makes it clear.
I have run a few test simulations in the last few days.
We already talked about the land versus sea precipitation, and with that I have to say that soil moisture accumulation looks fine in terms of dynamics, despite the overall amount of precipitation on land.

I have been looking at vorticity fields, and response to mountain ranges looks very realistic, including low-level flows in places like the East of the Andes. Extra-Tropical Cyclones and frontal systems look good. What is immediately noticeable is that there are no Tropical Cyclones. I have run the model all the way to T255 and there are none. This may be due to lack of feedback between precipitation formation and diabatic heat sources in mid-troposphere? But it could also be radiative balance.

PR #674, opened 2025-02-13
Changes for Reactant.jl

milankl:
@giordano

PR #679, opened 2025-02-19
SigmaCoordinates double-defined () constructor

milankl:
remove double defined `()` constructor for `SigmaCoordinates`

follow up from #674


maximilian-gelbrecht (commented):
Looks good. 

Yeah, the interactions with the `@kwdef` can be hard to see. I also would have thought it's okay like before. 

If the tests pass without warnings, this can be merged.

src/dynamics/vertical_coordinates.jl
```diff
@@ -5,21 +5,27 @@ export SigmaCoordinates
...
     SigmaCoordinates{T, V}(nlayers::Integer, σ_half::AbstractVector) where {T, V} = sigma_okay(nlayers, σ_half) ?
     new{T, V}(nlayers, σ_half) : error("σ_half = $σ_half cannot be used for $nlayers-level SigmaCoordinates")
 end
 
-# obtain nlayers from length of predefined σ_half layers
-SigmaCoordinates(σ_half::AbstractVector) = SigmaCoordinates{eltype(σ_half), typeof(σ_half)}(nlayers=length(σ_half)-1; σ_half)
-SigmaCoordinates(σ_half::AbstractRange) = SigmaCoordinates(collect(σ_half))
-SigmaCoordinates(nlayers::Integer = DEFAULT_NLAYERS) = SigmaCoordinates{Float64, Vector{Float64}}(; nlayers)
+# constructor using default sigma coordinates if only nlayers provided, also collect to allow for AbstractRange
+SigmaCoordinates(nlayers::Integer, σ_half::AbstractVector = default_sigma_coordinates(nlayers)) =
```
maximilian-gelbrecht, line 17: isn't this one overwriting the internal constructor in l 12?

maximilian-gelbrecht: There's no warning coming though, so I am wrong apparently

milankl: No because the original constructor has no type constraints I believe. It's only that `@kwdef` defines keyword constructors that include the empty constructor `()` that should not be defined again.

PR #680, opened 2025-02-19
SpeedyTransforms scratch memory part of DiagnosticVariables 

maximilian-gelbrecht:
Work in progress


maximilian-gelbrecht:
The basics of putting the relevant scratch memory part in the DiagnosticVariables works. 

This is done mainly so that we can - if desired - annotate the `model` with `Const` activity for Enzyme. This in turn makes it easier to interface with Lux, but also makes gradients that are not wrt system parameters much faster (think ~2x-3x). 

While the scratch memory part seems to work, there still seem to other issues that stop `model` to be used `Const` activity. I'll use this PR to continue to work on those

test/differentiability/timestepping.jl
```diff
@@ -70,5 +70,18 @@
...
+
+        # test that we can differentiate wrt to the IC while keeping model activity const 
+
+        d_progn = zero(progn)
+        d_diag = make_zero(diagn)
+
+        progn_new = zero(progn)
+        dprogn_new = one(progn) # seed 
+    
+        autodiff(set_runtime_activity(Reverse), timestep_oop!, Const, Duplicated(progn_new, dprogn_new), Duplicated(progn_copy_2, d_progn), Duplicated(diagn_copy_2, d_diag), Const(dt), Const(model))
```
maximilian-gelbrecht, line 82: The aim of this PR is basically that his here is possible. That we denote `model` with `Const` activity. 

Ideally, this would also work without turing on the `set_runtime_activity` of Enzyme, but currently it doesn't for reasons I don't fully understand. That being said, this runtime activity mode just gives a minor speed penalty.

maximilian-gelbrecht:
So, the basics of this seem to work now. 

In terms of performance of differentiating through the time stepping of a `PrimitiveWetModel` on my laptop: 

```Julia 
@time autodiff(Reverse, timestep_oop!, Const, Duplicated(progn_new, dprogn_new), Duplicated(progn, d_progn), Duplicated(diagn, d_diag), Const(dt), Duplicated(model, dmodel))
``` 
takes 0.047421 seconds and 

with `Const` activity as is now possible with this PR
```Julia 
@time  autodiff(set_runtime_activity(Reverse), timestep_oop!, Const, Duplicated(progn_new, dprogn_new), Duplicated(progn_copy_2, d_progn), Duplicated(diagn_copy_2, d_diag), Const(dt), Const(model))
```
takes just 0.017982 seconds. 

I haven't yet looked further into directly combining it with Lux, but I will do in the next days.

src/SpeedyTransforms/scratch_memory.jl
```diff
@@ -0,0 +1,50 @@
...
+ScratchMemory holds scratch memory for the `SpectralTransform` that's used both by the Fourier and Legendre transform. Fields are
+$(TYPEDFIELDS)"""
+struct ScratchMemory{
+    NF,
+    ArrayComplexType,           # <: ArrayType{Complex{NF}, 3},
+}
+    # SCRATCH MEMORY FOR FOURIER NOT YET LEGENDRE TRANSFORMED AND VICE VERSA
+    # state is undetermined, only read after writing to it
+    north::ArrayComplexType
+    south::ArrayComplexType
```
maximilian-gelbrecht, line 11: Currently only `north` and `south` are here because only those parts of the scratch memory are used in function calls which is what counts for Enzyme (apparently). 

I actually also tried having all parts of the scratch memory in here, but I had some problems with it that I couldn't quite figure out and doing it like it is now with just `north` and `south` also means that `_fourier!` and `_legendre!` don't need to be changed at all.

maximilian-gelbrecht:
So, in principle the PR is more or less what I wanted. 

There are still some other hurdles when interfacing with Lux (primarily that it's really slow at the moment) that I need to figure out. While I do that, I'll this PR here as a draft because I am not 100% how we'll do the whole Lux+Speedy thing in the end.

src/SpeedyTransforms/scratch_memory.jl
```diff
@@ -0,0 +1,50 @@
...
+    return ScratchMemory{
+        NF,
+        ArrayType{Complex{NF}, 3}
+    }(scratch_memory_north, scratch_memory_south)
+end 
+
+"""$(TYPEDSIGNATURES)
+Generator function for a `ScratchMemory` that holds the scratch memory for SpeedyTransforms.
+"""
+function ScratchMemory(
```
milankl, line 34: should there be a constructor method to create a 0-element `ScratchMemory` which could be in `SpectralTransform` if the scratch space in there is not needed? E.g. in the normal case of running a simulation the scratch memory can just be in the diagnostic variables and so the scratch memory in the spectral transform shouldn't be allocated?

src/SpeedyTransforms/spectral_transform.jl
```diff
@@ -55,8 +55,7 @@ struct SpectralTransform{
     
     # SCRATCH MEMORY FOR FOURIER NOT YET LEGENDRE TRANSFORMED AND VICE VERSA
     # state is undetermined, only read after writing to it
-    scratch_memory_north::ArrayComplexType
-    scratch_memory_south::ArrayComplexType
+    scratch_memory::ScratchMemory{NF, ArrayComplexType} 
     scratch_memory_grid::VectorType                 # scratch memory with 1-stride for FFT output
     scratch_memory_spec::VectorComplexType
     scratch_memory_column_north::VectorComplexType  # scratch memory for vertically batched Legendre transform
```
milankl, line 61: But these should be part of the scratch memory too no? Like they are actually used, though much smaller, I agree.

src/dynamics/time_integration.jl
```diff
@@ -285,6 +285,8 @@ function timestep!(
     # PARTICLE ADVECTION (always skip 1st step of first_timesteps!)
     not_first_timestep = lf2 == 2
     not_first_timestep && particle_advection!(progn, diagn, model.particle_advection)
+
+    return nothing
```
milankl, line 289: Is that what Cthulu.jl pointed out that the return value here isn't type stable? It's good!

src/dynamics/tendencies.jl
```diff
@@ -112,15 +112,18 @@ function pressure_gradient_flux!(
...
     # PRESSURE GRADIENT
     pres = progn.pres[lf]                   # log of surface pressure at leapfrog step lf
     ∇lnp_x_spec = diagn.dynamics.a_2D       # reuse 2D work arrays for gradients
     ∇lnp_y_spec = diagn.dynamics.b_2D       # in spectral space
     (; ∇lnp_x, ∇lnp_y) = diagn.dynamics     # but store in grid space
 
     ∇!(∇lnp_x_spec, ∇lnp_y_spec, pres, S)                   # CALCULATE ∇ln(pₛ)
-    transform!(∇lnp_x, ∇lnp_x_spec, S, unscale_coslat=true) # transform to grid: zonal gradient
-    transform!(∇lnp_y, ∇lnp_y_spec, S, unscale_coslat=true) # meridional gradient
+    transform!(∇lnp_x, ∇lnp_x_spec, scratch_memory, S, unscale_coslat=true) # transform to grid: zonal gradient
```
milankl, line 125: I like it, this is neat, not as complicated as I thought it might be!

test/differentiability/timestepping.jl
milankl: `Const` isn't actually constant it just has to be constant wrt to the differentiation right?

test/differentiability/primitivewet.jl
```diff
@@ -39,7 +39,7 @@
     ddiagn_copy = deepcopy(ddiagn)
     diagn_copy = deepcopy(diagn)
 
-    autodiff(Reverse, SpeedyWeather.parameterization_tendencies!, Const, Duplicated(diagn, ddiagn), Duplicated(progn, dprogn), Const(progn.clock.time), Duplicated(model, make_zero(model)))
+    autodiff(Reverse, SpeedyWeather.parameterization_tendencies!, Const, Duplicated(diagn, ddiagn), Duplicated(progn, dprogn), Const(progn.clock.time), Const(model))
```
milankl, line 42: okay I see, `Duplicated(model, make_zero(model))` is awkward ...

milankl (commented):
This looks great! I was always unhappy with the scratch space when we first introduced that last summer as it's a clear violation of our "model is constant" philosophy, so glad that you found another solution!

src/SpeedyTransforms/spectral_transform.jl
maximilian-gelbrecht: Yeah, as I wrote above, I had a version where I had those in as well first. Strictly speaking it's not necessary for Enzyme's purposes. Would also require (minor) changes to `_fourier!` and `_legendre!`. I am happy to look into it again though, I also find it a bit inconsistent like that.

src/dynamics/time_integration.jl
maximilian-gelbrecht: Yeah, Cthulhu flagged this as "yellow", I think type instabilities like this aren't always bad, and the compiler can often figure them out though. 

In my initial attempt I had some other instabilities from the scratch memory that it help me find.

maximilian-gelbrecht:
There's no direct feedback here in the PR but the command above to launch the extended differentiability tests above, did in fact launch those tests

src/SpeedyTransforms/scratch_memory.jl
maximilian-gelbrecht: This is now resolved through reference to the scratch memory

maximilian-gelbrecht:
Ok, doing the scratch memory as a reference works. 

From my side, this is ready to merge (when all tests pass)

milankl:
This just needs updating right? Merge before or after #732 ?

maximilian-gelbrecht:
I'll do it after all the GPU PRs that are currently open.

maximilian-gelbrecht:
I hope that I adjusted everything now to the last PRs, if any of the tests fail, I'll fix it next week. 

@bgroenks96 I wrote this PR so that there's less temporary storage in the `model`, in principle with this PR we should be able to annotate a model with `Const` activity in Enzyme

maximilian-gelbrecht:
I think this is good now.

maximilian-gelbrecht:
# Benchmarks

created for SpeedyWeather.jl v0.15.0 on Thu, 03 Jul 2025 15:37:01. 

### Machine details

```julia
julia> versioninfo()
Julia Version 1.10.9
Build Info:
Platform Info:
  CPU: 8 × Apple M3
  WORD_SIZE: 64
  LIBM: libopenlibm
```

### Explanation

- Δt: time step [s].

### Running the benchmarks

```> julia manual_benchmarking.jl```

## Models, default setups

| - | - | - | - | - | - | - |
| BarotropicModel | 31 | 1 | false | 2400 | 60809 | 1.66 MB |
| ShallowWaterModel | 31 | 1 | false | 2400 | 41166 | 1.70 MB |
| PrimitiveDryModel | 31 | 8 | true | 2400 | 2676 | 4.86 MB |
| PrimitiveWetModel | 31 | 8 | true | 2400 | 1652 | 5.19 MB |

## Grids

| PrimitiveWetModel | 63 | 8 | FullGaussianGrid | 96 | 1200 | 123 | 27.09 MB |
| PrimitiveWetModel | 63 | 8 | FullClenshawGrid | 95 | 1200 | 149 | 26.85 MB |
| PrimitiveWetModel | 63 | 8 | OctahedralGaussianGrid | 96 | 1200 | 195 | 18.58 MB |
| PrimitiveWetModel | 63 | 8 | OctahedralClenshawGrid | 95 | 1200 | 262 | 18.33 MB |
| PrimitiveWetModel | 63 | 8 | HEALPixGrid | 95 | 1200 | 362 | 13.58 MB |
| PrimitiveWetModel | 63 | 8 | OctaHEALPixGrid | 95 | 1200 | 183 | 16.53 MB |

| - | - | - | - | - | - | - |
| PrimitiveWetModel | 31 | 8 | 48 | 2400 | 1255 | 5.19 MB |
| PrimitiveWetModel | 42 | 8 | 64 | 1800 | 782 | 8.75 MB |
| PrimitiveWetModel | 63 | 8 | 96 | 1200 | 197 | 18.58 MB |
| PrimitiveWetModel | 85 | 8 | 128 | 900 | 105 | 32.32 MB |
| PrimitiveWetModel | 127 | 8 | 192 | 600 | 26 | 71.26 MB |
| PrimitiveWetModel | 170 | 8 | 256 | 450 | 9 | 126.62 MB |

| PrimitiveWetModel | 31 | 8 | true | true | 2400 | 1921 | 5.19 MB |
| PrimitiveWetModel | 31 | 8 | true | false | 2400 | 4490 | 5.19 MB |
| PrimitiveWetModel | 31 | 8 | false | true | 2400 | 2654 | 5.19 MB |

| - | - | - | - |
| pressure_gradient_flux! | 48.500 μs| 184.28 KiB| 2898 |
| linear_virtual_temperature! | 1.225 μs| 0 bytes| 0 |
| temperature_anomaly! | 3.156 μs| 0 bytes| 0 |
| geopotential! | 2.806 μs| 0 bytes| 0 |
| vertical_integration! | 14.292 μs| 0 bytes| 0 |
| surface_pressure_tendency! | 17.375 μs| 65.16 KiB| 1390 |
| vertical_velocity! | 4.292 μs| 0 bytes| 0 |
| linear_pressure_gradient! | 1.387 μs| 0 bytes| 0 |
| vertical_advection! | 86.333 μs| 4.81 KiB| 21 |
| vordiv_tendencies! | 174.625 μs| 19.06 KiB| 254 |
| temperature_tendency! | 238.750 μs| 21.53 KiB| 319 |
| humidity_tendency! | 231.000 μs| 21.53 KiB| 319 |
| bernoulli_potential! | 80.958 μs| 7.02 KiB| 114 |

| - | - | - | - | - | - | - |
| ShallowWaterModel | 31 | 1 | 48 | 2400 | 40823 | 1.70 MB |
| ShallowWaterModel | 42 | 1 | 64 | 1800 | 18570 | 2.89 MB |
| ShallowWaterModel | 63 | 1 | 96 | 1200 | 5089 | 6.24 MB |
| ShallowWaterModel | 85 | 1 | 128 | 900 | 1940 | 11.03 MB |
| ShallowWaterModel | 127 | 1 | 192 | 600 | 513 | 25.09 MB |
| ShallowWaterModel | 170 | 1 | 256 | 450 | 134 | 45.91 MB |
| ShallowWaterModel | 255 | 1 | 384 | 300 | 37 | 110.22 MB |

| - | - | - | - | - | - | - |
| PrimitiveWetModel | Float32 | 31 | 8 | 2400 | 1637 | 5.19 MB |
| PrimitiveWetModel | Float64 | 31 | 8 | 2400 | 192 | 9.61 MB |

| PrimitiveDryModel | 31 | 8 | true | true | 2400 | 2657 | 4.86 MB |
| PrimitiveDryModel | 31 | 8 | true | false | 2400 | 6796 | 4.86 MB |
| PrimitiveDryModel | 31 | 8 | false | true | 2400 | 3396 | 4.86 MB |

## Number of vertical layers

| - | - | - | - | - | - |
| PrimitiveWetModel | 31 | 4 | 2400 | 2438 | 3.57 MB |
| PrimitiveWetModel | 31 | 8 | 2400 | 1936 | 5.19 MB |
| PrimitiveWetModel | 31 | 12 | 2400 | 1487 | 6.82 MB |
| PrimitiveWetModel | 31 | 16 | 2400 | 1668 | 8.46 MB |

maximilian-gelbrecht:
Benchmarks look okay to me, @milankl there weren't really that many changes to PR since your intial review. I would merge this if all tests pass.

milankl:
Go for it!

PR #684, opened 2025-03-04
run! with steps argument

milankl:
I thought it would be neat if
```julia
run!(simulation, steps=5)
```

was also possible, so that also some specified numer of time steps can be run. I find myself in tests often in the position where I just want to make sure that the model runs for some time steps without throwing an error, so something like the above would be neat.

### DIFF
--- a/src/dynamics/clock.jl
+++ b/src/dynamics/clock.jl
@@ -52,46 +52,35 @@ function Base.copy!(clock::Clock, clock_old::Clock)
     return nothing 
 end 
 
-"""
-$(TYPEDSIGNATURES)
-Initialize the clock with the time step `Δt` in the `time_stepping`."""
-function initialize!(
-    clock::Clock,
-    time_stepping::AbstractTimeStepper;
-    n_timesteps::Int = -1,
-)
-    # if >=0 use to set n_timesteps, otherwise calculate from period
-    if n_timesteps >= 0
-        clock.n_timesteps = n_timesteps
-        clock.period = Second(time_stepping.Δt_sec * n_timesteps)
-    else
-        clock.n_timesteps = ceil(Int, clock.period.value/time_stepping.Δt_sec)
-    end
-    
+"""$(TYPEDSIGNATURES)
+Initialize the clock with the time step `Δt` from `time_stepping`."""
+initialize!(clock::Clock, time_stepping::Leapfrog, args...) =
+    initialize!(clock, time_stepping.Δt_millisec, args...)
+
+"""$(TYPEDSIGNATURES)
+Initialize the clock with the time step `Δt` and `period` to integrate for."""
+function initialize!(clock::Clock, Δt::Period, period::Period)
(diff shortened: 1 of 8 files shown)


src/dynamics/clock.jl
```diff
@@ -53,14 +55,30 @@ end
...
+    end
+    
     clock.start = clock.time    # store the start time
     clock.timestep_counter = 0  # reset counter
     clock.Δt = time_stepping.Δt_millisec
     return clock
 end
 
+function set_period!(clock::Clock, steps::Integer)
+    clock.period = Second(period)
```
maximilian-gelbrecht, line 78: There's a mistake here

src/models/simulation.jl
```diff
@@ -51,14 +55,25 @@ callbacks and performs the first two initial time steps to spin up the
...
+    period::Period = DEFAULT_PERIOD,
+    steps::Int = DEFAULT_TIMESTEPS,
     output::Bool = false,
 )
     progn, diagn, model = unpack(simulation)
 
     # SET THE CLOCK
-    set_period!(progn.clock, period)                            # set how long to integrate for
-    initialize!(progn.clock, simulation.model.time_stepping)    # store the start date, reset counter
+    (; clock) = progn
```
maximilian-gelbrecht, line 65: To me it seems like the cleaner solution to put all of this in `initialize!(::Clock)`

maximilian-gelbrecht (commented):
Yes, I like this. I also often just want to take the first few steps of the model and this makes it easier. Also good for any future ML/DA things.

src/dynamics/clock.jl
milankl: ah yes, ooops, haha. 🙈 removed that now

src/models/simulation.jl
milankl: Yes, just did that with the latest commit. Removed the `set_period!` function+methods and it's all in `initialize!` now with several methods for that

src/models/simulation.jl
```diff
@@ -51,14 +55,23 @@ callbacks and performs the first two initial time steps to spin up the
...
+    (; time_stepping) = model
+    if steps != DEFAULT_TIMESTEPS
+        # sets the steps, calculate period from it, store the start date, reset counter
+        @assert period == DEFAULT_PERIOD "Period and steps cannot be set simultaneously"
+        initialize!(clock, time_stepping, steps)    
+    else
+        # set period = how long to integrate for, tore the start date, reset counter
+        initialize!(clock, time_stepping, period)
```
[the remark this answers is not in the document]
milankl: Added a test `@test_throws AssertionError run!(simulation, steps=1, period=Day(1))` that this is indeed the case (it is)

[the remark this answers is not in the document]
maximilian-gelbrecht: Oh, yes, I overlooked that.

PR #686, opened 2025-03-06
Make OctaminimalGaussianGrid default

milankl:
Because it's faster at low resolution. Testing whether this passes all tests actually.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Make OctaminimalGaussianGrid default [#686](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/686)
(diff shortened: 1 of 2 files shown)


milankl:
```julia
Test Summary:                      |   Pass   Total  Time
Transform: Orography (exact grids) | 139112  139112  2.5s
NF = Float32: Test Failed at /home/runner/work/SpeedyWeather.jl/SpeedyWeather.jl/test/transforms/spectral_gradients.jl:42
  Expression: abs(div_lm) < sqrt(eps(NF))
   Evaluated: 0.0006569333f0 < 0.00034526698f0
```
the spectral tests are failing because not set to test on the exact OctahedralGaussianGrid!

milankl:
I don't think we should do this at this stage

Issue #689, opened 2025-03-07
Differentiability and NaNs


maximilian-gelbrecht:
@milankl This is continuing a bit from our conversation on Zulip 

So, currently we have `NaN` in all land and ocean variables for the masked grid points. That's not a problem in itself for the regular model but for the differentiability it becomes a problem. All mathematical operations that involve a `NaN` yield `NaN`. This includes the operations that are performed when accumulating derivatives. 

So currently, when differentiating the model, the derivative with respect to any parameter that's involved in some land and ocean calculations just yields `NaN`. 

One quite uncomplicated way to deal with it is to always check explicitly with `infinite`. We already do this in some functions, but not everywhere. 

Take for example the soil moisture: 

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/6d277fcc00aee8bfe1df9f8b1a59232211c3eba9/src/physics/land/vegetation.jl#L156-L166

If we add a 

```Julia 
isfinite(soil_moisture[ij])
``` 
 
in l. 94. Then we don't get `NaN` gradients anymore when differentiating. 

I am writing this as an issue and not a PR at the moment because I don't have a good overview at the moment of all the cases in the model where we do such operations and because it's also affecting #677

milankl:
So soil moisture isn't that much of a problem as in the current state of #677 we have with `LandBucketMoisture` simply 0s over the ocean (which makes sense, there's no soil so the moisture in the soil is zero). So while I agree we can add an `isfinite` check here, I don't necessarily see it as super important as there's a higher level solution, simply make soil moisture zero. However, for temperature that's a little trickier, as zero temperature clearly means something else than just the absence of, say, sea surface temperature over land or soil temperature over the ocean. But in general I'm probably more in favour of simply extending all fields across coastlines, replace NaNs/masks from the netCDF files with zeros or, say, 0C=273.15K and then we can still mask things for the output?

maximilian-gelbrecht:
Yeah, you are right for soil moisture it's also easily solvable with just having it zero over the ocean, this makes sense. 

Will this work correctly for temperatures though? Because then you get some kind of dynamics for those temperature as well and they won't stay at this value. I don't immediately see how we would solve that like this. Maybe it's just easier to do the `isfinite` check there e.g. in `ocean_timestep!`. We could also define a `eachfinitegridpoint` or `eachoceangridpoint`, something like this.

maximilian-gelbrecht:
[...]

PR #692, opened 2025-03-14
Isfinites in ocean and land to avoid NaNs 

maximilian-gelbrecht:
This is partially addressing #689 

I also tried doing it with the land sea mask, but this still yields `NaN` in the derivatives in the end. I suspect that's because the `mask[ij] > 0` isn't the same as the `NaN` in `soil_moisture_availability`, but I might be wrong, I was running out of time now to check that (I'll look into it later).

### DIFF
--- a/src/physics/land/vegetation.jl
+++ b/src/physics/land/vegetation.jl
@@ -49,7 +49,9 @@ function soil_moisture_availability!(
     r = 1/(D_top*W_cap + D_root*(W_cap - W_wilt))
 
     @inbounds for ij in eachindex(soil_moisture_availability)
-        soil_moisture_availability[ij] = r*D_top*soil_moisture[ij, 1]
+        if isfinite(soil_moisture_availability[ij])
+            soil_moisture_availability[ij] = r*D_top*soil_moisture[ij, 1]
+        end 
     end
 
     return nothing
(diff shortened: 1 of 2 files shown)


src/physics/land/vegetation.jl
```diff
@@ -49,7 +49,9 @@ function soil_moisture_availability!(
     r = 1/(D_top*W_cap + D_root*(W_cap - W_wilt))
 
     @inbounds for ij in eachindex(soil_moisture_availability)
-        soil_moisture_availability[ij] = r*D_top*soil_moisture[ij, 1]
+        if isfinite(soil_moisture_availability[ij])
+            soil_moisture_availability[ij] = r*D_top*soil_moisture[ij, 1]
+        end
```
milankl, line 54: The `isfinite` check should be on `soil_moisture` not `soil_moisture_availability` given that `soil_moisture` is overwritten anyway!

src/physics/land/vegetation.jl
```diff
@@ -169,11 +171,13 @@ function soil_moisture_availability!(
 
     for ij in eachgridpoint(soil_moisture_availability, high_cover, low_cover)
         
-        # Fortran SPEEDY source/land_model.f90 line 111 origin unclear
-        veg = max(0, high_cover[ij] + low_veg_factor*low_cover[ij])
-
-        # Fortran SPEEDY documentation eq. 51
-        soil_moisture_availability[ij] = r*(D_top*soil_moisture[ij, 1] +
-            veg*D_root*max(soil_moisture[ij, 2] - W_wilt, 0))
+        if isfinite(soil_moisture_availability[ij])
```
milankl, line 174: same here

src/physics/ocean.jl
```diff
@@ -161,8 +161,10 @@ function ocean_timestep!(
     weight = convert(NF, Dates.days(time-Dates.firstdayofmonth(time))/Dates.daysinmonth(time))
 
     @inbounds for ij in eachgridpoint(sea_surface_temperature)
-        sea_surface_temperature[ij] = (1-weight) * monthly_temperature[ij, this_month] +
-                                          weight  * monthly_temperature[ij, next_month]
+        if isfinite(sea_surface_temperature[ij])
+            sea_surface_temperature[ij] = (1-weight) * monthly_temperature[ij, this_month] +
+                                            weight  * monthly_temperature[ij, next_month]
+        end
```
milankl, line 167: and here too. `monthly_temperature` is read from file which may contain NaNs, or at least following our current logic the mask of that netCDF variable should be replaced with NaNs!

maximilian-gelbrecht:
I adjusted the land sea mask test to use the `SlabOcean` instead of a climatology based one because of #694 

But somehow I get errors from the long wave radiation tests now, even though this PR does nothing to those.

milankl:
@maximilian-gelbrecht should this be closed? We've generally adopted a philosophy that while data that is read in may contain NaNs these are replaced with a fallback value and the responsibility of masking is solely left to the land-sea mask. So there shouldn't be any (masked) NaNs within `simulation.diagnostic_variables`

maximilian-gelbrecht:
The motivation for this was differentiability where `NaN` quickly propagate. I haven't tested it again (but will do eventually). Anyway this particular PR is now so outdated that we'd have to start fresh again anyway. So closing it is fine.

PR #702, opened 2025-04-05
Documentation of convection

sunmoumou1:
Hey awesome team!

First off, I just wanted to express my sincere thanks for all the hard work you’ve put into this project. It’s been a pleasure going through the Physics section today. I took a second look and caught some errors, so I’ve fixed those. Additionally, there were a few spots that left me scratching my head, and I’ve expanded on them a bit for better clarity. Oh, and I’ve also added some images. I believe these visuals will provide a clearer overview of the parameterization process, which could be super helpful for other readers. Feel free to consider including them in the documentation – I think they’ll fit right in!

That said, I might’ve missed a few things, so please double-check the changes to ensure everything is in top shape.

Once again, thank you all for your incredible work. I’d love to continue contributing to this project as a contributor! Looking forward to collaborating more with all of you.


milankl:
This is fantastic @sunmoumou1 ! Quick check: You produced all visualisations yourself? Would love to add them, can I request though that you don't add them to  `/docs` but instead, e.g. copy them in here, and then only put the link into the docs? That way we exclude graphs from the git history which is nice to not overload that with updated pictures ...

milankl:
Oh and can you add a line to CHANGELOG.md? That one has to be updated of tests to pass

docs/src/large_scale_condensation.md
```diff
@@ -127,7 +127,7 @@ The tendencies ``\delta q`` in units of  kg/kg/s are vertically
 integrated to diagnose the large-scale precipitation ``P`` in units of meters
 
 ```math
-P = -\int \frac{\Delta t}{g \rho} \delta q dp
+P = - \int_{top}^{bottom} \frac{\Delta t}{g \rho} \delta q dp
```
milankl, line 130: I think this is bottom to top if written down mathematically. $\delta q$ is negative in the occurence of precipitation, $dp$ is positive when integrating top to bottom (i.e. $p=0 to p=p_0$) and the fraction is positive too. So it's only the negative sign of $\delta q$ that needs to be switched to have a positive amount of precipitation accordingly.

milankl (commented):
Looks great, I need to see the derivation of the vertical velocity compiled ...

milankl:
> I’d love to keep contributing to SpeedyWeather long-term

You're already making a very good impresssion, many thanks for your help! Feel free to play with the model, address any issues or suggest features as you see fit and have capacity, happy to help out with any contribution you'd like to make!

sunmoumou1:
> > I’d love to keep contributing to SpeedyWeather long-term
> 
> You're already making a very good impresssion, many thanks for your help! Feel free to play with the model, address any issues or suggest features as you see fit and have capacity, happy to help out with any contribution you'd like to make!


Thanks! I would like to contribute in a long term. You can merge any commits that you think are right into the main branch~

milankl:
Copying in the figure from @sunmoumou1 to have links independent of other repositories

image

image

image

image

milankl:
@sunmoumou1 I changed your derivation of $P_q = 0$ for shallow convection a little bit as you started with that objective and made use of it too. I thought it was cleaner to not do that but then still arrive at $= 0$. I hope you don't mind. Let me know if it's not clear

PR #726, opened 2025-05-05
Leapfrog (re)start, Implicit reinitialize option, RestartFile, ProgressTxt, ParametersTxt as callback

### LINKED ISSUE #703: Loading a previously-saved state and continuing a simulation
Hi,

Thanks for developing SpeedyWeather! It's brilliant - I've been using it with procedurally-generated planets and getting some really good results.

I'm very new to Julia (I only started using it with this project), so apologies if I've missed something obvious, but one thing I'd really like to be able to do is generate a NetCDF file for several hours or days of the running simulation, shut down Julia and then restart it again later to continue the simulation where it left off, essentially so I can save another NetCDF file with later days of the simulation.

I notice that the NetCDFOutput can create a 'restart.jld2' file, and I was assuming that would allow me to do what I'm looking for, but when I attempt to load it via Julia I can't figure out how to get things back running again.

Is this something that's currently possible?

Thanks!

Chris.

milankl:
fixes #703 

TODO

- [x] add `start_with_euler` option to `Leapfrog`, rework `first_timesteps!`, `timestep!` to use that
- [x] no more lossy compression of prognostic variables for bitreproducible restart 
- [x] ImplicitShallowWater, ImplicitPrimitiveEquation has `reinitialize` option
- [x] RestartFile, ProgressTxt, ParametersTxt reimplemented as callback  
- [x] test this
- [x] check bitreproducibility with/without restart when not reinitializing the implicit solver


milankl:
Buildkite fails because of

```julia
GPU Barotropic: Error During Test at /var/lib/buildkite-agent/builds/gpuci-2/julialang/speedyweather-dot-jl/test/GPU/barotropic.jl:1
  MethodError: no method matching mul!(::CuArray{ComplexF32, 1, CUDA.DeviceMemory}, ::CUDA.CUFFT.CuFFTPlan{ComplexF32, Float32, -1, false, 1, 1, CuArray{ComplexF32, 1, CUDA.DeviceMemory}}, ::SubArray{Float32, 1, Vector{Float32}, Tuple{UnitRange{Int64}}, true}, ::Bool, ::Bool)
  The function `mul!` exists, but no method is defined for this combination of argument types.
```

milankl:
I've checked the bit reproducibility with leapfrog restarts now but can't quite get it to be fully exact, so marked this test as `@test_broken` for now. Simulations are bitreproducible when executing the same simulation twice but currently not if one interrupts half way and restarts with leapfrog (not Euler) unfortunately. Haven't fully figured this out yet

src/dynamics/initial_conditions.jl
```diff
@@ -144,8 +144,14 @@ function initialize!(   progn::PrognosticVariables{NF},
     (; GridVariable2D, grid, nlayers) = model.spectral_grid
     (; radius) = model.planet
     A = convert(NF, initial_conditions.max_speed) * 2
-    u = 2A*rand(GridVariable2D, grid) .- A
-    v = 2A*rand(GridVariable2D, grid) .- A
+
+    # sample vector to use RNG (not implemented for RingGrids)
+    npoints = RingGrids.get_npoints(grid)
+    u_data = on_architecture(grid.architecture, rand(RNG, NF, npoints))
+    v_data = on_architecture(CPU(), rand(RNG, NF, npoints))
```
maximilian-gelbrecht: The error from buildkite was caused by these arrays not being allocated on GPU.

maximilian-gelbrecht:
I think I fixed the error you saw in buildkite, but let's see what it says now. 

For the proper GPU version we have to rework and adapt all the initial conditions as well. I just fixed the default one now, so that our one example in the CI can run. 

The problem is, even if we want to do the initialisation still only on CPU for now, this causes problems with the initial conditions because they often use the transforms, and the transforms of the model do only work on the selected architecture, so when initialisating a GPU model, the transforms can only transform GPU arrays.

maximilian-gelbrecht:
The error from the documentation CI also isn't trivial but related to some change in the initial conditions. I didn't look into fixing that one now though. Just reading the error kind of suggests there's something going wrong with 2d vs 3d `curl!` application there.

milankl:
> I think I fixed the error you saw in buildkite, but let's see what it says now.

Awesome, yeah! I ran into the problem that reproducibility wasn't given because we don't have a `rand(rng, grid)` generator function for field, so a specific random number generator `rng` cannot be used, only the global one. I make a mental note to somehow add this to RingGrids. And then I didn't realise that I would need to move this to the right architecture 🙈 

> The error from the documentation CI also isn't trivial

I had created a ShallowWaterModel with nlayers=8 ... the error looked hard, but it's not. I guess we currently throw a warning for nlayers > 1 with the ShallowWaterModel but we should just throw an error instead. Fixed it now

src/dynamics/time_integration.jl
```diff
@@ -220,6 +229,29 @@ function leapfrog!(
...
+    (; Δt) = time_stepping
+
+    # decide whether to start with 1x Euler then 1x Leapfrog at Δt
+    if time_stepping.start_with_euler
+        first_timesteps!(progn, diagn, model)
+        time_stepping.start_with_euler = false      # disable for subsequent restarts
+        
+    else    # or continue with leaprog steps at 2Δt (e.g. restart)
+            # but make sure that implicit solver is initialized in that situation
+        initialize!(model.implicit, 2Δt, diagn, model)
```
maximilian-gelbrecht, line 245: This does nothing if it's already initialised, correct?

src/dynamics/time_integration.jl
```diff
@@ -119,6 +122,12 @@ function initialize!(L::Leapfrog, model::AbstractModel)
     if nΔt != output_dt
         @warn "$n steps of Δt = $(L.Δt_millisec.value)ms yield output every $(nΔt.value)ms (=$(nΔt.value/1000)s), but output_dt = $(output_dt.value)s"
     end
+
+    # reset to start with Euler step because
+    # simulation = initialize! also reallocates the prognostic variables
+    # by default without setting the 2nd step
+    L.start_with_euler = true
```
maximilian-gelbrecht: the start_with_euler is always true, except for when the restart without reinialization is chosen?

src/models/simulation.jl
```diff
@@ -5,12 +5,12 @@ $(TYPEDSIGNATURES)
 Simulation is a container struct to be used with `run!(::Simulation)`.
 It contains
 $(TYPEDFIELDS)"""
-struct Simulation{Model<:AbstractModel} <: AbstractSimulation{Model}
+struct Simulation{Model<:AbstractModel, Progn, Diagn} <: AbstractSimulation{Model}
```
maximilian-gelbrecht, line 8: good catch!

src/output/restart_file.jl
```diff
@@ -1,44 +1,53 @@
...
-    
-    # move 2nd leapfrog to 1st to compress restart file
-    get_step(progn.vor, 1)   .= get_step(progn.vor, 2)
-    get_step(progn.div, 1)   .= get_step(progn.div, 2)
-    get_step(progn.temp, 1)  .= get_step(progn.temp, 2)
-    get_step(progn.humid, 1) .= get_step(progn.humid, 2)
-    get_step(progn.pres, 1)  .= get_step(progn.pres, 2)
-
-    # bitround 1st leapfrog step to output precision
-    if eltype(progn) <: Base.IEEEFloat  # currently not defined for other formats...
```
maximilian-gelbrecht, line 24: why was that here, and now it's not needed anymore?

src/models/simulation.jl
```diff
@@ -98,6 +98,5 @@ function finalize!(simulation::AbstractSimulation)
     unscale!(progn)                                 # undo radius-scaling for vor, div from the dynamical core
     unscale!(diagn)                                 # undo radius-scaling for vor, div from the dynamical core
     finalize!(model.output, simulation)             # possibly post-process output, then close netCDF file
-    write_restart_file!(model.output, progn)         # as JLD2
```
maximilian-gelbrecht, line 101: that's now part of the callbacks, that's why it's not needed as anymore, correct?

src/output/netcdf_output.jl
```diff
@@ -278,13 +285,16 @@ function initialize!(
...
-    # also export parameters into run????/parameters.txt
-    parameters_txt = open(joinpath(output.run_path, "parameters.txt"), "w")
-    for property in propertynames(model)
-        println(parameters_txt, "model.$property")
-        println(parameters_txt, getfield(model, property,), "\n")
-    end
-    close(parameters_txt)
+    # CALLBACKS
+    # add ParametersTxt callback
+    output.write_parameters_txt && add!(model.callbacks, :parameters_txt => ParametersTxt())
```
maximilian-gelbrecht, line 290: I like that, how it's now done with the Callbacks!

maximilian-gelbrecht (approved):

src/dynamics/time_integration.jl
milankl: good call, this shouldn't happen. RestartFile doesn't do any lossy compression on the prognostic variables now so this line should go. But maybe we add a warning somewhere in case someone tries to restart from a file that doesn't contain any 2nd leapfrog step

src/dynamics/time_integration.jl
milankl: Yes we have a line like

```julia
    # option to skip reinitialization at restart
    (implicit.initialized && !implicit.reinitialize) && return nothing
```

now for `ImplicitPrimitiveEquation`.

milankl: So you can force it to reinitialize but normally skipped. Only the ImplicitShallowWater is always "reinitialized" as it just needs to precompute one constant

src/models/simulation.jl
milankl: correct

src/output/restart_file.jl
milankl: There was some lossy compression for `restart.jld2` which I thought was just consistent with the NetCDFOutput compression but actually it makes very little difference in terms of overall output data size and interferes with being able to restart with leapfrog so I just took it out and there's no lossy compression anymore for the restart file. Only lossless but I don't it does much

src/dynamics/time_integration.jl
milankl: Okay changed this, it's a little more complicated/confusing so should add a paragraph in the docs to explain this better

justinfocus12:
Is there clarity at this point on bit-reproducibility of restarting an interrupted simulation? I have found that 
`run!(sim, 10)`
`run!(sim, 10)`
is not equivalent to 
`run!(sim, 20)`
even with `start_with_euler=false`. (I've also verified the two chunks of 10 is repeatable by writing and reading a restart file in between, so the issue is not I/O with JLD2). 

I'm using v0.21.1, with the Primitive Dry equations with Held-Suarez forcing. I can paste a more complete minimum failing example, but it's a bit involved so just want to ask if it's resolved. Apologies if I've misinterpreted the status of the thread above.

maximilian-gelbrecht:
This PR is already outdated by now, since then we had a major revision of the time stepping (e.g. introducing an NCycle Lorenz solver and generally providing a flexible interface for alternative solvers). 

Ultimately we favoured reasonable simplicity of our time stepping approach over bit-reprodcucibility of interrupted simulations. Of course they are still _approximately_ equal, but not bit-identical. 

Can you make a good case why you'd absolutely need bit-reproducibility of restarting an interrupted simulations?

milankl:
Hi Justin!! Yeah we had introduced a restart functionality that avoids discretization errors from the Euler warmup step before Leapfrog continues when the 2nd leapfrog step is available in the second (or later) `run!` calls. But this has been since removed again when generalising our time stepping scheme. For the future the idea is that this should again be possible with the NCycleLorenz time stepper (which doesn't need a warmup step, it's more like a RK method) for the primitive models, however, the primitive models currently don't support this time stepper yet. For barotropic/shallow water this is already possible and part of CI here

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/d2349a5d9e49519c606f7540b5c3d4aea52863bd/SpeedyWeather/test/dynamics/time_stepping.jl#L227-L272

As you see bitreproducibility (in the sense of one `run!` call vs two covering in total the same `period`) is still not fully given. This feature/property is at the moment not high priority for us but do let us know if it is for you. With a bit of testing it should be easy to isolate why it's not but at the moment it's not urgent for us hence we never bothered.

justinfocus12:
Thanks both Max and Milan for the update! It's very helpful to know I'm not missing something already. I'll explain why this feature would be nice to have for my research, but I can survive without it. 

I'm working on sampling and control algorithms for extreme weather events, in particular how to make optimal perturbations to enhance (or weaken) the severity of a heat wave, blocking event, rainstorm, etc. This involves creating datasets of simulations that branch off from each other at systematic times (e.g., before or during an extreme event to change its course), and studying how effective are different kinds of perturbations. For example, is a pulse of diabatic heating over Kamchatka, or a wave-2 wiggle in the jet stream, more potent at altering the 2021 Pacific Northwest heat wave? 

I want to examine this "response function" that maps initial perturbations (e.g., `dP` as a change in surface pressure) to final temperature changes (`dT`), and have it be deterministic unless I impose randomness explicitly through a seeded pseudo-random number generator. Procedurally, I would follow something like the Ensemble Boosting procedure (numbers chosen arbitrarily):
* run a "parent" simulation for 500 days, in chunks of 100 with restarts saved in between (`restart100.jld2`, ... `restart400.jld2`, `restart500.jld2`).
* find the biggest temperature spike. Suppose it happens at day 313. 
* Determine that day 293 (20 days in advance of the spike) is an appropriate time to launch an ensemble of alternatives. Suppose I want 9 ensemble members with perturbations `dP1`,...,`dP9`. 

What I do next depends on whether I have bit-reproducibility or not. 

## Procedure 1 (if restarting is bit-reproducible)
* Set up a "launch pad" in the form of a special restart file `restart293.jld2`, by repeating a 93-day simulation from day 200 to day 293. 
* Launch my 9 members from `restart293.jld2`, each with a small perturbation, and run it through day 313. 
* Analyze the function `dP->dT` to my heart's content 

## Procedure 2 (if restarting is not bit-reproducible)
* determine exactly the effect of "restart noise", by running one special un-interrupted simulation from `restart200.jld2` to day 301, and calculating the full state difference between that and what comes out of `restart300.jld2`
* Launch the 9 members from `restart200.jld2`
* At day 293, add the deliberate perturbation `dP`
* At day 300,  artificially add the restart noise
* Run the members through day 313

Clearly, procedure 2 is much less efficient and also dubious---I don't know if this restart noise should be considered additive or multiplicative, for example. Of course, you could argue reasonably that any conclusions I draw from this map `dP->dT` should be robust to restart noise. Ultimately I would agree, but for prototyping and exploration purposes, bit-reproducibility would be a huge practical advantage. 

Of course, I could work around restart noise by carefully designing the ensembles to prevent such an awkward situation, e.g., with smaller chunks in accordance with the error growth timescale. That might well be less hassle than the tedious engineering of ensuring bit-reproducibility on your end. I have no idea about how to do that, but would be happy to contribute to this effort insofar as I can.

Feel free to suggest alternative strategies, and of course it's ultimately your prerogative as the developers!

PR #727, opened 2025-05-05
Restructure (GPU) utilities and work towards spectral gradients GPU 

maximilian-gelbrecht:
I basically adapted the structure that Oceananigans uses. 

The kernel launching is still very much WIP and will change based on what exactly we need when we tackle the actual KA kernels.


maximilian-gelbrecht:
Still a lot of work left to do, but there's now a draft of how I imagine the code should be restructured, and the `∇²!` kernel abstractions version runs on CPU, and is somehow even faster on CPU than the old version.

maximilian-gelbrecht:
@milankl So, this is a bit preliminary draft, there’s no thorough review necessary, but already ready to be briefly looked at I think or more importantly just discuss some basics 

What did I do here? 
1. Restructure regular utitilties and device utilities into separate submodules (Architectures, Utilis [very similar as Oceananigans])
2. Wrote general kernel launching functions (very similar to Oceananigans) 
3. Started with the kernels for spectral gradients (there are only two missing when I write this, maybe I add them already tomorrow or so). So far all of these are quite naive implementations and not yet really optimised. At the moment these are just separate versions in the `spectral_gradients.jl` file called e.g. `∇²_KA!` instead of `∇²!`. In the end, we hopefully just have to keep the KA version. 

I tested everything on CPU and GPU, but no really comprehensive benchmarks yet. 

1. “Normal” kernels that just run over regular indices are even on CPU faster than the old version
2. Kernels that use recursive relations are slower on CPU than the old implementations, most likely because we need so many if checks or indices to be computed. It’s not easy to make those faster. I tested two variants so far:
    1. Just “naively” directly port them, this means a lot of if-clauses in the kernel 
    2. Split the kernel into three kernels: diagonal, last row and all remaining  -> so far this makes little difference on CPU, on GPU the three kernel version is a bit slower 

On GPU in high resolution (I tested T128L64) all GPU KA are much, much faster than the CPU version (~100x) on the same machine, but I haven’t done really comprehensive benchmarks yet. 

Some other general thoughts about the GPU version: 

The way we have our structs initialisation at the moment, all fields are constructed on CPU first and then transferred automatically to GPU when the actual struct constructor is called. I think this is actually a good default, because it should always work. Transferring all arrays to GPU in the end, makes the initialisation much slower though. 

We should review the constructors to replace the CPU specific arrays with the `ArrayType_` where this doesn’t need extra kernels and just runs on GPU without adjustments as well. I haven’t done any of that yet. As this only affects initialisation it does have a bit lower priority. 

Btw: there are also some problems with circular dependencies with CUDA at the moment, I couldn’t fix those yet.

maximilian-gelbrecht:
With #732, #734 already half using the changes here with `Architectures` introduced, I'll restructure this PR a little bit to make it mergeable quicker. Basically, I'll put those kernels that are fast already in the main model, but everything else stays out of the main code. 

So, let's start with `∇²!` for this the KA version seems to be flat out faster on CPU (and GPU of course). There's a benchmark in `benchmark/KA/laplacian_benchmark.jl`. So this will become the only version in the main model code. Here are the results: 

```
Size (L×M×N)      ∇²! (CPU)          ∇²_KA! (CPU{KernelAbstractions.CPU})     Speedup
 33× 32× 1        0.006 ms         0.002 ms          3.10x
 33× 32× 8        0.047 ms         0.007 ms          6.80x
 65× 64× 1        0.022 ms         0.004 ms          5.31x
 65× 64× 8        0.179 ms         0.019 ms          9.55x
129×128× 1        0.086 ms         0.012 ms          7.06x
129×128× 8        0.696 ms         0.070 ms          9.98x
257×256× 8        2.773 ms         0.271 ms         10.24x
513×512× 8       11.744 ms         1.110 ms         10.58x
```

maximilian-gelbrecht:
After a bit of optimisation, `∇!` is basically right on the edge of what we deemed as acceptable (x0.5)

``` 
Size (L×M×N)      old ∇! (CPU)          new (KA) ∇! (SpeedyWeather.Architectures.CPU{CPU})     Speedup
 33× 32× 1        0.001 ms         0.002 ms          0.24x
 33× 32× 8        0.004 ms         0.009 ms          0.46x
 65× 64× 1        0.002 ms         0.005 ms          0.38x
 65× 64× 8        0.016 ms         0.032 ms          0.51x
129×128× 1        0.008 ms         0.017 ms          0.45x
129×128× 8        0.063 ms         0.118 ms          0.53x
257×256× 8        0.243 ms         0.459 ms          0.53x
513×512× 8        0.964 ms         1.901 ms          0.51x
```

jackleland:
Looks great Max! The kernel launching logic is lovely, if I'm understanding correctly `worksize` is just the size of the indexing array in a given case? 

Also, is there a reason there's some KA kernels in the CUDA extension at the moment (SpeedyWeatherCUDAExt.spectral_gradients.jl) - presumably just living there temporarily?

maximilian-gelbrecht:
> Looks great Max! The kernel launching logic is lovely, if I'm understanding correctly `worksize` is just the size of the indexing array in a given case?

Yes, that's all currently. 

> 
> Also, is there a reason there's some KA kernels in the CUDA extension at the moment (SpeedyWeatherCUDAExt.spectral_gradients.jl) - presumably just living there temporarily?

Yes, it's kind of temporarily, but some may also end up there in the end. I haven't fully decided it yet. Basically, ideally we have just one version of a function as a KA kernel in the main code, but in some cases the KA version might be significantly slower on CPU than the current version. In these cases the KA version might just go into the CUDA extension. We could also put the actual kernel in the main code and just a function barrier in the extension, so that CuArrays use the KA version, not 100% settled yet on how to do that best.

maximilian-gelbrecht:
divergence and curl are performance critical. We'll keep the old version as the CPU version. 

For GPU, for the KA version there's also a bit of a trade off to be made: currently we do this very elegantly with anonymous functions `kernel` that decide whether it's a curl, divergence and whether the sign is flippend and it's added to the input or not. For KA, this results in worse performance than hard coding a kernel for every single of these cases. That's a bit unfortunate, because it's a lot of code redundancies. I tried instead to define a type `Operator{div, add, flip}` and dispatch based on that, but this didn't result in a better performance. 

For now, as a first draft I include a working (but somewhat slower) version with just a single kernel.

maximilian-gelbrecht:
The merge with the Spectrum PR got a bit messy. I need to continue to clean this up and fix it next week.

maximilian-gelbrecht:
Here are now the benchmark results. I decided to not run everything on my M3 / ARM as well, because the performance on x86 is already decent in itself, the ARM is only better. Generally the difference between the old CPU version and the CPU KA version is smaller on my M3, and in some cases KA+M3 is even faster than the old version on the M3. 

To sum up the tests very briefly: 

x86 - CPU old vs CPU KA: Almost everywhere x0.5 - x1 range that we deemed acceptable 

x86 CPU old (single core) vs GPU: The exact scaling differs slightly by function but in the high resolutions (T512L32) we get everywhere at least x100, the break even point is somewhere around the T64L8 resolution, at medium high resolutions such as T256L8 we get roughly x25 across the board. 

I did try out a few things with splitting kernels, and a few other things, but the GPU kernels are certainly not yet completely optimised. 

What to do: 

**Based on these results I would keep only the KA version for all functions in spectral_gradients.jl** I'll do this clean up in the next days. 


## Divergence

### CPU vs CPU Benchmarks

**Benchmarking divergence implementations**  
Standard implementation architecture: SpeedyWeather.Architectures.CPU{CPU}  
KA implementation architecture: SpeedyWeather.Architectures.CPU{CPU}

| Size (L×M×N) | Standard divergence! (ms) | KA divergence_KA! (ms) | Speedup |
|--------------|---------------------------|------------------------|---------|
| 33× 32× 1    | 0.002                     | 0.006                  | 0.31x   |
| 33× 32× 8    | 0.011                     | 0.023                  | 0.48x   |
| 65× 64× 1    | 0.006                     | 0.011                  | 0.56x   |
| 65× 64× 8    | 0.045                     | 0.076                  | 0.59x   |
| 129×128× 1   | 0.021                     | 0.039                  | 0.54x   |
| 129×128× 8   | 0.180                     | 0.289                  | 0.62x   |
| 257×256× 8   | 0.736                     | 1.133                  | 0.65x   |
| 513×512× 8   | 2.899                     | 4.547                  | 0.64x   |
| 513×512×16   | 5.793                     | 19.415                 | 0.30x   |
| 513×512×32   | 11.626                    | 45.876                 | 0.25x   |

### CPU vs GPU Benchmarks

**Benchmarking divergence implementations**  
Standard implementation architecture: SpeedyWeather.Architectures.CPU{CPU}  
KA implementation architecture: SpeedyWeather.Architectures.GPU{CUDABackend}

| Size (L×M×N) | Standard divergence! (ms) | KA divergence_KA! (ms) | Speedup  |
|--------------|---------------------------|------------------------|----------|
| 33× 32× 1    | 0.002                     | 0.033                  | 0.06x    |
| 33× 32× 8    | 0.011                     | 0.033                  | 0.33x    |
| 65× 64× 1    | 0.006                     | 0.033                  | 0.18x    |
| 65× 64× 8    | 0.045                     | 0.033                  | 1.34x    |
| 129×128× 1   | 0.021                     | 0.033                  | 0.63x    |
| 129×128× 8   | 0.179                     | 0.033                  | 5.38x    |
| 257×256× 8   | 0.730                     | 0.035                  | 21.07x   |
| 513×512× 8   | 2.888                     | 0.042                  | 68.80x   |
| 513×512×16   | 5.793                     | 0.060                  | 96.66x   |
| 513×512×32   | 11.747                    | 0.085                  | 138.38x  |

## Gradient

### CPU vs CPU Benchmarks

**Benchmarking ∇ implementations**  
Standard implementation architecture: SpeedyWeather.Architectures.CPU{CPU}  
KA implementation architecture: SpeedyWeather.Architectures.CPU{CPU}

| Size (L×M×N) | Old ∇! (ms) | New (KA) ∇! (ms) | Speedup |
|--------------|-------------|------------------|---------|
| 33× 32× 1    | 0.002       | 0.005            | 0.39x   |
| 33× 32× 8    | 0.014       | 0.019            | 0.72x   |
| 65× 64× 1    | 0.007       | 0.010            | 0.68x   |
| 65× 64× 8    | 0.056       | 0.059            | 0.94x   |
| 129×128× 1   | 0.025       | 0.030            | 0.85x   |
| 129×128× 8   | 0.225       | 0.228            | 0.99x   |
| 257×256× 8   | 0.902       | 0.908            | 0.99x   |
| 513×512× 8   | 3.564       | 3.568            | 1.00x   |
| 513×512×16   | 7.172       | 7.523            | 0.95x   |
| 513×512×32   | 14.367      | 15.132           | 0.95x   |

### CPU vs GPU Benchmarks

**Benchmarking ∇ implementations**  
Standard implementation architecture: SpeedyWeather.Architectures.CPU{CPU}  
KA implementation architecture: SpeedyWeather.Architectures.GPU{CUDABackend}

| Size (L×M×N) | Old ∇! (ms) | New (KA) ∇! (ms) | Speedup  |
|--------------|-------------|------------------|----------|
| 33× 32× 1    | 0.002       | 0.036            | 0.05x    |
| 33× 32× 8    | 0.014       | 0.036            | 0.38x    |
| 65× 64× 1    | 0.007       | 0.037            | 0.18x    |
| 65× 64× 8    | 0.056       | 0.036            | 1.54x    |
| 129×128× 1   | 0.026       | 0.036            | 0.71x    |
| 129×128× 8   | 0.221       | 0.036            | 6.08x    |
| 257×256× 8   | 0.902       | 0.037            | 24.40x   |
| 513×512× 8   | 3.572       | 0.041            | 86.15x   |
| 513×512×16   | 7.142       | 0.051            | 138.74x  |
| 513×512×32   | 14.374      | 0.084            | 171.40x  |

## Laplacian

### CPU vs CPU Benchmarks

**Benchmarking ∇² implementations**  
Standard implementation architecture: SpeedyWeather.Architectures.CPU{CPU}  
KA implementation architecture: SpeedyWeather.Architectures.CPU{CPU}

| Size (L×M×N) | Old ∇²! (ms) | New (KA) ∇²! (ms) | Speedup |
|--------------|--------------|-------------------|---------|
| 33× 32× 1    | 0.001        | 0.003             | 0.22x   |
| 33× 32× 8    | 0.006        | 0.013             | 0.42x   |
| 65× 64× 1    | 0.003        | 0.006             | 0.40x   |
| 65× 64× 8    | 0.020        | 0.043             | 0.47x   |
| 129×128× 1   | 0.011        | 0.018             | 0.58x   |
| 129×128× 8   | 0.088        | 0.162             | 0.55x   |
| 257×256× 8   | 0.346        | 0.652             | 0.53x   |
| 513×512× 8   | 1.382        | 2.549             | 0.54x   |
| 513×512×16   | 2.790        | 6.318             | 0.44x   |
| 513×512×32   | 5.582        | 16.218            | 0.34x   |

### CPU vs GPU Benchmarks

**Benchmarking ∇² implementations**  
Standard implementation architecture: SpeedyWeather.Architectures.CPU{CPU}  
KA implementation architecture: SpeedyWeather.Architectures.GPU{CUDABackend}

| Size (L×M×N) | Old ∇²! (ms) | New (KA) ∇²! (ms) | Speedup |
|--------------|--------------|-------------------|---------|
| 33× 32× 1    | 0.001        | 0.020             | 0.04x   |
| 33× 32× 8    | 0.006        | 0.020             | 0.31x   |
| 65× 64× 1    | 0.003        | 0.020             | 0.13x   |
| 65× 64× 8    | 0.020        | 0.020             | 1.04x   |
| 129×128× 1   | 0.011        | 0.020             | 0.55x   |
| 129×128× 8   | 0.090        | 0.020             | 4.53x   |
| 257×256× 8   | 0.346        | 0.021             | 16.72x  |
| 513×512× 8   | 1.387        | 0.025             | 55.63x  |
| 513×512×16   | 2.807        | 0.033             | 85.49x  |
| 513×512×32   | 5.643        | 0.059             | 95.78x  |

## UV_from_vordiv

### CPU vs CPU vs CPU Benchmarks

**Benchmarking UV_from_vordiv implementations**  
Standard implementation architecture: SpeedyWeather.Architectures.CPU{CPU}  
KA implementation architecture: SpeedyWeather.Architectures.CPU{CPU}  
KA split implementation architecture: SpeedyWeather.Architectures.CPU{CPU}

| Size (L×M×N) | Standard (ms) | KA original (ms) | KA split (ms) | Speedup (KA/std) | Speedup (split/KA) |
|--------------|---------------|------------------|---------------|------------------|-------------------|
| 33× 32× 1    | 0.002         | 0.006            | 0.010         | 0.42x            | 0.61x             |
| 33× 32× 8    | 0.020         | 0.027            | 0.045         | 0.75x            | 0.59x             |
| 65× 64× 1    | 0.009         | 0.012            | 0.019         | 0.74x            | 0.66x             |
| 65× 64× 8    | 0.081         | 0.100            | 0.143         | 0.81x            | 0.70x             |
| 129×128× 1   | 0.039         | 0.047            | 0.069         | 0.84x            | 0.67x             |
| 129×128× 8   | 0.318         | 0.387            | 0.560         | 0.82x            | 0.69x             |
| 257×256× 8   | 1.287         | 1.556            | 2.190         | 0.83x            | 0.71x             |
| 513×512× 8   | 5.125         | 6.338            | 8.673         | 0.81x            | 0.73x             |
| 513×512×16   | 10.373        | 17.640           | 38.974        | 0.59x            | 0.45x             |
| 513×512×32   | 20.565        | 36.584           | 95.154        | 0.56x            | 0.38x             |

### CPU vs GPU vs GPU Benchmarks

**Benchmarking UV_from_vordiv implementations**  
Standard implementation architecture: SpeedyWeather.Architectures.CPU{CPU}  
KA implementation architecture: SpeedyWeather.Architectures.GPU{CUDABackend}  
KA split implementation architecture: SpeedyWeather.Architectures.GPU{CUDABackend}

| Size (L×M×N) | Standard (ms) | KA original (ms) | KA split (ms) | Speedup (KA/std) | Speedup (split/KA) |
|--------------|---------------|------------------|---------------|------------------|-------------------|
| 33× 32× 1    | 0.003         | 0.037            | 0.060         | 0.07x            | 0.62x             |
| 33× 32× 8    | 0.020         | 0.038            | 0.060         | 0.51x            | 0.64x             |
| 65× 64× 1    | 0.010         | 0.037            | 0.061         | 0.25x            | 0.62x             |
| 65× 64× 8    | 0.080         | 0.038            | 0.060         | 2.12x            | 0.63x             |
| 129×128× 1   | 0.039         | 0.038            | 0.061         | 1.01x            | 0.63x             |
| 129×128× 8   | 0.318         | 0.039            | 0.060         | 8.16x            | 0.65x             |
| 257×256× 8   | 1.276         | 0.040            | 0.062         | 32.14x           | 0.64x             |
| 513×512× 8   | 5.106         | 0.051            | 0.069         | 100.26x          | 0.74x             |
| 513×512×16   | 10.226        | 0.073            | 0.089         | 140.89x          | 0.81x             |
| 513×512×32   | 21.077        | 0.102            | 0.130         | 207.48x          | 0.78x             |

maximilian-gelbrecht:
The PR is almost finished now, I just have to fix the differentiability issue. It's caused by the new kernel version of UV_from_vordiv!. In isolation it works, so this will need a bit deeper investigation next week.

benchmark/KA/spectral_gradients/README.md
maximilian-gelbrecht, line 1: We might also delete all of these benchmarking files again before merging.

src/RingGrids/field.jl
```diff
@@ -501,7 +501,13 @@ function KernelAbstractions.get_backend(
     return KernelAbstractions.get_backend(field.data)
 end
 
-function Adapt.adapt_structure(to, field::F) where {F <: AbstractField}
-    # TODO this reuses the same grid but adapt can change the array type which should also change the architecture?
-    return Field(Adapt.adapt(to, field.data), field.grid)
-end
\ No newline at end of file
+function Adapt.adapt_structure(to, field::Field{T, N, ArrayType, Grid}) where {T, N, ArrayType, Grid}
```
maximilian-gelbrecht, line 504: In order for a Field to acceptable for a GPU kernel, the Grid also has to be on GPU, that's why we do these adapts, and `on_architecture` here. 

At the same time I changed the `Vector{UnitRange}` for `rings`, `orders`, etc in `Grid` and `Spectrum` to tuples. This is more architecture agnostic and also how it's implemented e.g. for `SubArray` in JuliaBase

src/SpeedyTransforms/spectral_transform.jl
```diff
@@ -210,21 +219,34 @@ function SpectralTransform(
             grad_y1[l, m] = -(l-2)*ϵlms[l, m]
             grad_y2[l, m] = (l+1)*ϵlms[l+1, m]
         end
+        # explicitly set the last row to zero, so that kernels yield correct gradient in last row 
+        grad_y2[lmax, m] = 0
     end
 
     # meridional gradient used to get from u, v/coslat to vorticity and divergence
+    grad_x_vordiv = zeros(Complex{Int}, spectrum)
```
maximilian-gelbrecht: this is a pretty minor optimisation to include the `im` here in the array. A `Complex{Int}` matrix isn't big, but I am also okay with reversing that again

maximilian-gelbrecht:
The merge with #732 was successful. 

There are two things I still have to figure out

- [x] What's best to do with the pre-computed indices on GPU, that's a bit of a tricky thing. It wasn't much of an issue for `Spectrum` but with RingGrids that's more important. -> solved by converting `Vector{UnitRange}` into `Tuple`s
- [ ] What's wrong with `UV_from_vordiv!` with Enzyme (both CPU and GPU) 
- [ ] Also update the documentation: make a WIP page about Architectures

maximilian-gelbrecht:
@milankl The only remaining issue here is making the KA version of `UV_from_vordiv!` differentiable by Enzyme. I don't think I'll have much time in the next days (or maybe week) to work on this. So maybe that's already a good point for an initial review if you have some spare time. It's a pretty big PR and otherwise I think it's finished (except for updating the documentation).

maximilian-gelbrecht:
ok, I spoke too early. With the merge and the changes to adapt everything to it and make it suitable for GPU, a big performance regression came. I'll have to investigate this...

maximilian-gelbrecht:
Cthulhu suggests there might be a problem with type inference for `DiagnosticVariables` and `PrognosticVariables` now. So far I failed isolating and fixing it.

maximilian-gelbrecht:
~~The problem here is that I turned `orders` and `rings` into tuples, so that they can be on the CPU (as we want), even when the other precomputed fields are transferred to the GPU and adapted for kernel usage.~~

~~I tried to model this after how `SubArray` is implemented, the indices are saved as a tuple of indices there as well.~~

~~This turned to be quite hard to make type stable though. I am using a `ntuple(... , Val(N))` call to initialise the tuple at compile time like SubArray. Not quite sure what to do to improve this.~~

Solved by just keeping everything as Arrays und in case we really need it for an iteration is explicitly transferred to Vector if it's on GPU. This is only needed during pre-computations anyway, so it's not really a problem.

maximilian-gelbrecht:
I did resolve the performance problems, the differentiability of `UV_from_vordiv!` is still an issue. I tried to narrow it down, but despite some efforts, it's hard to understand what the problem is here. 

 I would think that we can also resolve this differentiability problem in another PR, and merge this with the old pure CPU version from `UV_from_vordiv!`.

maximilian-gelbrecht:
# Benchmarks

created for SpeedyWeather.jl v0.15.0 on Mon, 30 Jun 2025 11:12:09. 

### Machine details

```julia
julia> versioninfo()
Julia Version 1.10.9
Build Info:
Platform Info:
  CPU: 8 × Apple M3
  WORD_SIZE: 64
  LIBM: libopenlibm
```

### Explanation

- Δt: time step [s].

### Running the benchmarks

```> julia manual_benchmarking.jl```

## Models, default setups

| - | - | - | - | - | - | - |
| BarotropicModel | 31 | 1 | false | 2400 | 57294 | 1.67 MB |
| ShallowWaterModel | 31 | 1 | false | 2400 | 43685 | 1.71 MB |
| PrimitiveDryModel | 31 | 8 | true | 2400 | 2709 | 4.87 MB |
| PrimitiveWetModel | 31 | 8 | true | 2400 | 1861 | 5.20 MB |

## Grids

| PrimitiveWetModel | 63 | 8 | FullGaussianGrid | 96 | 1200 | 142 | 27.13 MB |
| PrimitiveWetModel | 63 | 8 | FullClenshawGrid | 95 | 1200 | 150 | 26.89 MB |
| PrimitiveWetModel | 63 | 8 | OctahedralGaussianGrid | 96 | 1200 | 241 | 18.62 MB |
| PrimitiveWetModel | 63 | 8 | OctahedralClenshawGrid | 95 | 1200 | 250 | 18.36 MB |
| PrimitiveWetModel | 63 | 8 | HEALPixGrid | 95 | 1200 | 394 | 13.61 MB |
| PrimitiveWetModel | 63 | 8 | OctaHEALPixGrid | 95 | 1200 | 258 | 16.56 MB |

| - | - | - | - | - | - | - |
| PrimitiveWetModel | 31 | 8 | 48 | 2400 | 1924 | 5.20 MB |
| PrimitiveWetModel | 42 | 8 | 64 | 1800 | 800 | 8.76 MB |
| PrimitiveWetModel | 63 | 8 | 96 | 1200 | 251 | 18.62 MB |
| PrimitiveWetModel | 85 | 8 | 128 | 900 | 97 | 32.39 MB |
| PrimitiveWetModel | 127 | 8 | 192 | 600 | 27 | 71.39 MB |
| PrimitiveWetModel | 170 | 8 | 256 | 450 | 8 | 126.85 MB |

| PrimitiveWetModel | 31 | 8 | true | true | 2400 | 1911 | 5.20 MB |
| PrimitiveWetModel | 31 | 8 | true | false | 2400 | 4420 | 5.20 MB |
| PrimitiveWetModel | 31 | 8 | false | true | 2400 | 2579 | 5.20 MB |

| - | - | - | - |
| pressure_gradient_flux! | 48.083 μs| 184.28 KiB| 2898 |
| linear_virtual_temperature! | 1.229 μs| 0 bytes| 0 |
| temperature_anomaly! | 3.167 μs| 0 bytes| 0 |
| geopotential! | 2.806 μs| 0 bytes| 0 |
| vertical_integration! | 14.375 μs| 0 bytes| 0 |
| surface_pressure_tendency! | 17.125 μs| 65.16 KiB| 1390 |
| vertical_velocity! | 4.518 μs| 0 bytes| 0 |
| linear_pressure_gradient! | 1.387 μs| 0 bytes| 0 |
| vertical_advection! | 86.375 μs| 4.81 KiB| 21 |
| vordiv_tendencies! | 176.208 μs| 19.06 KiB| 254 |
| temperature_tendency! | 241.000 μs| 21.53 KiB| 319 |
| humidity_tendency! | 232.917 μs| 21.53 KiB| 319 |
| bernoulli_potential! | 80.250 μs| 7.02 KiB| 114 |

| - | - | - | - | - | - | - |
| ShallowWaterModel | 31 | 1 | 48 | 2400 | 41387 | 1.71 MB |
| ShallowWaterModel | 42 | 1 | 64 | 1800 | 18780 | 2.90 MB |
| ShallowWaterModel | 63 | 1 | 96 | 1200 | 5186 | 6.27 MB |
| ShallowWaterModel | 85 | 1 | 128 | 900 | 1974 | 11.09 MB |
| ShallowWaterModel | 127 | 1 | 192 | 600 | 516 | 25.23 MB |
| ShallowWaterModel | 170 | 1 | 256 | 450 | 154 | 46.15 MB |
| ShallowWaterModel | 255 | 1 | 384 | 300 | 33 | 110.76 MB |

| - | - | - | - | - | - | - |
| PrimitiveWetModel | Float32 | 31 | 8 | 2400 | 1467 | 5.20 MB |
| PrimitiveWetModel | Float64 | 31 | 8 | 2400 | 185 | 9.62 MB |

| PrimitiveDryModel | 31 | 8 | true | true | 2400 | 2670 | 4.87 MB |
| PrimitiveDryModel | 31 | 8 | true | false | 2400 | 6195 | 4.87 MB |
| PrimitiveDryModel | 31 | 8 | false | true | 2400 | 3374 | 4.87 MB |

## Number of vertical layers

| - | - | - | - | - | - |
| PrimitiveWetModel | 31 | 4 | 2400 | 2694 | 3.58 MB |
| PrimitiveWetModel | 31 | 8 | 2400 | 1928 | 5.20 MB |
| PrimitiveWetModel | 31 | 12 | 2400 | 1478 | 6.83 MB |
| PrimitiveWetModel | 31 | 16 | 2400 | 1671 | 8.47 MB |

File: ext/SpeedyWeatherCUDAExt/SpeedyWeatherCUDAExt.jl [hunk 3]
```diff
@@ -1,34 +1,41 @@
...
+# DEVICE SETUP FOR CUDA
+# extend functions from main SpeedyWeather 
 
 # for RingGrids and LowerTriangularArrays:
 # every Array needs this method to strip away the parameters
 RingGrids.nonparametric_type(::Type{<:CuArray}) = CuArray
 LowerTriangularArrays.nonparametric_type(::Type{<:CuArray}) = CuArray
 
-SpeedyWeather.default_array_type(::Type{GPU}) = CuArray
+array_type(::GPU) = CuArray
```
milankl, line 17: So if we want other GPUs than CUDA then loading another extension, say SpeedyWeatherMetalExt, would simply define `array_type(::GPU) = MtlArray` ?

src/Architectures.jl
```diff
@@ -0,0 +1,110 @@
+module Architectures 
+    
+    import KernelAbstractions 
+
+    export AbstractArchitecture
+    export CPU, CPUStatic, GPU, CUDAGPU
+    export array_type, on_architecture, architecture, device, convert_to_device
```
milankl: We also have `array_type` in RingGrids defined to return the array type of a field. It's not exported so probably fine to export this here! But question is it important to export?

src/Architectures.jl
```diff
@@ -0,0 +1,110 @@
...
+        array_type(arch::AbstractArchitecture)
+
+    Return the array type that's used with the architecture `arch`.
+    """
+    array_type(::CPU) = Array
+    array_type(::Type{<:CPU}) = Array
+
+    ismatching(arch::Type{<:AbstractArchitecture}, array_T::Type{<:AbstractArray}) = array_T <: array_type(arch)
+    ismatching(arch::AbstractArchitecture, array_T::Type{<:AbstractArray}) = ismatching(typeof(arch), array_T)
+    ismatching(arch::AbstractArchitecture, array::AbstractArray) = ismatching(arch, typeof(array))
```
milankl: We also use `ismatching` in SpeedyTransforms to check whether transform and field or transform and lower triangular. I don't see a likely conflict (and given there is no dependency on SpeedyTransforms couldn't be extended only vice versa) and one could still do `SpeedyTransforms.ismatching(transform, field)` etc to avoid the clashing namespace. Also `SpeedyTransforms.ismatching` is currently not exported so probably fine!

src/LowerTriangularArrays/LowerTriangularArrays.jl
```diff
@@ -4,17 +4,18 @@ module LowerTriangularArrays
 using DocStringExtensions
 
 # GPU
-import Adapt
+import Adapt: Adapt, adapt
 import GPUArrays
 import KernelAbstractions
+import ..Architectures: AbstractArchitecture, on_architecture, array_type, ismatching, CPU, GPU, architecture
```
milankl: So `ismatching` is imported into the LowerTriangularArrays namespace which however is not exported when `using LowerTriangularArrays` in SpeedyTransforms, so still okay, just tracking potential conflicts!

src/LowerTriangularArrays/lower_triangular_array.jl
```diff
@@ -14,15 +14,19 @@ struct LowerTriangularArray{T, N, ArrayType <: AbstractArray{T, N}, S<:AbstractS
     LowerTriangularArray{T, N, ArrayType, S}(data::AbstractArray, spectrum::S) where {T, N, ArrayType <: AbstractArray{T, N}, S <: AbstractSpectrum} =
         check_lta_input_array(data, spectrum, N) ? 
         new(data, spectrum)  :
-        error(lta_error_message(data, spectrum, T, N, ArrayType))
+        error(lta_error_message(data, spectrum, T, N, ArrayType, S))
 end
 
 check_lta_input_array(data, spectrum, N) =
-    (ndims(data) == N) & (length(data) == prod(size(data)[2:end]) * nonzeros(spectrum)) 
+    (ndims(data) == N) & (length(data) == prod(size(data)[2:end]) * nonzeros(spectrum)) # & ismatching(spectrum, typeof(data)) TODO: reactivate this? problem for constructors
```
milankl, line 21: I would leave this commented out for now I believe having an architecture in spectrum that doesn't match the type of the data will trigger an error in some kernels anyway so maybe it's nice to not unnecessarily overconstrain here before we know it's really needed!

src/LowerTriangularArrays/lower_triangular_array.jl
```diff
@@ -14,15 +14,19 @@ struct LowerTriangularArray{T, N, ArrayType <: AbstractArray{T, N}, S<:AbstractS
...
 
-function lta_error_message(data, spectrum, T, N, ArrayType) 
-    return "$(Base.dims2string(size(data))) $(typeof(data)) cannot be used to create "*
-            "a $(Base.dims2string(matrix_size(data, spectrum))) LowerTriangularArray{$T, $N, $ArrayType}"
+function lta_error_message(data, spectrum, T, N, ArrayType, S) 
+    if ismatching(spectrum, typeof(data))
+        return "Architecture mismatch: Spectrum has architecture $(spectrum.architecture), but array type $(typeof(data)) is used"
```
milankl: given this is not checked for in `check_lta_input_array` this error message can be thrown without it being the triggering error in `check_lta_input_array` !?

src/LowerTriangularArrays/lower_triangular_array.jl
```diff
@@ -137,6 +141,18 @@ function Base.array_summary(io::IO, L::LowerTriangularMatrix{T}, inds::Tuple{Var
     print(io, Base.dims2string(length.(inds)), ", $(mn[1])x$(mn[2]) LowerTriangularMatrix{$T}")
 end
 
+function Base.show(io::IO, ::MIME"text/plain", L::LowerTriangularArray)
```
milankl, line 140: what's this for? Short comment how that `show`s differently?

src/LowerTriangularArrays/spectrum.jl
```diff
@@ -47,11 +58,28 @@ Spectrum(trunc::Integer; one_degree_more=false, kwargs...) =
...
+            spectrum.m_indices, 
+            spectrum.lm_orders))
+
 triangle_number(m::Integer) = m*(m+1)÷2
 nonzeros(l::Integer, m::Integer) = l*m - triangle_number(m-1)
 nonzeros(s::Spectrum) = nonzeros(s.lmax, s.mmax)
 resolution(s::Spectrum) = (s.lmax, s.mmax)
 truncation(s::Spectrum) = s.mmax - 1
+orders(s::Spectrum) = s.orders
+orders(s::Spectrum{<:GPU}) = Vector(s.orders) # on GPU transfer orders back to CPU first
```
milankl, line 80: Would that not convert to `Array` in every GPU kernel where `orders(::Spectrum)` is used?

milankl: Ah no, only for the kernel launching which is done on the CPU, right?!

src/LowerTriangularArrays/spectrum.jl
```diff
@@ -97,4 +125,9 @@ function Base.show(io::IO, S::Spectrum)
     print(io,   "└ architecture: $(typeof(S.architecture))")
 end
 
-eachorder(s::Spectrum) = s.lm_orders
\ No newline at end of file
+ismatching(s::Spectrum, array_type::Type{<:AbstractArray}) = ismatching(s.architecture, array_type)
+ismatching(s::Spectrum, array::AbstractArray) = ismatching(s.architecture, typeof(array))
```
milankl: Related to comments above: This tells me ideally we should extend `Architecture.ismatching` in SpeedyTransforms!

src/RingGrids/field.jl
```diff
@@ -501,7 +501,13 @@ function KernelAbstractions.get_backend(
...
+    adapted_data = adapt(to, field.data)
+    if ismatching(field.grid, typeof(adapted_data))
+        return Field(adapted_data, adapt(to, field.grid))
+    else # if not matching, create new grid with other architecture
+        return Field(adapted_data, adapt(to, Grid(field.grid, architecture(typeof(adapted_data)))))
```
milankl: This would silently create independent `field.grid` instances though if we accidentally create a grid with an architecture that doesn't match the type of the data array?

milankl: We could throw an error/warning for testing things temporarily?

src/RingGrids/grid.jl
```diff
@@ -169,15 +175,17 @@ and then each grid point per ring. To be used like
 
 Accesses precomputed `grid.rings`."""
 @inline eachring(grid::AbstractGrid) = grid.rings
+@inline eachring(grid::AbstractGrid{<:GPU}) = Vector(grid.rings) # on GPU transfer indices back to CPU first
```
milankl, line 178: Same here, this is for kernel launching which requires `Array` right?

src/RingGrids/grid.jl
```diff
@@ -169,15 +175,17 @@ and then each grid point per ring. To be used like
...
 """$(TYPEDSIGNATURES)
 Computes the ring indices `i0:i1` for start and end of every longitudinal point
 on a given ring `j` of `Grid` at resolution `nlat_half`. Used to loop
 over rings of a grid. These indices are also precomputed in every `grid.rings`."""
 function eachring(Grid::Type{<:AbstractGrid}, nlat_half::Integer)
-    rings = Vector{UnitRange{Int}}(undef, get_nlat(Grid, nlat_half))    # allocate
-    each_index_in_ring!(rings, Grid, nlat_half)                         # calculate iteratively
-    return rings
+    nlat = get_nlat(Grid, nlat_half)
+    rings = Vector{UnitRange{Int}}(undef, nlat)    # allocate
```
milankl, line 186: Wait didn't you say somewhere above that you wanted indices to be tuples not `Vector`s?

src/RingGrids/grid.jl
```diff
@@ -240,7 +248,7 @@ end
 
 """$(TYPEDSIGNATURES) Obtain ring index `j` from gridpoint `ij` and `rings`
 describing rind indices as obtained from `eachring(::Grid)`"""
-function whichring(ij::Integer, rings::AbstractVector)
+function whichring(ij::Integer, rings)
```
milankl, line 251: This is so that `rings` can be a tuple too?

src/RingGrids/grids/full_clenshaw.jl
```diff
@@ -45,4 +45,6 @@ get_latd(::Type{<:FullClenshawGrid}, nlat_half::Integer) = [90 - 90j/nlat_half f
 get_lond(::Type{<:FullClenshawGrid}, nlat_half::Integer) = get_lond(FullGaussianGrid, nlat_half)
 
 # QUADRATURE
-get_quadrature_weights(::Type{<:FullClenshawGrid}, nlat_half::Integer) = clenshaw_curtis_weights(nlat_half)
\ No newline at end of file
+get_quadrature_weights(::Type{<:FullClenshawGrid}, nlat_half::Integer) = clenshaw_curtis_weights(nlat_half)
+
+Adapt.@adapt_structure FullClenshawGrid
```
milankl, line 54: What does this do?

src/RingGrids/reverse.jl
```diff
@@ -4,7 +4,7 @@ function Base._reverse!(field::AbstractField, ::Val{:lat})
 
     rings = eachring(field)
     nrings = length(rings)
-    rings_north = view(rings, 1:nrings ÷ 2)
+    rings_north = rings[1:nrings ÷ 2]
```
milankl: removed because `rings` can be a tuple? For the `reverse` function it's probably fine to allocate something it's not used inside the dynamical core anyway

src/SpeedyTransforms/spectral_gradients.jl
```diff
@@ -1,5 +1,42 @@
...
 
+"""
+    KernelOP{mode, flipsign, add}
+
+Type for dispatching on kernel operations in spectral gradient calculations.
+- `mode`: `true` for curl, `false` for divergence
+- `flipsign`: `true` or `false` to negate the result
+- `add`: `true` or `false` to add to the output instead of overwriting
+"""
+struct KernelOP{mode, flipsign, add} end
```
milankl, line 11: Ah interesting approach! 🎉

src/SpeedyTransforms/spectral_transform.jl
```diff
@@ -10,26 +10,33 @@ to perform a spectral transform. Fields are
...
+    MatrixComplexType,                    # <: ArrayType{Complex{NF}, 2},
+    ArrayComplexType,                     # <: ArrayType{Complex{NF}, 3},
+    LowerTriangularMatrixType,            # <: LowerTriangularArray{NF, 1, ArrayType{NF}},
+    ComplexIntLowerTriangularMatrixType,  # <: LowerTriangularArray{Complex{Int}, 1, ArrayType{Complex{Int}}},
+    ComplexLowerTriangularMatrixType,     # <: LowerTriangularArray{Complex{NF}, 1, ArrayType{Complex{NF}}},
+    LowerTriangularArrayType,             # <: LowerTriangularArray{NF, 2, ArrayType{NF}},
 } <: AbstractSpectralTransform
 
+    # Architecture
+    architecture::AR
```
milankl, line 27: Is it preferred to have `spectral_transform.architecture` rather than `spectral_transform.spectrum.architecture` or `spectral_transform.grid.architecture`, they all point to the same architecture anyway no?

milankl: (just feels like a duplication)

src/SpeedyTransforms/spectral_transform.jl
```diff
@@ -205,37 +214,52 @@ function SpectralTransform(
     grad_y1 = zeros(NF, spectrum)                           # term 1, mul with harmonic l-1, m
     grad_y2 = zeros(NF, spectrum)                           # term 2, mul with harmonic l+1, m
 
-    for (m, degrees) in enumerate(spectrum.orders)          # 1-based degree l, order m
+    for (m, degrees) in enumerate(orders(spectrum))          # 1-based degree l, order m
```
milankl, line 215: Because this is the precomputation on the CPU the `orders` here is returning a vector right? (see comment above)

src/SpeedyTransforms/spectral_transform.jl
milankl: I think when I originally wrote this my thought was that it would be faster the other way around: Because `Complex{Int}` is 128bits instead of 64 it's better to load less from memory and move to `im` to the compute kernel!

src/SpeedyTransforms/spectral_transform.jl
```diff
@@ -205,37 +214,52 @@ function SpectralTransform(
...
             grad_y_vordiv2[l, m] = (l-1)*ϵlms[l+1, m]
         end
+        # explicitly set the last row to zero, so that kernels yield correct gradient in last row 
+        grad_y_vordiv1[lmax, m] = 0
+        grad_y_vordiv2[lmax, m] = 0
     end
 
     # zonal integration (sort of) to get from vorticity and divergence to u, v*coslat
-    vordiv_to_uv_x = LowerTriangularMatrix([-m/(l*(l+1)) for l in 0:(lmax-1), m in 0:(mmax-1)])
+    vordiv_to_uv_x = LowerTriangularMatrix([-m/(l*(l+1))*im for l in 0:(lmax-1), m in 0:(mmax-1)], spectrum)
```
milankl: same here, isn't it better to reduce memory access when something like `im` can be just inside the kernel? Also maybe an LTA like this one can be just recomputed on the fly? Includes a division which is expensive but maybe on a GPU better than reading 2x the memory in?

src/Utils/Utils.jl
```diff
@@ -0,0 +1,16 @@
+module Utils
```
milankl, line 1: new module just to tidy things up? Happy with that!

src/dynamics/diagnostic_variables.jl
```diff
@@ -165,16 +165,16 @@ $(TYPEDFIELDS)"""
...
     SpectralVariable2D,     # <: LowerTriangularArray
     SpectralVariable3D,     # <: LowerTriangularArray
     GridVariable2D,         # <: AbstractField
     GridVariable3D,         # <: AbstractField
 } <: AbstractDiagnosticVariables
     
-    spectrum::SP            # spectral resolution: maximum degree and order of spherical harmonics
-    grid::Grid              # grid resolution: number of latitude rings on one hemisphere (Eq. incl.)
+    spectrum::SpectrumType            # spectral resolution: maximum degree and order of spherical harmonics
+    grid::GridType              # grid resolution: number of latitude rings on one hemisphere (Eq. incl.)
```
milankl, line 177: Thanks for doing all of these consistently!

src/dynamics/prognostic_variables.jl
```diff
@@ -163,13 +163,13 @@ get_step(coeffs::LowerTriangularArray{T, 3}, i) where T = lta_view(coeffs, :, :,
 
 """$(TYPEDSIGNATURES)
 Generator function."""
-function PrognosticVariables(SG::SpectralGrid; nsteps=DEFAULT_NSTEPS)
+function PrognosticVariables(SG::SpectralGrid{Architecture, SpectrumType, GridType}; nsteps=DEFAULT_NSTEPS) where {Architecture, SpectrumType, GridType}
```
milankl, line 166: I think both parametric and non-parametric function signature are somewhat equivalent but I tend to write `typeof` now just because it avoids a `where`. Maybe that was just useful when we ended up changing the parameters of types which requires you to change the function signature all the time ...

src/dynamics/spectral_grid.jl
```diff
@@ -61,7 +62,7 @@ struct SpectralGrid{
     SpectralVariable4D::Type{<:AbstractArray}
     
     # SIZE OF GRID from trunc, Grid, dealiasing:
-    "[OPTION] how to match spectral with grid resolution: dealiasing factor, 1=linear, 2=quadratic, 3=cubic grid"
+    "[OPTION] how to match spectral with grid resolution: dealiasing factor,0=automatically chosen based on Grid, 1=linear, 2=quadratic, 3=cubic grid"
```
milankl: Don't see the `0` being used anywhere!

src/dynamics/spectral_grid.jl
```diff
@@ -119,48 +123,52 @@ function Base.show(io::IO, SG::SpectralGrid)
...
     nlat = RingGrids.get_nlat(grid)
     npoints = RingGrids.get_npoints(grid)
-    dealiasing = SpeedyTransforms.get_dealiasing(trunc, nlat_half)
+
+    # default dealiasing or user-defined one? 
+    dealiasing = SpeedyTransforms.get_dealiasing(trunc, grid.nlat_half)
     
+    # Convert numeric parameters to Float64
+    dealiasing_f64 = Float64(dealiasing)
+    radius_f64 = Float64(radius)
```
milankl, line 162: Why are these needed these fields are `::Float64` in `SpectralGrid` anyway, so conversion happens automatically?

milankl: (I think you did that before and I removed these lines in another PR again ... 🙈 )

milankl (commented):
Awesome, looks fantastic!! Very excited to get this merged 🚀

ext/SpeedyWeatherCUDAExt/SpeedyWeatherCUDAExt.jl
maximilian-gelbrecht: Possible, but might cause troubles. I would rather define this then more like we already do for `JLArray`, then we have `GPU{MetalDevice}` or something similar and have `array_type(::GPU{MetalDevice})=MetalArray`

src/Architectures.jl
maximilian-gelbrecht: Even after years scopes in Julia are sometimes still not 100% clear to me. I thought the export here, in a submodule, is needed so that we can use it in other submodules without explicitly importing it. It's not exported to the user when just doing `using SpeedyWeather` like that (just tested that again).

src/LowerTriangularArrays/LowerTriangularArrays.jl
maximilian-gelbrecht: As I wrote before, scopes are sometimes confusing for me, especially with submodules. I thought it's very "Julia" that we define this everywhere as `ismatching`. But to be sure, when maybe we need to import it explicitly in `SpeedyTransforms` to properly extend it?

src/LowerTriangularArrays/lower_triangular_array.jl
maximilian-gelbrecht: Yes, absolutely. Model construction on GPU otherwise won't work as well. We might consider this again in some future where even our initialisation is on GPU

src/Utils/Utils.jl
maximilian-gelbrecht: Shamelessly copied from Oceananigans ;)

src/LowerTriangularArrays/lower_triangular_array.jl
maximilian-gelbrecht: Oh yes, you are totally right. This is a relic from when I tried to actually enforce that.

src/LowerTriangularArrays/lower_triangular_array.jl
maximilian-gelbrecht: The other one is just the pretty one for `LowerTriangularMatrix`, for higher dimensions we didn't have so far. But with the default one from AbstractArrays, I got some errors on GPU. So this one is here, so that you get a nice return on GPU as well.

src/LowerTriangularArrays/spectrum.jl
maximilian-gelbrecht: This is basically my solution to this problem with the Tuple type stability in the end. 

So, for the GPU kernels, everything that needs to be handed over to a kernel needs to be adapted for CUDA use. You can't have a `Vector` in `Spectrum` or `Grid`. However, we want the iterators still to work, for this purpose the iterators have to be on CPU, because you can't iterate with scalars over GPU arrays. 

This is a bit of a compromise and looks bad, but actually isn't really that bad. The thing is we would never iterate over orders and harmonics in GPU kernels, so this is never actually called during GPU computations. We only need that for the pre-computation / initialisation that is done on CPU anyway.

src/RingGrids/grid.jl
maximilian-gelbrecht: See above my comments, only needed for precomputations / initialisation

src/RingGrids/grid.jl
maximilian-gelbrecht: After thinking about this for a while, I realised that in our situation it's not possible to get tuples type stable. For that you would need to now the tuple size at compile time. You could construct a way how to do this for `Spectrum`, but this wouldn't work for `Spectrum` and `Grid` at the same time. 

So, as described a bit above. All these things stay Arrays.

src/RingGrids/grid.jl
maximilian-gelbrecht: This is a bit of a relic from trying it with Tuples. I could revert that change, because we now still have Arrays, but then, function type declarations like this don't really change anything in this case. So I kept it. But if you have a strong opinion about this, I can reverse it again.

src/RingGrids/grids/full_clenshaw.jl
maximilian-gelbrecht: Everything that's handed over to a GPU kernel needs to be adapted for GPU use. So when you hand over a `Field` the `Grid` also needs to be adapted. This is just a macro that initialises the default for that.

src/RingGrids/reverse.jl
maximilian-gelbrecht: Good catch, yeah, I'll reverse that again to the Array version.

src/SpeedyTransforms/spectral_gradients.jl
maximilian-gelbrecht: Yeah, it turns out to be much, much faster with KernelAbstractions than doing anonymous functions.

src/SpeedyTransforms/spectral_transform.jl
maximilian-gelbrecht: Yes, this also already crossed my mind. They all do point to the same one, so it's not really redundant in terms of memory usage, but I also get the feeling that it feels like duplication.

src/SpeedyTransforms/spectral_transform.jl
maximilian-gelbrecht: Yes

src/SpeedyTransforms/spectral_transform.jl
maximilian-gelbrecht: Yeah, good call. I think I'll reverse that again. uv_from_vordiv is not really performance crucial anyway

src/dynamics/spectral_grid.jl
maximilian-gelbrecht: good catch, this was from an older commit and forgot to reverse it in the comment.

src/dynamics/spectral_grid.jl
maximilian-gelbrecht: there's some weird problem that the kwdef init didn't work with ints anymore, so I diid this here explicitly.

src/LowerTriangularArrays/spectrum.jl
maximilian-gelbrecht: Okay, done that now. I agree that's more proper to do it like this.

src/RingGrids/field.jl
maximilian-gelbrecht: That's a very good point. I still have to check something there. 

Because when launching kernels, all data gets adapted to CuDeviceArray. I have to check if this doesn't cause any allocations here

src/LowerTriangularArrays/LowerTriangularArrays.jl
maximilian-gelbrecht: See blow, done that now.

src/Architectures.jl
maximilian-gelbrecht: See below, hopefully properly extended now.

src/Architectures.jl
milankl: because Architecture exports it into SpeedyWeather but SpeedyWeather doesn't export it to global scope!

milankl: All fine here, CI would flag it if it was an actual problem!

src/dynamics/spectral_grid.jl
milankl: Okay that is weird but happy to leave that in for now we can still change that later if we're annoyed by it

src/LowerTriangularArrays/spectrum.jl
milankl: Okay great thanks for the clarification!

src/RingGrids/grid.jl
milankl: No all good, if not required for dispatch I'm happy to leave out type declarations!

src/SpeedyTransforms/spectral_transform.jl
milankl: It just also means that the same object's type `AR` shows up three types in the parameters of `SpectralTransform` not sure we would want to keep these numbers low unless necessary?

ext/SpeedyWeatherCUDAExt/legendre.jl
```diff
@@ -86,7 +86,7 @@ function SpeedyTransforms._legendre!(
     lmax = lmax-1                       # 0-based max degree l of spherical harmonics
     mmax = mmax-1                       # 0-based max order m of spherical harmonics
 
-    @boundscheck SpeedyTransforms.ismatching(S, specs) || throw(DimensionMismatch(S, specs))
+    @boundscheck SpeedyWeather.Architectures.ismatching(S, specs) || throw(DimensionMismatch(S, specs))
```
milankl: We could also `export Architectures` in SpeedyWeather then it would be `Architectures.ismatching` here or even `import Architectures: ismatching` then it would be simply `ismatching` again?

src/RingGrids/grid.jl
```diff
@@ -272,8 +272,8 @@ whichring(grid::AbstractGrid) = whichring(typeof(grid), grid.nlat_half, grid.rin
 whichring(Grid::Type{<:AbstractGrid}, nlat_half::Integer) = whichring(Grid, nlat_half, eachring(Grid, nlat_half))
 
 # for architectures / adapt 
-ismatching(grid::AbstractGrid, array_type::Type{<:AbstractArray}) = ismatching(grid.architecture, array_type)
-ismatching(grid::AbstractGrid, array::AbstractArray) = ismatching(grid.architecture, typeof(array))
+Architectures.ismatching(grid::AbstractGrid, array_type::Type{<:AbstractArray}) = ismatching(grid.architecture, array_type)
+Architectures.ismatching(grid::AbstractGrid, array::AbstractArray) = ismatching(grid.architecture, typeof(array))
```
milankl, line 276: Yeah, like that because that makes it clearer that we're extending insteand of the `import Architectures: ismatching` statement elsewhere disconnected from this line!

src/SpeedyTransforms/spectral_transform.jl
maximilian-gelbrecht: At the moment, we do launch kernels though with the `S.architecture`. But I guess also generally throughout the model it will okay to just take `architecture(array)` of the first array type that's used in the kernel.

src/RingGrids/field.jl
maximilian-gelbrecht: Yes, this was a good call, I had to adjust the logic a little bit for the `CuDeviceArray`s. I have warnings now turned on at least for some further tests, but I hope my adjustments were already sufficient.

maximilian-gelbrecht:
I feel like this is moving close to being mergeable. I think at the moment I would write nothing about these things in the documentation. Basically, as soon as I have the Barotropic model running on GPU, I'd start documenting this in the proper documentation.

milankl:
Very very happy with this. Go for it 🎉  then a v0.16 release?

maximilian-gelbrecht:
Let's wait for the tests to finish, I want to be sure the warning in the adapt really doesn't trigger anywhere I am not expecting it.

src/SpeedyTransforms/spectral_transform.jl
maximilian-gelbrecht: I adjusted this no so it always takes `architecture(array)` instead and remove the architecture field from the `SpectralTransform`

maximilian-gelbrecht:
Ok, as far as I can see it, the only warning I get is from where we explicitly adapt to `JLArray` in the tests. I'll remove those warnings again for now.

maximilian-gelbrecht:
I'll go over everything once more tomorrow calmly, and then merge it if I don't find anything.

milankl:
Max just flagged this
```julia
julia> GridVariable2D
OctahedralGaussianField{Float32, 1}

julia> GridVariable2D == OctahedralGaussianField{Float32, 1}
false
```

which is a problem of `show` currently as
```julia
Base.show(io::IO, F::Type{<:OctahedralGaussianField{T, N}}) where {T, N} =
    print(io, "OctahedralGaussianField{$T, $N}")
```

but should be removed and only adapted for showing a field with a maybe shorter summary like
```julia
julia> field = zeros(OctahedralGaussianGrid, 2)
88-element, 4-ring OctahedralGaussianField as Vector{Float32} on CPU:
 0.0
 ⋮
```
which can be implemented as outlined in `?Base.showarg`

maximilian-gelbrecht:
Ok, so when we define `Base.showarg` instead of `Base.show` for the Field short hands we get 
```repl
julia> spectral_grid.GridVariable2D
Field{Float32, 1, Vector{Float32}, OctahedralGaussianGrid{SpeedyWeather.Architectures.CPU{KernelAbstractions.CPU}, Vector{UnitRange{Int64}}, Vector{Int64}}}

julia>a = rand(spectral_grid.GridVariable2D, 20)
2320-element, 40-ring OctahedralGaussianField{Float32, 1} with Array on SpeedyWeather.Architectures.CPU{KernelAbstractions.CPU}:
0.22857863
 ⋮
```

I am happy with that, I think

milankl:
"with Array" -> "as Array" ? Looks great to me!!!

src/RingGrids/grids/full_clenshaw.jl
```diff
@@ -24,15 +24,19 @@ end
 nonparametric_type(::Type{<:FullClenshawGrid}) = FullClenshawGrid
 
 # FIELD
-const FullClenshawField{T, N} = Field{T, N, Architecture, Grid} where {Architecture, Grid<:FullClenshawGrid}
+const FullClenshawField{T, N} = Field{T, N, ArrayType, Grid} where {ArrayType, Grid<:FullClenshawGrid}
```
milankl, line 27: Haha oh gosh, I didn't even notice 🙈

src/RingGrids/grids/full_clenshaw.jl
```diff
@@ -24,15 +24,19 @@ end
...
 
 # define grid_type (i) without T, N, (ii) with T, (iii) with T, N but not with <:?Field
 # to not have precendence over grid_type(::Type{Field{...})
 grid_type(::Type{FullClenshawField}) = FullClenshawGrid
 grid_type(::Type{FullClenshawField{T}}) where T = FullClenshawGrid
 grid_type(::Type{FullClenshawField{T, N}}) where {T, N} = FullClenshawGrid
 
-Base.show(io::IO, F::Type{<:FullClenshawField{T, N}}) where {T, N} = print(io, "FullClenshawField{$T, $N}")
+function Base.showarg(io::IO, F::Field{T, N, ArrayType, Grid}, toplevel) where {T, N, ArrayType, Grid<:FullClenshawGrid{A}} where A <: AbstractArchitecture
+    print(io, "FullClenshawField{$T, $N}")
```
milankl, line 36: I've been wondering how to best define `field_type(::Type{Grid})` so that `FullGaussianGrid` can be turned into a string of `"FullGaussianField"` here ... maybe we should just define `prefix(::Type{FullGaussianGrid}) = "FullGaussian"` or something?

milankl: But all good, we can check whether we would want to remove this code redundancy at some point later ...

PR #732, opened 2025-05-12
Introduce Field for data on grid

### LINKED ISSUE #709: Introduce `Field` to separate data from `Grid`
At the moment our "grids" are types and instances of these contain both data and grid information. This means that for four variables (vorticity, divergence, temperature, pressure, say) we actually also have four grids -- that just happen to be the same. In fact the `rings` of each grid are defined over and over again. This was originally motivated by the analogy to vector/matrix which are types `Vector`/`Matrix` and an instance of these also contain data because there's no grid information really required (1:m, 1:n, ... is the space but that's it). 

Motivated by a discussion with @glwagner we could introduce `Field` as the type holding `data` (some array) and `grid` being our current grids minus the data. I'm drafting this here

```julia
abstract type AbstractGrid end

abstract type AbstractFullGrid <: AbstractGrid end
abstract type AbstractReducedGrid <: AbstractGrid end

struct ExampleGrid <: AbstractReducedGrid
    nlat_half::Int                  # number of latitudes on one hemisphere
    rings::Vector{UnitRange{Int}}
end
```

So we have an `AbstractGrid` tree, and `OctahedralGaussianGrid` would be a concrete type here, an instance is defined by the resolution `nlat_half`, `rings` for indexing and other grid information like `lon`, `lat` could be added (but could also be computed on the fly as currently done). Now we introduce `Field` which is for all grids

```julia
abstract type AbstractField{T, N, ArrayType <: AbstractArray{T, N}, Grid} <: AbstractArray{T, N} end

struct Field{T, N, ArrayType <: AbstractArray{T, N}, Grid} <: AbstractField{T, N, ArrayType, Grid}
    data::ArrayType                 # data array, ring by ring, north to south, then other dimensions
    grid::Grid
end

Base.size(field::Field, args...) = size(field.data, args...)
Base.getindex(field::Field, args...) = getindex(field.data, args...)
```

Then we can do things like

```julia
# create a grid (but no data on it)
grid = ExampleGrid(24, [1:20, 21:40])

# create a field with data based on `grid`
field = Field(rand(2,3), grid)

# create another field that shares the same `grid`
field2 = Field(rand(2,2), grid)
```

I don't think this is too much work to refactor RingGrids.jl based on this logic, but I'm not sure it's worth the effort. @maximilian-gelbrecht what do you think? @simone-silvestri maybe you also have an opinion on this?

milankl:
Fixes #709


glwagner:
It looks great!

glwagner:
what do "full" and "reduced" mean from the standpoint of the computations? I understand that reduced somehow means that the grid has fewer quadrature points than a corresponding full grid

milankl:
> what do "full" and "reduced" mean from the standpoint of the computations? I understand that reduced somehow means that the grid has fewer quadrature points than a corresponding full grid

Actually the quadrature points are the same, as its the same number of latitudes. But reduced grids are those that have towards the poles fewer points per latitude, so are generally more equal area than full grids that have, whichever latitudes you choose, the same number of longitude points per latitude ring. So they are representable as a matrix which puts too many grid points around the poles. 

That destinction is important, also for dispatch, as matrix-representable grids have some shortcuts in some algorithms which one would want to make use of. For consistency we store all those in a vector (even though "full" could be stored using 2 instead of 1 dimension) -- but you could reshape them, or view them differently.

maximilian-gelbrecht:
We have something for that already in LTA, this is a bit hacky solution, but I copied it more or less directly from the Julia documentation: 

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/3c27d4b81738542ace4b745bc691b23d043c2b0f/src/LowerTriangularArrays/lower_triangular_array.jl#L701-L708

https://docs.julialang.org/en/v1/manual/interfaces/#man-interfaces-broadcasting

Or do you need something else than that, here?

milankl:
@maximilian-gelbrecht with the latest commits I've implemented `find_field` similar to `find_L` and that works great, however while you use `similar(L.data, T)` I don't seem to be able to use that as that would force the broadcasted array to have the same size as the first field argument found with `find_field`, so an operation like `.==` between an `(N, )` and `(N,1)`-sized array wouldn't work as the array created by `similar` wouldn't have the correct `size(bc)` size! So I do this now

```julia
function Base.similar(bc::Broadcasted{FieldStyle{N, Grid}}, ::Type{T}) where {N, Grid, T}
    field = find_field(bc)
    ArrayType_ = nonparametric_type(typeof(field.data))
    new_data = ArrayType_{T}(undef, size(bc))
    old_grid = field.grid
    return Field(new_data, old_grid)
end
```

now this is possible
```julia
julia> field1 = zeros(FullGaussianField, 2)
32-element, 4-ring FullGaussianField{Float64, 1}:
 0.0
 ...
 0.0

julia> field2 = zeros(FullGaussianField, 2, 1)
32×1, 4-ring FullGaussianField{Float64, 2}:
 0.0
 ...
 0.0

julia> field1 .== field2
32×1, 4-ring FullGaussianField{Bool, 2}:
 1
 ...
 1
```

previously I got an error like this
```julia
Test threw exception
  Expression: all(F1 .== F4)
  DimensionMismatch: destination axes (Base.OneTo(32),) are not compatible with source axes (Base.OneTo(32), Base.OneTo(1))
```

maximilian-gelbrecht:
Currently all Grids have `architecture`, `nlat_half` and `rings` as their fields, right? 

Similar to how we have the precomputed `l_indices` and `m_indices` in `Spectrum`, I think for the GPU kernels it would be good to have a precomputed `whichring` in the `Grid`. Then we can do the kernels just over `(n_gridpoints, n_layers)`. I could also do that later in #727 or #733 though as well.

milankl:
`whichring` as being the inverse of `rings` essentially? Yeah, happy to include that! Once all other tests pass that should be easy to do as it's just a `Vector{Int}` ?

maximilian-gelbrecht:
Yes exactly, we already `whichring` function as well, but it goes with a while loop through all rings, so it's not very efficient. 

With that, it's trivial to write kernels for typical grids. For example in the vorticity flux, 
```Julia
rings = eachring(u_tend_grid, v_tend_grid, u, v, vor)
    @inbounds for k in eachgrid(u)
        for (j, ring) in enumerate(rings)
            coslat⁻¹j = coslat⁻¹[j]
            f_j = f[j]
            for ij in ring
                ω = vor[ij, k] + f_j                   # absolute vorticity
                u_tend_grid[ij, k] = (u_tend_grid[ij, k] + v[ij, k]*ω)*coslat⁻¹j
                v_tend_grid[ij, k] = (v_tend_grid[ij, k] - u[ij, k]*ω)*coslat⁻¹j
            end
        end
    end
```

Can be done as 

```Julia
@kernel inbounds=true function _vorticity_flux_kernel!(
    u_tend_grid, v_tend_grid, u, v, vor, @Const(f), @Const(coslat⁻¹), @Const(whichring)
)
    # Get indices
    j = whichring[ij]
    
    # Get the coriolis parameter and cosine latitude factor for this latitude
    f_j = f[j]
    coslat⁻¹j = coslat⁻¹[j]
    
    # Calculate absolute vorticity
    ω = vor[ij, k] + f_j
    
    # Update tendencies
    u_tend_grid[ij, k] = (u_tend_grid[ij, k] + v[ij, k] * ω) * coslat⁻¹j
    v_tend_grid[ij, k] = (v_tend_grid[ij, k] - u[ij, k] * ω) * coslat⁻¹j
end
```

Which should run decently on both CPU and GPU

milankl:
On which note I want to propose also to look into FixedSizeArrays / StaticArrays for these indices vectors as their mutability isn't supposed to be used. For the compiler this might be important because when several threads read that vector in it can assume it to be constant and create as many copies of it as it likes to lower memory access.

maximilian-gelbrecht:
Yes, good call.

milankl:
Last commit implements

```julia
struct OctahedralGaussianGrid{A, V, W} <: AbstractReducedGrid{A}
    nlat_half::Int                  # number of latitudes on one hemisphere
    architecture::A                 # information about device, CPU/GPU
    rings::V                        # precomputed ring indices (ring j -> grid point ij)
    whichring::W                    # precomputed ring index for each grid point ij
end
```

etc with

```julia
"""$(TYPEDSIGNATURES) Vector of ring indices for every grid point in `grid`."""
function whichring(Grid::Type{<:AbstractGrid}, nlat_half, rings::AbstractVector)
    w = zeros(Int, get_npoints(Grid, nlat_half))
    @inbounds for (j, ring) in enumerate(rings)
        w[ring] .= j
    end
    return w
end

whichring(grid::AbstractGrid) = whichring(typeof(grid), grid.nlat_half, grid.rings)
```

milankl:
Just introduced also `field_view` the analogue to `lta_view` but for fields.

```julia
julia> field = zeros(FullGaussianField, 2, 2)
32×2, 4-ring FullGaussianField{Float64, 2}:
 0.0  0.0
 0.0  0.0
...
 0.0  0.0
 0.0  0.0

julia> view(field, :, 1)    # returns a SubArray
32-element view(::FullGaussianField{Float64, 2}, :, 1) with eltype Float64:
 0.0
 0.0
...
 0.0
 0.0

julia> field_view(field, :, 1)   # returns a Field of a SubArray
32-element, 4-ring FullGaussianField{Float64, 1}:
 0.0
 0.0
...
 0.0
 0.0
```

milankl:
Inside JLD2 when writing the prognostic variables to file I get this error
```julia
  MethodError: no method matching OctahedralGaussianField{Float32, 1}(::Vector{Float32}, ::OctahedralGaussianGrid{Main.SpeedyWeather.RingGrids.DummyArchitecture, Vector{UnitRange{Int64}}, Vector{Int64}})
  The type `OctahedralGaussianField{Float32, 1}` exists, but no method is defined for this combination of argument types when trying to construct it.
```

maybe it's something about reconstructing instances with `typeof(x)`? Probably just add that method? Like `(::Type{<:Field{T, N}})(args...) where {T, N} = ...` or similar?

docs/src/ringgrids.md
```diff
@@ -28,203 +39,298 @@ to a `Matrix` (i.e. they are *rectangular* grids) but not the `OctahedralGaussia
...
-loop over all in-ring indices `i` (changing longitudes) to do something on the grid.
+
+`eachring(field)` accesses `field.grid.rings` a precomputed vector of `UnitRange` indices, such that
+we can loop over the ring index `j` (`j=1` being closest to the North pole) pull the coriolis parameter
+at that latitude and then loop over all in-ring indices `i` (changing longitudes) to do something on the grid.
 Something similar can be done to scale/unscale with the cosine of latitude for example.
-We can always loop over all grid-points like so
+
+We can always loop over all horizontal grid points with `eachgridpoint` and over every other dimensions
+with `eachlaye`, e.g. for 2D fields you can do
```
maximilian-gelbrecht: `eachlayer`

docs/src/land_sea_mask.md
```diff
@@ -98,10 +98,10 @@ Every (custom) land-sea mask has to be a subtype of `AbstractLandSeaMask`.
 A custom land-sea mask has to be defined as a new type (`struct` or `mutable struct`)
 
 ```julia
-CustomMask{NF<:AbstractFloat, Grid<:AbstractGrid{NF}} <: AbstractLandSeaMask{NF, Grid}
+CustomMask{NF<:AbstractFloat, GridVariable2D} <: AbstractLandSeaMask{NF, Grid}
```
maximilian-gelbrecht: Is this correct with `GridVariable2D` and `<:AbstractLandSeaMask{NF, Grid}`? Just reads odd to me. 

Also below the `GridVariable2D` just by itself might be confusing to users.

docs/src/ringgrids.md
```diff
@@ -28,203 +39,298 @@ to a `Matrix` (i.e. they are *rectangular* grids) but not the `OctahedralGaussia
...
+Data on a grid is called a `Field`, many variables, like temperature are a field. Surface temperature
+would be a 2D field (though represented as a vector), temperature on several vertical layers of
+the atmosphere would be 3D (data represented as a matrix, horizontal x vertical), including
+time would make it 4D. Several fields can share the same grid. Given that the grid is always
+two-dimensional, a 2D and 3D field can also share the same grid, leaving the 3rd dimension
+not further specified for flexibility. 
+
+## Creating a grid
+
+All grids are specified by name and the resolution parameter `nlat_half::Integer`, i.e. and instance
```
maximilian-gelbrecht: sentence reads a bit off to me, 

"and an instance [...]"?

docs/src/ringgrids.md
```diff
@@ -28,203 +39,298 @@ to a `Matrix` (i.e. they are *rectangular* grids) but not the `OctahedralGaussia
...
 
-Every `grid` in RingGrids has a `grid.data` field, which is a vector containing the data on the grid.
-The grid points are unravelled west to east then north to south, meaning that it starts at
-90˚N and 0˚E then walks eastward for 360˚ before jumping on the next latitude ring further south,
-this way circling around the sphere till reaching the south pole. This may also be called _ring order_.
+```@example ringgrids
+grid = FullGaussianGrid(24)
+```
+
+As a second argument `architecture` can be provided which helps to share information on the
```
maximilian-gelbrecht, line 74: Just a note to ourselves: In #727 we should probably also write a documentation page on the whole architecture concept and link it here and in the LTA docs as well

docs/src/ringgrids.md
```diff
@@ -28,203 +39,298 @@ to a `Matrix` (i.e. they are *rectangular* grids) but not the `OctahedralGaussia
...
+field = HEALPixField(undef, 2)              # using undef initializor
+field = HEALPixField{Float16}(undef, 2, 3)  # using Float16 as eltype
 ```
-Which can be reshaped to reobtain `map` from above. Alternatively you can `Matrix(grid)` to do this in one step
+
+## Creating a Field from data
+
+A `field` has `field.data` (some `AbstractArray{T, N}`) and `field.grid` (some `AbstractGrid` as described above).
+The first dimension of `data` describes the horizontal as the grid points on every grid (full and reduced)
+are unravlled west to east then north to south, meaning that it starts at
```
maximilian-gelbrecht: typo

docs/src/ringgrids.md
```diff
@@ -28,203 +39,298 @@ to a `Matrix` (i.e. they are *rectangular* grids) but not the `OctahedralGaussia
...
 ```
-You can also use `zeros`, `ones`, `rand`, `randn` to create a grid, whereby `nlat_half`, i.e. the number of latitude
-rings on one hemisphere, Equator included, is used as a resolution parameter and here as a second argument.
+Both return a field (there is no data in the grid itself) and create an instance of
+`FullGaussianGrid` on the fly. The `input_as=Matrix` is required to denote that the horizontal
+dimension is not unravelled into a vector, triggering a `reshape` internally, `input_as=Vector`
+is the default. A full Gaussian grid has always ``2N`` x ``N`` grid points, but a `FullClenshawGrid`
+has ``2N`` x ``N-1``, if those dimensions don't match, the creation will throw an error.
+
+If you have the data and know which data it comes one you can also create
```
maximilian-gelbrecht: typo

docs/src/ringgrids.md
```diff
@@ -28,203 +39,298 @@ to a `Matrix` (i.e. they are *rectangular* grids) but not the `OctahedralGaussia
...
-grid = randn(OctahedralGaussianGrid, nlat_half)
-heatmap(grid)
+grid = OctahedralGaussianGrid(24)
+field = randn(grid)
+heatmap(field)
 ```
 
-Reduced grids are interpolated to the corresponding full grids so that they can be visualised as
+Reduced fields are automatically interpolated to the corresponding full fields so that they can be visualised as
 as matrix.
```
maximilian-gelbrecht: double as / missing a

ext/SpeedyWeatherMakieExt.jl
```diff
@@ -5,31 +5,30 @@ using DocStringExtensions
...
-    heatmap(full_grid; title, kwargs...)
+    full_field = RingGrids.interpolate(RingGrids.full_grid_type(field.grid), field.grid.nlat_half, field)
+    heatmap(full_field; title, kwargs...)
 end
 
 """
 $(TYPEDSIGNATURES)
-Defines Makie's `heatmap` function for a`grid::AbstractGrid` via interpolation
-to `::AbstractFullGrid` (which can be reshaped into a matrix.)"""
+Defines Makie's `heatmap` function for a`field::AbstractFullField2D` which can be reshaped into a matrix."""
```
maximilian-gelbrecht: blank space missing

src/RingGrids/abstract_types.jl
```diff
@@ -0,0 +1,61 @@
+"""Abstract supertype for all grids in RingGrids.jl, representing a discretization (particularly a tessellation or tiling)
+of the sphere. A "grid" does not contain any data only its resolution is defined by `nlat_half`
+(number of latitude rings on one hemisphere including the equator). A grid does not contain the coordinates of the grid points,
+vertices, latitudes, longtiudes etc but they can be recomputed from the grid object (or its type and resolution) at any time.
+
+A grid is regarded as 2D but a field (data on a grid) can have N additional dimensions, e.g. for vertical levels or time.
+In that sense, a grid is does not have a number format / eltype. This is a property of a `Field` which can be different for
```
maximilian-gelbrecht: typo, "is" too much

src/RingGrids/field.jl
```diff
@@ -0,0 +1,480 @@
...
+# simply propagate all indices forward
+Base.@propagate_inbounds Base.getindex(field::AbstractField, ijk...) = getindex(field.data, ijk...)
+Base.@propagate_inbounds Base.setindex!(field::AbstractField, x, ijk...) = setindex!(field.data, x, ijk...)
+Base.fill!(field::AbstractField, x) = fill!(field.data, x)
+
+# make [:, k...] not escape the Field
+@inline Base.getindex(field::AbstractField, col::Colon, k...) = Field(field.data[col, k...], field.grid)
+@inline eachring(field::AbstractField) = eachring(field.grid)
+
+function eachring(field1::AbstractField, fields::AbstractField...; horizontal_only=true, kwargs...)
```
maximilian-gelbrecht, line 117: no docstring for `eachring` or is it somewhere else? the `horizontal_only` isn't immediately clear to me

src/RingGrids/field.jl
```diff
@@ -0,0 +1,480 @@
...
+                field[ij, k]"""
+eachlayer(field::AbstractField) = CartesianIndices(size(field)[2:end])
+
+# several arguments to check for matching grids
+function eachlayer(field1::AbstractField, fields::AbstractField...; vertical_only=true, kwargs...)
+    fields_match(field1, fields...; vertical_only, kwargs...) || throw(DimensionMismatch(field1, fields...))
+    return eachlayer(field1)
+end
+
+eachgridpoint(field::AbstractField) = eachgridpoint(field.grid)
```
maximilian-gelbrecht: docstring?

src/RingGrids/abstract_types.jl
```diff
@@ -0,0 +1,61 @@
...
+"""Abstract supertype for all fields, i.e. data on a grid. A field is an `AbstractArray` with a number format
+`T`, a number of dimensions `N`, an `ArrayType` (e.g. `Vector`, `Matrix`, `Tensor`) and a grid type `Grid`.
+Fields can be full or reduced, 2D, 3D, or 4D, depending on the grid and the number of dimensions.
+Fields are used to store data on the grid, e.g. temperature, pressure, or any other variable.
+Every `Field` has `data` the grid-free array of the data and `grid` which contains information about the grid
+the field is defined on."""
+abstract type AbstractField{T, N, ArrayType, Grid} <: AbstractArray{T, N} end
+
+"""Abstract supertype for all fields on full grids, i.e. grids with a constant number of longitude points across latitude rings."""
+const AbstractFullField = AbstractField{T, N, ArrayType, Grid} where {T, N, ArrayType, Grid<:AbstractFullGrid}
```
maximilian-gelbrecht, line 39: I like this! this is a good way to make the code much more readable for the remaining RingGrids module!

src/RingGrids/field.jl
```diff
@@ -0,0 +1,480 @@
...
+    similar_data = similar(field.data, Tnew, get_npoints(new_grid), k...)
+    return Field(similar_data, new_grid)
+end
+
+# general version with ArrayType{T, N}(undef, ...) generator
+function (::Type{F})(
+    ::UndefInitializer,
+    nlat_half::Integer,
+    k::Integer...,
+) where {F<:AbstractField{T, N, ArrayType, Grid}} where {T, N, ArrayType, Grid<:AbstractGrid{Architecture}} where Architecture
```
maximilian-gelbrecht, line 312: ha! nice tripple `where`! took me a second to get it, but yeah seems like that's the best way to do this here

src/RingGrids/field.jl
```diff
@@ -0,0 +1,480 @@
...
+function (::Type{F})(
+    ::UndefInitializer,
+    nlat_half::Integer,
+    k::Integer...,
+) where {F<:AbstractField{T, N, ArrayType}} where {T, N, ArrayType}
+    Grid = grid_type(F)
+    grid = nonparametric_type(Grid)(nlat_half, DEFAULT_ARCHITECTURE())
+    data = nonparametric_type(ArrayType){T}(undef, get_npoints(grid), k...)
+    return Field(data, grid)
+end
```
maximilian-gelbrecht, line 377: Did you check that we really need all the undef inits? Hard for me to get a good idea of that just looking at the code

src/RingGrids/grid.jl
```diff
@@ -0,0 +1,267 @@
+architecture_type(grid::AbstractGrid) = typeof(grid.architecture)
+architecture_type(::Type{<:AbstractGrid{Architecture}}) where Architecture = Architecture
```
maximilian-gelbrecht: Where are you using oder needing that? In #727, we only have `architecture(...)`.

src/RingGrids/grids/full_healpix.jl
```diff
@@ -1,41 +1,37 @@
-"""A `FullHEALPixArray` is an array of full grids, subtyping `AbstractFullGridArray`, that use
-HEALPix latitudes for each ring. This type primarily equists to interpolate data from
-the (reduced) HEALPixGrid onto a full grid for output.
-
-First dimension of the underlying `N`-dimensional `data` represents the horizontal dimension,
-in ring order (0 to 360˚E, then north to south), other dimensions are used for the vertical
-and/or time or other dimensions. The resolution parameter of the horizontal grid is
-`nlat_half` (number of latitude rings on one hemisphere, Equator included) and the ring indices
-are precomputed in `rings`. Fields are
+"""A `FullHEALPixGrid` is like a `HEALPixGrid` but with every latitude ring having the same number of longitude
```
maximilian-gelbrecht, line 1: Thinking about it again, it's not obvious to what a FullHealpix grid even is. Healpix is defined to be equal area, how does this work in this case? It's not equal area anymore, right? Maybe just add a comment about that?

src/RingGrids/grids/octahealpix.jl
```diff
@@ -113,87 +109,87 @@ function each_index_in_ring!(   rings::Vector{<:UnitRange{<:Integer}},
...
-    Matrix!(MGs::Tuple{AbstractMatrix{T}, OctaHEALPixGrid}...; kwargs...)
-
-Like `Matrix!(::AbstractMatrix, ::OctaHEALPixGrid)` but for simultaneous
-processing of tuples `((M1, G1), (M2, G2), ...)` with matrices `Mi` and grids `Gi`.
-All matrices and grids have to be of the same size respectively."""
-function Matrix!(   MGs::Tuple{AbstractMatrix{T}, OctaHEALPixGrid}...;
-                    quadrant_rotation::NTuple{4, Integer}=(0, 1, 2, 3),     # = 0˚, 90˚, 180˚, 270˚ anti-clockwise
-                    matrix_quadrant::NTuple{4, Tuple{Integer, Integer}}=((2, 2), (1, 2), (1, 1), (2, 1)),
-                    ) where T
+# ## CONVERSION
```
maximilian-gelbrecht, line 112: What's the reason for the matrix conversion here commented out?

src/RingGrids/interpolation.jl
```diff
@@ -264,69 +282,74 @@ function interpolate!(
...
+    kwargs...
+)
+    I = interpolator(grid, A.grid; kwargs...)
+    Aout = Field(grid, size(A)[2:end]...)
+    interpolate!(Aout, A, I)    # returns an array
+    return Aout                 # returns the field wrapped around that array
 end
 
+# if only the grid type is provided, create a grid with nlat_half from the input field
+interpolate(Grid::Type{<:AbstractGrid}, A::AbstractField; kwargs...) = interpolate(Grid(A.grid.nlat_half), A; kwargs...)
```
maximilian-gelbrecht: Might need to inherit the architecture as well?

src/dynamics/diagnostic_variables.jl
```diff
@@ -101,58 +102,59 @@ $TYPEDFIELDS."""
 @kwdef struct GridVariables{
```
maximilian-gelbrecht, line 102: I like the look of this. Just handing over and initialising with the grid like this!

maximilian-gelbrecht:
These are the benchmark results on my laptop (same machine as what's currently in the readme file). 

Some fluctuations, as expected as it's just a single run, but overall this looks good I think. 

# Benchmarks

created for SpeedyWeather.jl v0.15.0 on Wed, 18 Jun 2025 09:55:59. 

### Machine details

```julia
julia> versioninfo()
Julia Version 1.10.9
Build Info:
Platform Info:
  CPU: 8 × Apple M3
  WORD_SIZE: 64
  LIBM: libopenlibm
```

### Explanation

- Δt: time step [s].

### Running the benchmarks

```> julia manual_benchmarking.jl```

## Models, default setups

| - | - | - | - | - | - | - |
| BarotropicModel | 31 | 1 | false | 2400 | 58352 | 1.67 MB |
| ShallowWaterModel | 31 | 1 | false | 2400 | 36267 | 1.72 MB |
| PrimitiveDryModel | 31 | 8 | true | 2400 | 2191 | 4.88 MB |
| PrimitiveWetModel | 31 | 8 | true | 2400 | 1691 | 5.21 MB |

## Grids

| PrimitiveWetModel | 63 | 8 | FullGaussianGrid | 96 | 1200 | 119 | 27.15 MB |
| PrimitiveWetModel | 63 | 8 | FullClenshawGrid | 95 | 1200 | 139 | 26.92 MB |
| PrimitiveWetModel | 63 | 8 | OctahedralGaussianGrid | 96 | 1200 | 186 | 18.64 MB |
| PrimitiveWetModel | 63 | 8 | OctahedralClenshawGrid | 95 | 1200 | 192 | 18.39 MB |
| PrimitiveWetModel | 63 | 8 | HEALPixGrid | 95 | 1200 | 256 | 13.64 MB |
| PrimitiveWetModel | 63 | 8 | OctaHEALPixGrid | 95 | 1200 | 188 | 16.59 MB |

| - | - | - | - | - | - | - |
| PrimitiveWetModel | 31 | 8 | 48 | 2400 | 1883 | 5.21 MB |
| PrimitiveWetModel | 42 | 8 | 64 | 1800 | 805 | 8.78 MB |
| PrimitiveWetModel | 63 | 8 | 96 | 1200 | 241 | 18.64 MB |
| PrimitiveWetModel | 85 | 8 | 128 | 900 | 61 | 32.43 MB |
| PrimitiveWetModel | 127 | 8 | 192 | 600 | 21 | 71.50 MB |
| PrimitiveWetModel | 170 | 8 | 256 | 450 | 8 | 127.04 MB |

| PrimitiveWetModel | 31 | 8 | true | true | 2400 | 1902 | 5.21 MB |
| PrimitiveWetModel | 31 | 8 | true | false | 2400 | 3300 | 5.21 MB |
| PrimitiveWetModel | 31 | 8 | false | true | 2400 | 2594 | 5.21 MB |

| - | - | - | - |
| pressure_gradient_flux! | 47.750 μs| 182.81 KiB| 2876 |
| linear_virtual_temperature! | 989.583 ns| 0 bytes| 0 |
| temperature_anomaly! | 3.276 μs| 0 bytes| 0 |
| geopotential! | 2.907 μs| 0 bytes| 0 |
| vertical_integration! | 13.333 μs| 0 bytes| 0 |
| surface_pressure_tendency! | 18.083 μs| 65.16 KiB| 1390 |
| vertical_velocity! | 4.548 μs| 0 bytes| 0 |
| linear_pressure_gradient! | 970.588 ns| 0 bytes| 0 |
| vertical_advection! | 89.791 μs| 4.16 KiB| 21 |
| vordiv_tendencies! | 169.167 μs| 15.75 KiB| 210 |
| temperature_tendency! | 243.666 μs| 19.88 KiB| 297 |
| humidity_tendency! | 235.625 μs| 19.88 KiB| 297 |
| bernoulli_potential! | 153.209 μs| 216.00 KiB| 13536 |

| - | - | - | - | - | - | - |
| ShallowWaterModel | 31 | 1 | 48 | 2400 | 35367 | 1.72 MB |
| ShallowWaterModel | 42 | 1 | 64 | 1800 | 16138 | 2.92 MB |
| ShallowWaterModel | 63 | 1 | 96 | 1200 | 4696 | 6.30 MB |
| ShallowWaterModel | 85 | 1 | 128 | 900 | 1741 | 11.14 MB |
| ShallowWaterModel | 127 | 1 | 192 | 600 | 329 | 25.33 MB |
| ShallowWaterModel | 170 | 1 | 256 | 450 | 159 | 46.33 MB |
| ShallowWaterModel | 255 | 1 | 384 | 300 | 38 | 111.17 MB |

| - | - | - | - | - | - | - |
| PrimitiveWetModel | Float32 | 31 | 8 | 2400 | 1908 | 5.21 MB |
| PrimitiveWetModel | Float64 | 31 | 8 | 2400 | 1471 | 9.63 MB |

| PrimitiveDryModel | 31 | 8 | true | true | 2400 | 2200 | 4.88 MB |
| PrimitiveDryModel | 31 | 8 | true | false | 2400 | 4271 | 4.88 MB |
| PrimitiveDryModel | 31 | 8 | false | true | 2400 | 3279 | 4.88 MB |

## Number of vertical layers

| - | - | - | - | - | - |
| PrimitiveWetModel | 31 | 4 | 2400 | 1579 | 3.59 MB |
| PrimitiveWetModel | 31 | 8 | 2400 | 1462 | 5.21 MB |
| PrimitiveWetModel | 31 | 12 | 2400 | 1157 | 6.84 MB |
| PrimitiveWetModel | 31 | 16 | 2400 | 1088 | 8.47 MB |

src/RingGrids/field.jl
milankl: `horizontal_only` is to check for matching fields. `eachring(field1, field2)` throws no error if field1 is 3D and field2 is 2D as long as the horizontal dimension matches.

src/RingGrids/field.jl
milankl: added

src/RingGrids/field.jl
milankl: Very proud of myself to succesfully write a triple `where`. Maybe that means I finally mastered multiple dispatch.

src/RingGrids/field.jl
milankl: I've mostly added them when I hit an error. Some of them are used in broadcasting. Doesn't hurt I think

src/RingGrids/grid.jl
milankl: defined them as I thought they were needed but you are right they aren't. Removing these

src/RingGrids/grids/full_healpix.jl
milankl: Added that comment. Intended use is for output from a HEALPixGrid.

src/RingGrids/grids/octahealpix.jl
milankl: It's an algorithm to change the order of grid points to become directly representable as matrix as therefore also for visualisation. It's not a default order like the `NestedOrder` HEALPix grids define. I have uncommented these at this functionality was experimental, it's no longer used but could be reintroduced. I thought about just leaving this here in commented form but happy for other suggestions.

src/RingGrids/interpolation.jl
milankl: sure

PR #733, opened 2025-05-12
Minimal barotropic model GPU compat

milankl:
@maximilian-gelbrecht I haven't understood yet how to choose `@index(Global, Linear)` vs `@index(Global, Cartesian)` and the `:lmk` argument to `launch!`. But this does work, is however, at T31L8 2x slower than the non-kernel version

```julia
julia> @btime SpeedyWeather.leapfrog!($L, $L2, $dL, 1, 1, $leapfrogg)
  4.083 μs (0 allocations: 0 bytes)

julia> @btime SpeedyWeather.leapfrog!($L, $L2, $dL, 1, 1, $leapfrogg)
  9.625 μs (26 allocations: 944 bytes)
```

maybe because of the `CPU()` allocation?


maximilian-gelbrecht:
When we just have a kernel that treats really every element the same, just use the linear index. But in some kernels, we need the `lm` and `k` indices individually for the computation. 

I am not sure about the CPU performance at the moment. For GPU performance things the memory bandwidth matter and could experiment with optimising that, one could also experiment with doing the real/imag part manually. But for CPU performance this shouldn't really matter, I think?

maximilian-gelbrecht:
@milankl I think as soon as I am finished with the draft of #727, I want to take over this PR/branch and turn it into making a minimal barotropic model work. 

As far as I see, we only need three more kernels for that: 
* vorticity_flux_curldiv!
* horizontal_diffusion!
* spectral_truncation!
* Leapfrog! 

Leapfrog is already here, so I'd just continue here.

milankl:
Not necessarily for this PR here but @michel2323 is keen to help out with KernelAbstractions and AD stuff! Once we start the kernel translation party we can loop Michel in and distribute the efforts?

maximilian-gelbrecht:
Ok, the basic, default barotropic model is running. 

Roadmap ahead here 

- [ ] Clean up 
- [ ] Benchmarking 

In principle, I want to merge this PR rather sooner than later because it contains a lot of basic adjustments to our initialisation etc to make it more agnostic of the array type / using `Architectures`. More kernels for other model setups should follow in other PRs.

milankl:
Oh yeah! So exciting!! Agree no need for completeness or performance here just to experiment with a working example of GPU kernels! If this is merged also others should be able to understand the logic and we can distribute the efforts to translate the various models with and without physics!

src/output/feedback.jl
```diff
@@ -137,10 +137,10 @@ function nar_detection!(feedback::Feedback, progn::PrognosticVariables)
 
     feedback.nars_detected && return nothing    # escape immediately if nans already detected
     i = feedback.progress_meter.counter         # time step
-    vor0 = progn.vor[2][1, end]                 # only check 0-0 mode of surface vorticity
+    vor0 = get_step(progn.vor, 2)[1:1, end]     # only check 0-0 mode of surface vorticity
```
milankl: should be `[1:1, end, 2:2]` to avoid the view business?

src/dynamics/time_integration.jl
```diff
@@ -173,13 +173,27 @@ function leapfrog!(
...
+    #     A_new[lm] = a_new - w2*a_update         # Williams filter: A_new[lm] becomes 1xfiltered value at t+Δt
+    # end
+
+    launch!(architecture(tendency), :lmk, size(tendency), leapfrog_kernel!, A_old, A_new, A_lf, tendency, dt_NF, w1, w2)
+    synchronize(architecture(tendency))
+end
+
+@kernel inbounds=true function leapfrog_kernel!(A_old, A_new, A_lf, tendency, @Const(dt), @Const(w1), @Const(w2))
+
+    lmk = @index(Global, Linear)    # every harmonic lm, every vertical layer k
```
milankl, line 182: `Linear` is the `eachindex` equivalent right?

src/dynamics/horizontal_diffusion.jl
```diff
@@ -157,6 +157,49 @@ function horizontal_diffusion!(
...
+    
+    return tendency
+end
+
+@kernel inbounds=true function _horizontal_diffusion_kernel!(
+    tendency, var, @Const(expl), @Const(impl))
+
+    I = @index(Global, Cartesian)
+    lm = I[1]
+    k = ndims(var) == 1 ? 1 : I[2]
```
milankl, line 160: Wouldn't `CartesianIndex(1)` be better here? Because afaik `@index(Global, Cartesian)` would return that type?

src/dynamics/diagnostic_variables.jl
```diff
@@ -513,10 +513,13 @@ function DiagnosticVariables(
...
 )
-    (; spectrum, nparticles, NF, nlayers) = SG
+    (; spectrum, grid, nparticles, NF, nlayers) = SG
+    (; SpectralVariable2D, SpectralVariable3D) = SG
+    (; GridVariable2D, GridVariable3D) = SG
+    (; ArrayType, VectorType, MatrixType, ParticleVector) = SG
 
     tendencies = Tendencies(SG)
-    grid = GridVariables(SG)
+    grid_variables = GridVariables(SG)
```
milankl, line 522: I had a similar naming conflict elsewhere and I ended up calling `the_grid = SG.grid` ... 🙈

milankl: I like `grid_variables` more

ext/SpeedyWeatherCUDAExt/legendre.jl
```diff
@@ -122,10 +122,12 @@ function SpeedyTransforms._legendre!(
...
     )
     
+    #TODO: suboptimal adjustment to GPU made to avoid scalar indexing here. 
+    #TODO: either integrate it into the kernel or write a new kernel for it
     if unscale_coslat
         @inbounds for j in 1:nlat_half          # symmetry: loop over northern latitudes only
-            g_north[:, nlayers, j] .*= coslat⁻¹[j]        # scale in place
-            g_south[:, nlayers, j] .*= coslat⁻¹[j]
+            g_north[:, nlayers, j] .*= coslat⁻¹[j:j]        # scale in place
+            g_south[:, nlayers, j] .*= coslat⁻¹[j:j]
```
milankl: Is `j:j` a temporary hack or actually the proposed solution 🤨

ext/SpeedyWeatherCUDAExt/function_barriers.jl
```diff
@@ -0,0 +1,11 @@
+# function barrier for GPU
+function SpeedyWeather.vorticity_flux_curldiv!(
```
milankl: This could also live in `src/tendencies.jl` no? Just checking whether you generally think it's good to collect function barriers for GPU stuff separately

docs/src/architectures_gpu.md
```diff
@@ -0,0 +1,30 @@
+# GPU & Architectures
+
+!!! warn Work in progress
+    The GPU support of SpeedyWeather.jl is still work in progress and some parts of this documentation might be not be always updated to the latest state. We will extend this documentation over time. Don't hesitate to contact us via GitHub issues or mail when you have questions or want to colloborate.
+
+Some of SpeedyWeather.jl already supposts GPU acceleration, e.g. the barotropic model. Our development focusses on CUDA GPUs, but other architectures are thinkable in the future as well, as our approach relies on the device agnostic `KernelAbstractions.jl`. The SpeedyWeather.jl submodule `Architectures` encodes all the information of the device we run our models on. In order to initialize a model on a GPU, we need to load the `CUDA` package and pass the architecture to the model constructor. For example, to initialize a barotropic model on a GPU, we can do the following:  
+
+```julia
+using SpeedyWeather, CUDA 
+spectral_grid = SpectralGrid(trunc=41, nlayers=1, architecture=SpeedyWeather.GPU()())
```
milankl: only `()`

ext/SpeedyWeatherCUDAExt/function_barriers.jl
milankl: Ah because `GPU` is only defined when `using CUDA`

src/dynamics/horizontal_diffusion.jl
```diff
@@ -157,6 +157,49 @@ function horizontal_diffusion!(
     end
 end
 
+# temp. function barrier for GPU version 
+function horizontal_diffusion!(
+    tendency::LowerTriangularArray{NF, N, AT, <:Spectrum{<:GPU}},     # tendency of a
+    var::LowerTriangularArray{NF, N, AT, <:Spectrum{<:GPU}},          # spectral horizontal field to diffuse
```
milankl: Due to `GPU` this would need to live in the extension again no? For dispatch could we not have a generic `AbstractGPU` that's defined without `using CUDA`?

src/dynamics/set.jl
```diff
@@ -183,15 +183,22 @@ function set!(
     add::Bool=false,
 )
     (; londs, latds, σ_levels_full) = geometry
-    kernel = add ? (a,b) -> a+b : (a,b) -> b
-    for k in eachlayer(var)
-        for ij in eachgridpoint(var)
-            var[ij, k] = kernel(var[ij, k], f(londs[ij], latds[ij], σ_levels_full[k]))
-        end
-    end
+    kernel_func = add ? (a,b) -> a+b : (a,b) -> b
```
milankl, line 186: You said there were issues with anonymous functions but I think the concept of `KernelOP` could also be defined for two arguments. Not sure what we all need but the one here could also just be independent from the one in spectral_gradients.jl?

```julia
struct KernelOP{add} end

@inline (::KernelOP{true})( a, b) = a+b
@inline (::KernelOP{false})(a, b) = b
```

milankl (commented):
Super nice weekend present 🎁

src/dynamics/set.jl
maximilian-gelbrecht: To be honest I don't fully understand when anonymous functions are bad and when not, I haven't benchmarked this yet, but I'll do next week.

src/dynamics/horizontal_diffusion.jl
maximilian-gelbrecht: `GPU` is actually defined in the main module. But nonetheless is is probably just temporaryly, because I think the performance of the KA version is probably good enough to just keep one version.

ext/SpeedyWeatherCUDAExt/legendre.jl
maximilian-gelbrecht: temporary

src/dynamics/time_integration.jl
maximilian-gelbrecht: yes

src/dynamics/horizontal_diffusion.jl
maximilian-gelbrecht: I am not sure I understand, where exactly would you put `CartesianIndex(1)`? I also don't think this is optimal how I've done it so far, but I couldn't think of a better way.

ext/SpeedyWeatherCUDAExt/legendre.jl
maximilian-gelbrecht: I think I'll just check Jack's KA version next week and then this one here might probably be deleted anyway

src/dynamics/set.jl
maximilian-gelbrecht: Ok, so I just did a benchmark and in this case the anonymous function is fine. It's also fine for the laplacian where it also caused no performance problems. Maybe for `curl` the problem is that it's more complex and also inplace?

maximilian-gelbrecht:
# CPU vs GPU Benchmarks 

AMD EPYC 9554 vs Nvidia H100 

| Truncation | CPU Time (SYPD) | GPU Time (SYPD) | Speedup | 
| ----- | ----- | -----| ------ | 
|        31 |     16274.37 |      1153.31 |   0.07x | 
|       63 |      2154.03 |       296.71 |   0.14x | 
|       127 |       261.92 |        77.82 |   0.30x | 
|       255 |        24.77 |        18.16 |   0.73x | 
|       511 |         2.09 |         4.81 |   2.31x |
|      1023 |         0.17 |         1.02 |   6.04x|

image

maximilian-gelbrecht:
So, the results pretty much mirror the benchmarks for the transforms, which is expected. For the future there's probably still some potential in optimising our transforms on GPU, but so far this is already quite good. Moving to primitive equation models on 8 layers or more should move that break even point to much smaller truncations. 

I think this PR is already close to being finished.

maximilian-gelbrecht:
@milankl I can't request a review because it's your PR, but please do one when you have some time.

src/dynamics/horizontal_diffusion.jl
```diff
@@ -146,15 +146,24 @@ function horizontal_diffusion!(
...
+
+@kernel inbounds=true function _horizontal_diffusion_kernel!(
+    tendency, var, @Const(expl), @Const(impl))
+
+    I = @index(Global, Cartesian)
+    lm = I[1]
+    k = ndims(var) == 1 ? 1 : I[2]
+    
+    # Get the degree l for this coefficient
+    l = var.spectrum.l_indices[lm]
```
michel2323, line 163: It would be good to get rid of this and precompute `expl` and `impl` to be accessed via `expl[I]` and `impl[I]`.

maximilian-gelbrecht: Hi Michel, 
Nice to see you here, great to get some more advice! 
So, in this situation `l` is ranging within basically of the size of the truncation of the model, so ~30 for very coarse resolution, ~1500 for 10km scale. So `expl` and `impl`are vectors of this size like this, so let's say `(l_max)`-length vectors. With `lm`/`I` we loop over all indices, which are much more, roughly `l_max * l_max / 2`. So they are flattened triangular matrixes of the size. 

Given that, would you still rather precompute `expl` and `impl` in the full matrices despite it being much more memory when directly accessing them with `expl[I]`?

src/dynamics/set.jl
```diff
@@ -183,15 +183,22 @@ function set!(
...
+
     return var
 end
 
+@kernel inbounds=true function set_field_3d_kernel!(var, @Const(londs), @Const(latds), @Const(σ_levels_full), @Const(f), @Const(kernel_func))
+    # Get indices
+    ij, k = @index(Global, NTuple)
+    
+    # compute 
+    var[ij, k] = kernel_func(var[ij, k], f(londs[ij], latds[ij], σ_levels_full[k]))
```
michel2323, line 199: Just brainstorming... but I wonder whether avoiding the anonymous function here works better.

```julia
y = f(londs[ij], latds[ij], σ_levels_full[k])
var[ij, k] = add ? var[ij,k] + y : var[ij,k]
```

michel2323: Or just dispatch on `Val(add)`.

src/dynamics/tendencies.jl
```diff
@@ -586,38 +587,49 @@ function vorticity_flux_curldiv!(   diagn::DiagnosticVariables,
...
     div && divergence!(div_tend, u_tend, v_tend, S; add)    # ∂D/∂t = ∇⋅(u_tend, v_tend)
     return nothing       
 end
 
+@kernel inbounds=true function _vorticity_flux_kernel!(
+    u_tend_grid, v_tend_grid, u, v, vor, @Const(f), @Const(coslat⁻¹), @Const(whichring)
+)
+    # Get indices
+    ij, k = @index(Global, NTuple)
+    j = whichring[ij]
```
michel2323, line 619: Precomputing the `j`s might help.

src/dynamics/tendencies.jl
```diff
@@ -586,38 +587,49 @@ function vorticity_flux_curldiv!(   diagn::DiagnosticVariables,
...
+    
+    # Get the coriolis parameter and cosine latitude factor for this latitude
+    f_j = f[j]
+    coslat⁻¹j = coslat⁻¹[j]
```
michel2323, line 623: Or these two as new arrays.

src/dynamics/tendencies.jl
maximilian-gelbrecht: The `whichring[ij]` is already pre-computed, how would you precompute it even further?

It's a bit similar to above, when we would precompute `f` and `coslat⁻¹` on the `ij` indices, the precomputed-arrays would be much larger because `ij` loops over a triangular matrix and `j` just over a vector that's basically just one column of that matrix. Do you think that's worth it?

src/dynamics/set.jl
milankl: Wouldn't inlined code dispatched on `Val(add)` be compiled to `? :`?

src/SpeedyTransforms/legendre.jl
```diff
@@ -176,4 +176,27 @@ function _legendre!(                        # GRID TO SPECTRAL
...
+end
+
+"""
+$(TYPEDSIGNATURES)
+Unscale by cosine of latitude on the fly.
+"""
+function unscale_coslat!(g_north::AbstractArray{<:Complex, 3}, 
+    g_south::AbstractArray{<:Complex, 3}, 
+    coslat⁻¹::AbstractArray{<:Real, 1})
+    launch!(architecture(g_north), :array_3d, size(g_north), unscale_costlat_kernel!,
```
milankl: `g_north` does't have a grid or a spectrum, so `architecture(g_north)` will create a new `CPU()` or `GPU()` is that a problem?

src/SpeedyTransforms/spectral_transform.jl
```diff
@@ -529,7 +529,7 @@ function transform(             # GRID TO SPECTRAL
     S::SpectralTransform,       # precomputed spectral transform
 )
     ks = size(field)[2:end]     # the non-horizontal dimensions
-    coeffs = zeros(Complex{eltype(S)}, S.spectrum, ks...)
+    coeffs = on_architecture(architecture(field), zeros(Complex{eltype(S)}, S.spectrum, ks...))
```
milankl, line 532: this allocates on the CPU and then moves data across instea of directly allocating on the GPU, correct?

src/SpeedyTransforms/spectral_truncation.jl
```diff
@@ -8,27 +8,39 @@ function spectral_truncation!(
...
-
+    (; l_indices, m_indices) = alms.spectrum
+    
+    # Convert to 1-based indexing
     ltrunc += 1     # 0-based to 1-based
     mtrunc += 1
+    
+    # TODO: there's currently a bug that prevents this from working on GPU without the .data
+    # that's mostly related to the custom broadcasting
+    alms.data[(l_indices .> ltrunc) .|| (m_indices .> mtrunc), :] .= 0
```
milankl, line 19: fine with me, we can revisit this at some point, good comment

src/dynamics/horizontal_diffusion.jl
milankl: Maybe I'm wrong but having `one(I[2])` would be a safe bet to guarantee a type stability here despite the `?`?

src/dynamics/horizontal_diffusion.jl
milankl: Because `@Const(expl)` I was hoping that `expl` is considered immutable for the kernel meaning that all threads can freely copy its memory hence it makes sense to keep `expl` smaller if possible? Is that not a good mental model here, Michel?

src/dynamics/set.jl
milankl: Maybe @vchuravy has some insights when anonymous functions inside kernels are fine performance-wise and when not?

milankl (commented):
Corrected some typos, removed the commented out old leapfrog kernel and renamed the `unscale_costlat_kernel!` to `unscale_coslat_kernel!`

src/SpeedyTransforms/legendre.jl
maximilian-gelbrecht: oh, good catch! yes that is a problem, and I'll fix it.

src/SpeedyTransforms/spectral_transform.jl
maximilian-gelbrecht: Yeah, it does. I admit that's not ideal, but it's also just a convenience function anyway. 

In principle, one could define a mapping of `zeros(architecture, ...)` for every architecture and array type for a device specific `zeros` routine. But do we need that really?

maximilian-gelbrecht:
Aside from potential feedback about the kernels, this could be merged in my opinion.

milankl:
@michel2323 @vchuravy do you have any advice on the performance rule of thumbs raised above regarding recomputing vs precomputing, anonymous functions and dispatch over `Val`?

maximilian-gelbrecht:
I merge this PR now as other PRs depend on it as well. I am very happy to address any further comments and tips that are made here in follow-up PRs though!

PR #734, opened 2025-05-12
Introduce Spectrum for LowerTriangularArrays

milankl:
@maximilian-gelbrecht this allows one to do something like

```julia
spectrum = Spectrum(5, 5)
data = rand(nonzeros(spectrum.lmax, spectrum.mmax))
L = LowerTriangularArray(spectrum, data)
```

returning

```julia
15-element LowerTriangularArray{Float64, 1, Spectrum{CPU, Vector{UnitRange{Int64}}, Vector{Tuple{Int64, Int64}}}, Vector{Float64}}:
 0.9711349715841152
 0.5848979776185497
...
 0.3162930223303433
 0.676597687850843
```

So `L.spectrum` would contain the size mxn, but also the `architecture` as well as some precomputed indices in case we want to use them to speed up some kernels.


milankl:
For example one could do
```julia
for (m, order) in enumerate(spectrum.orders)
    for l in order
        println("l: $l, m: $m")
    end
end
```
returning 

```julia
l: 1, m: 1
l: 2, m: 1
l: 3, m: 1
l: 4, m: 1
l: 5, m: 1
l: 2, m: 2
l: 3, m: 2
l: 4, m: 2
l: 5, m: 2
l: 3, m: 3
l: 4, m: 3
l: 5, m: 3
l: 4, m: 4
l: 5, m: 4
l: 5, m: 5
```

or

```julia
for lm in eachindex(L)
    l, m = L.spectrum.degrees_orders[lm]
    println("lm: $lm, l: $l, m: $m")
end
```

yielding

```julia
lm: 1, l: 1, m: 1
lm: 2, l: 2, m: 1
lm: 3, l: 3, m: 1
lm: 4, l: 4, m: 1
lm: 5, l: 5, m: 1
lm: 6, l: 2, m: 2
lm: 7, l: 3, m: 2
lm: 8, l: 4, m: 2
lm: 9, l: 5, m: 2
lm: 10, l: 3, m: 3
lm: 11, l: 4, m: 3
lm: 12, l: 5, m: 3
lm: 13, l: 4, m: 4
lm: 14, l: 5, m: 4
lm: 15, l: 5, m: 5
```

maximilian-gelbrecht (commented):
Yeah, I like it. If you want to we can also split up the work, and I continue this PR and you do the RingGrids one.

glwagner:
What is the abstraction for the `SpectralSpace` in this context (which is the analog of "Grid")? I believe you're proposing `Spectrum` as basically the spectral analogue of a physical-space `Field`, which I think makes sense.

milankl:
> What is the abstraction for the `SpectralSpace` in this context (which is the analog of "Grid")? I believe you're proposing `Spectrum` as basically the spectral analogue of a physical-space `Field`, which I think makes sense.

Yes so like a grid discretizes the space (in our case sphere) a `spectrum` discretizes that same space with wavenumbers. So technically this discretization is just storing the number of waves in either direction (degree and order of the spherical harmonics) so apart from `lmax` (maximum degree l) and `mmax` (maximum order m) there's not much information to store here. But given that we realised that the archcitecture (CPU/GPU) should be stored within a field, particularly in the grid, the analogy here would be to have that same architecture stored in `Spectrum` that's the "grid" for a LowerTriangularArray. I guess someone could say `SpectralField` instead of `LowerTriangularArray` but that's no change of structure and a pure renaming of the "spectral field" that takes the form of a lower triangular array -- so essentially equivalent and I think it's better to communicate the array shape being triangular

glwagner:
Okay, well the key thing is to provide an abstraction that sharply distinguishes between the discretization itself and "data that uses the discretization". For example a "power spectrum" refers to the "spectral density" --- not the specific wavenumber bins that are used. So do we feel that "spectrum" conveys the idea of the discretization?

In FourierFlows, the "grid" combines both spectral and physical information since one implies the other; eg if we decide on a domain extend + number of points, this uniquely determines the wavenumber space associated with the Fourier transform too:

https://github.com/FourierFlows/FourierFlows.jl/blob/1fcca1d9bf4e74b40643264bba1be8d71e78b174/src/domains.jl#L111

It's hazy to me whether this also holds here though.

milankl:
> In FourierFlows, the "grid" combines both spectral and physical information since one implies the other; eg if we decide on a domain extend + number of points, this uniquely determines the wavenumber space associated with the Fourier transform too:

Yeah no unfortunately it doesn't... With the dealiasing actually being an option to the user there's quite some flexibility in combining grid and spectral space, e.g. T31 can be used in combination with a 96x48 FullGaussianGrid, but also with a 64-ring HEALPixGrid which again could also be used with a T42 spectral truncation 🙈. So I feel it's best to have a "grid" for the grid space defining the discretization of the sphere and a "spectrum" to define the spectral space "discretization". Sure that space is technically already discretized so only truncation wavenumber has to be defined but such a grid/spectrum object would also contain architecture CPU/GPU and precomputed indices -- essentially the stuff that several spectral fields (that contain data) would share.

glwagner:
Ok, then I see the motivation to have separate objects for the spatial grid and the wavenumber grid.

src/dynamics/spectral_grid.jl
```diff
@@ -44,14 +47,17 @@ $(TYPEDFIELDS)
     "[OPTION] horizontal resolution as the maximum degree of spherical harmonics"
     trunc::Int = DEFAULT_TRUNC
 
+    "[DERIVED] spectral spectrum"
+    spectrum::SP = Spectrum(trunc+2, trunc+1)
```
maximilian-gelbrecht, line 51: @milankl I now added the `Spectrum` here in the `SpectralGrid` to have this as its "main location" that the other `LowerTriangularArray` then point to (didn't check yet that all of these are really references and not copies the `Spectrum`). Is this what you were thinking about with #732 as well?

milankl: That's what I thought yes. But then `architecture` should also be taken as argument! The indices should be computeed at creation

milankl: I'd do the same for `grid` and then use those as field constructors `field = zeros(grid, ...)`, `lta = zeros(spectrum, ...)` I guess both should then be considered only 2D so that we'd need to supply the vertical dimension somhow, e.g. `field = zeros(grid, nlayers)`

milankl:
>     * [ ]  Really check that it is referenced and not copied

Use the `===` egal operator for this?

```julia
julia> a = [1, 2]
2-element Vector{Int64}:
 1
 2

julia> b = a
2-element Vector{Int64}:
 1
 2

julia> b === a    # returns true because pointing to same memory
true

julia> b === deepcopy(a)   # returns false even though identical because independent objects, not same memory
false
```

maximilian-gelbrecht:
> > Clean up. We now have a lot of different constructors of all kinds. Do we really need all of them?
> 
> I'd say we need constructors as before, like `zeros(LowerTriangularMatrix{ComplexF32}, 5, 5)` as in most cases you don't want to deal with what a `Spectrum` is. But then once you have one, one should be able to do `zeros(spectrum)` to create a lower triangularmatrix based on that `spectrum`, no?

Yep, so far I have this as `zeros(LowerTriangularMatrix{ComplexF32}, spectrum)` because the type is still needed. But we could also change all these constructors to `zeros(ComplexF32, spectrum, ...)` 

The main part of this works now and I adjusted hopefully everything (maybe some model component somewhere with a pre-computed field is left). It also does work that the `Spectrum` is really referenced/egal `===` and not copied.

maximilian-gelbrecht:
Differentiability is back, so this is ready for review now and hopefully merge soon.

maximilian-gelbrecht:
@milankl I can't request a review from you via the GitHub interface because you started the PR, but please do one ;)

docs/src/speedytransforms.md
```diff
@@ -121,17 +121,17 @@ create a `SpectralTransform` is to start with a `SpectralGrid`, which already de
 which spectral resolution is supposed to be combined with a given grid.
 ```@example speedytransforms
 using SpeedyWeather
-spectral_grid = SpectralGrid(NF=Float32, trunc=5, Grid=OctahedralGaussianGrid, dealiasing=3)
+spectral_grid = SpectralGrid(NF=Float32, trunc=5, Grid=OctahedralGaussianGrid, dealiasing=3.)
```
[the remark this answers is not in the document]
milankl: I can look at that.

maximilian-gelbrecht:
No hard feelings. In the `PrognosticVariables` and `DiagnosticVariables` I also put the `Spectrum` before the all the array types because I thought the `Spectrum` and more so the `Grid` is a bit more "basic" and one might wanna do dispatch on it in some rare situation. So I did it in LTA as well to be consistent throughout. But yeah, no strong feelings about the order.

maximilian-gelbrecht:
Ah wait, I did it after the `ArrayType` but before the `SpectralVariables...` etc. So yeah, then forgot what I just write, I think it makes more sense to put it last in LTA, dispatches on the `ArrayType` are obviously also occurring.

maximilian-gelbrecht:
Ok done that now. 

With both #734 and #732, I would leave it to #727 then to actually put the architecture the whole model uses into the `Spectrum` and `Grid` while constructing the model, so I'll do that then in #727 after both PRs are merged.

src/LowerTriangularArrays/LowerTriangularArrays.jl
```diff
@@ -11,10 +11,13 @@ import KernelAbstractions
 # NUMERICS
 import LinearAlgebra: tril!
 
+export AbstractSpectrum, Spectrum, resolution, truncation
+ 
 export LowerTriangularMatrix, LowerTriangularArray
 export eachharmonic, eachmatrix
```
milankl, line 17: In RingGrids, given that the `grid` is now inherently 2D we have
- `eachgridpoint` returning a 2D iterator
- `eachindex` for the 3D iterator
- `eachlayer` returning an ND iterator for vertical, time, ...
- `eachring` for the latitude iterator 

Can we revisit that `eachharmonic`, `eachindex`, `eachmatrix`, `eachorder` (<- does not exist yet? but possible with the new indices in `Spectrum`) are equivalent for consistency?

milankl (commented):
Haven't finished the review but already some comments

docs/src/examples_3D.md
```diff
@@ -12,7 +12,7 @@ See also [Examples 2D](@ref Examples) for examples with the
 
 ```@example jablonowski
 using SpeedyWeather
-spectral_grid = SpectralGrid(trunc=31, nlayers=8, Grid=FullGaussianGrid, dealiasing=3)
+spectral_grid = SpectralGrid(trunc=31, nlayers=8, Grid=FullGaussianGrid, dealiasing=3.)
```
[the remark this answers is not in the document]
maximilian-gelbrecht: Somehow, now it works again when I just tried it  🤷 

Not sure what's the issue there and what cause it, because automatic promotion does work at least in some cases.

ext/SpeedyWeatherCUDAExt/legendre.jl
```diff
@@ -1,10 +1,10 @@
 # convert i, j indices of a matrix (here 0-based l,m though...) to a single 1-based running index
-import SpeedyWeather.LowerTriangularMatrices: ij2k
+import SpeedyWeather.LowerTriangularArrays: lm2k
```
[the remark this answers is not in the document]
maximilian-gelbrecht: Yes, true, it's still not ideal. The `k` should be changed if we also use it in other ways. `n` is also a bit unusual for a running index, no? Maybe `I` or `i`? I use this in the kernels now as well for the CartesianIndices that index the LTAs in a "flat" style. So `lm2i` or `lm2I`?

src/LowerTriangularArrays/LowerTriangularArrays.jl
maximilian-gelbrecht: Ok, I look into it. But just mentioning: with the introduction of the KA kernels, we won't use these iterators a lot. Most of these loops will be replaced by a kernel.

docs/src/examples_3D.md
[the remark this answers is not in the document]
maximilian-gelbrecht: Ah, sorry, I was in the wrong branch. Yes, it still doesn't work. With this PR `SpectralGrid` is now a parametric type, but this really can't be the main issue, because automatic promotion does work for those in other cases as well.

ext/SpeedyWeatherCUDAExt/legendre.jl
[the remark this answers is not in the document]
milankl: happy with that too. `i` otherwise refers to the in-ring index (zonal), `j` to the ring index (meridional) and `ij` is the running index for all 2D grid points, given this is in RingGrids I find the confusing small and happy to just stick to a general `i`

docs/src/examples_3D.md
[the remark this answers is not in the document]
maximilian-gelbrecht: If we can't figure it out, I agree, we should write our own constructor. I'll add one, if the problem still exist before merging.

src/dynamics/spectral_grid.jl
```diff
@@ -117,14 +123,91 @@ function Base.show(io::IO, SG::SpectralGrid)
...
 
+# Constructor that takes all [OPTION] parameters as keyword arguments
+# and calculates all derived fields
+function SpectralGrid(;
+    NF::Type{<:AbstractFloat} = DEFAULT_NF,
+    device::AbstractDevice = DEFAULT_DEVICE,
+    ArrayType::Type{<:AbstractArray} = default_array_type(device),
+    trunc::Int = DEFAULT_TRUNC,
+    Grid::Type{<:AbstractGrid} = DEFAULT_GRID,
+    dealiasing::Number = 2.0,
```
milankl, line 134: `Real`

src/dynamics/spectral_grid.jl
```diff
@@ -117,14 +123,91 @@ function Base.show(io::IO, SG::SpectralGrid)
...
+    dealiasing_f64 = Float64(dealiasing)
+    radius_f64 = Float64(radius)
+    
+    # Calculate derived fields
+    VectorType = ArrayType{NF, 1}
+    MatrixType = ArrayType{NF, 2}
+    TensorType = ArrayType{NF, 3}
+    
+    # Spectral space
+    spectrum = Spectrum(trunc+2, trunc+1)
```
milankl, line 150: take on `architecture`?

src/SpeedyTransforms/spectral_transform.jl
```diff
@@ -196,65 +196,66 @@ function SpectralTransform(
...
     # RECURSION FACTORS
-    ϵlms = get_recursion_factors(lmax+1, mmax)
+    ϵlms = get_recursion_factors(spectrum)
 
     # GRADIENTS (on unit sphere, hence 1/radius-scaling is omitted)
     # meridional gradient for scalars (coslat scaling included)
-    grad_y1 = zeros(LowerTriangularMatrix, lmax+1, mmax+1)      # term 1, mul with harmonic l-1, m
-    grad_y2 = zeros(LowerTriangularMatrix, lmax+1, mmax+1)      # term 2, mul with harmonic l+1, m
+    grad_y1 = zeros(LowerTriangularMatrix, spectrum)      # term 1, mul with harmonic l-1, m
+    grad_y2 = zeros(LowerTriangularMatrix, spectrum)      # term 2, mul with harmonic l+1, m
```
milankl, line 204: For RingGrids the equivalent is now just `zeros(grid)`, given that `grid` implies the actual type. Similar I could think one could just write `zeros(spectrum)` here, or `zeros(NF, spectrum)`?

src/SpeedyTransforms/spectral_transform.jl
```diff
@@ -272,48 +273,58 @@ function SpectralTransform(
...
+    vargs...; kwargs...                             # grid resolution, latitude rings on one hemisphere incl equator
+) where NF = SpectralTransform(NF, Spectrum(lmax, mmax), vargs...; kwargs...)
+    
 """$(TYPEDSIGNATURES)
 Generator function for a `SpectralTransform` struct based on the size of the spectral
 coefficients `specs`. Use keyword arguments `nlat_half`, `Grid` or `deliasing` (if `nlat_half`
 not provided) to define the grid."""
 function SpectralTransform(
-    specs::LowerTriangularArray{NF, N, ArrayType};  # spectral coefficients
+    specs::LowerTriangularArray{NF, N, ArrayType, S};  # spectral coefficients
```
milankl, line 291: btw I ended up dropping many of these so just `field::AbstractField` and then use `eltype(field)`, `ndims(field)`, etc. instead

src/LowerTriangularArrays/lower_triangular_array.jl
```diff
@@ -130,95 +141,138 @@ for f in (:zeros, :ones, :rand, :randn)
...
+            ::Type{LowerTriangularArray{T}},
+            spectrum::AbstractSpectrum,
+            I::Vararg{Integer, M},
+        ) where {T, M}
+            return LowerTriangularArray($f(T, nonzeros(spectrum), I...), spectrum)
+        end
+
+        function Base.$f(
+            ::Type{LowerTriangularMatrix{T}},
+            spectrum::AbstractSpectrum,
```
milankl, line 190: `spectrum` implies a `LowerTriangularArray` so could be `zeros(spectrum)` too

src/LowerTriangularArrays/lower_triangular_array.jl
```diff
@@ -526,31 +579,31 @@ function LowerTriangularMatrix{T}(M::LowerTriangularMatrix{T2}) where {T,T2}
...
 end
 
 function Base.convert(::Type{LowerTriangularMatrix{T}}, L::LowerTriangularMatrix) where T
-    return LowerTriangularMatrix{T}(L.data, L.m, L.n)
+    return LowerTriangularMatrix{T,typeof(L.spectrum)}(L.data, L.spectrum)
 end
 
-function Base.similar(::LowerTriangularArray{T, N, ArrayType}, I::Integer...) where {T, N, ArrayType}
-    return LowerTriangularArray{T,N,ArrayType}(undef, I...)
+function Base.similar(::LowerTriangularArray{T, N, ArrayType, SP}, I::Integer...) where {T, N, ArrayType, SP}
```
milankl, line 592: I believe `similar` methods should share the `spectrum` of the similar lowertriangular array?

src/LowerTriangularArrays/lower_triangular_array.jl
```diff
@@ -617,27 +670,28 @@ find_L(a::LowerTriangularArray, rest) = a
...
 
 function Base.similar(
-    bc::Broadcasted{LowerTriangularStyle{N, ArrayType}},
+    bc::Broadcasted{LowerTriangularStyle{N, ArrayType, S}},
     ::Type{T},
-) where {N, ArrayType, T}
+) where {N, ArrayType, S, T}
     L = find_L(bc)
-    return LowerTriangularArray{T, N, ArrayType{T,N}}(undef, size(L; as=Matrix))
+    return LowerTriangularArray{T, N, ArrayType{T,N}, S}(undef, size(L; as=Matrix))
```
milankl, line 677: same here, I think the broadcasted spectrum should be reused to construct a new lowertriangular array

src/dynamics/spectral_grid.jl
maximilian-gelbrecht: Will be done in #727 , we don't have the architectures here yet.

src/SpeedyTransforms/spectral_transform.jl
maximilian-gelbrecht: Yes, those constructors already exist. I just didn't use them everywhere in the code yet. 
* `zeros(LowerTriangularMatrix, spectrum)`
* `zeros(spectrum)`
*  `zeros(LowerTriangularMatrix, 5,5)`

Is all possible

src/LowerTriangularArrays/lower_triangular_array.jl
maximilian-gelbrecht: That's defined just a few lines below :)

src/LowerTriangularArrays/lower_triangular_array.jl
```diff
@@ -130,95 +141,138 @@ for f in (:zeros, :ones, :rand, :randn)
...
+
+        Base.$f(::Type{LowerTriangularArray}, spectrum::AbstractSpectrum, I::Vararg{Integer, M}) where {M} =
+            $f(LowerTriangularArray{Float64}, spectrum, I...)
+        Base.$f(::Type{LowerTriangularMatrix}, spectrum::AbstractSpectrum) =
+            $f(LowerTriangularArray{Float64}, spectrum)
     end
 end
 
 Base.zero(L::LTA) where {LTA <: LowerTriangularArray} = zeros(LTA, size(L; as=Matrix)...)
 Base.one(L::LTA) where {LTA <: LowerTriangularArray} = ones(LTA, size(L; as=Matrix)...)
```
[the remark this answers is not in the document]
maximilian-gelbrecht: Good call!

src/LowerTriangularArrays/lower_triangular_array.jl
maximilian-gelbrecht: You are absolutely right in principle, I also adjusted this for the third similar constructor now. But it's a bit unclear to me how handle this with the `I` here. So far you were able to allocate a similar LTA with different dimensions with this routine. In these cases you couldn't share/reuse the `spectrum`. We could also remove that possibility and just interpret `I` as the trailing batch dimensions. What do you think?

src/LowerTriangularArrays/lower_triangular_array.jl
milankl: For RingGrids I solved this like
```julia
function Base.similar(bc::Broadcasted{FieldGPUStyle{N, ArrayType, Grid}}, ::Type{T}) where {N, ArrayType, Grid, T}
    for maybe_field in bc.args
        if maybe_field isa AbstractField
            return similar(maybe_field, T)
        end
    end
end
```

So this looks for the first field in broadcasted and reuses that `grid`. This is necessary because you can write `1 .+ field` or `field .+ 1` or `field1 .+ field2`. So which `grid` to use in either case? I thought it would be an easily rule to use the first `grid` respectively

src/LowerTriangularArrays/lower_triangular_array.jl
milankl: e.g. I wrote this
```julia
Base.similar(field::AbstractField) = Field(similar(field.data), field.grid)
```

src/LowerTriangularArrays/lower_triangular_array.jl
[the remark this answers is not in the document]
milankl: I wrote
```julia
Base.zero(field::AbstractField) = Field(zero(field.data), field.grid)
```

src/LowerTriangularArrays/lower_triangular_array.jl
milankl: For in-place operations like `L .= L1 .+ L2 .+ 1` I don't think this applies as each `L, L1, L2` retain their respective `spectrum`

src/dynamics/drag.jl
```diff
@@ -98,11 +98,11 @@ function drag!(
 end
 
 export JetDrag
-@kwdef struct JetDrag{NF, SpectralVariable2D} <: SpeedyWeather.AbstractDrag
+@kwdef struct JetDrag{NF, SP, SpectralVariable2D} <: SpeedyWeather.AbstractDrag
 
     # DIMENSIONS from SpectralGrid
     "Spectral resolution as max degree of spherical harmonics"
-    trunc::Int
+    spectrum::SP
```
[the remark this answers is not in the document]
maximilian-gelbrecht: I am not sure I understand what you mean by externalise the constructor.

src/LowerTriangularArrays/lower_triangular_array.jl
maximilian-gelbrecht: Ok, I've done that now. I also added some tests that test that's really the same spectrum.

src/LowerTriangularArrays/LowerTriangularArrays.jl
maximilian-gelbrecht: Okay, added that now

maximilian-gelbrecht:
Thanks for the great review! 

I went over all the comments now, added some new tests, constructors and docstrings for `Spectrum`. Everything related to the `architecture` I intend to finish off in #727. So from my side this is ready to merge when all tests pass.

milankl:
Awesome! I'm still battling all the breaking changes with fields ... 🙈

maximilian-gelbrecht:
Benchmarks looks fine as well, I committed them as well because the old ones are now already quite outdated. I'll press the green button ones these tests are passed now.

milankl:
@jackleland with this pull request we have now several precomputed indices in `spectrum`, e.g.

```julia
ulia> spectrum.orders
32-element Vector{UnitRange{Int64}}:
 1:33
 2:33
 3:33
...
 30:33
 31:33
 32:33

julia> spectrum.m_indices
560-element Vector{Int64}:
  1
  1
  1
...
 31
 32
 32

julia> spectrum.lm_orders
32-element Vector{UnitRange{Int64}}:
 1:33
 34:65
 66:96
 ...
 552:555
 556:558
 559:560
```

Are any of these useful to replace or reuse for any of the following

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/3c27d4b81738542ace4b745bc691b23d043c2b0f/src/SpeedyTransforms/spectral_transform.jl#L174-L189

PR #738, opened 2025-06-04
Array-agnostic LowerTriangularMatrix

milankl:
We had
```julia
const LowerTriangularMatrix{T, S} = LowerTriangularArray{T, 1, Vector{T}, S} where S <: AbstractSpectrum
```
maybe we can do this instead to not hardcode the `Vector`
```julia
const LowerTriangularMatrix = LowerTriangularArray{T, 1} where T
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Array-agnostic LowerTriangularMatrix [#738](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/738)
 - Major rework of LowerTriangularArrays: Introduce Spectrum type, changes in indexing and constructors [#734](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/734)
 - Reduce CI time [#731](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/731)
 - Remove DocsPreviewCleanup workflow [#730](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/730)
--- a/src/LowerTriangularArrays/lower_triangular_array.jl
+++ b/src/LowerTriangularArrays/lower_triangular_array.jl
@@ -28,19 +28,22 @@ end
 """2D `LowerTriangularArray` of type `T`"""
 const LowerTriangularMatrix = LowerTriangularArray{T, 1} where T
 
-LowerTriangularArray(data::ArrayType, spectrum::S) where {T, N, ArrayType <: AbstractArray{T,N}, S <: AbstractSpectrum} = LowerTriangularArray{T, N, ArrayType, S}(data, spectrum)
+# construct LTA from data and spectrum
+LowerTriangularArray( data::ArrayType, spectrum::S) where {T, N, ArrayType <: AbstractArray{T, N}, S <: AbstractSpectrum} =
+    LowerTriangularArray{T, N, ArrayType, S}(data, spectrum)
 
-LowerTriangularMatrix(data::AbstractVector{T}, spectrum::S) where {T, S <: AbstractSpectrum} =
-    LowerTriangularMatrix{T, typeof(spectrum)}(data, spectrum)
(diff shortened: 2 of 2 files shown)


milankl:
Now this works, one can view a `LowerTriangularArray` (LTA) and you get an LTA of a view not a view of an LTA.

```julia
julia> specs = zeros(LowerTriangularArray, 5, 5, 2)
15×2 LowerTriangularArray{Float64, 2, Matrix{Float64}, Spectrum{Main.LowerTriangularArrays.CPU, Vector{UnitRange{Int64}}, Vector{Int64}, Vector{Int64}}}:
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0

julia> view(specs, :, 1)
15-element view(::LowerTriangularArray{Float64, 2, Matrix{Float64}, Spectrum{Main.LowerTriangularArrays.CPU, Vector{UnitRange{Int64}}, Vector{Int64}, Vector{Int64}}}, :, 1) with eltype Float64:
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0

julia> spec_view(L::LowerTriangularArray, args...) = LowerTriangularArray(view(L.data, args...), L.spectrum)
spec_view (generic function with 1 method)

julia> spec_view(specs, :, 1)
15-element, 5x5 LowerTriangularMatrix{Float64}
 0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0
```

calling this `spec_view` for now to not extend `Base.view`

milankl:
@maximilian-gelbrecht this is ready to be merged from my side. While the type LowerTriangularMatrix doesn't have the ArrayType hardcoded anymore, i.e.

```julia
julia> LowerTriangularMatrix
LowerTriangularMatrix{T} where T (alias for LowerTriangularArray{T, 1, ArrayType} where {T, ArrayType<:AbstractArray{T, 1}})
```

and types based on other arrays can be defined

```julia
julia> LowerTriangularMatrix{Float32, SubArray{Float32, 1}}
LowerTriangularMatrix{Float32, SubArray{Float32, 1}} (alias for LowerTriangularArray{Float32, 1, SubArray{Float32, 1}})
```

many constructors don't allow it, e.g. we still have

```julia
function LowerTriangularMatrix{T}(::UndefInitializer, spectrum::AbstractSpectrum) where T
    return LowerTriangularMatrix(Vector{T}(undef, nonzeros(spectrum)), spectrum)
end
```

hardcoding the `Vector{T}` in there (which was reasonable before). But don't see the need to generalise this now, we can add it if needed.

maximilian-gelbrecht:
I'm honestly surprised that this didn't break something somewhere else 😅 

With regards to `view`, I am just wondering if this causes any trouble anywhere when `view` doesn't return a `SubArray` as it does usually.

maximilian-gelbrecht (approved):
So, yeah, if all tests pass, I'm also happy with this. 

I like this more than the old version.

milankl:
> I'm honestly surprised that this didn't break something somewhere else 😅

In contrast to the 3000 lines PRs quite refreshing that something can work with 17 lines of changes 😄 

> With regards to `view`, I am just wondering if this causes any trouble anywhere when `view` doesn't return a `SubArray` as it does usually.

We can just call it `spec_view` or whatever for now so that `view` itself is unchanged. Or `view(L, as=LowerTriangularArray, :, 1)` for example so that `view(L)` is unchanged

milankl:
Follow-up PR #739

PR #739, opened 2025-06-04
Add view that returns a LowerTriangularArray

milankl:
Introduces `lta_view` (happy for naming suggestions)

```julia
julia> L = zeros(LowerTriangularArray, 3, 3, 2)
6×2 LowerTriangularArray{Float64, 2, Matrix{Float64}, Spectrum{Main.LowerTriangularArrays.CPU, Vector{UnitRange{Int64}}, Vector{Int64}, Vector{Int64}}}:
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0

julia> view(L, :, 1)
6-element view(::LowerTriangularArray{Float64, 2, Matrix{Float64}, Spectrum{Main.LowerTriangularArrays.CPU, Vector{UnitRange{Int64}}, Vector{Int64}, Vector{Int64}}}, :, 1) with eltype Float64:
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0

julia> lta_view(L, :, 1)
6-element, 3x3 LowerTriangularMatrix{Float64}
 0.0  0.0  0.0
 0.0  0.0  0.0
 0.0  0.0  0.0
```

so `view` works as before returning a `SubArray` of a `LowerTriangularArray` but `lta_view` returns a `LowerTriangularArray` of a view on the data of a `LowerTriangularArray`, in other words `lta_view` preserves the `LowerTriangularArray`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- LowerTriangularArray-preserving view [#739](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/739)
 - Array-agnostic LowerTriangularMatrix [#738](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/738)
 - Major rework of LowerTriangularArrays: Introduce Spectrum type, changes in indexing and constructors [#734](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/734)
 - Reduce CI time [#731](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/731)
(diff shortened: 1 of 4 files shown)


milankl:
This is introduced so that these guys

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/a57494d4fefe0e00f358f5fad57b84623222f76a/src/dynamics/prognostic_variables.jl#L94-L95

Can be written as `vor::SpectralVariable4D = zeros(SpectralVariable4D, spectrum, nlayers, nsteps)` with the view used here instead of unpacking the tuple

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/a57494d4fefe0e00f358f5fad57b84623222f76a/src/dynamics/tendencies.jl#L844

which would then be `vor = view(progn.vor, :, :, lf)`

milankl:
@maximilian-gelbrecht for now I'd keep the name as `lta_view` not extend `Base.view` and when everything works (in another pull request that actually uses `lta_view` in the dynamical core) we can still think about simply extending `Base.view` checking that we don't introduce new problems when doing that.

maximilian-gelbrecht (approved):
Yeah, looks good. 

I also got no suggestions for a better name than `lta_view`.

milankl:
#732 introduces `field_view` the analogue of `lta_view` for fields

PR #745, opened 2025-06-13
Use view instead of tuple for prognostic variables

milankl:
Also introduces `get_step` to neatly write

```julia
vor = get_step(progn.vor, lf)   # relative vorticity at leapfrog step lf
```

instead of potentially being confused how many `:` to add

```julia
vor = lta_view(progn.vor, :, :, lf)
```


milankl:
#732 should be merged first

milankl:
@maximilian-gelbrecht representing the prognostic variables as `SpectralVariable4D` requires `lta_view(vor, :, :, 1)` for example to get a LowerTriangularArray again. So instead of

```julia
vor = progn.vor[1]
```

to unpack the tuple we now do a lot of 

```julia
vor = get_step(vor, 1)
```

which I thought was an easier interface to get a specific leapfrog step (or any last index) for that matter. Defined as

```julia
export get_step
get_step(specs::LowerTriangularArray{T, 2}, i) where T = lta_view(specs, :, i)
get_step(specs::LowerTriangularArray{T, 3}, i) where T = lta_view(specs, :, :, i)
```

You think that's a good idea?

milankl:
@maximilian-gelbrecht with the `lta_view`s I had to define

```julia
nonparametric_type(::Type{<:SubArray{T, N, A}}) where {T, N, A} = nonparametric_type(A)
```

So the `nonparametric_type` of a `SubArray` (created by a view) is the nonparametric type of the underlying array and does not return simply `SubArray`. This is because we use `nonparameteric_type` exclusively in array creation to do things like `ArrayType_{NF, 1}` which works for actual arrays but not for SubArrays (as the array type it's viewing is undefined). This might be confusing, but is probably not that far from what's done in Base anyway, e.g.
```julia
julia> similar(view(zeros(2), :))
2-element Vector{Float64}:
 0.0
 0.0
```
So a `similar` of a view on an `Array` is an `Array` not a `SubArray`. This problem was triggered because broadcasting for `lta_view` didn't work, e.g.

```julia
julia> coeffs = zeros(LowerTriangularArray, 2, 2, 2)
3×2 LowerTriangularArray{Float64, 2, Matrix{Float64}, Spectrum{Main.SpeedyWeather.LowerTriangularArrays.CPU, Vector{UnitRange{Int64}}, Vector{Int64}, Vector{Int64}}}:
 0.0  0.0
 0.0  0.0
 0.0  0.0

julia> coeff1 = SpeedyWeather.get_step(coeffs, 1)    # a view
3-element, 2x2 LowerTriangularMatrix{Float64}
 0.0  0.0
 0.0  0.0

julia> 5*coeff1
ERROR: MethodError: no method matching nonparametric_type(::Type{SubArray{Float64, 1, Matrix{Float64}, Tuple{Base.Slice{Base.OneTo{Int64}}, Int64}, true}})
The function `nonparametric_type` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  nonparametric_type(::Type{<:Array})
   @ Main.SpeedyWeather ~/git/SpeedyWeather.jl/src/LowerTriangularArrays/lower_triangular_array.jl:718

Stacktrace:
 [1] Base.Broadcast.BroadcastStyle(::Type{LowerTriangularMatrix{Float64, SubArray{…}, Spectrum{…}}})
   @ Main.SpeedyWeather.LowerTriangularArrays ~/git/SpeedyWeather.jl/src/LowerTriangularArrays/lower_triangular_array.jl:699
 [2] combine_styles(c::LowerTriangularMatrix{Float64, SubArray{…}, Spectrum{…}})
   @ Base.Broadcast ./broadcast.jl:426
 [3] combine_styles(c1::Int64, c2::LowerTriangularMatrix{Float64, SubArray{…}, Spectrum{…}})
   @ Base.Broadcast ./broadcast.jl:431
 [4] broadcasted
   @ ./broadcast.jl:1331 [inlined]
 [5] broadcast_preserving_zero_d
   @ ./broadcast.jl:860 [inlined]
 [6] *(A::Int64, B::LowerTriangularMatrix{Float64, SubArray{…}, Spectrum{…}})
   @ Base ./arraymath.jl:21
 [7] top-level scope
   @ REPL[40]:1
```

But defining the above works then as expected, creating a new array using the array type from within the subarray on the fly
```julia
julia> 5*coeff1
3-element, 2x2 LowerTriangularMatrix{Float64}
 0.0  0.0
 0.0  0.0
```

maximilian-gelbrecht:
As you write it, this definitely makes sense. 

Also shows though that we probably need some more unit tests for `lta_view` and `field_view` (or what's it called now?).

milankl:
> Also shows though that we probably need some more unit tests for `lta_view` and `field_view` (or what's it called now?).

Yes but also tested not as a unit but during model integrations, anyway! But will add some more!

milankl:
@maximilian-gelbrecht this is also ready to go. It's based off #732, so when merging that the changed lines of codes here should massively drop as most commits in here are also in #732 (Don't merge this one first). Sorry for that.

maximilian-gelbrecht:
Could you resolve the merge conflicts, then I can do a review with the much smaller difference.

milankl:
Down to ~200 lines of changes!

docs/src/initial_conditions.md
```diff
@@ -73,7 +73,7 @@ One may filter out low values of spectral vorticity with some cut-off amplitude
 c = 1e-10       # cut-off amplitude
 
 # 1 = first leapfrog timestep of spectral vorticity
-vor = simulation.prognostic_variables.vor[1]
+vor = get_step(simulation.prognostic_variables.vor, 1)      # get the first leapfrog step
```
maximilian-gelbrecht, line 76: we are using both here in the docs: get_step and the full 3d/2d indexing. Maybe we could add a little explanation somewhere? might be a bit confusing otherwise

src/dynamics/implicit.jl
```diff
@@ -71,10 +71,8 @@ function implicit_correction!(  diagn::DiagnosticVariables,
                                 implicit::ImplicitShallowWater)
 
     (; div_tend, pres_tend) = diagn.tendencies # tendency of divergence and pressure/η
-    div_old = progn.div[1]      # divergence at t
-    div_new = progn.div[2]      # divergence at t+dt
-    pres_old = progn.pres[1]    # pressure/η at t
-    pres_new = progn.pres[2]    # pressure/η at t+dt
+    div_old, div_new   = get_steps(progn.div)   # divergence at t, t+dt
+    pres_old, pres_new = get_steps(progn.pres)  # pressure/η at t, t+dt
```
maximilian-gelbrecht, line 75: yeah, that looks good. I like this more than our old approach

src/dynamics/prognostic_variables.jl
```diff
@@ -137,20 +135,35 @@ export PrognosticVariables
...
+function get_steps(coeffs::LowerTriangularArray{T, 3}) where T
+    nsteps = size(coeffs, 3)
+    return ntuple(i -> lta_view(coeffs, :, :, i), nsteps)
+end
+
+export get_step
+get_step(coeffs::LowerTriangularArray{T, 2}, i) where T = lta_view(coeffs, :, i)
+get_step(coeffs::LowerTriangularArray{T, 3}, i) where T = lta_view(coeffs, :, :, i)
+
+get_nsteps(progn::PrognosticVariables{T, A, S, G, nsteps}) where {T, A, S, G, nsteps} = nsteps
```
maximilian-gelbrecht: does this work? nsteps isn't a parameter anymore? one could just do `=progn.nsteps` instead

docs/src/initial_conditions.md
milankl: `get_step` is really just to avoid thining about whether this is a 2D or 3D variable, otherwise you have to know whether it's `:, 1` or `:, :, 1` ... and it also means that there's only one place where we call `lta_view` and can maybe just rename this to `view` eventually (if we want to)

src/dynamics/prognostic_variables.jl
milankl: Good catch, this was the old method, deleting this

docs/src/initial_conditions.md
milankl: Added some docstrings to explain this with [b13cce1](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/745/commits/b13cce112085842988871a2ba4d583884dc7cf19)

PR #746, opened 2025-06-13
More shameless piracy on the high seas of time

bgroenks96:
If you're gonna be bad, you might as well go all the way :pirate_flag: 

Adds `Century` and `Millenium` types as well as additional conversion and promotion rules.

I think we could probably add more tests but happy to hear what you think @milankl

### DIFF
--- a/src/utility_functions.jl
+++ b/src/utility_functions.jl
@@ -96,11 +96,66 @@ function print_fields(io::IO, A, keys;arrays::Bool=false)
     end
 end
 
+"""
+    Century <: Period
+
+Convenience time type representing a 100-year period.
+"""
+struct Century <: Period
+    value::Int
+    Century(value::Int) = new(value)
+end
+
+Dates._units(m::Century) = m.value == 1 ? " century" : " centuries"
+
+# convert Century -> Year
+Base.convert(::Type{Year}, c::Century) = Year(c.value*100)
+
+# promotion rules
+Base.promote_rule(::Type{Century}, ::Type{Year}) = Year
+Base.promote_rule(::Type{Century}, ::Type{Month}) = Month
+Base.promote_rule(::Type{Century}, ::Type{Day}) = Day
+Base.promote_rule(::Type{Century}, ::Type{Hour}) = Hour
+Base.promote_rule(::Type{Century}, ::Type{Second}) = Second
+
+"""
+    Millenium <: Period
+
+Convenience time type representing a 1000-year period.
+"""
+struct Millenium <: Period
+    value::Int
+    Millenium(value::Int) = new(value)
+end
+
+Dates._units(m::Millenium) = m.value == 1 ? " millenium" : " millenia"
+
+# convert Millenium -> Century and Year
+Base.convert(::Type{Century}, m::Millenium) = Century(m.value*10)
+Base.convert(::Type{Year}, m::Millenium) = Year(Century(m))
+
+# promotion rules
+Base.promote_rule(::Type{Millenium}, ::Type{Century}) = Century
+Base.promote_rule(::Type{Millenium}, ::Type{Year}) = Year
+Base.promote_rule(::Type{Millenium}, ::Type{Month}) = Month
+Base.promote_rule(::Type{Millenium}, ::Type{Day}) = Day
+Base.promote_rule(::Type{Millenium}, ::Type{Hour}) = Hour
+Base.promote_rule(::Type{Millenium}, ::Type{Second}) = Second
+
+
+# add coarserperiod dispatches for Century and Millenium
+Dates.coarserperiod(::Type{Year}) = (Century, 100)
+Dates.coarserperiod(::Type{Century}) = (Millenium, 10)
+
 Dates.Second(x::AbstractFloat) = convert(Second, x)
 Dates.Minute(x::AbstractFloat) = Second(60x)
 Dates.Hour(  x::AbstractFloat) = Minute(60x)
 Dates.Day(   x::AbstractFloat) = Hour(24x)
 Dates.Week(  x::AbstractFloat) = Day(7x)
+Dates.Month( x::AbstractFloat) = Day(30x)  # approximate
+Dates.Year(  x::AbstractFloat) = Day(365x) # approximate
+Century(     x::AbstractFloat) = Year(100x)
+Millenium(   x::AbstractFloat) = Century(10x)
 
 # use Dates.second to round to integer seconds
 Dates.second(x::Dates.Nanosecond) = round(Int, x.value*1e-9)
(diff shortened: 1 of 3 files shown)


src/utility_functions.jl
```diff
@@ -96,11 +96,66 @@ function print_fields(io::IO, A, keys;arrays::Bool=false)
...
 end
 
+"""
+    Century <: Period
+
+Convenience time type representing a 100-year period.
+"""
+struct Century <: Period
+    value::Int
+    Century(value::Int) = new(value)
```
milankl: Does this inner constructor do anything? (It has the same signature as the default constructor)

src/utility_functions.jl
```diff
@@ -96,11 +96,66 @@ function print_fields(io::IO, A, keys;arrays::Bool=false)
...
+
+# convert Century -> Year
+Base.convert(::Type{Year}, c::Century) = Year(c.value*100)
+
+# promotion rules
+Base.promote_rule(::Type{Century}, ::Type{Year}) = Year
+Base.promote_rule(::Type{Century}, ::Type{Month}) = Month
+Base.promote_rule(::Type{Century}, ::Type{Day}) = Day
+Base.promote_rule(::Type{Century}, ::Type{Hour}) = Hour
+Base.promote_rule(::Type{Century}, ::Type{Second}) = Second
```
milankl, line 121: When do these kick in? century plus year would be a compound period no?

src/utility_functions.jl
```diff
@@ -110,11 +165,39 @@ Dates.second(x::Dates.Millisecond) = round(Int, x.value*1e-3)
...
\ No newline at end of file
+end
+
+# month conversions
+Base.convert(::Type{Dates.Day}, m::Dates.Month) = Day(Second(m))
+Base.convert(::Type{Dates.Hour}, m::Dates.Month) = Hour(Second(m))
+function Base.convert(::Type{Dates.Second}, m::Dates.Month)
+    @warn "Month is assumed to be approximately equal to 30 days. Use Minute, Hour, or Day otherwise."
+    return Second(m.value * 30 * 24 * 60 * 60) # approximate
+end
```
milankl: This is one way of doing it but I'd also be happy to quietly convert this following a logic of if you use Month, Year (which aren't exact periods by definition anway) you get them only approximately. It's not that we are redefining the length of a Second ...

src/utility_functions.jl
```diff
@@ -110,11 +165,39 @@ Dates.second(x::Dates.Millisecond) = round(Int, x.value*1e-3)
...
+
+# additional century conversions
+Base.convert(::Type{Month}, c::Century) = Month(Year(c))
+Base.convert(::Type{Day}, c::Century) = Day(Year(c))
+Base.convert(::Type{Hour}, c::Century) = Hour(Year(c))
+Base.convert(::Type{Second}, c::Century) = Second(Year(c))
+
+# additional month conversions
+Base.convert(::Type{Month}, m::Millenium) = Month(Year(m))
+Base.convert(::Type{Day}, m::Millenium) = Day(Second(Year(m)))
```
milankl: `Day(Second(m))` ?

test/utility_functions.jl
```diff
@@ -62,4 +62,35 @@ end
...
+    @test convert(Second, 1.4) == Second(1)
+    @test convert(Second, 1.5) == Second(2)
+    # Pirated conversions for month and year
+    @test Second(Year(1)) == Second(365 * 24 * 60 * 60)
+    @test Day(Year(1)) == Day(365)
+    @test Second(Month(1)) == Second(30 * 24 * 60 * 60)
+    @test Day(Month(1)) == Day(30)
+    # Test warnings
+    @test_logs (:warn, "Month is assumed to be approximately equal to 30 days. Use Minute, Hour, or Day otherwise.") Second(Month(1))
+    @test_logs (:warn, "Year is assumed to be approximately equal to 365 days. Use Minute, Hour, Day, or Month otherwise.") Second(Year(1))
```
milankl: That test fails if the warning is a different one no? In case someone is updating the warning

src/utility_functions.jl
bgroenks96: Yes. It overrides the default secondary (untyped) constructor, which until now I didn't actually know was a thing. It must have been added in a recent Julia release. See here:

```julia
julia> """
           Century <: Period                                                  
                                                                              
       Convenience time type representing a 100-year period.                  
       """
       struct Century <: Period
           value::Int
       end
Century

julia> methods(Century)                                                       
# 2 methods for type constructor:                                             
 [1] Century(value::Int64)
     @ REPL[3]:7
 [2] Century(value)
     @ REPL[3]:7
```

The default untyped constructor internally calls `convert`; but for `Period` types, it also causes an ambiguous dispatch unless explicitly overloaded.

src/utility_functions.jl
bgroenks96: They are relevant when automatically promoting to a common type. So for example when making a list:

```julia
julia> [Century(1),Year(2)]
2-element Vector{Year}:
 100 years
 2 years
```

It also makes things like `Year(100) == Century(1)` work correctly.

test/utility_functions.jl
bgroenks96: Yes

src/utility_functions.jl
bgroenks96: So you mean remove the warning?

src/utility_functions.jl
bgroenks96: I wouldn't use `CompoundPeriod` in this case because a century is exactly equal to 100 years, following the logic of redefining a year to have an approximate fixed length.

src/utility_functions.jl
bgroenks96: I actually think `Day(Year(m))` is better?

src/utility_functions.jl
milankl: Didn't know about the ambiguity. You could make it `::Integer` though.

bgroenks96: Does that potentially introduce type instability anywhere? I guess not because everything gets converted to `Second`...?

src/utility_functions.jl
milankl: Yeah I think that's reasonable given that `Month`/`Year` aren't exact periods anyway so a user can't expect them to be. Also "Month is assumed to be approximately equal to 30 days" technically should be "Month is assumed to be equal to 30 days", a month is always approximately 30 days be we assume it to be *exactly* 30 days.

src/utility_functions.jl
bgroenks96: Ok, I'm not seeing the ambiguous dispatch error anymore... so I removed the constructor. It may have had something to do with me having added `ConstructionBase` before. Will find out in the next PR.

bgroenks96: Oops, nevermind... it was because of `promote`, not the direct use of the constructor. Re-added them :man_facepalming:.

milankl: What's wrong with
```julia
julia> struct T
           value::Int
           T(value::Integer) = new(Int(value))
       end

julia> methods(T)
# 1 method for type constructor:
 [1] T(value::Integer)
     @ REPL[1]:3
```

bgroenks96: I guess you could do that? You just need to get rid of the untyped constructor. Here is the error:

```
  Test threw exception
  Expression: promote(Year(1), Millenium(1)) == (Year(1), Year(1000))         
  MethodError: Century(::Millenium) is ambiguous.                             

  Candidates:
    Century(value)
      @ SpeedyWeather ~/projects/HybridESM/SpeedyWeather.jl/src/utility_functions.jl:105
    (::Type{T})(p::Dates.Period) where T<:Dates.Period                        
      @ Dates ~/.julia/juliaup/julia-1.10.9+0.x64.linux.gnu/share/julia/stdlib/v1.10/Dates/src/periods.jl:384

  Possible fix, define
    Century(::Dates.Period)

  Stacktrace:
   [1] convert(::Type{Year}, m::Millenium)                                    
     @ SpeedyWeather ~/projects/HybridESM/SpeedyWeather.jl/src/utility_functions.jl:133
   [2] _promote
     @ ./promotion.jl:370 [inlined]
   [3] promote(x::Year, y::Millenium)
     @ Base ./promotion.jl:393
```

But what is the reason for this? To make the constructor less restrictive to other potential integer types?

bgroenks96: Fixed in [fbfd2e0](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/746/commits/fbfd2e0668bddac27841f880c72eaa8a89c763fd)

milankl: What happens if you define, as it says, `Century(p::Period) = convert(Century, p)` ?

milankl: That way `Century(::Any)` tries to convert to `Int` as per default constructor (works for all integers but also `1.0`) but `Century(::Period)` uses the conversion methods you defined.

bgroenks96: That's fine too :)

milankl: > To make the constructor less restrictive to other potential integer types?

Yeah, why be restrictive if you don't have to. `Century(p::Period) = convert(Century, p)` sounds perfectly reasonable to me given that's where the actual conversion is defined. I don't think it's the job of `Century` to define that's convertible to `::Int` and what's not.

milankl:
```julia
julia> run!(simulation, period=Month(1))
ERROR: MethodError: Cannot `convert` an object of type Month to an object of type Millisecond
The function `convert` exists, but no method is defined for this combination of argument types.
```
were missing, just added them.

bgroenks96:
Great, thanks, I'll merge now then.

PR #754, opened 2025-06-25
Add new parameter handling scheme for Barotropic model

### LINKED ISSUE #753: Implement parameter handling for Barotropic model
The first step for the parameter handling scheme is the Barotropic model. The current concept is a non-invasive "opt-in" scheme where model components provide dispatches for a new method:

```julia
parameters(obj; kwargs...) = parameters(SpeedyParam, obj; kwargs...)
parameters(::Type{PT}, obj; kwargs...) where {PT<:ModelParameters.AbstractParam} = (;)
parameters(::Type{PT}, param::ModelParameters.AbstractParam; kwargs...) where {PT<:ModelParameters.AbstractParam} = PT(merge(parent(param), kwargs))
parameters(::Type{PT}, x::Union{Number,AbstractArray}; kwargs...) where {PT<:ModelParameters.AbstractParam} = PT(x; kwargs...)
```

where `SpeedyParam` is a custom implementation of `AbstractParam` that provides concrete-typed fields.

An initial prototype of this for the `Earth` type is as follows:

```julia
parameters(earth::Earth{NF}; kwargs...) where {NF} =
    (
        rotation = parameters(earth.rotation; desc="angular frequency of Earth's rotation [rad/s]", kwargs...),
        gravity = parameters(earth.gravity; bounds=HalfLine(), desc="gravitational acceleration [m/s^2]", kwargs...),
        axial_tilt = parameters(earth.axial_tilt; bounds=-180..180, desc="angle [˚] rotation axis tilt wrt to orbit", kwargs...),
        solar_constant = parameters(earth.solar_constant; bounds=HalfLine(), desc="Total solar irradiance at the distance of 1 AU [W/m²]", kwargs...),
    )
```
The `reconstruct` method (should probably be renamed so as not to conflict with `Flatten.jl`) can then be called with the returned (possibly nested) `NamedTuple` to recreate the type with new parameter values:

```julia
earth = Earth(Float64)
ps = ComponentArray(parameters(earth, category="planet"));
new_ps = stripparams(ps)*2;
earth_updated = SpeedyWeather.reconstruct(earth, new_ps)
```
```
Earth{Float64} <: SpeedyWeather.AbstractPlanet
├ rotation::Float64 = 7.29e-5
├ gravity::Float64 = 9.81
├ daily_cycle::Bool = true
├ length_of_day::Second = 86400 seconds
├ seasonal_cycle::Bool = true
├ length_of_year::Second = 31557600 seconds
├ equinox::DateTime = 2000-03-20T00:00:00
├ axial_tilt::Float64 = 23.4
└ solar_constant::Float64 = 1365.0

Earth{Float64} <: SpeedyWeather.AbstractPlanet
├ rotation::Float64 = 0.0001458
├ gravity::Float64 = 19.62
├ daily_cycle::Bool = true
├ length_of_day::Second = 86400 seconds
├ seasonal_cycle::Bool = true
├ length_of_year::Second = 31557600 seconds
├ equinox::DateTime = 2000-03-20T00:00:00
├ axial_tilt::Float64 = 46.8
└ solar_constant::Float64 = 2730.0
```

Note that all operations here except for the `ComponentArray` constructor are type-stable thus far. However, the initial construction of the `ComponentArray` could be done before or during initialization of the simulation.

One rather unfortunate thing here is the duplication of the docstring in the description field. We could avoid this by either not including a description field in the parameters or figuring out a way to programmatically extract the docstring; I tried that but haven't found a solution yet. @milankl @maximilian-gelbrecht let me know if you are aware of a way to do this.

A draft PR will follow shortly.

bgroenks96:
Resolves #753 (see comments therein)

Note that I still need (?) to go back through and add bounds to the parameters. I am not sure if this is explicitly needed at the moment, but I think it's usually helpful to have.

@milankl and @maximilian-gelbrecht please let me know if you agree with (1) the general scheme, and (2) the choice of which types/fields to mark as parameters. I did this relatively naively so it would be good to get feedback on that.

Remaining tasks
- [x] Add optional parameter argument to `initialize` which automatically updates the model struct in the simulation with the given parameters.
- [x] Add `@parameterized` macro for more user-friendly parameter specification
- [x] Move parameter types and methods into new `Utils` module
- [ ] ~~Update I/O to write parameter table instead of `parameters.txt`~~ (wait until next PR)
- [x] Test compatibility with Enzyme


src/SpeedyWeather.jl
```diff
@@ -24,6 +27,11 @@ import JLD2: jldopen, jldsave, JLDFile
 import CodecZlib
 import BitInformation: round, round!
 import ProgressMeter
+import ModelParameters
+import ConstructionBase: constructorof, getproperties, setproperties
+
+# DOMAINS
+using DomainSets
```
bgroenks96: @milankl I know you prefer `import` generally but I think `using` is easier for `DomainSets`. Also, VS Code incorrectly highlights `import DomainSets: :(..)` (the interval operator) as a syntax error which is kind of annoying...

src/parameters.jl
```diff
@@ -0,0 +1,90 @@
...
+    value::NF = NaN
+    
+    "numerical domain on which the parameter is defined"
+    bounds::Domain = RealLine()
+
+    "human-readable description of the parameter"
+    desc::String = ""
+
+    "additional attributes with unspecified types"
+    attrs::NamedTuple
```
maximilian-gelbrecht: `NamedTuple` isn't a concrete type. 

Generally, we _really_ need concrete types everywhere as this is also important for Enzyme (and Reactant some time in the future)

maximilian-gelbrecht (commented):
Really cool. This looks good so far! 

So, with this, how would the work flow look like if we only want to make one specific of the parameters of a certain struct tunable?

src/parameters.jl
bgroenks96: I am aware, but I created SpeedyParam to get around needing to add type parameters to the model types. I opted for the abstract type here because it (presumably) will not be used or invoked during model execution since we would have already reconstructed the type before that. I think we would be able to avoid the Enzyme code ever seeing SpeedyParam at all, actually. It's really here just as a user interface feature.

Since I adopted the non-invasive approach, though, I guess SpeedyParam is no longer strictly necessary since we're not assigning Param types to struct fields directly.

maximilian-gelbrecht: Okay, understood!

bgroenks96:
> So, with this, how would the work flow look like if we only want to make one specific of the parameters of a certain struct tunable?

Good question. That's not fully fleshed out yet... though it would already be possible by considering subarrays of the `ComponentArray`, e.g.

```julia
spectral_grid = SpectralGrid(trunc=31, nlayers=1)   # define resolution
model_ps = ComponentVector(parameters(model))
model_ps_sub = model_ps[ComponentArrays.label2index("planet.solar_constant")]
new_model = reconstruct(model, model_ps_sub)
```

`ModelParameters` does not explicitly support working with subsets of parameters, although it was something I had long wanted to add. If you follow the more typical `ModelParameters` workflow of assigning `Param` directly to struct fields, you can just assign them only to the fields your interested in tuning (see [here](https://github.com/CryoGrid/CryoGrid.jl/blob/master/examples/cglite_parameter_ensembles.jl) for an example).

However, I think we probably want to avoid this since it then requires every field to always have its own type parameter, and it's a bit awkward from the user point-of-view because, if the user has to specify the `Param` types, they would lose all of the potentially valuable metadata about the parameters (e.g. bounds, description, etc.) unless they already know it...

If you have any ideas about how you would like to be able to do this, please let me know and I'll try to think of how to implement it.

My current idea is to add a `SpeedyParams` wrapper over the named-tuples that would provide APIs for taking subsets.

bgroenks96:
@maximilian-gelbrecht Related question, is `initialize!` parameter-dependent? i.e. when optimizing/sampling/etc. would you include a call to `initialize!` in each evaluation or pre-initialize and then only call `run!`?

maximilian-gelbrecht:
> @maximilian-gelbrecht Related question, is `initialize!` parameter-dependent? i.e. when optimizing/sampling/etc. would you include a call to `initialize!` in each evaluation or pre-initialize and then only call `run!`?

That's a very good question that I haven't considered yet in detail. 

`initialize!` is parameter dependent in general (for some component it's not, but in these cases it also usually just does nothing) 

So, yes, we need to run it again. But we should have a system that only calls the initialize! of the components that are updated. Like, you don't need to reinitialise the transforms for example. If we were to do it like this, it also would be fast I assume.

maximilian-gelbrecht:
I think the best cause of action would be set up a simple test case with Enzyme to test all of these things. 

We should probably also merge #680 for that. I can update it as soon as I find some time.

bgroenks96:
> But we should have a system that only calls the initialize! of the components that are updated. Like, you don't need to reinitialise the transforms for example. If we were to do it like this, it also would be fast I assume.

Maybe this can just be done with `if` checks, but otherwise we could introduce a new method `reinitialize!` that defaults to calling `initialize!` but can be overridden for special cases.

maximilian-gelbrecht:
Just mentioning that in order for this work properly, then one would also need to compute gradients of not just the time stepping / `run!`, but this `reinitialize!` and `run!` together in the general case. So I think we really don't want to call the initialisation of the whole model, but just where we insert parameters and it's needed.  

On the other hand.... for an early proof of concept we could also just focus on those parameters that don't need an `initialize`, so that are directly used, espacially because this applies to any kind of ANN with Lux.

bgroenks96:
Ok still need tests but subsetting parameters is now supported via `SpeedyParams`, which is a customized implementation of the `AbstractModel` interface from ModelParameters (confusing, I know...).

## Table interface

```julia
spectral_grid = SpectralGrid(trunc=31, nlayers=1)   # define resolution
model_ps = SpeedyParams(parameters(model))
```

```
SpeedyParams{@NamedTuple{planet::@NamedTuple{rotation::SpeedyParam{Float32}, gravity::SpeedyParam{Float32}, axial_tilt::SpeedyParam{Float32}, solar_constant::SpeedyParam{Float32}}, atmosphere::@NamedTuple{mol_mass_dry_air::SpeedyParam{Float32}, mol_mass_vapour::SpeedyParam{Float32}, heat_capacity::SpeedyParam{Float32}, R_gas::SpeedyParam{Float32}, R_dry::SpeedyParam{Float32}, R_vapour::SpeedyParam{Float32}, mol_ratio::SpeedyParam{Float32}, μ_virt_temp::SpeedyParam{Float32}, κ::SpeedyParam{Float32}, water_density::SpeedyParam{Float32}, latent_heat_condensation::SpeedyParam{Float32}, latent_heat_sublimation::SpeedyParam{Float32}, stefan_boltzmann::SpeedyParam{Float32}, pres_ref::SpeedyParam{Float32}, temp_ref::SpeedyParam{Float32}, moist_lapse_rate::SpeedyParam{Float32}, dry_lapse_rate::SpeedyParam{Float32}, layer_thickness::SpeedyParam{Float32}}, forcing::@NamedTuple{strength::SpeedyParam{Float32}, wavenumber::SpeedyParam{Float32}}, drag::@NamedTuple{c::SpeedyParam{Float32}}}} with parent object of type: 

@NamedTuple{planet::@NamedTuple{rotation::SpeedyParam{Float32}, gravity::SpeedyParam{Float32}, axial_tilt::SpeedyParam{Float32}, solar_constant::SpeedyParam{Float32}}, atmosphere::@NamedTuple{mol_mass_dry_air::SpeedyParam{Float32}, mol_mass_vapour::SpeedyParam{Float32}, heat_capacity::SpeedyParam{Float32}, R_gas::SpeedyParam{Float32}, R_dry::SpeedyParam{Float32}, R_vapour::SpeedyParam{Float32}, mol_ratio::SpeedyParam{Float32}, μ_virt_temp::SpeedyParam{Float32}, κ::SpeedyParam{Float32}, water_density::SpeedyParam{Float32}, latent_heat_condensation::SpeedyParam{Float32}, latent_heat_sublimation::SpeedyParam{Float32}, stefan_boltzmann::SpeedyParam{Float32}, pres_ref::SpeedyParam{Float32}, temp_ref::SpeedyParam{Float32}, moist_lapse_rate::SpeedyParam{Float32}, dry_lapse_rate::SpeedyParam{Float32}, layer_thickness::SpeedyParam{Float32}}, forcing::@NamedTuple{strength::SpeedyParam{Float32}, wavenumber::SpeedyParam{Float32}}, drag::@NamedTuple{c::SpeedyParam{Float32}}}


Parameters:
┌───────┬──────────────────────────┬──────────┬──────────────────────────────┬────────────┬─────────────────────────────────────┬────────────────────────────────────────────────────────────────────┐
│   idx │                fieldname │      val │                componenttype │  component │                              bounds │                                                               desc │
│ Int64 │                   Symbol │  Float32 │                     DataType │     Symbol │ IntervalSets.TypedEndpointsInterval │                                                             String │
├───────┼──────────────────────────┼──────────┼──────────────────────────────┼────────────┼─────────────────────────────────────┼────────────────────────────────────────────────────────────────────┤
│     1 │                 rotation │  7.29e-5 │               Earth{Float32} │     planet │       -Inf .. Inf (open) (RealLine) │                      angular frequency of Earth's rotation [rad/s] │
│     2 │                  gravity │     9.81 │               Earth{Float32} │     planet │ 0.0 .. Inf (closed-open) (HalfLine) │                                 gravitational acceleration [m/s^2] │
│     3 │               axial_tilt │     23.4 │               Earth{Float32} │     planet │                         -180 .. 180 │                          angle [˚] rotation axis tilt wrt to orbit │
│     4 │           solar_constant │   1365.0 │               Earth{Float32} │     planet │ 0.0 .. Inf (closed-open) (HalfLine) │              Total solar irradiance at the distance of 1 AU [W/m²] │
│     5 │         mol_mass_dry_air │  28.9649 │     EarthAtmosphere{Float32} │ atmosphere │       -Inf .. Inf (open) (RealLine) │                                      molar mass of dry air [g/mol] │
│     6 │          mol_mass_vapour │  18.0153 │     EarthAtmosphere{Float32} │ atmosphere │       -Inf .. Inf (open) (RealLine) │                                 molar mass of water vapour [g/mol] │
│     7 │            heat_capacity │   1004.0 │     EarthAtmosphere{Float32} │ atmosphere │       -Inf .. Inf (open) (RealLine) │                     specific heat at constant pressure cₚ [J/K/kg] │
│     8 │                    R_gas │   8.3145 │     EarthAtmosphere{Float32} │ atmosphere │       -Inf .. Inf (open) (RealLine) │                                   universal gas constant [J/K/mol] │
│     9 │                    R_dry │  287.054 │     EarthAtmosphere{Float32} │ atmosphere │       -Inf .. Inf (open) (RealLine) │                         specific gas constant for dry air [J/kg/K] │
│    10 │                 R_vapour │  461.524 │     EarthAtmosphere{Float32} │ atmosphere │       -Inf .. Inf (open) (RealLine) │                    specific gas constant for water vapour [J/kg/K] │
│    11 │                mol_ratio │  0.62197 │     EarthAtmosphere{Float32} │ atmosphere │       -Inf .. Inf (open) (RealLine) │ Ratio of gas constants: dry air / water vapour, often called ε [1] │
│    12 │              μ_virt_temp │ 0.607794 │     EarthAtmosphere{Float32} │ atmosphere │       -Inf .. Inf (open) (RealLine) │                          Virtual temperature Tᵥ calculation factor │
│    13 │                        κ │ 0.285911 │     EarthAtmosphere{Float32} │ atmosphere │       -Inf .. Inf (open) (RealLine) │                   = R_dry/cₚ, gas const for air over heat capacity │
│    14 │            water_density │   1000.0 │     EarthAtmosphere{Float32} │ atmosphere │       -Inf .. Inf (open) (RealLine) │                                              water density [kg/m³] │
│    15 │ latent_heat_condensation │  2.501e6 │     EarthAtmosphere{Float32} │ atmosphere │       -Inf .. Inf (open) (RealLine) │                                 latent heat of condensation [J/kg] │
│    16 │  latent_heat_sublimation │  2.801e6 │     EarthAtmosphere{Float32} │ atmosphere │       -Inf .. Inf (open) (RealLine) │                                  latent heat of sublimation [J/kg] │
│    17 │         stefan_boltzmann │  5.67e-8 │     EarthAtmosphere{Float32} │ atmosphere │       -Inf .. Inf (open) (RealLine) │                                stefan-Boltzmann constant [W/m²/K⁴] │
│    18 │                 pres_ref │ 100000.0 │     EarthAtmosphere{Float32} │ atmosphere │       -Inf .. Inf (open) (RealLine) │                                    surface reference pressure [Pa] │
│    19 │                 temp_ref │    288.0 │     EarthAtmosphere{Float32} │ atmosphere │       -Inf .. Inf (open) (RealLine) │                                  surface reference temperature [K] │
│    20 │         moist_lapse_rate │    0.005 │     EarthAtmosphere{Float32} │ atmosphere │       -Inf .. Inf (open) (RealLine) │             reference moist-adiabatic temperature lapse rate [K/m] │
│    21 │           dry_lapse_rate │   0.0098 │     EarthAtmosphere{Float32} │ atmosphere │       -Inf .. Inf (open) (RealLine) │               reference dry-adiabatic temperature lapse rate [K/m] │
│    22 │          layer_thickness │   8500.0 │     EarthAtmosphere{Float32} │ atmosphere │       -Inf .. Inf (open) (RealLine) │                    layer thickness for the shallow water model [m] │
│    23 │                 strength │  3.0e-12 │      KolmogorovFlow{Float32} │    forcing │       -Inf .. Inf (open) (RealLine) │                                         Strength of forcing [1/s²] │
│    24 │               wavenumber │      8.0 │      KolmogorovFlow{Float32} │    forcing │       -Inf .. Inf (open) (RealLine) │       Wavenumber of forcing in meridional direction (pole to pole) │
│    25 │                        c │   1.0e-7 │ LinearVorticityDrag{Float32} │       drag │       -Inf .. Inf (open) (RealLine) │                            Linear vorticity drag coefficient [1/s] │
└───────┴──────────────────────────┴──────────┴──────────────────────────────┴────────────┴─────────────────────────────────────┴────────────────────────────────────────────────────────────────────┘

```

Note that this also means you can do things like:

```julia
import CSV

CSV.write("parameters.csv", model_ps)
```

One quick note about the table interface: `ModelParameters` doesn't fully implement tabular behavior; e.g. it only supports column indexing and not row indexing, which would in principle also allow for flexibly selecting subsets of parameters based on their attributes. I had a [PR](https://github.com/rafaqz/ModelParameters.jl/pull/44) for this a long time ago but it never got merged. But I think @rafaqz is open to putting some development effort into ModelParameters again so that's something I can consider taking another look at.

## Parameter subsets

For now, parameter subsets can be extracted with the `ComponentArrays` label syntax:

```julia
model_ps_subset = model_ps[["planet.solar_constant","atmosphere.heat_capacity"]]
```

```
Parameters:
┌───────┬────────────────┬─────────┬──────────────────────────┬────────────┬─────────────────────────────────────────────────────┬───────────────────────────────────────────────────────┐
│   idx │      fieldname │     val │            componenttype │  component │                                              bounds │                                                  desc │
│ Int64 │         Symbol │ Float32 │                 DataType │     Symbol │ DomainSets.FixedInterval{L, :open, Float64} where L │                                                String │
├───────┼────────────────┼─────────┼──────────────────────────┼────────────┼─────────────────────────────────────────────────────┼───────────────────────────────────────────────────────┤
│     1 │ solar_constant │  1365.0 │           Earth{Float32} │     planet │                 0.0 .. Inf (closed-open) (HalfLine) │ Total solar irradiance at the distance of 1 AU [W/m²] │
│     2 │  heat_capacity │  1004.0 │ EarthAtmosphere{Float32} │ atmosphere │                       -Inf .. Inf (open) (RealLine) │        specific heat at constant pressure cₚ [J/K/kg] │
└───────┴────────────────┴─────────┴──────────────────────────┴────────────┴─────────────────────────────────────────────────────┴───────────────────────────────────────────────────────┘
```

A `CompoentVector` can be constructed from SpeedyParams with `vec`:

```julia
pvec = vec(model_ps)
```

```
ComponentVector{Float32}(planet = (rotation = 7.29f-5, gravity = 9.81f0, axial_tilt = 23.4f0, solar_constant = 1365.0f0), atmosphere = (mol_mass_dry_air = 28.9649f0, mol_mass_vapour = 18.0153f0, heat_capacity = 1004.0f0, R_gas = 8.3145f0, R_dry = 287.05432f0, R_vapour = 461.52438f0, mol_ratio = 0.62197006f0, μ_virt_temp = 0.60779446f0, κ = 0.2859107f0, water_density = 1000.0f0, latent_heat_condensation = 2.501f6, latent_heat_sublimation = 2.801f6, stefan_boltzmann = 5.67f-8, pres_ref = 100000.0f0, temp_ref = 288.0f0, moist_lapse_rate = 0.005f0, dry_lapse_rate = 0.0098f0, layer_thickness = 8500.0f0), forcing = (strength = 3.0f-12, wavenumber = 8.0f0), drag = (c = 1.0f-7))
```

I will add tests and reasonable `bounds` for all parameters tomorrow. Also, if @milankl is up for it, I think it would be nice to add `Unitful` as another dependency (sorry for adding so many...) and have first-class units as a column in the parameter table.

rafaqz:
Yep I am very open to changing ModelParameters.jl to fit needs here. I'm hoping do some kind of second round overhaul to make it work for more use cases.

src/SpeedyWeather.jl
milankl: Fine with me

milankl:
1. We currently write `parameters.txt` with output, `CSV.write("parameters.csv", model_ps)` looks like a neater approach feel free to use that instead (in another PR probably though)

2. > Also, if @milankl is up for it, I think it would be nice to add Unitful as another dependency (sorry for adding so many...) and have first-class units as a column in the parameter table.

I'm not strictly against Unitful just careful. So question can we first keep the units in the docstrings/description and then if everything else works get the units in? Unitful seems to be relatively light weight so don't see big problems adding this as a dependency

src/dynamics/atmosphere.jl
```diff
@@ -64,4 +64,26 @@ end
...
+        water_density = parameterof(atm, :water_density; desc="water density [kg/m³]", kwargs...),
+        latent_heat_condensation = parameterof(atm, :latent_heat_condensation; desc="latent heat of condensation [J/kg]", kwargs...),
+        latent_heat_sublimation = parameterof(atm, :latent_heat_sublimation; desc="latent heat of sublimation [J/kg]", kwargs...),
+        stefan_boltzmann = parameterof(atm, :stefan_boltzmann; desc="stefan-Boltzmann constant [W/m²/K⁴]", kwargs...),
+        pres_ref = parameterof(atm, :pres_ref; desc="surface reference pressure [Pa]", kwargs...),
+        temp_ref = parameterof(atm, :temp_ref; desc="surface reference temperature [K]", kwargs...),
+        moist_lapse_rate = parameterof(atm, :moist_lapse_rate; desc="reference moist-adiabatic temperature lapse rate [K/m]", kwargs...),
+        dry_lapse_rate = parameterof(atm, :dry_lapse_rate; desc="reference dry-adiabatic temperature lapse rate [K/m]", kwargs...),
+        layer_thickness = parameterof(atm, :layer_thickness; desc="layer thickness for the shallow water model [m]", kwargs...),
+    )
```
milankl: So as far as I understand this, every model component now needs to have this defined additionally. While it feels a bit like code duplication as we mean essentially the same when writing
```julia
"Some doc string"
R_gas::NF = 8.3145
```
and so in an ideal world I would have simply marked this somehow as a parameter similar to how one can put `const` in front of a struct field to mean immutability. But if this is the current best way to implement this I'd be happy with this, it feels indeed non-invasive, just requires one to keep two lines of code updated when making changes.

milankl: Also just checking the `heat_capacity` and `:heat_capacity` always need to be the same yeah?

```julia
heat_capacity = parameterof(atm, :heat_capacity; desc="specific heat at constant pressure cₚ [J/K/kg]", kwargs...)
```

src/dynamics/planet.jl
```diff
@@ -44,3 +44,12 @@ end
 
 Earth(SG::SpectralGrid; kwargs...) = Earth{SG.NF}(; kwargs...)
 Earth(::Type{NF}; kwargs...) where NF = Earth{NF}(; kwargs...)
+
+parameters(earth::Earth; kwargs...) =
+    (
+        rotation = parameterof(earth, :rotation; desc="angular frequency of Earth's rotation [rad/s]", kwargs...),
+        gravity = parameterof(earth, :gravity; bounds=HalfLine(), desc="gravitational acceleration [m/s^2]", kwargs...),
+        axial_tilt = parameterof(earth, :axial_tilt; bounds=-180..180, desc="angle [˚] rotation axis tilt wrt to orbit", kwargs...),
```
milankl: should be `bounds=-90..90`, axial tilt is the tropic of cancer, i.e. the latitude the sun reaches during solstice

src/models/barotropic.jl
```diff
@@ -61,6 +61,13 @@ end
 prognostic_variables(::Type{<:Barotropic}) = (:vor,)
 default_concrete_model(::Type{Barotropic}) = BarotropicModel
 
+parameters(model::Barotropic; kwargs...) = (
```
milankl: these can always be stacked? Because in the primitive models we have one more layer, e.g. `model.land.soil_moisture`

src/parameters.jl
```diff
@@ -0,0 +1,196 @@
...
+    $SIGNATURES
+
+Returns a "parameterized" copy of the given data structure. For numeric types, this simply wraps
+the given value in a `Param`-like type. Dispatches can be added for compound data types that
+reconstruct the data type with `Param`-like wrappers on all relevant parameters.
+"""
+parameters(obj; kwargs...) = parameters(SpeedyParam, obj; kwargs...)
+parameters(::Type{PT}, obj; kwargs...) where {PT<:AbstractParam} = (;)
+parameters(::Type{PT}, param::AbstractParam; kwargs...) where {PT<:AbstractParam} = PT(merge(parent(param), kwargs))
+parameters(::Type{PT}, x::Union{Number,AbstractArray}; kwargs...) where {PT<:AbstractParam} = PT(x; kwargs...)
```
milankl: These are not defined in ModelParameters.jl? Feels like type piracy ...

src/parameters.jl
```diff
@@ -0,0 +1,196 @@
+const AbstractParam = ModelParameters.AbstractParam
```
milankl: Don't you want to `import ModelParameters: AbstractParam`?

milankl: (Does that even make a difference?)

src/parameters.jl
```diff
@@ -0,0 +1,196 @@
...
+"""
+    $SIGNATURES
+
+Reconstructs the given data structure with the given `values`. If `values` is a `NamedTuple`, the structure
+must match that of `obj`. This function is used to reconstruct model types from `SpeedyParams`.
+"""
+reconstruct(obj::T, value::T) where {T} = value
+reconstruct(obj::AbstractParam, value::T) where {T} = ModelParameters.update(obj, Tuple(value))
+reconstruct(obj::NamedTuple{keys,V}, values::NamedTuple{keys,V}) where {keys,V<:Tuple} = values
+reconstruct(obj, values::ComponentArray) = reconstruct(obj, NamedTuple(values))
```
milankl: similar here

milankl:
I really like this, as said in #753 I'm happy with defining `parameters` for every model component we have and just duplicate some code there to have a separate list of what's a parameter and what's not. Seems worth the effort.

However I would love for the high-level user interface to be as simple as

```julia
earth = Earth(...)
ps = parameters(earth)
new_earth = SpeedyWeather.reconstruct(earth, ps*2)
```

(which should throw an error if some parameters cannot be `*2`... Is there anything that blocks such a neat interface?

src/parameters.jl
bgroenks96: No. `parameters` is currently being defined in Speedy. ModelParameters has a method `params` but this does something subtly different, i.e. it returns a flattened tuple of all `Param` types in the object.

src/models/barotropic.jl
bgroenks96: Yes, they can always be stacked, as many times as you want :)

src/dynamics/planet.jl
bgroenks96: Yes, I actually wanted your input on setting bounds for all of the parameters. I am currently making an attempt but I will rely on you to tell me if I do anything wrong (like here).

src/dynamics/atmosphere.jl
bgroenks96: > Also just checking the heat_capacity and :heat_capacity always need to be the same yeah?

Yes, not great, I know. But I wanted to do this so that we could extract the component type. from the object (here `atm`) and automatically store it in the parameter metadata.

> So as far as I understand this, every model component now needs to have this defined additionally. 

Yes. I also would like to avoid duplicating the docstring, but as far as I can tell, there is no way to programmatically extract it, except while in the REPL.

> and so in an ideal world I would have simply marked this somehow as a parameter similar to how one can put const in front of a struct field to mean immutability.

There is [FieldMetadata.jl](https://github.com/rafaqz/FieldMetadata.jl) which would allow for something like this, but then you would still need to duplicate the docstring, or move it to the description meatdata. I don't love FieldMetadata though... @rafaqz doesn't support it anymore (I think) and it's a bit ugly and cumbersome. I guess I also don't like macros that hide too many things from the user/programmer.

>  just requires one to keep two lines of code updated when making changes.

It doesn't even strictly require it in all cases; e.g. if you add a new parameter, it won't break anything not to include it in the named tuple returned by `parameters`. It just won't show up there until you add it.

src/parameters.jl
bgroenks96: Could also do that, but then you have to write `import ModelParameters: ModelParameters, AbstractParam` which always kinds of annoys me.

Then again, I guess it's not worse than this...

src/parameters.jl
bgroenks96: No type piracy here, just bad naming. Haven't thought of something better yet. `rebuild` is taken... maybe `remake` but that's what SciML uses everywhere.

src/parameters.jl
milankl: In practice there isn't that big of a difference I agree but I think I'd prefer to have things that are, in fact, *imported* from other modules to be labelled as `import`. Reads just more consistent to me. `const` sounds like a local (as in independent from other modules) constant to me...

bgroenks96:
@milankl Since I value your happiness, I have attempted to implement your ideal solution with a macro `@parameterized`.

Here is an example for `Earth`:

```julia
@parameterized @kwdef mutable struct Earth{NF<:AbstractFloat} <: AbstractPlanet

    "angular frequency of Earth's rotation [rad/s]"
    @param rotation::NF = DEFAULT_ROTATION
    
    "gravitational acceleration [m/s^2]"
    @param gravity::NF = DEFAULT_GRAVITY (bounds=Nonnegative,)
    
    "switch on/off daily cycle"
    daily_cycle::Bool = true
    
    "Seconds in a daily rotation"
    length_of_day::Second = Hour(24)

    "switch on/off seasonal cycle"
    seasonal_cycle::Bool = true

    "Seconds in an orbit around the sun"
    length_of_year::Second = Day(365.25)
    
    "time of spring equinox (year irrelevant)"
    equinox::DateTime = DateTime(2000, 3, 20) 

    "angle [˚] rotation axis tilt wrt to orbit"
    @param axial_tilt::NF = 23.4 (bounds=-90..90,)

    "Total solar irradiance at the distance of 1 AU [W/m²]"
    @param solar_constant::NF = 1365 (bounds=Nonnegative,)
end
```
The `@param` is not a real macro just a marker... which is maybe confusing. I have mixed feelings about this since I generally don't love introducing non-standard Julia syntax... but maybe the win here is big enough to merit it. One particularly big advantage is that the docstring is automatically extracted without duplication.

```julia
model_ps = parameters(Earth(Float16))

Parameters:
┌───────┬────────────────┬─────────┬────────────────┬─────────────────────────────────────┬───────────────────────────────────────────────────────┐
│   idx │      fieldname │     val │  componenttype │                              bounds │                                                  desc │
│ Int64 │         Symbol │ Float16 │       DataType │ IntervalSets.TypedEndpointsInterval │                                                String │
├───────┼────────────────┼─────────┼────────────────┼─────────────────────────────────────┼───────────────────────────────────────────────────────┤
│     1 │       rotation │ 7.29e-5 │ Earth{Float16} │       -Inf .. Inf (open) (RealLine) │         angular frequency of Earth's rotation [rad/s] │
│     2 │        gravity │    9.81 │ Earth{Float16} │ 0.0 .. Inf (closed-open) (HalfLine) │                    gravitational acceleration [m/s^2] │
│     3 │     axial_tilt │    23.4 │ Earth{Float16} │                           -90 .. 90 │             angle [˚] rotation axis tilt wrt to orbit │
│     4 │ solar_constant │ 1.365e3 │ Earth{Float16} │ 0.0 .. Inf (closed-open) (HalfLine) │ Total solar irradiance at the distance of 1 AU [W/m²] │
└───────┴────────────────┴─────────┴────────────────┴─────────────────────────────────────┴───────────────────────────────────────────────────────┘
```

What do you think?

Edit: I guess I should also clarify that the `parameters` dispatch is automatically generated by the macro.

maximilian-gelbrecht:
I get your concerns about non-Julia syntax and too many macros, but this looks really almost exactly like what we want to have. 

If it works reliably and doesn't hurt performance, I would be in favour of it. 

Btw: I think we should probably put these parameter related functions into another new submodule, or in the `Utils` submodule once #727 is merged

src/SpeedyWeather.jl
```diff
@@ -27,13 +27,17 @@ import JLD2: jldopen, jldsave, JLDFile
...
 import ConstructionBase: constructorof, getproperties, setproperties
 
 # DOMAINS
-import DomainSets: Domain, RealLine, HalfLine, UnitInterval
+import DomainSets: Domain, RealLine, NonnegativeRealLine, PositiveRealLine, NegativeRealLine, UnitInterval
 using DomainSets.IntervalSets
 
+const StrictlyPositive = PositiveRealLine()
+const Nonnegative = NonnegativeRealLine()
+const Negative = NegativeRealLine()
```
milankl: `StrictlyPositive` but not `StrictlyNegative`? I wouldn't even mind if we stick to the terminology introduced by DomainSets? `bounds=Negative` sounds great to me though!

src/parameters.jl
```diff
@@ -11,7 +9,7 @@ Base.@kwdef struct SpeedyParam{NF<:AbstractFloat} <: AbstractParam{NF}
     value::NF = NaN
     
     "numerical domain on which the parameter is defined"
-    bounds::Domain = RealLine()
+    bounds::Domain = ℝ
```
milankl: FANCY! I think I do would prefer `RealLine()` though as R or any other fancy R is just ambiguous across the whole naming scope a model like SpeedyWeather tries to cover (radius? gas constant? drag coefficient?)

src/SpeedyWeather.jl
milankl: Also because you do use `RealLine()` or $\mathbb{R}$ below without this `const` renaming

src/parameters.jl
```diff
@@ -125,6 +124,9 @@ Base.getindex(ps::SpeedyParams, nm::Symbol) = getindex(ps, :, nm)
     end
 end
 
+## non-mutating setindex!
+## consider using a method instead of setindex! to avoid confusion?
+Base.setindex!(ps::SpeedyParams, x, nm::Symbol) = setindex!(ps, x, :, nm)
```
milankl: yeah I find `index` which takes a `Symbol` as argument confusing, `set_parameters!` simply?

milankl: I see it's because `ModelParameters._setindex` uses that terminology, why?

src/parameters.jl
```diff
@@ -172,6 +174,82 @@ function reconstruct(obj, values::NamedTuple{keys}) where {keys}
...
+            :($fieldname::$FT = $defval)
+        # case 3: non-parameterized field definition
+        elseif @capture(ex, fieldname_::FT_ = defval_) || @capture(ex, fieldname_::FT_)
+            # reset lastdoc variable to prevent confusing docstrings between parameteter and non-parameter fields
+            lastdoc = nothing
+            ex
+        # case 4: field documentation
+        elseif @capture(ex, docs_String)
+            # store in lastdoc field
+            lastdoc = docs
```
milankl: This technically would allow

```julia
    "doc string"
    # empty line here
    a::T
```

no? Because I think DocStringExtensions.jl doesn't allow that?!

milankl (commented):
This is awesome! And I agree with Max likely exactly what we want!

milankl:
> If it works reliably and doesn't hurt performance, I would be in favour of it.

If I understand this correctly `parameters(::Earth)` is now defined *every time* one constructs an `earth = Earth(Float16)` right? But the `earth` is then exactly the same as without `@parameterized` so there cannot be any performance penalty?

milankl:
Somewhat unrelated but I struggled with this for a while
```julia
julia> Earth()
ERROR: MethodError: no method matching Earth(::Float64, ::Float64, ::Bool, ::Hour, ::Bool, ::Second, ::DateTime, ::Float64, ::Int64)
The type `Earth` exists, but no method is defined for this combination of argument types when trying to construct it.
```

while the constructor `Earth()` is defined automatically through `@kwdef` (and therefore cannot be redefined manually without causing precompilation issues) it can't do the type conversion Hour -> Second and Int64 -> Float64 without the type parameter `NF` in `Earth{NF}` being specified. That's why we have `Earth(::Type{NF})` and `Earth(::SpectralGrid)` methods. But I feel it counterintuitive that one cannot just create an `Earth()` ... 😠

src/parameters.jl
bgroenks96: Yes, it would allow that, but I don't think you should do it?

bgroenks96:
> If I understand this correctly parameters(::Earth) is now defined every time one constructs an earth = Earth(Float16) right?

No, it gets defined once when the macro is expanded prior to compilation, which should happen when you do `using SpeedyWeather`.

But yes, it really cannot possibly have any effect on performance. That was the point of making it non-invasive!

bgroenks96:
> the constructor Earth() is defined automatically through @kwdef (and therefore cannot be redefined manually without causing precompilation issues) it can't do the type conversion Hour -> Second and Int64 -> Float64 without the type parameter NF in Earth{NF} being specified. That's why we have Earth(::Type{NF}) and Earth(::SpectralGrid) methods. But I feel it counterintuitive that one cannot just create an Earth() ... 😠

Why can't it be redefined...? It's not obvious to me why that would cause precompilation issues.

src/parameters.jl
bgroenks96: Yes this is just because of `ModelParameters`. I would like to change it, and we probably will change it soon. But I think this needs to stay the way it is for now in order to be consistent with the `Model` interface.

src/parameters.jl
bgroenks96: I thought I removed this... but yes I will change it back.

src/SpeedyWeather.jl
bgroenks96: Oops, yes that was my mistake. We can just do `Positive` and `Negative`, I think; that would be the most clear within the context of Speedy.

milankl:
Oh you can't redefine a method otherwise you get precompilation warnings! Try it 😉

src/SpeedyWeather.jl
milankl: Sounds good!! Even though `Positive` sounds ambiguous somehow out of context it's not clear it refers to bounds?

src/parameters.jl
milankl: Yeah no. Just thought out loud here. I think it's fine don't see this causing problems

bgroenks96:
> Oh you can't redefine a method otherwise you get precompilation warnings! Try it 😉

~~That should only be the case if the redefinition causes recompilation of other functions~~, which shouldn't happen in this case...? But maybe I'm wrong.... it's often quite unintuitive in Julia what triggers recompilation. I think the compiler takes a very conservative approach most of the time.

Edit: Ok yeah... I see. I haven't dealt with the problem of redefining methods in a long time... and I guess they made this much stricter at some point. I definitely remember doing method redefinition for hotfixes in earlier versions of Julia (probably 1.6)  and did not get this warning.

bgroenks96:
@milankl Please review the choices of  `bounds` on all parameters!

src/SpeedyWeather.jl
bgroenks96: Hmm... it doesn't necessarily have to refer to bounds actually. You could use it in other contexts. It's just a domain.

src/Utils/parameters.jl
```diff
@@ -0,0 +1,343 @@
...
+
+# Override internal ModelParameters method _columntypes to condense type names in table schema (just to look nicer)
+# TODO: propose this change upstream in ModelParameters.jl
+ModelParameters._columntypes(ps::SpeedyParams) = map(k -> promote_type(map(typeof, getindex(ps, k))...), keys(ps)) 
+
+# parameters method interface
+"""
+    $SIGNATURES
+
+Returns a "parameterized" copy of the given data structure. For numeric types, this simply wraps
```
bgroenks96: TODO: fix this out-of-date docstring

src/dynamics/atmosphere.jl
```diff
@@ -6,62 +6,62 @@ $(TYPEDSIGNATURES)
...
     "molar mass of water vapour [g/mol]"
-    mol_mass_vapour::NF = 18.0153
+    @param mol_mass_vapour::NF = 18.0153 (bounds=Positive,)
 
     "specific heat at constant pressure cₚ [J/K/kg]"
-    heat_capacity::NF = 1004
+    @param heat_capacity::NF = 1004 (bounds=Nonnegative,)
 
     "universal gas constant [J/K/mol]"
     R_gas::NF = 8.3145
```
milankl, line 20: I removed `@param` here as I think some of these true physical constants shouldn't be regarded as a parameter. Sure you can change them if you really want when constructing an atmosphere but do we really want to flag them as a tunable parameter??

src/dynamics/atmosphere.jl
```diff
@@ -6,62 +6,62 @@ $(TYPEDSIGNATURES)
...
 
     "specific gas constant for dry air [J/kg/K]"
-    R_dry::NF = 1000*R_gas/mol_mass_dry_air
+    @param R_dry::NF = 1000*R_gas/mol_mass_dry_air (bounds=Positive,)
```
milankl: So many models would probably just use `R_dry` as parameter instead of calculating it on the fly from the universal gas constant but so here flagging `R_dry` and `mol_mass_dry_air` as independently tunable parameters is also weird as they should clearly be coupled following that formula.

src/dynamics/drag.jl
```diff
@@ -25,12 +25,13 @@ end
 
 # Quadratic drag
 export QuadraticDrag
-@kwdef mutable struct QuadraticDrag{NF} <: AbstractDrag
+@parameterized @kwdef mutable struct QuadraticDrag{NF} <: AbstractDrag
     "[OPTION] drag coefficient [1]"
-    c_D::NF = 1e-12         # TODO is this a good default?
+    @param c_D::NF = 1e-12 (bounds=Nonnegative,)    # TODO is this a good default?
```
milankl, line 30: drag coefficients should generally be non negative. Zero will just switch off the drag term that is fine but negative would accelerate the flow in the opposite direction which has generally a diverging exponential as solution. Not a good idea!

src/models/barotropic.jl
```diff
@@ -61,6 +61,13 @@ end
 prognostic_variables(::Type{<:Barotropic}) = (:vor,)
 default_concrete_model(::Type{Barotropic}) = BarotropicModel
 
+parameters(model::Barotropic; kwargs...) = SpeedyParams(
+    planet = parameters(model.planet; component=:planet, kwargs...),
+    atmosphere = parameters(model.atmosphere; component=:atmosphere, kwargs...),
+    forcing = parameters(model.forcing; component=:forcing, kwargs...),
+    drag = parameters(model.drag; component=:drag, kwargs...),
+)
```
milankl, line 69: Just to clarify: While structs with parameters now can just use the new `@parameterized, @param` logic structs of those structs still need to be defined manually collecting several parameters together? That sounds reasonable to me!

milankl (commented):
Changed some bounds with the last commit!

bgroenks96:
Thanks @milankl :)

src/Utils/parameters.jl
```diff
@@ -0,0 +1,343 @@
...
+    $SIGNATURES
+
+Unpacks and strips all `AbstractParam` types from the given `params` object, replacing them with their numeric `val`.
+This is a type-stable alternative to the (currently unstable) `stripparams` in `ModelParameters`.
+"""
+stripparams(param::AbstractParam) = value(param)
+stripparams(params::NamedTuple) = map(stripparams, params)
+stripparams(params::SpeedyParams) = stripparams(unpack_params(params))
+
+# Base overridese for SpeedyParams
```
maximilian-gelbrecht: typo

src/Utils/parameters.jl
```diff
@@ -0,0 +1,343 @@
...
+"""
+    $SIGNATURES
+
+Reconstructs the given data structure with the given `values`. If `values` is a `NamedTuple`, the structure
+must match that of `obj`. This function is used to reconstruct model types from `SpeedyParams`.
+"""
+reconstruct(obj::T, value::T) where {T} = value
+reconstruct(obj::AbstractParam, value::T) where {T} = ModelParameters.update(obj, Tuple(value))
+reconstruct(obj::NamedTuple{keys,V}, values::NamedTuple{keys,V}) where {keys,V<:Tuple} = values
+reconstruct(obj, values::ComponentArray) = reconstruct(obj, NamedTuple(values))
```
maximilian-gelbrecht, line 178: Just wondering, if we generally keep the parameters in a ComponentArray in ML tasks, maybe it would be better to have a routine that directly reconstructs from a ComponentArray as well. Or is `NamedTuple(::ComponentArray)` trivial and non-allocating?

maximilian-gelbrecht (commented):
I am not so familiar with DomainSets. What happens when trying to set a parameter with an out of bounds value? Just a straight out of bounds error? I am even wondering what kind of behaviour we would desire when using this in optimisation tasks. Maybe I'd even prefer it staying at the lower/upper bound and just raising a warning.

bgroenks96:
@maximilian-gelbrecht 
> What happens when trying to set a parameter with an out of bounds value? Just a straight out of bounds error?

At the moment, nothing. DomainSets is just being used to define the bounds types which are then available as parameter metadata. This metadata can then be later used to automatically specify priors or constraints/transforms when setting up a sampler/optimizer.

In principle, we could also augment `@parameterized` to override the default constructor and add checks based on the bounds... but that seems like a lot of work for a pretty minor gain...?

maximilian-gelbrecht:
No, that's fine for now definitely. I think in the long run, some kind of check might be nice, but that's not something we'll have to do right now. 

I just imagine that it would be pretty cool to also have an automatic optimization within physical bounds.

bgroenks96:
@milankl and @maximilian-gelbrecht I wrote a simple test for differentiability with Enzyme which seems to be working. If everything still passes, would be good to merge this finally, I think?

bgroenks96:
Ok, don't merge this yet actually, I have some pending refactoring changes to the Barotropic differentiability tests, but I need to make sure they pass locally first, which takes a long time...

test/parameters.jl
```diff
@@ -0,0 +1,123 @@
+using ComponentArrays: ComponentVector
+using DomainSets: Domain, RealLine, UnitInterval
+using SpeedyWeather: value, bounds, description, attributes, parameters, reconstruct, stripparams
+
+import ModelParameters: ModelParameters, Model, Param, params, update
```
milankl, line 5: `parameters`, `stripparams`, `Param`, `params` are somewhat inconsistetly named. Happy to keep the current if there's a good reason for it but maybe there isn't?

milankl (commented):
Looks great to me! I'd be happy to merge this without putting too much additional thought or effort into the design. It's already waaaay beyond an initial draft of how parameter handling could be done and seems to have some fantastic potential! I believe the first use cases of this will more easily reveal how to adapt the parameter handling and then this can be updated in another PR!

milankl:
Maybe one last thing: Documentation of the parameter handling would come in another pull request when it's actually implemented for all models? Otherwise feel free to add a section already, probably just after the [differentiability](https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/differentiability/) section? Could be as short as "this is how to retrieve all parameters of a model, how to change them and reconstruct a new model from those" and then "more to come"?

bgroenks96:
> Maybe one last thing: Documentation of the parameter handling would come in another pull request when it's actually implemented for all models?

Yes.

> Could be as short as "this is how to retrieve all parameters of a model, how to change them and reconstruct a new model from those" and then "more to come"?

Sure, but do you think we should export `reconstruct`? I would rather not because it conflicts with Flatten.jl.

I'll try to think of another name...

bgroenks96:
How about `reparam` or `withparams`?

bgroenks96:
> How about reparam or withparams?

How about we worry about this in the next PR and just merge this now?

Edit: 
> Otherwise feel free to add a section already, probably just after the [differentiability](https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/differentiability/) section? Could be as short as "this is how to retrieve all parameters of a model, how to change them and reconstruct a new model from those" and then "more to come"?

Btw @milankl I added a preliminary description to the docs in the differentiability section as you suggested.

milankl:
@torsten20 maybe relevant for what you want ☝🏼

maximilian-gelbrecht:
I am in favour of agreeing on the names of the routines now, but otherwise feel free to merge :)

src/Utils/parameters.jl
maximilian-gelbrecht: @bgroenks96 You resolved this without commentating. So this is a non-issue?

bgroenks96:
> I am in favour of agreeing on the names of the routines now

It's fine, just a find-replace! The main open question is whether or not `reconstruct` (or whatever we call it) should actually be a user-facing API or just something we do internally.

Issue #737, opened 2025-05-27
Code availability for `*.mp4` in Gallery section of README.md?


jfdev001:
Are there any full codes available for the examples (i.e., specific humidity animation in `humid_L15.mp4`, relative vorticity in ` vorticity_clouds_T1023.mp4`) shown in the README.md gallery? I think the code for these examples would be really useful for showing a full model setup pipeline plus visualization! 

The following are examples for which codes should be included (if you'd like to be soft assigned an example, leave a comment and then I will write your user name next to the example of interest): 

- [ ] Specific humidity in primitive equation model (`humid_L15.mp4`), @jfdev001 some work complete (see https://github.com/SpeedyWeather/SpeedyWeather.jl/issues/737#issuecomment-2961973881), not actively working on issue at the moment, so someone else could take over as needed
- [ ] Relative vorticity in shallow water model (`vorticity_clouds_T1023.mp4`)
- [ ] Surface temperature in primitive equation model (`out11.mp4`)
- [ ] 2D barotropic vorticity with stochastic stirring (`stochastic_stirring_T340_ortho.mp4`)
- [ ]  Gravity waves (`random_waves_T255.mp4`)

milankl:
This is a great idea! Unfortunately, both of these videos use older versions of SpeedyWeather, so these videos aren't directly reproducible. I'm pretty sure the humidity video was the default model setup so I would start with that. The vorticity clouds video was the unforced shallow water model with some manually set spectral coefficients as initial conditions that kicked off the turbulence you can see. If those initial conditions are important we can look back in the git history to find them. But they were just pulled out of the heat to produce something pretty. I'm copying in though the plotting script for those that I just found. It's matplotlib+cartopy as I didn't use Makie yet back then for plotting. If you're looking for Makie scripts for visualisation let me know

```python
import numpy as np
import matplotlib.pyplot as plt
import cartopy.crs as ccrs
import netCDF4 as NetCDF
import cartopy.feature as cfeature
from cmcrameri import cm
from matplotlib.colors import ListedColormap
from scipy.special import expit
import os

# for the blue marble background 
os.environ["CARTOPY_USER_BACKGROUNDS"] = "/path/to/blue/marble/background/files/"

# load in data
ncfile = NetCDF.Dataset("run0006/output.nc")
vor = ncfile.variables["vor"]
vor0 = ncfile.variables["vor"][240,0,:,:]
lat = ncfile.variables["lat"][:]
[...]

PR #755, opened 2025-06-27
GeoMakie extension: animate simulation function 

maximilian-gelbrecht:
This puts the function from #737 into the GeoMakie extension. Note that the functions needs a Makie backend loaded to work as well. 

This produces animations like this: 

https://github.com/user-attachments/assets/81f9e814-d95c-4db7-bd01-3085b97e2614

Do we want any more customisation of the plot? Probably? Let me know what I should add. 

I'll write something in the documentation after everything else is finished


ext/SpeedyWeatherGeoMakieExt.jl
```diff
@@ -121,4 +122,180 @@ function SpeedyWeather.globe(
...
+            # Add a small buffer to the range
+            range_buffer = (data_max - data_min) * 0.05
+            colorrange = (data_min - range_buffer, data_max + range_buffer)
+        else
+            colorrange = (-1, 1)  # Fallback if all data is NaN
+        end
+    end
+    
+    # Create the surface plot
+    hm = surface!(ax, lon, lat, data_init; colormap=colormap, colorrange=colorrange)
```
milankl: I think there's a better function here than `surface!` which is meant for 3D plots and therefore has some additional lines that dim the colours of the plot through transparency. Maybe check `meshimage`?

ext/SpeedyWeatherGeoMakieExt.jl
```diff
@@ -121,4 +122,180 @@ function SpeedyWeather.globe(
...
+
+# Animate surface pressure with a specific colormap
+animate_simulation("output.nc", variable="pres", colormap=:thermal)
+
+# Create an animation with Orthographic projection
+animate_simulation("output.nc", variable="temp", projection="+proj=ortho +lon_0=0 +lat_0=30")
+```
+"""
+function SpeedyWeather.animate_simulation(
+    nc_file::String;
```
milankl: Not necessarily for this PR but should we have a lower-level version of this function taking in a 3D ring grid which animates the last dimension? A netCDF array could then be wrapped again into that.

milankl: But for here, shouldn't we allow for a NCDataset as input rather than a string? Just wondering whether there's an advantage of doing the open and close internally?

ext/SpeedyWeatherGeoMakieExt.jl
```diff
@@ -121,4 +122,180 @@ function SpeedyWeather.globe(
...
+    record(fig, output_file, (transient_timesteps + 1):length(time); framerate=framerate) do frame
+        # Update the data
+        if is_3d
+            new_data = ds[variable][:, :, level, frame]
+        else
+            new_data = ds[variable][:, :, frame]
+        end
+        
+        # Update the surface plot
+        hm[3] = new_data
```
milankl: Shouldn't this be done with observables?

milankl:
Also what do you think about adding coastlines and removing tick labels and grid instead?

ext/SpeedyWeatherGeoMakieExt.jl
maximilian-gelbrecht: > Not necessarily for this PR but should we have a lower-level version of this function taking in a 3D ring grid which animates the last dimension? A netCDF array could then be wrapped again into that.

On the other hand, with a netCDF dataset you don't need to load the whole dataset into RAM like this. I can see the point though, maybe something for another PR.

maximilian-gelbrecht: > But for here, shouldn't we allow for a NCDataset as input rather than a string? Just wondering whether there's an advantage of doing the open and close internally?

The advantage is not having to open and close it when you just want a quick animation. But I'll do a the low level version with a NCDataset as an input and then have a convinience version with a string input and Simulation input.

ext/SpeedyWeatherGeoMakieExt.jl
maximilian-gelbrecht: Yes, that's the more proper way

ext/SpeedyWeatherGeoMakieExt.jl
maximilian-gelbrecht: GeoMakie docs mostly use `surface!` for these kind of plots as well. I think `heatmap` and `meshimage` would be alternatives. But I don't have a strong opinion on that. Maybe just make it an option? With `heatmap` and `meshimage` we have to reorder lat/lons and data, I think, which is just a bit annoying... (to -180 - +180 degrees format)

maximilian-gelbrecht:
Ok, added those and a lot more customisation. You can just hand over kwargs to `GeoAxis` and `Colorbar` for things like this now.

ext/SpeedyWeatherGeoMakieExt.jl
```diff
@@ -2,7 +2,8 @@ module SpeedyWeatherGeoMakieExt
 
 using SpeedyWeather
 using GeoMakie
-using DocStringExtensions
+using SpeedyWeather.DocStringExtensions
```
milankl, line 5: Is this general practice btw?

ext/SpeedyWeatherGeoMakieExt.jl
maximilian-gelbrecht: Ok, added that now. Both `heatmap!` and `surface!` can be used. According to the GeoMakie docs what's recommended depends on the backend: https://geo.makie.org/dev/introduction#Gotchas-2

maximilian-gelbrecht:
I am happy with this, from my side this could be merged. But I am also open for more suggestions.

milankl:
I would still make `meshimage!` default, with `surface!`

image

with `meshimage!`

image

maximilian-gelbrecht:
Ok, added `meshimage` as well. All three are now possible.

maximilian-gelbrecht:
Ups, forgot about the documentation, I'll add a short paragraph to the NetCDF output section tomorrow

milankl (approved):

ext/SpeedyWeatherGeoMakieExt.jl
maximilian-gelbrecht: I read somewhere that this helps to avoid circular dependencies. Can't find that reference anymore at the moment.

milankl:
Happy for this to be merged! Or do you want to "test" it by including `animate` in the documentation somewhere actually creating a small video of one of the examples on every docs build?

maximilian-gelbrecht:
Can we run it in the docs and then directly include the generated mp4 file?

milankl:
I think that's how it's done in the Oceananigans docs too!!!

maximilian-gelbrecht:
So, in the CI this seems to have worked. I hope it also does when the docs are actually deployed. I'll press the merge button as soon as the tests are finished.

milankl:
You found this in the Oceananigans docs? https://github.com/CliMA/Oceananigans.jl/blob/73ddc2b93361378ece386ae9c15fd4f0e083a29a/examples/two_dimensional_turbulence.jl#L165-L176

maximilian-gelbrecht:
Ok, it worked https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/output/#Visualizing-output

PR #757, opened 2025-07-03
Adjust particles for GPU

maximilian-gelbrecht:
Still work in progress


milankl:
What about we move `active` from being a type parameter to being a field, i.e.

```julia
struct Particle{
    NF<:AbstractFloat,  # number format of coordinates      
} <: AbstractParticle{NF}
    "indicating whether particle is active (being advected)"
    active::Bool

    "longitude in [0,360˚]"
    lon::NF

    "latitude [-90˚,90˚]"
    lat::NF

    "vertical sigma coordinate [0 (top) to 1 (surface)]"
    σ::NF
end
```

That way a particle is type stable throughout integration. I somehow thought that deciding on the `active` parameter would neatly allow for multiple dispatch to decide what to do with a particle as it's a immutable struct anyway but didn't consider that this means a particle Vector `Vector{Particle{NF}}` would be type unstable! My bad 🙈 

This wouldn't require a `particle_activity` vector but neatly store coordinate and active information next to each other in memory and hopefully would just mean that the `isactive(p::Particle) = p.active` method would change? And probably some `if isactive(particle)` else do nothing lines?

maximilian-gelbrecht:
I think that's already it. 

This isn't a full GPU version yet, but just to make sure everything is type stable and won't cause any problems during model initialisation on GPU. 

For a proper working GPU version, we'd also need to adjust the interpolation.

src/dynamics/particle_advection.jl
```diff
@@ -121,10 +121,7 @@ function particle_advection!(
     clock.timestep_counter % n == (n-1) || return nothing   
 
     # also escape if no particle is active
-    any_active::Bool = false 
-    for particle in particles
-        any_active |= active(particle)
-    end
+    any_active::Bool = any(active.(particles))
```
milankl: Does that not allocate a vector? Or can the compiler take that out as the `any` reduces it back to a bool? That's why I wrote this as a loop but I see that for GPU you don't want that!

src/dynamics/particles.jl
```diff
@@ -85,54 +81,35 @@ function Base.isapprox(
...
-Only active particles are moved."""
-@inline function move(p::Particle{NF,true}, dlon, dlat, dσ) where NF
-    (;lon, lat, σ) = p
-    Particle{NF,true}(lon+dlon,lat+dlat,σ+dσ)
+Move a particle with increments (dlon, dlat, dσ) in those respective coordinates."""
+@inline function move(p::Particle{NF}, dlon, dlat, dσ) where NF
+    if p.active
+        (;lon, lat, σ) = p
+        Particle{NF}(p.active, lon+dlon, lat+dlat, σ+dσ)
+    end
```
milankl: I think move should always return a particle but be the identity if the particle is not active!

src/dynamics/particles.jl
```diff
@@ -85,54 +81,35 @@ function Base.isapprox(
...
-Only active particles are moved."""
-@inline function move(p::Particle{NF,true}, dlon, dlat) where NF
-    (;lon, lat, σ) = p
-    Particle{NF,true}(lon+dlon,lat+dlat,σ)
+Move a particle with increments (dlon, dlat) in 2D. No movement in vertical σ."""
+@inline function move(p::Particle{NF}, dlon, dlat) where NF
+    if p.active
+        (;lon, lat, σ) = p
+        Particle{NF}(p.active, lon+dlon, lat+dlat, σ)
+    end
```
milankl: Same here!

src/dynamics/particles.jl
```diff
@@ -155,4 +132,18 @@ function ismod(p::Particle)
     valid &= 0 <= p.lon < 360       # 360˚E excluded (=0˚)
     valid &= 0 <= p.σ <= 1          # top and ground included
     return valid
-end
\ No newline at end of file
+end
+
+export activate, deactivate, active
```
milankl: Wondering whether `isactive` would be a better name than `active`?

milankl (commented):
Looks great! Thanks Max!!

src/dynamics/particle_advection.jl
maximilian-gelbrecht: Yeah, it's the easiest way to already make this gpu compatible.

src/dynamics/particles.jl
maximilian-gelbrecht: ah yes, good call

maximilian-gelbrecht:
Ok, finally all tests and docs adjusted. When the tests are all finished I'd merge this.

maximilian-gelbrecht:
I think the last enzyme update broke something... we also get some warning we didn't get before. 

This has nothing to do with this PR

PR #759, opened 2025-07-04
Move full differentiability tests outside of regular tests

maximilian-gelbrecht:
@milankl 

This moves the full differentiability tests outside of the regular tests into a separate workflow. This way, also if users just do the `test SpeedyWeather` they don't get this very long test. Currently this is still triggered on every commit, but we might reconsider this or temporarily deactivate it when we get problems.


milankl:
Happy with that!

PR #762, opened 2025-07-08
Fix Enzyme CI 

maximilian-gelbrecht:
I made a stupid mistake in #759, the Enzyme CI just ran the regular tests, instead of the Enzyme tests. 

This PR fixes that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix Enzyme CI to run as expected [#762](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/762)
+
 ## v0.16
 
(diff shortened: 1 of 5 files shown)


maximilian-gelbrecht:
While I have this PR open, I'll also take a look at the extended test set tomorrow before merging this.

maximilian-gelbrecht:
Okay, also some problems with the regular Enzyme test as well. I'll use this PR to work on that as well

maximilian-gelbrecht:
ok, there still remains some issue with the extended tests at the moment (two fail at the moment). I am going to merge this PR regardless because it fixes the error in the CI workflow and adjusts syntax errors in the extended tests. 

@bgroenks96 let's have a chat about how to change the extended differentiability tests in the next days.

PR #763, opened 2025-07-09
Overhaul UV_from_vordiv!

### LINKED ISSUE #715: Zonal modes instability in shallow water
Setup by @pierluigividale

```julia
using SpeedyWeather
spectral_grid = SpectralGrid(trunc=127, nlayers=1)

implicit = ImplicitShallowWater(spectral_grid, α=0.5)
orography = NoOrography(spectral_grid)
initial_conditions = StartFromRest()
model = ShallowWaterModel(spectral_grid; orography, initial_conditions, implicit)

div(λ, θ, σ) = min(-1e-4*exp(-((λ-150)^2+θ^2)/(3^2)), 0)
set!(simulation, div=div)

run!(simulation, period=Hour(2))

using CairoMakie
heatmap(simulation.diagnostic_variables.grid.pres_grid)
```

yields

Image

The yellow ring is expected, the horizontal stripes excited as zonal modes in the spherical harmonics originating from the poles are not. With a 6x smaller time step the dispersion is different (expected) but the problem is not resolved

Image

milankl:
first attempts to fix #715

### DIFF
--- a/src/SpeedyTransforms/SpeedyTransforms.jl
+++ b/src/SpeedyTransforms/SpeedyTransforms.jl
@@ -13,10 +13,14 @@ import Adapt: adapt
 import KernelAbstractions: @kernel, @index, @Const, synchronize
 
 # SPEEDYWEATHER MODULES
-using ..Architectures
-using ..Utils
-using ..LowerTriangularArrays
-using ..RingGrids
+include("../Utils/Utils.jl")
+using .Utils
+include("../Architectures.jl")
+using .Architectures
+include("../LowerTriangularArrays/LowerTriangularArrays.jl")
+using .LowerTriangularArrays
+include("../RingGrids/RingGrids.jl")
+using .RingGrids
 
 # TRANSFORM
 export  SpectralTransform,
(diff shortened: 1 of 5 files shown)


milankl:
I set a `unscale_coslat=false` in the transfrom from spectral $V$ to grid space $v$. I don't think this is actually the solution but it clearly already looks much better

https://github.com/user-attachments/assets/54acc661-52ef-4064-a014-4ff572e714ce

maximilian-gelbrecht:
In #727 I left out `UV_from_vordiv!` and didn't write a GPU kernel yet because I had a problem with the kernel version and Enzyme that I couldn't figure out at the time. For #792 I really need a kernel version of ` UV_from_vordiv!` now. Do you plan any work on the CPU version of `UV_from_vordiv!` here soon? Then I don't have to adapt ` UV_from_vordiv!` twice. 

~~Or will the actual `UV_from_vordiv!` not change at all?~~ I actually also have to change some of the precomputations, so no matter what, an extra adjustment is needed.

milankl:
Partially superseded by #952

PR #736, opened 2025-05-22
Add KA kernels for spectral transform legendre methods

jackleland:
This builds off of @maximilian-gelbrecht's PR overhauling GPU implementation, switching the existing CUDA kernels for the legendre portion of the spectral transform over to the new KA framework. I had to add Atomix as an explicit dependency so we can use it for atomic operations in the inverse kernel, but I think it was an implicit dependency beforehand anyhow. 

Timing-wise the KA vs. CUDA kernels seem pretty similar, with KA coming out slightly faster on GPU and slower on CPU. The only stand-out was the forward KA kernel on CPU is massively slower than the non-KA algorithm (150ms vs. 2.5s for T255x64), so we probably want to look at maintaining the non-KA version going forward as it will be a bottleneck. 

Still to do:
- [ ] Add proper benchmarking
- [ ] Tidy up and merge these into `src/SpectralTransforms/legendre.jl`

Issue #648, opened 2025-01-06
Grid cell average via overlapping polygons


milankl:
In #644 the actual underlying problem is that `RingGrids.grid_cell_average!` assumes that the grid to interpolate from is much higher resolution than the grid to interpolate to. Note that this is independent of our `AnvilInterpolator` which is a bilinear interpolator from a grid to any given coordinate. However, for the land-sea mask we need an interpolation which takes an average of all grid cells of a higher resolution grid into a lower resolution grid. Especially because the high-resolution land-sea mask is binary whereas SpeedyWeather uses a fractional land-sea mask such that we need to know how high the fraction of land points is within one (coarse-resolution) grid cell. So instead of patching the current algorithm I suggest to actually calculate the polygon overlap, which is the right way of doing it.

@asinghvi17 suggested to use GeometryOps.jl via `GO.intersection(GO.GEOS(), geom1, geom2)`.

milankl:
Thanks to @asinghvi17 I drafted a solution

```julia
using Shapefile, SpeedyWeather
using GeometryBasics, GeoInterface, GeometryOps
import LibGEOS

# load Natural Earth land polygon from file
table = Shapefile.Table("Downloads/ne_110m_land/ne_110m_land.shp")
geoms = Shapefile.shapes(table)

# convert from 0 to 360 to -180 to 180
mod180(x) = -180 <= x < 180 ? x : mod180(x-sign(x)*360)
mod180(p::P) where P<:Point = P(mod180(p.data[1]), p.data[2])

function land_sea_mask(Grid, nlat_half, land)
    
    e, s, w, n = RingGrids.get_vertices(Grid, nlat_half)

    mask = zeros(Grid, nlat_half)

    for ij in eachindex(mask)
    
        faces = [   mod180(Point2(e[:, ij])),
                    mod180(Point2(s[:, ij])),
                    mod180(Point2(w[:, ij])),
                    mod180(Point2(n[:, ij])),
                    mod180(Point2(e[:, ij]))
                ]

        cell = GeoInterface.Polygon([GeoInterface.LinearRing(faces)])

        mask[ij] = fraction(cell, land)
    end

    return mask
end

function fraction(polygon, coastlines)
    A = GeometryOps.area(GeometryOps.intersection(GeometryOps.GEOS(), polygon, polygon))
    f = 0
    for shape in coastlines
[...]

PR #764, opened 2025-07-10
Interpolation on GPU 

maximilian-gelbrecht:
Work in progress, just started. 

Based on the `mk/leapfrogkernel` branch.


maximilian-gelbrecht:
Okay, I want to keep the PRs rather small, when it's possible to isolate specific things. 

So this PR will be mostly about the interpolation. Then in the next PRs I'll work on the actual primitive dry model. 

I finished drafting a rough version, but I have to properly check it and benchmark it next week. No review needed yet, here.

maximilian-gelbrecht:
@milankl After looking at the code a bit, I don't like how we handle the number format precision currently. I just find a bit messy to allow different precisions in input, output, and computations here. 

Would be fine if I simplify this to all be in one precision? I think this needs to be done for GPU anyway. 
What we could still do is as to convert to certain precision in the end after the interpolation is done.

milankl:
> @milankl After looking at the code a bit, I don't like how we handle the number format precision currently. I just find a bit messy to allow different precisions in input, output, and computations here.
> 
> Would be fine if I simplify this to all be in one precision? I think this needs to be done for GPU anyway. What we could still do is as to convert to certain precision in the end after the interpolation is done.

Do you really think this is a problem? I see that the interpolator and the input could have the same precision but the output array surely can just be written into with another precision converting on the fly? But also it's the same as with the `SpectralTransform` though that one, because of FFTW only, has to have the same precision as the grid space field (whether input or output). It's just against our flexibility philosophy if you suddenly need to recompute a new interpolator because you want to mix up some precision, whether in the dycore or interactively

maximilian-gelbrecht:
I think just outputting on a different NF (at least by converting after the interpolation) is fine. I was mainly a bit concerned about what we state in the docstrings: 

> The number format NF is the format used for the calculations within the interpolation, the input data
and/or output data formats may differ."""

I really don't want to allow all three to be different. That's also why the tests here are failing currently. It's some precision issue after the initial changes I had to make for the GPU version. 

After some fixes, the GPU version is now working and yielding correct results. I just need to decide after a little bit more benchmarks for the `update_locater!`,  and related functions to decide if just keeping the one KA version is sufficient. I also noticed we don't actually have unit tests for those.

maximilian-gelbrecht:
Ok, I did the benchmarks. On CPU vs CPU KA, `interpolate!` has the same performance. `update_locater!` is roughly 40% slower in CPU KA than in CPU old. But in accordance with our general rule for the KA translation that's still acceptable, so I'll keep only the KA version. 

I am off work for a few days. In this PR there's just a little bit of clean up to do. In general, I want to merge #733 first (then the code diff here also becomes much smaller for a proper review) 

I didn't do a GPU version of `grid_cell_average!`

maximilian-gelbrecht:
In terms of general GPU strategy here. 

All of the `get_lon`, `get_lat`, etc for RingGrids are hard coded to return CPU Arrays at the moment. I considered changing that, but in the end I didn't do it. I didn't seem worth the effort at this point for me. So, here in the interpolation we just transfer the result of `get_londslatds` when needed to GPU.

milankl:
> In terms of general GPU strategy here.
> 
> All of the `get_lon`, `get_lat`, etc for RingGrids are hard coded to return CPU Arrays at the moment. I considered changing that, but in the end I didn't do it. I didn't seem worth the effort at this point for me. So, here in the interpolation we just transfer the result of `get_londslatds` when needed to GPU.

Yes I think that's fine, in the end `get_lon` etc are based on the idea of recomputation: So instead of having a field in grid that says `grid.lon` which should live on the GPU we would just recompute that array when needed which in practice is only interactively. Otherwise `Geometry` or `GridGeometry` will contain those on the respective device.

maximilian-gelbrecht:
I think this is also ready in principal, but #733 needs to be merged first, then the code diff here also becomes much smaller for a proper review

maximilian-gelbrecht:
Some tests occasionally fail since I don't use Float64 anymore in every interpolation. We could just give a slightly larger tolerance in the tests though.

maximilian-gelbrecht:
@milankl Could you do a review here some time this week? 

Then, afterwards I can also merge #736 without further reviews necessary, as I already reviewed Jack's code there anyway.

src/RingGrids/RingGrids.jl
```diff
@@ -107,7 +110,7 @@ export  interpolate,
 export zonal_mean
 
 # CONSTANTS
-const DEFAULT_NF = Float64
+const DEFAULT_NF = Float32
```
milankl, line 113: Yeah I was wondering that! Initially I thought it would be good to have the base libraries like RingGrids, LTAs ... use Float64 as a default and only SpeedyWeather is Float32 by default but I guess that causes some inconsistencies?

src/RingGrids/field.jl
```diff
@@ -65,6 +66,18 @@ function Base.array_summary(io::IO, field::AbstractField, inds::Tuple{Vararg{Bas
...
+    Base.array_summary(io, field, axes(field))
+
+    if get(io, :limit, false)::Bool && displaysize(io)[1]-4 <= 0
+        return print(io, " …")
+    else
+        println(io)
+    end
+    
+    Base.print_array(io, field.data)
+end
```
milankl, line 79: What does this do? Cause I still get
```julia
julia> field
3168×2, 48-ring OctahedralGaussianField{Float64, 2} on Array on SpeedyWeather.Architectures.CPU{KernelAbstractions.CPU}:
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 0.0  0.0
 ⋮
```

src/RingGrids/RingGrids.jl
```diff
@@ -12,9 +12,12 @@ export rotate, rotate!
...
 import Adapt: Adapt, adapt, adapt_structure
 import GPUArrays
-import KernelAbstractions
-import ..Architectures: Architectures, AbstractArchitecture, on_architecture, array_type, 
-    ismatching, CPU, GPU, architecture
+import KernelAbstractions: KernelAbstractions, @kernel, @index, @Const, synchronize
+
+# SPEEDYWEATHER SUBMODULES
+import ..Architectures: Architectures, AbstractArchitecture, CPU, GPU, 
+        on_architecture, architecture, array_type, ismatching
```
milankl: should we also move `nonparametric_type` into Architectures? It's currently defined in both RingGrids/LowerTriangularArrays independently just because there's no one package from us that both depend on

src/SpeedyTransforms/spectral_transform.jl
```diff
@@ -1,6 +1,6 @@
 const DEFAULT_NLAYERS = 1
 const DEFAULT_GRID = FullGaussianGrid
-const DEFAULT_NF = Float64
+const DEFAULT_NF = Float32
```
milankl, line 3: Same here

src/Utils/kernel_launching.jl
```diff
@@ -40,6 +40,8 @@ function work_layout(dims_type::Symbol, worksize::NTuple{N, Int}) where N
         return heuristic_workgroup(worksize[2], worksize[3:end]...), (worksize[2], worksize[3:end]...)
     elseif dims_type == :array_3d # kernel over a regular 3D array
         return heuristic_workgroup(worksize...), worksize
+    elseif dims_type == :linear # kernel over eachindex / linear
+        return heuristic_workgroup(prod(worksize)), (prod(worksize),)
     else
```
milankl: is `if else` the best structure here? It feels like something that could be more nicely addressed and updated using multiple dispatch? I see that we currently define `dims_type::Symbol` but what about

```julia
abstract type AbstractWorkOrder end

# previously :linear
struct LInearWorkOrder <: AbstractWorkOrder end
work_layout(::LinearWorkOrder, worksize::NTuple{N, Int}) where N = heuristic_workgroup(prod(worksize)), (prod(worksize),)

# previously :lmk
struct SpectralWorkOrder <: AbstractWorkOrder end
work_layout(::SpectralWorkOrder, worksize::NTuple{N, Int}) where N = heuristic_workgroup(worksize...)
```

etc?

src/physics/land_sea_mask.jl
```diff
@@ -122,8 +122,14 @@ function initialize!(land_sea_mask::EarthLandSeaMask, model::PrimitiveEquation)
...
-    # average onto grid cells of the model
-    RingGrids.grid_cell_average!(land_sea_mask.mask, lsm_highres)
+    # average onto grid cells of the model 
+    if typeof(architecture(model.spectral_grid.architecture)) <: CPU
+        RingGrids.grid_cell_average!(land_sea_mask.mask, lsm_highres)
+    else
+        temp_mask = on_architecture(CPU(), land_sea_mask.mask)
+        RingGrids.grid_cell_average!(temp_mask, lsm_highres)
+        land_sea_mask.mask .= on_architecture(model.spectral_grid.architecture, temp_mask)
+    end
```
milankl, line 132: Happy with that for now. Longterm idea is anyway to compute this via ConservativeRegridding using the N largest polynomials of coastlines see https://github.com/SpeedyWeather/SpeedyWeather.jl/issues/648#issuecomment-2574141215 which would make the somewhat hacky `grid_cell_average!` function obsolete

src/RingGrids/interpolation.jl
```diff
@@ -50,15 +52,14 @@ function GridGeometry(
...
     nlons = get_nlons(grid)                                 # number of longitude per ring, pole to pole
-    lon_offsets = [londs[ring[1]] for ring in grid.rings]   # offset of the first point from 0˚E
+    lon_offsets = [londs[ring[1]] for ring in eachring(grid)]   # offset of the first point from 0˚E
 
     # vector type
-    ArrayType_ = nonparametric_type(ArrayType)
-    VectorType = ArrayType_{NF, 1}
-    VectorIntType = ArrayType_{Int, 1}
+    VectorType = array_type(architecture, NF, 1)
+    VectorIntType = array_type(architecture, Int, 1)
```
milankl, line 59: So we do enforce one array type per architecture now? I thought the initial plan was to allow there some flexibility? I'm easy though!

src/RingGrids/interpolation.jl
```diff
@@ -77,19 +78,19 @@ between two latitude rings."""
     VectorIntType,
 } <: AbstractLocator
 
-    npoints::Int            # number of points to interpolate onto (length of following vectors)
+    npoints_output::Int            # number of points to interpolate onto (length of following vectors)
```
milankl, line 81: 👍🏼 Yeah, I also found that confusing at times.

src/RingGrids/interpolation.jl
```diff
@@ -359,69 +388,112 @@ function update_locator!(
...
+        if θ <= latd[j_mid] && (j_mid == length(latd) - 1 || θ > latd[j_mid + 1])
+            # Found the correct ring
+            j = j_mid
+            break
+        elseif θ > latd[j_mid]
+            j_high = j_mid - 1
+        else
+            j_low = j_mid + 1
+        end
+    end
```
milankl: Very nice! I was lazy back then and implemented only linear search but this is obv much better. Would you mind adding some more comments here or provide a reference?

src/RingGrids/interpolation.jl
```diff
@@ -493,19 +598,21 @@ $(TYPEDSIGNATURES)
...
 on the northern-most ring. Similar for the South pole."""
-function average_on_poles(A::AbstractVector{NF}, rings) where {NF<:AbstractFloat}   
-    A_northpole = mean(view(A, rings[1]))     # average of all grid points around the north pole
-    A_southpole = mean(view(A, rings[end]))   # same for south pole
-    return convert(NF, A_northpole), convert(NF, A_southpole)
+@inline function average_on_poles(A::AbstractVector{NF}, rings) where {NF<:AbstractFloat} 
+    # TODO: doing the computation below with views causes scalarindexing on GPUs
+    A_northpole = mean(A[rings[1]])     # average of all grid points around the north pole
+    A_southpole = mean(A[rings[end]])   # same for south pole
+    return A_northpole, A_southpole
```
milankl: How do you want to solve that?

src/RingGrids/interpolation.jl
```diff
@@ -99,24 +100,23 @@ update the grid indices used for interpolation and their weights. The number for
...
+    NF::Type{<:AbstractFloat},                                 # number format
+    npoints::Integer; 
+    architecture::AbstractArchitecture = DEFAULT_ARCHITECTURE(), # architecture to use
 ) where {L<:AbstractLocator}
 
-    ArrayType_ = nonparametric_type(ArrayType)
-    VectorType = ArrayType_{NF, 1}
-    VectorIntType = ArrayType_{Int, 1}
+    VectorType = array_type(architecture, NF, 1)
+    VectorIntType = array_type(architecture, Int, 1)
```
milankl, line 109: So you did retain the idea that the interpolator can have a different number format to the input/output fields, nice!

milankl (commented):
very nice, I like! Doesn't look like there's much a breaking change here (if at all?) and you even improved the linear search to binary search!

src/RingGrids/field.jl
maximilian-gelbrecht: Without you'd get a scalar indexing error on GPU

maximilian-gelbrecht: It calls the `print_array` defined for GPU arrays in this case in line 78 (which actually just explicitly transfers the array to CPU before printing). Otherwise you'd call a generic abstract array `show` that uses scalarindexing directly.

src/RingGrids/RingGrids.jl
maximilian-gelbrecht: So far we didn't get any inconsistencies that I know of, but I am always a bit worried about mixing precisions, especially when moving to GPU.

src/RingGrids/RingGrids.jl
maximilian-gelbrecht: yeah, good idea. 

That being said: if we really want to be thorough here, we could also go through the code and revise the use of `nonparametric_type` a bit: since we now have this `array_type(arch, NF, N)` we also need the  `nonparametric_type` much less.

src/Utils/kernel_launching.jl
maximilian-gelbrecht: Yeah good idea. Initially I did it like this just to keep the analogy to Oceanigans, but using the multiple dispatch seems the more Julia way to actually do that.

src/RingGrids/interpolation.jl
maximilian-gelbrecht: I still think of `array_type` as the default one. We do also have `compatible_array_types` now. 

Both of these are needed because of how arrays on GPU are handled: 

1) `CuArray{NF, N}` isn't a concrete type, it's actually `CuArray{NF, N, CUDA.DeviceMemory}`, so `array_type` now makes sure that really the correct concrete array type is chosen: `Array{NF,N}` on CPU and `CuArray{NF, N, CUDA.DeviceMemory}` on CUDA GPU. 
2) GPUs have hosts and device arrays (`CuArray` is the host array), so we need to allow both to be compatible with the architecture, that's what `compatible_array_types` is for.

src/RingGrids/interpolation.jl
maximilian-gelbrecht: This confusion definitely cost me a two hours debugging my code 😅

src/RingGrids/interpolation.jl
maximilian-gelbrecht: To be transparent, Claude did that completely automatically when I asked for a GPU version of the kernel. But as binary search has a complexity of O(log n), instead of O(n), and our `latd` are already in order, I think this is certainly the preferred search algorithm here.

src/RingGrids/interpolation.jl
maximilian-gelbrecht: I mean this is already a solution, isn't it? Because before it was `mean(view(A, rings[1]))` which didn't work. The allocation isn't ideal, but it also isn't too bad I think.

src/RingGrids/interpolation.jl
maximilian-gelbrecht: Yes, I ended up not really changing that.

src/physics/land_sea_mask.jl
maximilian-gelbrecht: Yes, also, even in the longterm, we'd always need some kind of statements like that when dealing with input from files, because I/O is always CPU.

src/RingGrids/RingGrids.jl
maximilian-gelbrecht: Okay, done that now.

src/Utils/kernel_launching.jl
maximilian-gelbrecht: Ok, done that too!

maximilian-gelbrecht:
@milankl Are you okay with just increasing the tolerance in the tests below a bit, or is this rather a sign that we do actually need a higher precision interpolation? They are never failing with `Float64`, but occasionally for `Float32`. The change to before is that before the internal precision in the interpolator was always `Float64`, now by default it takes the precision from the input array. 

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/60deed7ff9a77db8666c2cd167c12ebd154aed4a/test/grids/interpolation.jl#L31-68

src/RingGrids/interpolation.jl
maximilian-gelbrecht: However, I did change that we do take the precision of the input array for the interpolator , instead of the default (used to be Float64, now Float32). See my comment/question below.

src/RingGrids/interpolation.jl
milankl: ~~we could make it a~~ `field_view` ? Ah no we can't. Allocation isn't the worst but we should at least flag it otherwise we'll have a hard time identifying where the allocation is from later? Just add `# TODO this allocations memory`?

src/RingGrids/interpolation.jl
milankl: Yes I agree and also it seems that it's quite standard so happy to not provide a reference but feel free to add some comments?

src/RingGrids/RingGrids.jl
milankl: Given some tests fail/need additional tolerance with Float32 I'm wondering whether we can still make the default Float64?

src/RingGrids/interpolation.jl
milankl: yeah maybe that's just the most consistent one, if one does `A\b` with Float32 arrays you also expect the LU decomposition to work in Float32 in the same way we should probably treat the interpolation: If you throw Float32 in you also get computations in Float32, if you want higher precision just for that you have to specify it

src/RingGrids/interpolation.jl
maximilian-gelbrecht: I added some comments

maximilian-gelbrecht:
Ok, I added a `rtol=1e-4` to the interpolation of the meridionally constant field test, and with this it should always pass with f32 as well. 

I think, otherwise everything's done here. If there's nothing else, I'd merge it. 

I'll also merge #736 quickly after that, if you don't mind. I don't think it needs another review necessarily, it's already my review and revision of Jack's code anyway.

milankl:
Top go for it 🎉

PR #767, opened 2025-07-16
Forward Rules for _fourier! 

maximilian-gelbrecht:
Forward diff rules defined as discussed with @bgroenks96 for Enzyme Forward mode for possible extra correctness checks. Currently there's still some problem here, that I have to figure out next week.


maximilian-gelbrecht:
Currently `Duplicated` activity is needed for the `SpectralTransform`. That's a bit odd, because it's not needed for the reverse mode. 

One could look into expanding the `ScratchMemory` to really include all temporary storage, however, as we use the forward mode so far only to cross-check with the reverse mode, I would make this a very low priority for the moment.

maximilian-gelbrecht:
Huh, the remaining temporary storage in `SpectralTransform`, might be a bigger problem than I initially thought for the forward mode. This needs a bit deeper look.

bgroenks96:
Why is that not a problem for reverse-mode? Don't the gradients also need to be tracked through the transform in this case? Or do the custom rules bypass the need for this?

maximilian-gelbrecht:
No idea, it also seems a bit strange to me.

maximilian-gelbrecht:
I made a separate PR now to address these issues #791

maximilian-gelbrecht:
Even with #791 merged, I still get an error from Enzyme complaining about temporary memory. That's really strange, because now there really isn't any temporary memory anymore in the `SpectralTransform`. I have to continue to investigate this.

maximilian-gelbrecht:
Claude to the rescue, this seems to work now (and is also needed for our sensitivity project, so it better should)

PR #782, opened 2025-07-17
Overwrite output folder option

### LINKED ISSUE #775: Allow output to overwrite existing runs
When an id is defined in NetCDFOutput (e.g. `output_writer = NetCDFOutput(spectral_grid, id = "control")`), the first time the experiment is run it will write the output in `run_id`, but the second time the same experiment is run, it will write in `run_0001` because `run_id` exists. While this behaviour is understandable to avoid data loss, I have two alternative suggestions to allow for re-running the same experiment while keeping track of the id we defined : 
1. Allow for overwrite with an argument like `overwrite = true`
2. Instead, name the first time `run_id_0001` so that if it is re-run, it can be called `run_id_0002`. This allows for not losing data due to clumsiness, while keeping the information of the id.

milankl:
fixes #775 

A new logic is implemented to determine the name of the run folder, i.e. the `run_0001` in `path/run_0001/output.nc`

1. We now distinguish between prefix (default `run`) an identification "id" (default "") and a number (default `1`), concenated with `_` they yield the `run_folder`
2. If `model.output.overwrite = true` (`false` by default) then prefix, id, number are used as is overwriting an existing run folder if it already exists
3. Otherwise we set the run number to 1, check whether the run folder already exists and count up until it doesn't

Some examples

- default remains `run_0001`, then `run_0002` and so on
- if you have `run_0001`, `run_0002`, `run_0003` and then delete `run_0002` then this logic would fill that gap with the next output
- if you change the `id` then the counter restarts from 1 to determine the number
- if you want the run folder to be called `best_simulation_ever_0123` then set manually `output.run_prefix = "best_simulation"; output.id = "ever"; output.run_number=123; output.overwrite = true`. Without the overwrite the counter would start again from 1 to determine the lowest non-existing run folder.

Also
- renames `nar` (NaR, Not-a-Real meaning both NaN & Inf) simply to `nan` as this caused confusion with users. We are still testing for `~isfinite(...)` so technically we are testing for NaR not NaN but in practice this is irrelavant as Infs quickly become NaN when they hit the first subtraction or division


milankl:
@stella-bourdin would you be happy with this logic? ☝🏼

milankl:
@maximilian-gelbrecht why did the Enzyme tests run here?

maximilian-gelbrecht:
Why not? As I wrote before, they run on every commit just split in another workflow. Of course it's possible to restrict CI workflows to certain branches, but we constantly create new branches. It would be super tedious to constantly add and remove new branches from the Enzyme CI workflow.

PR #783, opened 2025-07-17
Add @param markers for all other model components

### LINKED ISSUE #760: Implement new parameter handling system for ShallowWater and Primitive models
This is just an extension of #753 and #754 to both other model types.

Tasks
- [ ] `ShallowWaterModel` and all components therein
- [ ] `PrimitiveDryModel` and all components therein
- [ ] `PrimitiveWetModel` and all components therein

bgroenks96:
Applies `@parameterized` to all the ShallowWater, PrimitiveDry, and PrimitiveDry models as well as all subcomponents.

Also introduces the `@component` marker which functions the same as `@param` but automatically adds an attribute `:component` matching the field name. This makes it easier to declare parameterized subcomponents of a model.

Needs tests and docs. @milankl please review the choices of which fields to make parameters and also bounds!

Resolves #760


bgroenks96:
@milankl I will try to look at this again next week, but it would be good if you could provide feedback first?

bgroenks96:
Should I just close this PR (and start over)? Seems like it might be a lost cause.

milankl:
Ah shoot, yes, sorry! 🙈 Actually while there are conflicts I don't think they are in terms of `struct` definitions ... I can see whether I can reanimate this PR!

bgroenks96:
Ok, I'm going to steal some of the changes for #879 so that will also conflict.

milankl:
superseded by #935

PR #784, opened 2025-07-18
Restart from file for ocean and land

### LINKED ISSUE #774: Need to include Land Surface and Ocean states in StartFromFile (re-starts)
This is just a reminder to check whether these prognostics are read when starting from a previous run (using initial_conditions = StartFromFile)

- Soil Temperature
- Soil Moisture
- Snow Cover (when implemented)
- SST
- Sea Ice (when implemented)

milankl:
fixes #774 by moving the initial conditions call after initialize! land and ocean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Restart from file for ocean and land [#784](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/784)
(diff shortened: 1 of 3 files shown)


milankl:
Following the test from #774 I get

```julia
julia> simulation.prognostic_variables.ocean.sea_surface_temperature[1]
1231.0088f0
```

So this seems to solve the problem!

PR #788, opened 2025-07-21
Soil moisture field capacity

### LINKED ISSUE #771: Soil Moisture Availability going up to 2 instead of being capped at 1
This code in vegetation.jl reflects what is on the documentation pages:

The available of soil water over land is represented by
(over the ocean ``\alpha_{sw} = 1``)

```math
\alpha_{sw} = \frac{D_{top} W_{top} + f_{veg} D_{root} \max(0, W_{root} - W_{wil})}{
    D_{top}W_{cap} + D_{root}(W_{cap} - W_{wil})}
```

following the Fortran SPEEDY documentation[^SPEEDY] which follows Viterbo and Beljiars 1995
[^Viterbo95]. The variables (or spatially prescribed arrays) are water content in the top
soil layer ``W_{top}`` and the root layer below ``W_{root}`` using the vegetation
fraction ``f_{veg} = veg_{high} + 0.8 veg_{low}`` composed of a (dimensionless)
high and low vegetation cover per grid cell ``veg_{high}, veg_{low}``.
The constants are depth of top soil layer ``D_{top} = 7~cm``, depth of root layer
``D_{root} = 21~cm``, soil wetness at field capacity (volume fraction)
``W_{cap} = 0.3``, and soil wetness at wilting point (volume fraction) ``W_{wil} = 0.17``.

From inspection, the value of should be capped at 1, as low_cover and high_cover only go from 0 to 1, and their sum is always capped at 1.

However, SMA asymptotes to a value of 2, see the time series plot below instead, where, for this shorter run, it asymptotes to 1.75. In the long run, it goes all the way to 2.

    # precalculate
    r = 1/(D_top*W_cap + D_root*(W_cap - W_wilt))

    for ij in eachgridpoint(soil_moisture_availability, high_cover, low_cover)
        
        # Fortran SPEEDY source/land_model.f90 line 111 origin unclear
        veg = max(0, high_cover[ij] + low_veg_factor*low_cover[ij])

        # Fortran SPEEDY documentation eq. 51
        soil_moisture_availability[ij] = r*(D_top*soil_moisture[ij, 1] +
            veg*D_root*max(soil_moisture[ij, 2] - W_wilt, 0))
    end

Image

milankl:
Fixes #771 

@pierluigividale can you review this? Just invited you as collaborator. Maybe convenient: If you press `.` then `github.dev` automatically opens and you can add commits to the pull request directly without cloning/pulling anything!


milankl:
Just added and updated some docstrings for clarification I hope this is correct now?

milankl:
@pierluigividale I guess the question here is aren't `field_capacity` and `W_cap` supposed to be the same thing here?

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/ac17f39280017b7b1305cc49503e6b8dc5d47473/src/physics/land/thermodynamics.jl#L8-L18

pierluigividale:
I am confused wrt the units of field_capacity: per m soil. but what unit per m? 
I think that it is m/m, so unitless.

Unless that number, 0.24 is percentage of saturation?

Then the next one is w_cap and it is declared as m3/m3, so also unitless.

pierluigividale (commented):
Looking at the equations, I think that soil.field_capacity is m/m2 (a weird unit), then soil.f1 and soil.f2 are volumetric soil moisture content, in m/m (or m3/m3). 
This agrees with the SPEEDY documentation and is customary. By also looking for consistency with the fluxes, e.g. P-E, this seems correct. 
Rather than going through the strange conversion from soil.fiedl_capacity to W_cap, I would just use the value of W_cap that was set initially, which comes from Table B.10 of the SPEEDY documentation. In that documentation, W_cap is 0.3 and W_wilt is 0.17. Both are volume fraction, so m3/m3

pierluigividale:
OK, I have now read the MIT GCM documentation and I see that we have indeed a deadly mix of equations and coefficients. Because of the way they define field capacity, the definition of soil moisture is also something else entirely, that is, instead of being the volumetric fraction, it is the fraction of field capacity. This also affects the definition of runoff. I need to write out the equations by hand, put in the units, and make sure that it is all consistent. What I said last night was based on the SPEEDY documentation, but the equations used here are different.

PR #791, opened 2025-07-23
More ScratchMemory handling for SpectralTransforms 

maximilian-gelbrecht:
Still WIP, this is to fix an issue in #767

### DIFF
--- a/SpeedyTransforms/Project.toml
+++ b/SpeedyTransforms/Project.toml
@@ -1,7 +1,7 @@
 name = "SpeedyTransforms"
 uuid = "ed023a3f-7a3c-42dd-bab6-68c21c8c3105"
 authors = ["SpeedyTransforms contributors"]
-version = "0.1.0"
+version = "0.1.1"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 # Unreleased
 
 - Expand scratch memory handling in spectral transforms for better differentiability with Enzyme and revise FFT plans [#791](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/791)
+- Increase version of RingGrids, LowerTriangularArrays and SpeedyTransforms to v0.1.1 [#875](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/875)
 - Pluto notebook example in README [#872](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/872/)
 - Reimplement RestartFile, ProgressTxt, ParamatersTxt as callback [#726](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/726/)
 - Implicit has reinitialize option [#726](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/726/)
--- a/LowerTriangularArrays/Project.toml
+++ b/LowerTriangularArrays/Project.toml
@@ -1,7 +1,7 @@
 name = "LowerTriangularArrays"
 uuid = "ed20b33f-a660-45d0-a353-9c8c20bb5814"
 authors = ["LowerTriangularArrays contributors"]
-version = "0.1.0"
+version = "0.1.1"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/RingGrids/Project.toml
+++ b/RingGrids/Project.toml
@@ -1,7 +1,7 @@
 name = "RingGrids"
 uuid = "d1845624-ad4f-453b-8ff4-a8db365bf3a7"
 authors = ["RingGrids contributors"]
-version = "0.1.0"
+version = "0.1.1"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"


maximilian-gelbrecht:
The basics of this seem to work now, but this has to be merged after #764 and #733

bgroenks96:
I take it that a lot of these commits are from those unrelated PRs?

maximilian-gelbrecht:
I mean I wouldn't say that they are unrelated, as I needed some of the functions from the GPU PRs, but yes, most commits are from #764 and #733.

bgroenks96:
Yeah sorry, I guess I just meant the *other* PRs

maximilian-gelbrecht:
I would say we merge one after the other the next few days, and then we can have a look at the forward diff tests next week. I hope this is enough to make the forward mode work.

maximilian-gelbrecht:
Reminder to myself: this needs another adjustment that I forgot to actually use the correct scratch memory in the diagnostic variables in all occasions.

maximilian-gelbrecht:
There are now problems with the correctness with the rule that I didn't expect (see CI logs). Need to look into this further.

bgroenks96:
Ok, but the technical issues are largely resolved?

maximilian-gelbrecht:
Yes, everything's implemented as I thought it needs to be, but at the moment I really have no idea why the rule isn't correct.

maximilian-gelbrecht:
This PR might be needed to resolve #801 fully, so I'll have to have another look into it asap.

maximilian-gelbrecht:
I currently still can't quite get my head around why the rule is incorrect after this update. The result is just wrong for the ring at the equator, which points at a normalisation issue. However, the normalisation shouldn't change just because of restructuring the scratch memory...

The thing is, the GPU version seems to actually works because it doesn't use scratch memory in that part of the transform. 

@milankl In the transforms we have different versions for `_apply_serial_fft!` and `_apply_batched_fft!`. On CPU it uses scratch memory because you wrote that FFTW requires stride-1 outputs. I don't think that's actually true, the plans just need to be initialised correctly, see: https://discourse.julialang.org/t/plan-rfft-together-used-with-mul-and-view-complains-about-wrong-strides-output/58215/23?u=maximiliangelbrecht 

Is there another reason why the current CPU version is preferable? Because if not, I could also just rework the transform in a way that the scratch memory for the FFT isn't needed anymore there. This would also resolve the problem with the incorrect rule and we have less code redundancy because CPU and GPU can use the same code again.

maximilian-gelbrecht:
Ok, I just tried it, without the scratch memory it's even slightly faster, so I'll just do it like that now.

milankl:
Awesome yes! Maybe I just misunderstood the FFTW interface, glad I put a comment in to clarify the reason 😉

milankl:
I let you also resolve the conflicts, love to see more code go 🎉

maximilian-gelbrecht:
Uh... generally it works, but now the tolerances in a lot of tests are not kept anymore. That's annoying.

maximilian-gelbrecht:
Somehow this only seems to be an issue for Clenshaw grids though.

maximilian-gelbrecht:
@milankl Have you had any trouble with the numerical accuracy for Clenshaw grids before? I am not sure where this is coming from. 

I also realised FFTW.jl supports applying FFTs to strided data, but not writing into strided array with `mul!`. There is some low-level support for it though, so it might still be possible. I'll look into that.

milankl:
Clenshaw grids need `dealiasing=3` for the transform to be exact otherwise the quadrature is inexact. That's the only thing that should be different from Gaussian grids. No immediate idea why otherwise Clenshaw grids could cause problems with numerical accuracy

maximilian-gelbrecht:
So, calling the low level functions from FFTW to make a plan to write into strided arrays is possible. I opened an issue in FFTW.jl, but for the meanwhile I included the code here in `fourier.jl` 

https://github.com/JuliaMath/FFTW.jl/issues/324

I have to wrap up to use it correctly, then this PR is hopefully done.

maximilian-gelbrecht:
Ok, so also with using `mul!` there are the same numerical precision problems for the Clenshaw transforms. 

I think, I'll revert it to just using the allocating transform for FFTW, until we have a proper solution in FFTW.jl. It's not slower and we have to do it in CUFFT anyway as well. A custom FFTW plan here is just a bit complicated because then we also need to distinguish between GenericFFTs and FFTW. 

So, I am still left a bit puzzled why the numerical problems start to occur for some coefficients in the Crenshaw transforms. These problems are not really that bad probably... Can we just ignore them? @milank what do you think?

maximilian-gelbrecht:
@milankl Could you have a look at this and do you maybe have an idea what's the issue for some of the Clenshaw grids here?

milankl:
I need to check in more detail what's going on here but my best bet it's something with the `if not_equator` statements that were removed: Clenshaw, in comparison to Gaussian grids have a ring on the equator that doesn't have a north-south mirror symmetry. So the `if not_equator` statements were introduced to avoid it doing 2 FFTs over the same ring (the equator) in the forward transforms and avoid adding the iFFT twice in the backwards transform.

Because number of failed tests increases linearly with resolution (but are independent with NF/full vs octahedral) I suspect that's what's happening here

```julia
Test Summary:                              |   Pass  Fail   Total   Time
Transform: Individual Legendre polynomials | 530408  1816  532224  10.4s
  trunc = 31                               |  26616   264   26880   5.2s
    NF = Float32                           |  13308   132   13440   3.1s
      Grid = FullGaussianGrid              |   3360          3360   0.6s
      Grid = FullClenshawGrid              |   3294    66    3360   1.6s
      Grid = OctahedralGaussianGrid        |   3360          3360   0.4s
      Grid = OctahedralClenshawGrid        |   3294    66    3360   0.5s
    NF = Float64                           |  13308   132   13440   2.1s
      Grid = FullGaussianGrid              |   3360          3360   0.6s
      Grid = FullClenshawGrid              |   3294    66    3360   0.5s
      Grid = OctahedralGaussianGrid        |   3360          3360   0.4s
      Grid = OctahedralClenshawGrid        |   3294    66    3360   0.5s
  trunc = 63                               | 102392   520  102912   1.2s
    NF = Float32                           |  51196   260   51456   0.6s
      Grid = FullGaussianGrid              |  12864         12864   0.1s
      Grid = FullClenshawGrid              |  12734   130   12864   0.3s
      Grid = OctahedralGaussianGrid        |  12864         12864   0.1s
      Grid = OctahedralClenshawGrid        |  12734   130   12864   0.2s
    NF = Float64                           |  51196   260   51456   0.6s
      Grid = FullGaussianGrid              |  12864         12864   0.1s
      Grid = FullClenshawGrid              |  12734   130   12864   0.2s
      Grid = OctahedralGaussianGrid        |  12864         12864   0.1s
      Grid = OctahedralClenshawGrid        |  12734   130   12864   0.2s
  trunc = 127                              | 401400  1032  402432   3.9s
    NF = Float32                           | 200700   516  201216   1.8s
      Grid = FullGaussianGrid              |  50304         50304   0.3s
      Grid = FullClenshawGrid              |  50046   258   50304   0.6s
      Grid = OctahedralGaussianGrid        |  50304         50304   0.3s
      Grid = OctahedralClenshawGrid        |  50046   258   50304   0.6s
    NF = Float64                           | 200700   516  201216   2.1s
      Grid = FullGaussianGrid              |  50304         50304   0.3s
      Grid = FullClenshawGrid              |  50046   258   50304   0.6s
      Grid = OctahedralGaussianGrid        |  50304         50304   0.3s
      Grid = OctahedralClenshawGrid        |  50046   258   50304   0.9s
```

`

maximilian-gelbrecht:
The `not_equator` should still be used because it's present in the function that actually calls all fft plans, I just restructered that a bit. But that's already a good idea, and I'll double check that later.

maximilian-gelbrecht:
Okay, by partially reverting some changes, I solved the remaining issue here. 
@milankl Could you do a review of this now? 

In the end now, we do actually allocate in the CPU version as well the small work array, because while FFTW supports strided outputs FFTW.jl doesn't yet. I wouldn't really be that complicated to implement support for it, but then we'd also would have to distinguish between GenericFFTs and FFTW. Anyway CUFFT doesn't have it either, and as we seem to put a bit more priority on GPU now anyway, I think it's not to bad to get rid of the scratch array here. This definitely simplifies some things.

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
```diff
@@ -45,36 +46,36 @@ module SpeedyTransformsCUDAExt
...
     """$(TYPEDSIGNATURES)
-    (Forward) FFT, applied in zonal direction of `grids` provided. This is the 
+    (Forward) FFT, applied in zonal direction of `fields` provided. This is the 
     GPU/CUDA equivalent of the `apply_batched_fft!` function in the CPU version. 
     Uses indexing as we seemingly can't use views with the FFT planning with CUFFT.
     """
     function SpeedyTransforms._apply_batched_fft!(
         f_out::CuArray{<:Complex, 3},
-        grids::AbstractField{NF, N, <:CuArray},
+        fields::AbstractField{NF, N, <:CuArray},
```
milankl: should be `field` no? Maybe historically was called gridS because of several stacked grids with grid the old meaning?

SpeedyTransforms/src/fourier.jl
```diff
@@ -61,16 +56,9 @@ function _apply_batched_fft!(
...
-    # views and copies necessary for stride-1 outputs required by FFTW
-    out = reshape(view(S.scratch_memory_grid, 1:nlon*nlayers), (nlon, nlayers))
-
-    # PERFORM FFT, inverse complex to real, hence brfft
-    # FFTW is in-place writing into `out` via `mul`
-    # FFTW requires stride-1 output, hence view on scratch memory
     if not_equator  # skip FFT, redundant when north already did that latitude
-        LinearAlgebra.mul!(out, brfft_plan, view(g_in, 1:nfreq, 1:nlayers, j))
-        field[ilons, :] = out
+        # this version currently allocates a small array, FFTW supports strided outputs, but FFTW.jl doesn't
```
milankl, line 60: for every latitude this is a length nlayer vector correct?

milankl: So total allocated memory for one transform is nlayers x nlat?

SpeedyTransforms/Project.toml
```diff
@@ -31,7 +31,7 @@ RingGrids = {path = "../RingGrids"}
 AbstractFFTs = "1.5.0"
 Adapt = "4.3.0"
 AssociatedLegendrePolynomials = "1.0.1"
-Atomix = "1.1.2"
+Atomix = "1"
```
milankl, line 34: Why was this specific before? just checking what would happen if someone ended up installing a version between [1, 1.1.2)

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
milankl: fixed this in [710569e](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/791/commits/710569ec897f38dbcfe177949332eb29e4eb0c32)

SpeedyTransforms/Project.toml
maximilian-gelbrecht: I don't know, maybe from CompatHelper? I had some problem with it, that's why I made it more unspecific.

SpeedyTransforms/src/fourier.jl
maximilian-gelbrecht: I think it's ~ `nlat * nlayers * nfreq` all together for the complete transform, no? with `nfreq` and `nlat` varying per ring.

maximilian-gelbrecht:
> Looks all good to me! Glad the transform error for the odd grids somehow resolved?

I think your suspicion with the `not_equator` was right, but somehow I couldn't find the error directly and instead reverted back to an earlier version.

maximilian-gelbrecht:
Performance-wise the allocations make almost no difference. And on GPU we always had them anyway. 

```Julia 

trunc = 128 
NF = Float32

sp_grid = SpectralGrid(; trunc, NF)
S = SpectralTransform(sp_grid)

A = rand(NF, sp_grid.spectrum)
B = transform(A, S)
@benchmark transform($A, $S)
@benchmark transform($B, $S)
```

Results on `mg/scratch-v2` 

```
julia> @benchmark transform($A, $S)
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  327.458 μs …   7.317 ms  ┊ GC (min … max): 0.00% … 93.79%
 Time  (median):     374.041 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   398.425 μs ± 218.444 μs  ┊ GC (mean ± σ):  6.29% ± 10.27%

    █                                                            
  ▃▅█▄▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▂
  327 μs           Histogram: frequency by time          1.5 ms <

 Memory estimate: 564.75 KiB, allocs estimate: 1205.

julia> @benchmark transform($B, $S)
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  296.708 μs …  39.455 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     313.958 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   355.960 μs ± 657.232 μs  ┊ GC (mean ± σ):  4.69% ± 7.84%

  █▇▄▂▁                                                         ▁
  ██████▇▇▇▆▆▅▆▆▆▆▅▅▅▅▄▄▅▁▃▅▄▄▅▅▃▁▄▁▄▃▃▃▁▄▃▁▃▁▄▁▁▄▁▃▁▁▄▁▁▄▁▃▃▄▅ █
  297 μs        Histogram: log(frequency) by time       1.54 ms <

 Memory estimate: 360.38 KiB, allocs estimate: 1205.
```


Results on `main` 

``` 
julia> @benchmark transform($A, $S)

BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  327.334 μs …   8.731 ms  ┊ GC (min … max): 0.00% … 94.32%
 Time  (median):     348.167 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   365.431 μs ± 176.555 μs  ┊ GC (mean ± σ):  4.25% ±  8.67%

   ▇█                                                            
  ▄██▇▃▂▂▂▂▂▂▂▂▂▂▂▁▂▂▁▂▁▂▁▁▁▂▁▁▁▁▁▁▁▂▁▁▁▁▁▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▂
  327 μs           Histogram: frequency by time          898 μs <

 Memory estimate: 362.72 KiB, allocs estimate: 405.

julia> @benchmark transform($B, $S)
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  272.375 μs …   7.556 ms  ┊ GC (min … max): 0.00% … 94.83%
 Time  (median):     303.229 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   312.448 μs ± 158.829 μs  ┊ GC (mean ± σ):  2.52% ±  5.49%

                             ▄▄▄▆█▃▁▁▂▂▁▂▁                       
  ▂▁▁▁▂▁▁▂▁▂▁▁▂▁▁▁▂▂▁▁▁▂▃▃▅▄▅██████████████▆▅▅▅▄▄▄▄▃▃▃▃▃▂▂▂▂▂▂▂ ▄
  272 μs           Histogram: frequency by time          328 μs <

 Memory estimate: 152.09 KiB, allocs estimate: 405.

maximilian-gelbrecht:
Nonetheless when I find some time, I'd still look into writing a strided output plan for FFTW. All the low-level support for it is there. It's just missing a proper constructor in Julia.

PR #792, opened 2025-07-23
Towards PrimitiveDry on GPU 

maximilian-gelbrecht:
More WIP, currently the model initialisation is already working.


milankl:
For clarity: This PR focuses on the PrimitiveDryModel _without_ parameterizations, i.e. `physics=false`? Because we should be able to move both aspects to the GPU quite independently with `physics=false` or `dynamics=false` as these flags completely circumvent the entire respective pieces of code

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/ac17f39280017b7b1305cc49503e6b8dc5d47473/src/dynamics/time_integration.jl#L343-L348

milankl:
For the dynamics I'm very happy for you to go ahead and translate everything as you see fit. I wouldn't be concerned about kernel fusing for now, I believe we should tackle that later. It'll require some restructuring but I don't expect it to be inherently more difficult with KernelAbstractions, rather the opposite.

For the parameterizations I believe we should have a discussion of how to actually do that efficiently and maybe also test some things first before making a call on the required changes. All of these evolve around the idea of what to do with `ColumnVariables` and where to allocate/store the data it represents, whether to do these as views or whether small vectors fit into lower level caches or whether we should just transpose entire 3D arrays.

maximilian-gelbrecht:
Yes, I'll see how far I get with this PR, but I'll def won't include parametrizations here yet.

maximilian-gelbrecht:
@milankl I think after our discussion, we'll takle some other aspects of the GPU version separately in other PRs in the upcoming weeks. So, it'd just like to merge in these minor changes here, that we don't have to constantly update this PR or base the other ones on it

Issue #770, opened 2025-07-16
CoupledClimate configuration for multi-decadal runs


pierluigividale:
All, I attach a notebook with my current setting.
Please note that it is set to run from a previous simulation, but you can uncomment that part and just start from the standard ICs.

```julia
using SpeedyWeather
# components
spectral_grid = SpectralGrid(trunc=31, nlayers=8)
output = NetCDFOutput(spectral_grid, PrimitiveWetModel, path="/Users/vidale/SpeedyWeather", id="Climate_Ocean", output_dt=Minute(180))
add!(output, SpeedyWeather.TemperatureOutput(), SpeedyWeather.LargeScalePrecipitationOutput(), SpeedyWeather.ConvectivePrecipitationOutput())  # output also temperature and precip

# Some parameterisation choices
large_scale_condensation = ImplicitCondensation(spectral_grid)
convection = SimplifiedBettsMiller(spectral_grid)
#surface_evaporation = NoSurfaceEvaporation(spectral_grid)
thermodynamics = LandThermodynamics(spectral_grid, heat_conductivity=0.25)
temperature=LandBucketTemperature(spectral_grid)
soil_moisture=LandBucketMoisture(spectral_grid)
land = LandModel(spectral_grid; temperature, soil_moisture)
ocean = SlabOcean(spectral_grid, mixed_layer_depth=75)
#longwave_radiation = NBandRadiation(spectral_grid, nbands=5)

#Try to fix ocean albedo
#albedo = Albedo(AlbedoClimatology(spectral_grid), AlbedoClimatology(spectral_grid))
albedo = Albedo(GlobalConstantAlbedo(spectral_grid, albedo=0.3), AlbedoClimatology(spectral_grid))
# create model, initialize, run
### Entirely new run
#model = PrimitiveWetModel(spectral_grid; large_scale_condensation, convection, land, ocean, albedo, output=output)

### Continuation run
initial_conditions = StartFromFile(path="/Users/vidale/SpeedyWeather", id="0018")   # or specify `path` kwarg if not in this folder
model = PrimitiveWetModel(spectral_grid; initial_conditions, large_scale_condensation, convection, land, ocean, albedo, output=output)

add!(model, SpeedyWeather.AllOutputVariables()...)
simulation = initialize!(model)
run!(simulation, period=Day(30*365), output=true)
```

A stable setup also requires altering the long wave coefficients inside longwave_radiation.jl, like this:

```julia
[...]

PR #794, opened 2025-07-25
Jeevanjee surface fluxes with Stefan-Boltzmann and soil moisture field capacity

pierluigividale:
A number of fixes and changes, to make the model stable and reasonably realistic.

### DIFF
--- a/src/physics/longwave_radiation.jl
+++ b/src/physics/longwave_radiation.jl
@@ -41,25 +41,25 @@ $(TYPEDFIELDS)"""
 end
 
 UniformCooling(SG::SpectralGrid; kwargs...) = UniformCooling{SG.NF}(; kwargs...)
-initialize!(scheme::UniformCooling, model::PrimitiveEquation) = nothing
+initialize!(radiation::UniformCooling, model::PrimitiveEquation) = nothing
 
 function longwave_radiation!(
     column::ColumnVariables,
-    scheme::UniformCooling,
+    radiation::UniformCooling,
     model::PrimitiveEquation,
 )
-    longwave_radiation!(column, scheme)
+    longwave_radiation!(column, radiation)
 end
 
 function longwave_radiation!(
     column::ColumnVariables{NF},
-    scheme::UniformCooling,
+    radiation::UniformCooling,
 ) where NF
     (; temp, temp_tend) = column
-    (; temp_min, temp_stratosphere) = scheme
+    (; temp_min, temp_stratosphere) = radiation
     
-    cooling = -inv(convert(NF, scheme.time_scale.value))
-    τ⁻¹ = inv(scheme.time_scale_stratosphere.value)
+    cooling = -inv(convert(NF, radiation.time_scale.value))
+    τ⁻¹ = inv(radiation.time_scale_stratosphere.value)
 
     @inbounds for k in eachlayer(column)
         # Pauluis and Garner, 2006, eq (1) and (2)
@@ -102,31 +102,31 @@ $(TYPEDFIELDS)"""
 end
 
 JeevanjeeRadiation(SG::SpectralGrid; kwargs...) = JeevanjeeRadiation{SG.NF}(; kwargs...)
-initialize!(scheme::JeevanjeeRadiation, model::PrimitiveEquation) = nothing
+initialize!(radiation::JeevanjeeRadiation, model::PrimitiveEquation) = nothing
 
 # function barrier
 function longwave_radiation!(
     column::ColumnVariables,
-    scheme::JeevanjeeRadiation,
+    radiation::JeevanjeeRadiation,
     model::PrimitiveEquation,
 )
-    longwave_radiation!(column, model.atmosphere, scheme)
+    longwave_radiation!(column, radiation, model.atmosphere)
 end
 
 function longwave_radiation!(
     column::ColumnVariables{NF},
+    radiation::JeevanjeeRadiation,
     atmosphere::AbstractAtmosphere,
-    scheme::JeevanjeeRadiation,
 ) where NF
 
     (; nlayers, temp_tend) = column
     T = column.temp                 # to match Seeley, 2023 notation
     F = column.flux_temp_upward
-    (; α, time_scale) = scheme
-    ϵ_ocn = scheme.emissivity_ocean
-    ϵ_lnd = scheme.emissivity_land
+    (; α, time_scale) = radiation
+    ϵ_ocean = radiation.emissivity_ocean
+    ϵ_land = radiation.emissivity_land
     σ = atmosphere.stefan_boltzmann
-    Tₜ = scheme.temp_tropopause
+    Tₜ = radiation.temp_tropopause
     
     (; skin_temperature_sea, skin_temperature_land, land_fraction) = column
 
@@ -134,12 +134,12 @@ function longwave_radiation!(
     # between surface and lowermost air temperature
     # but zero flux if land/sea not available
     Fₖ_ocean = isfinite(skin_temperature_sea) ?
-        ϵ_ocn*σ*skin_temperature_sea^4 : 
+        ϵ_ocean*σ*skin_temperature_sea^4 : 
         zero(skin_temperature_sea)
     column.surface_longwave_up_ocean = Fₖ_ocean
 
     Fₖ_land = isfinite(skin_temperature_land) ?
-        ϵ_lnd*σ*skin_temperature_land^4 : 
+        ϵ_land*σ*skin_temperature_land^4 : 
         zero(skin_temperature_land)
     column.surface_longwave_up_land = Fₖ_land
 
@@ -170,11 +170,11 @@ export NBandRadiation
 end
 
 NBandRadiation(SG::SpectralGrid; kwargs...) = NBandRadiation(; kwargs...)
-initialize!(scheme::NBandRadiation, model::PrimitiveEquation) = nothing
+initialize!(radiation::NBandRadiation, model::PrimitiveEquation) = nothing
 
 function longwave_radiation!(
     column::ColumnVariables,
-    scheme::NBandRadiation,
+    radiation::NBandRadiation,
     model::PrimitiveEquation,
 )
 


src/physics/longwave_radiation.jl
```diff
@@ -123,12 +123,12 @@ function longwave_radiation!(
     # extension to Jeevanjee: Include temperature flux between surface and lowermost air temperature
     # but zero flux if land/sea not available
     Fₖ_ocean = isfinite(skin_temperature_sea) ?
-        (T[end] - skin_temperature_sea) * α * (Tₜ - skin_temperature_sea) : 
+        (T[end] - skin_temperature_sea) * 3.1 * α * (Tₜ - skin_temperature_sea) :
```
maximilian-gelbrecht: The continuous integration tests for differentiability with Enzyme are failing because this change introduces a type instability. `3.1` is a `Float64`, but we have `NF` (so e.g. `Float32`) precision here. So `Fₖ_ocean` can be either a `Float64` or `NF`.  

An easy fix should be to just write `NF(3.1)` here instead of `3.1`.

milankl: @pierluigividale can we move the 3.1 (what's this?) into the JeevanJee struct definition?

src/physics/longwave_radiation.jl
```diff
@@ -123,12 +123,12 @@ function longwave_radiation!(
...
     # but zero flux if land/sea not available
     Fₖ_ocean = isfinite(skin_temperature_sea) ?
-        (T[end] - skin_temperature_sea) * α * (Tₜ - skin_temperature_sea) : 
+        (T[end] - skin_temperature_sea) * 3.1 * α * (Tₜ - skin_temperature_sea) : 
         zero(skin_temperature_sea)
     column.surface_longwave_up_ocean = Fₖ_ocean
 
     Fₖ_land = isfinite(skin_temperature_land) ?
-        (T[end] - skin_temperature_land) * α * (Tₜ - skin_temperature_land) : 
+        (T[end] - skin_temperature_land) * 4.6 * α * (Tₜ - skin_temperature_land) :
```
milankl: Similar here. It looks like something that could be absorbed into alpha? But it seems like you want an alpha_ocean and an alpha_land?

src/physics/land/thermodynamics.jl
```diff
@@ -5,8 +5,8 @@ export LandThermodynamics
     "[OPTION] Heat conductivity λ of the soil [W/(m K)]" 
     heat_conductivity::NF = 0.42
 
-    "[OPTION] Field capacity γ per meter soil [1]"
-    field_capacity::NF = 0.24
+    "[OPTION] Field capacity γ per meter soil [1]" #CPL changed this to be the same as below
```
milankl: Shall we not remove one then if they are supposed to be the same anyway?

pierluigividale: Yes, we can just take field_capacity, as it is not used in this code, but we need to check elsewhere in the code.

src/physics/longwave_radiation.jl
pierluigividale: Yes, I could not use a single value. If we want a stable climate, until we have a cloud-aware radiation scheme, these coefficients were the only ones that allow land and ocean to remain stable. Took me over 200 simulations to test all possible combinations of albedo, LW settings, ML ocean depth, soil settings, to finally achieve a stable climate that looks good for several decades. Seems like the worst of the brute-force method, but I did learn a lot about Speedy in the process.

milankl: Okay I moved this into `α_ocean, α_land` parameters for JeevanJee radiation

milankl:
@pierluigividale how is this pull request related to #788 ?

pierluigividale:
Hi
I did not remove field_capacity from my pull request, as I sent that pull request before we discussed going for it, and also because I have been investigating the 16 layer and 32 layer configurations in the last couple of days. See my other issue, as what I have found is quite puzzling.

I can work on eliminating field_capacity later this week, please assign it to me.

milankl:
> Hi I did not remove field_capacity from my pull request, as I sent that pull request before we discussed going for it, and also because I have been investigating the 16 layer and 32 layer configurations in the last couple of days. See my other issue, as what I have found is quite puzzling.

Then let's keep these as two pull requests.

milankl:
Can we make this pull request just about the Jeevanjee surface fluxes and move the soil moisture changes to #788 ?

I did an analysis of surface longwave radiation fluxes comparing Jeevanjee to Stefan-Boltzmann i.e.

```julia
α = 0.025
Tₜ = 200

# ocean
Fo_up = @. sst * α * (sst - Tₜ)
Fo_down = @. - t * α * (sst - Tₜ)
Fo = Fo_up + Fo_down

# land
Fl_up = @. st * α * (st - Tₜ)
Fl_down = @. - t * α * (st - Tₜ)
Fl = Fl_up + Fl_down
```

versus $\sigma T^4$ where I took SST `sst`, land surface temperatue `st` and lower-most air temperature $t$ from a short simulation just to have some typical values without much drift that had a chance to occur yet. This yields

image

So Jeevanjee has a higher up and down flux compared the Stefan-Boltzmann but the net is pretty similar in the range up to maybe 30˚C or so but for higher temperatures the surface (ocean/land) can loose heat more efficiently with Stefan-Boltzmann due to the $T^4$-based formulation vs the $T^2$ from Jeenvanjee. I think this also somehow means we should just abandon Jeevanjee for non-prescribed surface fluxes which is also what the authors wrote in their paper. Maybe I was just naive and thought "how bad can it be".

milankl:
I moved all soil moisture related changes to #788

pierluigividale:
I was thinking the same thing: why not make the surface LW_up just Stefan-Botlzmann? It is not going to cost more than what is there now.

Can you tell me where we are with the soil moisture pull request? I tried to update the code, but nothing is available.

pierluigividale:
There is no way to make the Stefan-Boltzmann work on land: no reasonable radiative balance can be achieved. Sadly, it works really well over the ocean, and I was even able to use a realistic ocean albedo for the SW.

For land, at least, and for the time being, we must go back to the Jeevanjee approach to surface fluxes. How can I refuse this merge and go back to editing the code?

pierluigividale:
Forgot to say: there is a way to still use S-B on land, and that is, make the top soil layer considerably thicker than it is. That will increase its heat capacity and reduce the diurnal oscillation swings. I can experiment with that tomorrow.

milankl:
Now you are implementing the changes from #788 here again! If you feel both pull requests shouldn't be treated independently then you can merge #788 here first before merging both into main!

milankl:
Let me know if you want me to do that!

src/physics/longwave_radiation.jl
```diff
@@ -104,31 +104,36 @@ function longwave_radiation!(
...
     scheme::JeevanjeeRadiation,
 ) where NF
 
     (; nlayers, temp_tend) = column
     T = column.temp                 # to match Seeley, 2023 notation
     F = column.flux_temp_upward
     (; α, time_scale) = scheme
+    σ = atmosphere.stefan_boltzmann
+    ϵ_ocn = 0.6
+	ϵ_lnd = 0.65
```
milankl: Move these into the `JeevanjeeRadiation` struct declared as `::NF` otherwise you're hardcoding in double precision!

pierluigividale: AGHHH. This is the same issue I had with those alphas

pierluigividale: wrong place, half a minute

src/physics/longwave_radiation.jl
```diff
@@ -118,8 +118,9 @@ function longwave_radiation!(
     F = column.flux_temp_upward
     (; α, time_scale) = scheme
     σ = atmosphere.stefan_boltzmann
-    ϵ_ocn = 0.6
-	ϵ_lnd = 0.65
+#INTRODUCING SOME EFFECTIVE EMISSIVITIES AS TUNING PARAMETERS
+    ϵ_ocn::NF = 0.6    # for surface longwave calculation over the ocean
+    ϵ_lnd::NF = 0.65   # for surface longwave calculation over the land
```
milankl: technically you can do this too, this would set a correct number format `NF` but it's still hardcoded to those values so a user wouldn't easily be able to run with other emissivities, if you move this into the `JeevanjeeRadiation` struct though then a user can do

```julia
longwave_radition = JeevanjeeRadiation(spectral_grid, ϵ_ocn = 0.9)
model = PrimitiveWetModel(spectral_grid; longwave_radiation)
```

and so while you provide a default there's an easy interface to change that value!

milankl: also suggestion: `emissivity_ocean` and `emissivity_land` then you can do
```julia
ϵ_ocn = scheme.emissivity_ocean
```
inside `longwave_radiation!` for local renaming to something brief such that equations look like math!

milankl:
CI fails because of
```julia
UndefVarError: `ϵ_ocn` not defined
  Stacktrace:
    [1] longwave_radiation!(column::ColumnVariables{Float32, Vector{Float32}, Matrix{Float32}}, atmosphere::EarthAtmosphere{Float32}, scheme::JeevanjeeRadiation{Float32})
      @ SpeedyWeather ~/work/SpeedyWeather.jl/SpeedyWeather.jl/src/physics/longwave_radiation.jl:132
    [2] longwave_radiation!
      @ ~/work/SpeedyWeather.jl/SpeedyWeather.jl/src/physics/longwave_radiation.jl:111 [inlined]
```
😉

pierluigividale:
OK, as far as I can tell, I have put the definitions in the right places, but the pull request still fails on this:
UndefVarError: `ϵ_ocn` not defined

milankl:
I've changed

```julia
#INTRODUCING SOME EFFECTIVE EMISSIVITIES AS TUNING PARAMETERS
    "Effective emissivities of different surfaces []"
    emissivity_ocean::NF = 0.6    # for surface longwave calculation over the ocean
    emissivity_land::NF = 0.65   # for surface longwave calculation over the land
```

to

```julia
    "[OPTION] Effective emissivity for surface flux over ocean [1]"
    emissivity_ocean::NF = 0.6

    "[OPTION] Effective emissivity for surface flux over land [1]"
    emissivity_land::NF = 0.65
```

While both are equivalent comments for the code the latter is the structure that's needed to be automatically picked up by the documenter! So the pattern is

```julia
"Some docstring that's automatically read and used in the help via ?SomeType and written into the documentation"
name::T = default_value
```

So with this you can just type `?JeenvanjeeRadiation` and get

```julia
help?> JeevanjeeRadiation
search: JeevanjeeRadiation NBandRadiation

  Temperature flux longwave radiation from Jeevanjee and Romps, 2018,
  following Seeley and Wordsworth, 2023, eq (1)

  dF/dT = α*(T_t - T)

  with F the upward temperature flux between two layers with temperature
  difference dT, α = 0.025 W/m²/K² and T_t = 200K a prescribed tropopause
  temperature. Flux into the lowermost layer is 0. Flux out of uppermost layer
  also 0, but dT/dt = (T_t - T)/τ is added to relax the uppermost layer
  towards the tropopause temperature T_t with time scale τ = 24h (Seeley and
  Wordsworth, 2023 use 6h, which is unstable a low resolutions here). Fields
  are

    •  α::Any: [OPTION] Radiative forcing constant (W/m²/K²)

    •  emissivity_ocean::Any: [OPTION] Effective emissivity for surface
       flux over ocean [1]

    •  emissivity_land::Any: [OPTION] Effective emissivity for surface
       flux over land [1]

    •  temp_tropopause::Any: Tropopause temperature [K]

    •  time_scale::Second: Tropopause relaxation time scale to
       temp_tropopause
```

No need to look into the code to know what the options are!!

milankl:
And hopefully last point, we have a `NoVegatation` scheme which takes the soil moisture availability and simplifies the equation for veg=0 which also required some changes because of renaming `W_cap` to `field_capacity`

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/cac0660b6511f1bd422f60b3534f69d4574a0a90/src/physics/land/vegetation.jl#L35-L60

So you can also do `vegetation = NoVegetation()` and you have killed all trees and plants!

pierluigividale:
Amazing. I was tempted to do that, but did not want to sound pompous. Now I know what it is all for.
OK, shall deal with the novegetation in a second.

pierluigividale:
Ah, but you seem to have done that already!

pierluigividale:
Except that you did not do it correctly. This equation is inconsistent with the one at the very end:
    @inbounds for ij in eachindex(soil_moisture_availability)
        # Fortran SPEEDY documentation eq. 51 but veg=0
        soil_moisture_availability[ij] = r*D_top*soil_moisture[ij, 1]

Basically what we want here is the result of defining vegetation cover as 0, so that the veg coefficient is zero in this equation:

       soil_moisture_availability[ij] = r*(D_top*soil_moisture[ij, 1]*W_cap +
                                                      veg*D_root*max(soil_moisture[ij, 2]*W_cap - W_wilt, 0))

So:
        soil_moisture_availability[ij] = r*(D_top*soil_moisture[ij, 1]*W_cap

milankl:
Awesome, just did that and also removed the `eachindex` as scalar indexing isn't allowed for GPUs but a line like

```julia
soil_moisture_availability .= D_top*soil_moisture_top*W_cap/
                                    (D_top*W_cap + D_root*(W_cap - W_wilt))
```

translates automatically to a GPU kernel

pierluigividale:
> Awesome, just did that and also removed the `eachindex` as scalar indexing isn't allowed for GPUs but a line like
> 
> ```julia
> soil_moisture_availability .= D_top*soil_moisture_top*W_cap/
>                                     (D_top*W_cap + D_root*(W_cap - W_wilt))
> ```
> 
> translates automatically to a GPU kernel

OK, I have been dropped into the deep end of Julia, and I have had quite a few successes, but today I made many blunders. Now, anything that has to do with GPU is a mystery to me, but my ambition is to understand some of it by the end of CMSS.

milankl:
We use KernelAbstractions and put a lot of effort already in that transitioning from CPU to GPU is a very smooth experience. Boils down to a few rules

1. Array types should not be hardcoded
2. Scalar indexing is not allowed
3. If you want to express a kernel with scalar indexing you have to write a `@kernel`

For 1. We basically took care of with the `SpectralGrid`, e.g.

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/53d3d0be53a132e50916e7fce6a209d0c76e1ea3/src/dynamics/spectral_grid.jl#L175-L183

For 2. current loops like `for i in eachindex(...)` will only work on the CPU

For 3. See as an example the leapfrogging kernel that's already written. A kernel is defined, launched and then synchronized afterwards

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/53d3d0be53a132e50916e7fce6a209d0c76e1ea3/src/dynamics/time_integration.jl#L176-L189

milankl:
fixes #770

milankl:
@pierluigividale I give you the honor to squash and merge!!

pierluigividale:
Very satisfactory !

PR #809, opened 2025-08-06
show architecture as "CPU"/"GPU"

milankl:
Currently the default title doesn't even fit...!

image


maximilian-gelbrecht:
This kind of fixes the symptom and not the cause, right? We could also think about a custom short type print for the Grid / Spectrum. This would make the stack trace also more readable in many cases

milankl:
Yes and no. I reckon `typeof` shouldn't be changed which is currently what's used in the default title or do you think `Base.typeof(::Field)` and wouldn't have unforeseen consequences? Or do you mean just the show of type of fields? Yeah I don't know how to do that but that sounds reasonable!!

milankl:
Now

image

maximilian-gelbrecht:
> Yes and no. I reckon `typeof` shouldn't be changed which is currently what's used in the default title or do you think `Base.typeof(::Field)` and wouldn't have unforeseen consequences? Or do you mean just the show of type of fields? Yeah I don't know how to do that but that sounds reasonable!!

We recently had this discussion about `showarg`

```
>?Base.showarg 

showarg(io::IO, x, toplevel)

  Show x as if it were an argument to a function. This function is used by summary to display type information
  in terms of sequences of function calls on objects. toplevel is true if this is the direct call from summary
  and false for nested (recursive) calls.

  The fallback definition is to print x as "::$(typeof(x))", representing argument x in terms of its type. (The
  double-colon is omitted if toplevel=true.) However, you can specialize this function for specific types to
  customize printing.

  Example
  ≡≡≡≡≡≡≡

  A SubArray created as view(a, :, 3, 2:5), where a is a 3-dimensional Float64 array, has type

  SubArray{Float64, 2, Array{Float64, 3}, Tuple{Colon, Int64, UnitRange{Int64}}, false}

  The default show printing would display this full type. However, the summary for SubArrays actually prints as

  2×4 view(::Array{Float64, 3}, :, 3, 2:5) with eltype Float64

  because of a definition similar to

  function Base.showarg(io::IO, v::SubArray, toplevel)
      print(io, "view(")
      showarg(io, parent(v), false)
      print(io, ", ", join(v.indices, ", "))
      print(io, ')')
      toplevel && print(io, " with eltype ", eltype(v))
  end

  Note that we're calling showarg recursively for the parent array type, indicating that any recursed calls are
  not at the top level. Printing the parent as ::Array{Float64,3} is the fallback (non-toplevel) behavior,
  because no specialized method for Array has been defined.
```

Can't this be just in such a case because it can distinguish top level and not top level?

milankl:
I still don't fully understand how to do this. What do we want? Currently e.g.

```julia
julia> field = zeros(OctahedralGaussianGrid, 2)
88-element, 4-ring OctahedralGaussianField{Float32, 1} on Array on SpeedyWeather.Architectures.CPU{KernelAbstractions.CPU}
...
```

it would be neat if that could be shortened to

```julia
88-element, 4-ring OctahedralGaussianField{Float32, 1} as Array on CPU
```

We already defined a short-hand `show` for `<:AbstractGrid`

```julia
julia> field.grid
4-ring OctahedralGaussianGrid
├ nlat_half=2 (88 points, ~21.7˚, reduced)
└ architecture: SpeedyWeather.Architectures.CPU{KernelAbstractions.CPU}
```

but in some circumstances it might be nice to show

```julia
OctahedralGaussianGrid{CPU{...}, ...}
```

instead of 

```julia
julia> typeof(field.grid)
OctahedralGaussianGrid{SpeedyWeather.Architectures.CPU{KernelAbstractions.CPU}, Vector{UnitRange{Int64}}, Vector{Int64}}
```

but generally I guess `typeof(field)`, `typeof(field.grid)` and `typeof(field.grid.architecture)` should be unchanged, you agree? So one can always look up the exact type but summaries are printed in short form

maximilian-gelbrecht:
I am a bit confused about `showargs` docstring, so we also just need to test it. 

I think the crucial bit in their that we can manipulate what's displayed when the type information is printed / shown at top-level and what's shown when it's recursively called, not at top level.

milankl:
Also for these now shorter
```julia
julia> spectral_grid = SpectralGrid(trunc=31, nlayers=1)
SpectralGrid{Spectrum{...}, OctahedralGaussianGrid{...}}
├ Spectral:     T31 LowerTriangularMatrix{Complex{Float32}}, radius = 6.371e6 m
├ Grid:         Field{Float32} on 48-ring OctahedralGaussianGrid, 3168 grid points
├ Resolution:   401km (average)
├ Vertical:     1-layer atmosphere, 2-layer land
└ Architecture: CPU using Array

julia> field = zeros(OctaminimalGaussianGrid, 2)
24-element, 4-ring OctaminimalGaussianField{Float32, 1} as Array on CPU
 0.0f0
...
 0.0f0

julia> field.grid
4-ring OctaminimalGaussianGrid
├ nlat_half=2 (24 points, ~41˚, reduced)
└ architecture: CPU

julia> field.grid.architecture
CPU
```

milankl:
Postponing the `showarg` issue but merging this by defining show for architectures

PR #811, opened 2025-08-06
Thermodynamic sea ice model

milankl:
Implemented the following thermodynamic sea ice model of sea ice concentration $c$ in $[\text{m}^2/\text{m}^2]$. Using sea surface temperature $SST$ at the previous time step $i-1$ or the current $i$ and uncorrected marked with $*$ and time step size $\Delta t$.

First determine an insulation factor $r$ ($r=0$ equals full insulation, $r=1$ no insulation) linearly from sea ice concentration which reduces the air-sea flux $F$ used as the tendency for the slab ocean SST
```math
\begin{aligned}
r &= 1 - c_{i-1} \\
SST_i^* &= SST_{i-1} + \Delta t~r~F \\
SST_i &= max(SST_i^*, T_f)
\end{aligned}
```
And bound the uncorrected sea surface temperature $SST_i^*$ by freezing at $T_f = -1.8˚C$.

Now determine a tendency in sea ice concentration $\Delta c$ from the melting and freezing, both proportional to the difference of the sea surface temperature to freezing with possibility different freeze rate $f$ in $[\text{m}^2/\text{m}^2/K]$ and melt rate $m$ in $[\text{m}^2/\text{m}^2/s/K]$. The freeze rate misses a 1/s in the units as it's applied with $1/\Delta t$ below because $\min(SST_i^* - T_f, 0)$ estimates $\Delta t F$ from above not purely the (insulated) flux $rF$

```math
\begin{aligned}
\Delta c &= -m \max(SST_i^* - T_f, 0) - \frac{f}{\Delta t} \min(SST_i^* - T_f, 0) \\
c_i^* &= c_{i-1} + \Delta t \Delta c
\end{aligned}
```

But bound the uncorrected sea ice concentration $c_i^*$ in $[0, 1]$ by

```math
c_i = \max( \min( c_i^*, 1) , 0)
```

To be used like

```julia
...
sea_ice = ThermodynamicSeaIce(spectral_grid)
ocean = SlabOcean(spectral_grid, mixed_layer_depth=50)
albedo = Albedo(ocean=OceanSeaIceAlbedo(spectral_grid), land=AlbedoClimatology(spectral_grid))
model = PrimitiveWetModel(spectral_grid; land, ocean, sea_ice, albedo)

# choose initial conditions
set!(simulation, sea_ice_concentration=0)

# output both SST and sea ice concentration
add!(model, SpeedyWeather.OceanOutput()...)
```

Seasonal cycle of sea ice cover starting from 0 concentration in January

https://github.com/user-attachments/assets/e41fd6ae-b9c5-4f79-8176-f0600ff7b85b

~~not sure why we have sea ice over large parts of Antarctica, shouldn't happen~~ solved this with quantization of the land sea mask, so values can only be multiples of 1% now by default to avoid tiny lakes being considered ocean

TODO
- [x] units of sea ice model parameters (EDIT: formulated with melt/freeze rates)
- [x] write as KernelAbstractions kernel
- [x] mathematical formulation
- [x] documentation (here https://speedyweather.github.io/SpeedyWeatherDocumentation/previews/PR811/ocean_sea_ice/)
- [ ] tests


maximilian-gelbrecht:
Nice! Looks good. 

One thing though: Going forward, when adding new features, I think we should directly do them with KA. (intiailization as well)

milankl:
> One thing though: Going forward, when adding new features, I think we should directly do them with KA. (intiailization as well)

YES! I know I have to get used to this and just write kernels directly ... 🙈

milankl:
@simone-silvestri now your sea ice model in comparison!

simone-silvestri:
Oh nice, SpeedyWeather is becoming a full on ESM now :)

milankl:
sea ice growth after switching off the sun (`planet.solar_constant = 0`) over the course of a year

https://github.com/user-attachments/assets/a5c9cc3e-0abd-4a27-a6ba-2d501399b6f8

milankl:
So Antarctic sea ice melts ~1.4x faster than it grows (Mar to Oct, 7 months growth, 5 months melt)

image

But Arctic sea ice grows and melts relatively similarly (mid March to mid September; mid September till mid March)

image

glwagner:
Note most models use `a` for sea ice concentration; ClimaSeaIce uses `א`.

src/physics/sea_ice.jl
```diff
@@ -72,20 +72,22 @@ function sea_ice_timestep!( progn::PrognosticVariables,
...
-            F = -m*max(dT, 0) - f*min(dT, 0)        # melt if above freezing, freeze if below
-            sst[ij] = max(sst[ij], temp_freeze)     # cap sst at freezing
-
-            # update sea ice concentration, cap between [0, 1]
-            ice[ij] += Δt*F
-            ice[ij] = max(min(ice[ij], 1), 0)
-        end
-    end
+    launch!(architecture(ice), LinearWorkOrder, size(ice), sea_ice_kernel!, ice, sst, mask, temp_freeze, m, f, Δt)
+    synchronize(architecture(ice))
```
simone-silvestri, line 76: I don't think you need synchronization after a launch. If you are running on a GPU, all the kernels are executed on one stream, so the GPU is synchronized with itself. If you pass memory between the CPU and GPU, the GPU automatically synchronizes with the CPU. You need to explicitly synchronize only if you are using MPI, since that runs on a different stream. All other explicit synchronizations will lead to performance loss.

maximilian-gelbrecht: My understanding so far was that this is the case if you only use KA, but we do use some CUDA.jl functions/executations as well. CUFFT is unavoidable for us, then we also use e.g. some broadcasting here and there (which we could in theory replace). Or is this fine? Then we should really remove also other explicit synchronisation calls we already have.

simone-silvestri: We also use CUFFT and have never had problems with synchronization. I am not sure about broadcasts, also our broadcasts are kernels. However, if you never manually change out of the default CUDA stream, everything should run on the default stream I think.

maximilian-gelbrecht: Ok great to know.

milankl:
Introduced an `OceanSeaIceAlbedo` that scales linearly from ocean albedo to ice albedo given sea ice concentration. March Albedo

image


September Albedo

image

milankl:
@pierluigividale feel free to test this if you're keen. As you can see above there's actually three parts of using `ThermodynamicSeaIce`

1. Insulation of air-sea fluxes with increasing ice concentration (by default linear in the `SlabOcean`)
2. `sea_ice = ThermodynamicSeaIce(spectral_grid, ...)` itself, this controls the ice freeze/melt and will bound SST at freezing
3. Using `OceanSeaIceAlbedo` to change the albedo linearly with sea ice concentration between 0.06 (ocean) and 0.6 (ice) which are the defaults

You should then see the impact of sea ice on albedo, fluxes etc in the netCDF files.

pierluigividale:
I am really keen to test this. I have completed the land surface changes, all is stable, so I hope to just recommend them for the CTL run. However, if these sea ice tests work, I may have to once again tune the radiation to cope with these albedo changes.
@simone-silvestri: yes, many ESM bits coming in, but the highest priority continues to be radiation. Somewhat underwhelming  having quite realistic surface fluxes if then the atmosphere cannot react to them via the radiative balance.

milankl:
Ah! You merged already 😝 Sorry maybe I should have left this as "draft" there were still some tests to be added, well then they go in the next pull request. Merging small pull requests without much review is fine but new features should definitely have tests added before merging in so that we make sure they actually work. Like technically the sea ice model can be used with a PrimitiveDryModel but neither me nor github ever tested this. Normally we leave it to the pull request creator to merge!

milankl:
What I meant by asking for a review is for you to install this branch locally and play around with it not to merge it and then install the new version of main

pierluigividale:
Sorry, I am too eager. I was just thinking, as I went through, that previously we ran interim tests.
However, this one does nothing if no one calls that module, so it should not be too damaging?
I am already running the test...

milankl:
All good, the idea is just that this new features here, sea ice, also needs tests so that if someone else works one another feature that somehow interacts with the changes made here (e.g. ocean, radiation, surface fluxes) then suddenly the sea ice model could fail without us noticing 😉

milankl:
I'll add a test that `ThermodynamicSeaIce` and `NoSeaIce` run, with and without `OceanSeaIceAlbedo` and in combination with different oceans like `SlabOcean`, `SeasonalOceanClimatology`, `AquaPlanet` ... tomorrow. E.g. sea ice concentration should always remain within `[0, 1]` whether one starts at the defaults, with a super cold planet or in a hothouse! ... For a model being modular all these combinations somehow have to be tested!

pierluigividale:
OK, it runs fine, but it does not run if I try to do this:
add!(model, SpeedyWeather.AllOutputVariables()...)

pierluigividale:
I ran 10 years. It looks uncannily good.

milankl:
What do you think should the atmosphere "feel" of the sea ice? At the moment the atmosphere experiences a -1.8C ocean where there is sea ice. Meaning evaporation is like above liquid water (maybe not important though given the low temperatures), I wouldn't bother adjusting momentum fluxes although we could, and for heat fluxes is probably okay to assume a -1.8C surface?

pierluigividale:
I have put the output of a 30-year simulation with T31_L8 on JASMIN at this location:
/work/scratch-nopw2/plvidale/Speedy/run_Climate_Ocean_Land_SeaIce_0004

This is the script I have used for that simulation:

```Julia
using SpeedyWeather
# components
output = NetCDFOutput(spectral_grid, PrimitiveWetModel, path="/Users/vidale/SpeedyWeather", id="Climate_Ocean_Land_SeaIce", output_dt=Minute(360))

# Some parameterisation choices
vertical_advection = WENOVerticalAdvection(spectral_grid)
#large_scale_condensation = ImplicitCondensation(spectral_grid)
#convection = SimplifiedBettsMiller(spectral_grid)
temperature=LandBucketTemperature(spectral_grid)
soil_moisture=LandBucketMoisture(spectral_grid)
land = LandModel(spectral_grid; temperature, soil_moisture)
ocean = SlabOcean(spectral_grid, mixed_layer_depth=50)
sea_ice = ThermodynamicSeaIce(spectral_grid)

#Try to fix ocean albedo
#albedo = Albedo(AlbedoClimatology(spectral_grid), AlbedoClimatology(spectral_grid))
#albedo = Albedo(GlobalConstantAlbedo(spectral_grid, albedo=0.06), AlbedoClimatology(spectral_grid))
albedo = Albedo(ocean=OceanSeaIceAlbedo(spectral_grid), land=AlbedoClimatology(spectral_grid))
# create model, initialize, run

#time_stepping = Leapfrog(spectral_grid, Δt_at_T31=Minute(20))
#implicit = ImplicitPrimitiveEq(spectral_grid, \alpha=0.5...1)
#time_stepping = Leapfrog(spectral_grid, Δt_at_T31=Minute(40)) #CPL this is the standar time step
# change the time scheme to be more off-centred and enable other waves than Kelvin and WMMRGs.

### Entirely new run
model = PrimitiveWetModel(spectral_grid; vertical_advection, time_stepping, land, ocean,  sea_ice, albedo, output)

### Continuation run
#initial_conditions = StartFromFile(path="/Users/vidale/SpeedyWeather", id="Climate_Ocean_Land_SeaIce",run_number=15)   # or specify `path` kwarg if not in this folder
#model = PrimitiveWetModel(spectral_grid; initial_conditions, vertical_advection, time_stepping, land, ocean,  sea_ice, albedo, output)

#Make sure to add all diagnostics to the model
#add!(model, SpeedyWeather.AllOutputVariables()...)

# now offset via set! with add=true
set!(albedo.land, (λ, φ) -> 0.1, add=true)
model.land.geometry.layer_thickness[1]=0.2
model.land.geometry.layer_thickness[2]=2.0
model.land.soil_moisture.runoff_fraction=0.25
model.land.soil_moisture.mask=true
model.land.temperature.mask=true
#simulation.prognostic_variables.land.soil_moisture .= 0.3
set!(simulation, soil_moisture = 0.75)
set!(simulation, sea_ice_concentration=0)

initialize!(model.land, model)   # first arg: what you want to initialize, 2nd: always model (read only)
#simulation = initialize!(model)
# output both SST and sea ice concentration
add!(model, SpeedyWeather.OceanOutput()...)
add!(model, SpeedyWeather.LandOutput()...)
add!(model, SpeedyWeather.RadiationOutput()...)
add!(model, SpeedyWeather.PrecipitationOutput()...)
add!(model, SpeedyWeather.SurfaceFluxesOutput()...)
run!(simulation, period=Year(30), output=true)
```

PR #812, opened 2025-08-07
Keyword arguments for ImplicitPrimitiveEquation

milankl:
```julia
julia> ImplicitPrimitiveEquation(spectral_grid, α=0.5)
ERROR: MethodError: no method matching ImplicitPrimitiveEquation(::SpectralGrid{…}; α::Float64)
This method may not support any kwargs.
```

`, kwargs...` means any additional positional arguments (hence should be called `args...` not `kwargs...`) whereas `; kwargs...` means any additional keyword arguments. Replaces `,` with `;`. My bad!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Keyword arguments corrected for ImplicitPrimitiveEquation [#812](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/812)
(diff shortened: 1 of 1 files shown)


pierluigividale:
OK, I had tried both ; and , but it looks like the error is such that it does not matter?

milankl:
> OK, I had tried both ; and , but it looks like the error is such that it does not matter?

Yeah no it's a positional argument in that's tried to be passed on as keyword argument but you can't put the key in if it's not supposed to be a keyword argument!

pierluigividale:
I see. Should probably have realised from my knowledge of python, but when the code is this complex I struggle even in Python.
Waiting for the checks to complete.

PL

PR #813, opened 2025-08-07
Thermodynamic sea ice model testing

milankl:
followup from #811


milankl:
Starting from a clean branch otherwise this is confusing ...

PR #815, opened 2025-08-08
ColumnFields 

maximilian-gelbrecht:
Ok, I really quickly got the basics, but somehow this broke GPU broadcasting. So I'll have to finish this another time.

### DIFF
--- a/src/RingGrids/column_field.jl
+++ b/src/RingGrids/column_field.jl
@@ -1,0 +1,68 @@
+struct ColumnField{T, N, ArrayType <: AbstractArray, Grid <: AbstractGrid} <: AbstractField{T, N, ArrayType, Grid}
+    data::ArrayType
+    grid::Grid
+
+    function ColumnField(data, grid)
+        data_matches_grid(data, grid; horizontal_dim=2) || throw(DimensionMismatch(data, grid))
+        return new{eltype(data), ndims(data), typeof(data), typeof(grid)}(data, grid)
+    end
+end
+
+const ColumnField2D = ColumnField{T, 1} where T
+const ColumnField3D = ColumnField{T, 2} where T
+const ColumnField4D = ColumnField{T, 3} where T
+
+const FullColumnField = ColumnField{T, N, ArrayType, Grid} where {T, N, ArrayType, Grid<:AbstractFullGrid}
+const ReducedColumnField = ColumnField{T, N, ArrayType, Grid} where {T, N, ArrayType, Grid<:AbstractReducedGrid}
+
+# default constructors 
+ColumnField(grid::AbstractGrid, k...) = zeros(grid, k...)
+ColumnField(::Type{T}, grid::AbstractGrid, k...) where T = zeros(T, grid, k...)
+(::Type{<:ColumnField{T}})(data::AbstractArray, grid::AbstractGrid) where T = ColumnField(T.(data), grid)
+
+# TYPES
+Architectures.nonparametric_type(::Type{<:ColumnField}) = ColumnField
+grid_type(::Type{ColumnField{T, N, A, G}}) where {T, N, A, G} = G
+Architectures.array_type(::Type{ColumnField{T, N, A, G}}) where {T, N, A, G} = A
+
+# conversion from Field 
+transpose(field::Field) = transpose_safe(field)
+transpose!(field::Field) = transpose_unsafe!(field, similar(transpose(field.data)))
+
+# and back to Field 
+transpose(field::ColumnField) = transpose_safe(field)
+transpose!(field::ColumnField) = transpose_unsafe!(field, similar(transpose(field.data)))
+
+# safe version which allocates a new field
+function transpose_safe(field::Field)
+    data_transposed = collect(permutedims(field.data, (2, 1, 3:ndims(field)...)))
+    return ColumnField(data_transposed, field.grid)
+end
+
+# unsafe version that leaves the origional data corrupted (=transposed even though `field` doesn't indicate this)
+# but returns a ColumnField view onto the same memory
+function transpose_unsafe!(field::Field, scratch::AbstractArray)
+    @boundscheck size(transpose(scratch)) == size(field.data) || throw(BoundsError(scratch, field.data))
+    permutedims!(scratch, field.data, (2, 1, 3:ndims(field)...))    # transpose into scratch memory
+    vec(field.data) .= vec(scratch)               # copy back (scratch memory is free again after this)
+    return ColumnField(field.data, field.grid)    # view on the same data of field
+end
+
+# safe version which allocates a new field
+function transpose_safe(field::ColumnField)
+    data_transposed = collect(permutedims(field.data, (2, 1, 3:ndims(field)...)))
+    return Field(data_transposed, field.grid)
+end
+
+# unsafe version that leaves the origional data corrupted (=transposed even though `field` doesn't indicate this)
+# but returns a ColumnField view onto the same memory
+function transpose_unsafe!(field::ColumnField, scratch::AbstractArray)
+    @boundscheck size(transpose(scratch)) == size(field.data) || throw(BoundsError(scratch, field.data))
+    permutedims!(scratch, field.data, (2, 1, 3:ndims(field)...))    # transpose into scratch memory
+    vec(field.data) .= vec(scratch)               # copy back (scratch memory is free again after this)
+    return Field(field.data, field.grid)    # view on the same data of field
+end
+
+# full grids can be reshaped into a matrix, reduced grids cannot
+Base.Array(field::FullColumnField, as::Type{Matrix}) = Array(reshape(field.data, get_nlat(field), :, size(field)[2:end]...))
+
--- a/src/RingGrids/column_field.jl
+++ b/src/RingGrids/column_field.jl
@@ -1,3 +1,16 @@
+"""
+ColumnField is a version of `Field` with data stored in a column-major format. 
+It is used e.g. to represent data on a vertical column of a grid in column-based
+parametrizations. 
+$(TYPEDFIELDS)
+
+A ColumnField can be easily created from a `Field` by transposing it:
+
+```julia
+field = Field(grid, k...)
+column_field = transpose(field)
+```
+"""
 struct ColumnField{T, N, ArrayType <: AbstractArray, Grid <: AbstractGrid} <: AbstractField{T, N, ArrayType, Grid}
     data::ArrayType
     grid::Grid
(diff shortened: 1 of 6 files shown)


src/RingGrids/column_field.jl
```diff
@@ -0,0 +1,223 @@
...
+function transpose_safe(field::Field)
+    data_transposed = collect(permutedims(field.data, (2, 1, 3:ndims(field)...)))
+    return ColumnField(data_transposed, field.grid)
+end
+
+# unsafe version that leaves the origional data corrupted (=transposed even though `field` doesn't indicate this)
+# but returns a ColumnField view onto the same memory
+function transpose_unsafe!(field::Field, scratch::AbstractArray)
+    @boundscheck size(transpose(scratch)) == size(field.data) || throw(BoundsError(scratch, field.data))
+    permutedims!(scratch, field.data, (2, 1, 3:ndims(field)...))    # transpose into scratch memory
```
milankl: I'm wondering whether we could check for `length` only and `reshape` the `scratch` here to match the size we want as it just needs the same length, the shape is technically irrelevant. This way one can use `similar(transpose(field.data))` as scratch memory but also `similar(field.data)` and in general *any* other field of the same size as `field`. What do you think?

src/RingGrids/column_field.jl
```diff
@@ -0,0 +1,223 @@
...
+) where {F<:ColumnField{T, N, ArrayType}} where {T, N, ArrayType}
+    Grid = grid_type(F)
+    grid = nonparametric_type(Grid)(nlat_half, DEFAULT_ARCHITECTURE())
+    data = nonparametric_type(ArrayType){T}(undef, nlayers, get_npoints(grid), k...)
+    return ColumnField(data, grid)
+end
+
+## Some ARITHMETICS with regular Fields 
+add!(field::Field, column_field::ColumnField) = field .+= transpose(column_field)
+add!(column_field::ColumnField, field::Field) = column_field .+= transpose(field)
```
milankl: Do you not want to allow `field + column_field` or `field += column_field`? Not saying that we should just wondering if it's actually that bad!

milankl:
Looks great already!! I like it how you tried to reuse some of the field functionality but I guess in the end there's just some duplication we can't avoid or it would be too cumbersome to do. Big question for me is just though how we handle arithmetic operations between a field and a column field, like should `+`, `*`, etc just be allowed? With `A+B` one expects to allocate another array so +50% in memory what `A,B` already use. For `field + column_field` naively this would be +100% as the tranpose scratch memory also has to be allocated. I say "naively" because technically you can allocate the scratch and then accumulate field into it, so +50% also here. But then would we say field + column_field returns a `Field` so has precedence in type promotion?

src/RingGrids/column_field.jl
maximilian-gelbrecht: Should `field + column_field` be a `Field` or `ColumnField`? Probably a `Field`, but it's a bit poorly defined. Practically in our model code `add!` is probably what could be useful at times.

maximilian-gelbrecht:
I fixed the broadcasting issue. 

Wrt arithmetics: What do we need for the model? I would probably keep this quite minimal now, and just see what we need later. E.g. with multiplication it's also not clear to me immediately how we should define it. 

One thing I didn't adapt is the interpolation, I think that's just not necessary. But this means, I'll probably have to restrict the interpolation to only accept `Field` and not `AbstractField` anymore.

src/RingGrids/column_field.jl
maximilian-gelbrecht: Can be a bit unsafe in some situations to allow for an automatic reshape though, right?

milankl:
> Wrt arithmetics: What do we need for the model? I would probably keep this quite minimal now, and just see what we need later. E.g. with multiplication it's also not clear to me immediately how we should define it.

I don't think we need anything for the model because the idea isn't to mix field and column fields but use the latter only after explicit tranposes. We could completely skip the idea of `+`, `*` etc of field and column field to be honest. Makes our live easier and it would only be for convenient interactive playing around anyway

> One thing I didn't adapt is the interpolation, I think that's just not necessary. But this means, I'll probably have to restrict the interpolation to only accept Field and not AbstractField anymore.

Yes, I would do that too!

milankl:
Having said this, we could check whether another `tranpose` ... `tranpose` block could be put around the vertical advection. WENO is currently quite costly and it might be cheaper if the memory was reordered.

src/RingGrids/column_field.jl
milankl: I'm suggesting this because I was planning to use `diagnostic_variables.dynamics.a_grid`, `.b_grid` our existing scratch memories for that, they are a `Field` not a `ColumnField` ...

maximilian-gelbrecht:
Tests are passing now on my laptop, so this is almost ready. I wouldn't add a lot of arithmetics, at least not now. 

I am just not sure about the exact way to do the `transpose_unsafe!`, I took over your sketch from the issue, but I had to add a reshape, which I think made the whole idea with the scratch memory pointless? This is also related to your comment about reshaping in the constructor. 

I don't have further time today though, I'll look at it again later.

src/RingGrids/column_field.jl
milankl: yeah let's skip those for now

src/RingGrids/column_field.jl
milankl: What I meant is in [f0125f1](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/815/commits/f0125f122ebf51b3293d7fd7e44544da5124c06e) have a look whether you agree?

milankl:
> I am just not sure about the exact way to do the `transpose_unsafe!`, I took over your sketch from the issue, but I had to add a reshape, which I think made the whole idea with the scratch memory pointless? This is also related to your comment about reshaping in the constructor.

Did this in [f0125f1](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/815/commits/f0125f122ebf51b3293d7fd7e44544da5124c06e) I hope I didn't break anything! `reshape` is a view on the same data just with different dimensions as long as the length stays the same! So regardless the size of `scratch`, `reshape(scratch, size...)` will reshape the scratch memory how we need it -- as long as the length fits!

milankl:
Okay I understand that `transpose(transpose(vector))` is defined as unity, so a 3-element vector would become a 1x3 matrix but the singleton dimension is dropped in the backtranspose

```julia
julia> transpose(transpose([1,2,3]))
3-element Vector{Int64}:
 1
 2
 3

julia> transpose([1,2,3])
1×3 transpose(::Vector{Int64}) with eltype Int64:
 1  2  3
```

but the original size here is somehow stored, see how it says `Vector` despite being 1x3 ... we don't do that. In that sense maybe what we currently do, not removing the singleton trailing dimensions is best

```julia
julia> field
24-element, 4-ring OctaminimalGaussianField{Float32, 1} on Array on Main.RingGrids.Architectures.CPU{KernelAbstractions.CPU}
 0.0f0
...
 0.0f0

julia> transpose(field)
1×24, 4-ring ColumnField{Float32, 2, Matrix{Float32}, OctaminimalGaussianGrid{Main.RingGrids.Architectures.CPU{KernelAbstractions.CPU}, Vector{UnitRange{Int64}}, Vector{Int64}}}
 0.0f0  0.0f0  0.0f0  0.0f0  0.0f0  …  0.0f0  0.0f0  0.0f0  0.0f0  0.0f0

julia> transpose(transpose(field))
24×1, 4-ring OctaminimalGaussianField{Float32, 2} on Array on Main.RingGrids.Architectures.CPU{KernelAbstractions.CPU}
 0.0f0
...
 0.0f0
```
seee how 24-element transposes to 1x24, transposes to 24x1. Hence we have

```julia
julia> field == transpose(transpose(field))
false
```

which is consequent but counter intuitive. But it only arises with Field2D which transpose isn't made for anyway, so I'd be happy to just ignore that and note it here. One can always `transpose!` in-place and reuse the old reference again without the trailing singleton dimension.

maximilian-gelbrecht:
Ok, some minor issues still need to be resolved, but as far as I can see it this new structure should work in Julia 1.11 without any issues and SpeedyWeather (main module) always uses the local versions in the repo through the `[sources]`. 

In Julia 1.10 this causes an error at the moment, but as soon as we register `LowerTriangularArrays`, `RingGrids`, `SpeedyTransforms` it would install the registered version. That's better than nothing, but not really great because we would need to register new versions of those packages for every change we make that we want to use in the main package in Julia 1.10. This is kind of what we wanted to avoid. After trying out a lot of things this might be unavoidable though... 

@bgroenks96  Somehow through this restructuring broke two unit tests from SpeedyParamters though, and I really struggle to see why, because we didn't really do any changes to it that I am aware of.

bgroenks96:
Does it break only on 1.11? Or also on 1.10?

maximilian-gelbrecht:
Both

maximilian-gelbrecht:
Okay, seems as everything is resolved. As soon as this is merged, we should immediately register the submodules.

maximilian-gelbrecht:
I noticed one problem. This only works in 1.10 if we also register Architectures. I am not sure we really want that as it is. But how about we do a SpeedyWeatherCore in which we put Architectures and Utils and register that one? In the future we could also put some other stuff in there

milankl:
Happy with that! Though I would find "Core" confusing as it's not really core functionality. What about "Internals"?

maximilian-gelbrecht:
Ok... SpeedyInternals is there. 

So, are we happy with this now? Then, the docs would need to be adjusted to reflect this new structure. Then merge, then register.

milankl:
I can do a review later?

milankl:
Just checked, I can develop SpeedyWeather and any submodules independently in VS code on this branch with no package or scope issues. Fantastic!

maximilian-gelbrecht:
I revised the readme and the leading paragraph of the docs of the sub-packages. Anything else that we to adjust in the docs?

milankl:
Do we want to add a `README.md` at `src/LowerTriangularArrays/README.md` and similar for other packages too? Maybe they are even required by the registrator?

maximilian-gelbrecht:
Yeah, we could do that. I am also not sure what exactly the registrator even requires for packages within other repos, the manual doesn't even mention that it's possible to register submodules at all.

milankl:
Sorry not done yet with the readmes ...

maximilian-gelbrecht:
Ok, I finished the readmes now.

maximilian-gelbrecht:
I think what's still missing for registering the sub packages is a license file in each subfolder. Our current license is including e.g. the copyright notice for the fortran speedy parametrization. I don't want to copy that to every submodule, but we have to keep the MIT licence obviously.  

Shall we just use a standard MIT licence with 
"The SpeedyWeather.jl Contributors for SpeedyWeather.jl" in each of the submodules?

milankl:
Or is that the moment where we think about using EUPL? I would like to have some copyleft aspect in our license and even if it's a weak one

maximilian-gelbrecht:
We would need the agreement of all contributors to the submodules though, right?

bgroenks96:
Yes, unless you get the consent of all contributors, you need to keep it under the same license. I would recommend keeping MIT on everything in the Speedy monorepo for now.

milankl:
Okay happy with that

milankl:
Merge now? 😄

bgroenks96:
I think you should do the honors @milankl!

PR #816, opened 2025-08-08
Make dynamic Land/Ocean/SeaIce models the default

pierluigividale:
- Changes a few minor things for the land surface
- Makes the `LandBucketTemperature`, `LandBucketMoisture` defaults for land model
- the `SlabOcean` for ocean models
- and `ThermodynamicSeaIce` default for sea ice model


milankl:
~~I want to merge #814 into here first~~ well at least #814 needs to be in main first

milankl:
With #814 merged rerunning the tests here now. I also want to check for CI time. If CI takes too long it gets annoying to wait and so we have to "clean up" the tests by checking which take long and how we can test for the same things in shorter time. Most of what CI is spending time on is compile time so unfortunately that gets really tricky ... 🙈 but we aim for check to be completed in 15min. In contrast, 30min is already ... 😴

milankl:
https://speedyweather.github.io/SpeedyWeatherDocumentation/previews/PR816

We should also check the documentation now as many things now run ocean, land and sea ice by default!

milankl:
@maximilian-gelbrecht do you understand why Enzyme fails here?

pierluigividale:
> With #814 merged rerunning the tests here now. I also want to check for CI time. If CI takes too long it gets annoying to wait and so we have to "clean up" the tests by checking which take long and how we can test for the same things in shorter time. Most of what CI is spending time on is compile time so unfortunately that gets really tricky ... 🙈 but we aim for check to be completed in 15min. In contrast, 30min is already ... 😴

The worst of this is, the new generation of students shall never fully appreciate how such expectations are infuriatingly hilarious, and what it should have been all along.

pierluigividale:
Right, I have checked the documentation under Main and all looks as we set it in the branch, but there are a few inconsistencies with the different sections of the documentation, e.g. on what the ocean is under examples versus what it is under model dynamics/physics definitions. I think that we need to take a little time to think about it and come up with a strategy, also for the future, as I continue to want to add a (terrestrial) snow module.

milankl:
@pierluigividale is there anything you still want to change to have new default within the land model?

pierluigividale:
I am happy with the current setup.

PR #817, opened 2025-08-19
Snow fall from large-scale condensation

pierluigividale:
I have added snow diagnostic variables and a few rules for deciding whether precipitation is rain or snow.
My code is very likely full of syntax errors and/or naive programming, eg. introducing IF statements that are likely to compromise performance.

If someone can give me pointers on how to do this correctly, I can then try to see whether I see a diagnostic.
After that, I would like to introduce a parametrisation similar to the ice parametrisation, but the initial triggers must be BOTH: accumulated snow and soil temperature.

TODO before merge

- [x] documentation (maths + interface)
- [x] tests (`snow=false` has zero snow, otherwise some snow, and `reevaporation=0` also possible)
- [x] decide what to do with snow in convection (skip for now)
- [x] consistent naming of precipitation, rain, snow in variables and output


src/physics/large_scale_condensation.jl
```diff
@@ -112,5 +113,8 @@ function large_scale_condensation!(
 
     # convert to rain rate [m/s]
     column.precip_rate_large_scale = column.precip_large_scale / Δt_sec
+    if temp[:] < freezing_threshold
+        column.snow_rate_large_scale = column.precip_rate_large_scale
+        column.precip_rate_large_scale = 0.
```
milankl: Could you describe with words what is supposed to happen here? I read this as "if the temperature anywhere in the column is below the freezing threshold then declare precipitation to be snow and not water"

I'm wondering what you want is actually that if condensation in one layer occurs where the temperature is below the freezing threshold then it should come down as snow. That way you could have precipitation from one layer that is snow from another that is rain (=sleet?).

I'd probably write this as

```julia
# decide whether to declare precip (rain water) as snow
snow, precip = temp[k] < freezing_threshold ? (precip, zero(precip)) : (zero(precip), precip)
column.precip_large_scale += precip
column.snow_large_scale += snow
```

at the moment the problem is only the `temp[:]` syntax, that's python. `temp` is the vector of the vertical temperature profile, we don't `[:]` flatten it. And you're missing the `end` statement to the `if` clause.

src/output/variables/precipitation.jl
```diff
@@ -92,6 +92,105 @@ Fields are: $(TYPEDFIELDS)"""
...
+    shuffle::Bool = true
+    keepbits::Int = 7
+    transform::F = (x) -> 1000x     # [m] to [mm]
+end
+
+function output!(
+    output::NetCDFOutput,
+    variable::AbstractSnowRateOutputVariable,
+    acc_variable::AbstractOutputVariable,
+)
```
milankl: This is exactly the same as for rain precipitation no? Then we can just make this reuse that code and we should call things `AbstractRateOutputVariable` for consistency, whether its snow or rain, it doesn't matter.

src/dynamics/diagnostic_variables.jl
```diff
@@ -343,6 +343,19 @@ $(TYPEDFIELDS)"""
...
+
+    "Accumulated large-scale precipitation [m]"
+    snow_convection::GridVariable2D = zeros(GridVariable2D, grid)
+
+    "Rate of large-scale precipitation [m/s], instantaneous"
+    snow_rate_large_scale::GridVariable2D = zeros(GridVariable2D, grid)
+
+    "Rate of large-scale precipitation [m/s], instantaneous"
+    snow_rate_convection::GridVariable2D = zeros(GridVariable2D, grid)
+
```
milankl: yes this looks right!

pierluigividale:
Thanks, these are great help.

I was just trying to test out the insertion of new variables and output.
I have not yet thought what to do, but the main issue, as you can see below, is that I lack the syntax skills to implement ideas. The below helps greatly, I can now try and do something that makes sense.

In the mean time I have turned sea_ice.jl into snow,jl, but I am not convinced that it makes sense, as the snowfall trigger is a flickery one, unlike SSTs.

One idea could be to do this, but in order to do something meaningful I need to draw it on paper, can’t think physics from the code:

milankl:
(copying this in here as email responses do not support markdown)


```julia
@kernel inbounds=true function snow_kernel!(snow, st, mask, @Const(temp_freeze), @Const(m), @Const(f_Δt), @Const(Δt))
    ij = @index(Global, Linear)    # every grid point ij

    if mask[ij] = 1 && isfinite(sst[ij])       # Only INLAND, ST not NaN (=masked) 

        # Snow-ST flux as a relaxation term wrt to freezing, with different melt/freeze rates
        dT = st[1,ij] - temp_freeze              # uncorrected difference to freezing temperature
		if (snow_large_scale+snow_convective)  > snowfall_threshold      # check for significant snowfall
        	F = - f_Δt*min(dT, 0)           # freeze if below
			st[1,ij] = max(st[1,ij], temp_freeze)    # cap sst at freezing
		else                                # melt if above freezing, 
			F = -m*max(dT, 0)
		end if
        # Euler forward step to update snow concentration, cap between [0, 1]
        snow[ij] += Δt*F
        snow[ij] = max(min(snow[ij], 1), 0)
    end
end
```

milankl:
Some general questions: 

- Do we need to distinguish between snow fall/precipitation (units of m, or a rate of m/s) and snow cover on the ground (also units of m, even though that's probably meters of water if melted?). The latter I see as an analogy to a sea ice thickness.
- Given the `snow[ij] = max(min(snow[ij], 1), 0)` line it seems you are thinking of snow cover as an area fraction, but I am not sure that makes sense?
- Would albedo simply be set to snow if snow cover > 0? We wouldn't distinguish between fresh snow and old snow?
- Why would the soil temperature be kept at freezing? For SST, ocean shouldn't be supercooled, but soil temperatures could be below freezing? The problem I see at the moment in `st[1,ij] = max(st[1,ij], temp_freeze)` that you could have soil of -40C without snow, once it snows the soil temperature would jump up to freezing!
- For snow covered soil, would you also want to insulate the land by reducing the land-air fluxes?

milankl:
Also to be clear, I would let this pull request just implement the snow precipitation, if that works we can make a new pull request for snow cover that actually influences the land. Also at the moment writing the snow fall back into the gridded fields is missing, this needs to be added here

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/96357e66538e4247b3e310cba7a2d3b9f45d9239/src/physics/column_variables.jl#L134-L140

milankl:
And maybe a more general question, would we want to be able to switch snow on/off? I believe so yes, maybe the easiest to have a flag `snow::Bool = true` in `ImplicitCondensation` and `SimplifiedBettsMiller` that way we could disable snow at any time or disable it only for convection etc.

pierluigividale:
Hi

So, the ideal thing to do would be to create a snow depth variable, but I did not want to do that before consulting you: not sure what the computational cost is of creating more and more variables, and more and more output streams.

Then, if we have snow depth, there are quite well established formulas that go from snow depth to area coverage of snow: fraction of a grid box covered by snow.

Albedo on land will depend on the fraction of bare soil, the fraction of vegetation, and the fraction of snow cover. For now, the first two are not dealt with for now, they are constant in time, but if we introduce snow, we must have a transition from 0 to 100% snow cover.

We do not want to go snow / no-snow for albedo: that would introduce flickering heat fluxes, looks horrible in time series of the entire surface energy balance, and is prone to generate numerical instability.

With regards to snow freezing thawing, the right thing to do would be to:
a) change the thermal conductivity of the top layer if there is snow cover
b) compute the latent heat exchange when snow melts and to add that to the surface energy balance, but I am not terribly keen to do that to a 20cm soil. This is why I was thinking of an interval in which soil is kept at freezing, but prescribing that analytically is probably not good value for money.

Yes, for the time being I was thinking of thinking of insulation: some way for just damping the soil fluxes. But maybe the best way is to do it in a realistic way. Just worried about costs.

Cheers,
PL

pierluigividale:
Hi

Indeed, for now I just wanted to test having some snowfall. No surface snow scheme at this stage.
As you will see from the other message, I have not designed a surface snow scheme yet.

Thanks for the other pointers, the question on what else needs to be modified was primary, let me try and do all those things, then test that we have some snowfall.

Cheers,
PL

pierluigividale:
Yes, shall do that too.

Cheers,
PL

pierluigividale:
OK, I have added all sorts of changes, fixes, hopefully the way you suggested I do.

milankl:
You had the `temp[k] < temp_freezing` check outside of the vertical loop over `k`. I moved this in, so it only snows if the temperature in the layer `k` where convection/condensation occurs. This means that snow wouldn't melt when falling through warmer layers or rain doesn't turn into snow when falling through colder layers.

src/physics/convection.jl
```diff
@@ -45,16 +45,19 @@ export SimplifiedBettsMiller
...
 
-    "[OPTION] Relative humidity for reference profile"
+    "[OPTION] Relative humidity for reference profile [1]"
     relative_humidity::NF = 0.7
 
+    "[OPTION] Convert rain fall to snow below freezing?"
+    snow::Bool = true
+
+    "[OPTION] Freezing temperature below which to convert rain into snow [K]"
+    freezing_threshold::NF = 263
```
milankl: When introducing new parameters into a `struct` don't forget to add a descriptive string in the layer above. This will automatically be picked up by the documentation and by the `?` help!

src/physics/convection.jl
```diff
@@ -45,16 +45,19 @@ export SimplifiedBettsMiller
...
+@kwdef mutable struct SimplifiedBettsMiller{NF, SP} <: AbstractConvection
     "[OPTION] Relaxation time for profile adjustment"
     time_scale::Second = Hour(4)
 
-    "[OPTION] Relative humidity for reference profile"
+    "[OPTION] Relative humidity for reference profile [1]"
     relative_humidity::NF = 0.7
 
+    "[OPTION] Convert rain fall to snow below freezing?"
+    snow::Bool = true
```
milankl: this allows to return to `convection = SimplifiedBettsMiller(spectral_grid, snow=false)`

src/physics/define_column.jl
```diff
@@ -91,10 +91,16 @@ $(TYPEDFIELDS)"""
...
     precip_large_scale::NF = 0              # precipitation due to large-scale condensation [m]
-    precip_rate_convection::NF = 0          # Precipitation rate due to convection [m/s]
+    precip_rate_convection::NF = 0          # precipitation rate due to convection [m/s]
     precip_rate_large_scale::NF = 0         # precipitation rate due to large-scale condensation [m/s]
+    
+    # same for snow
+    snow_convection::NF = 0                 # snow due to convection [m]
+    snow_large_scale::NF = 0                # snow due to large-scale condensation [m]
+    snow_rate_convection::NF = 0            # snow rate due to convection [m/s]
+    snow_rate_large_scale::NF = 0           # snow rate due to large-scale condensation [m/s]
```
milankl, line 103: Every column also needed these!

milankl (commented):
Just some comments on what I changed compared to your version!

milankl:
> not sure what the computational cost is of creating more and more variables, and more and more output streams.

This is 2D and independent columns, so it scales linearly with the number of horizontal grid cells and is a constant with the vertical resolution, so don't worry about costs and memory. If we're worried about speed we have to invest more into the transforms or maybe write a semi-Lagrangian advection scheme one day.

src/physics/convection.jl
pierluigividale: Ah, yes, I should have remembered, you taught me about 10 days ago...

src/physics/convection.jl
pierluigividale: OK, thanks, looked around, was not sure where to insert it.

src/physics/define_column.jl
pierluigividale: So, are these just allocations, or is it an actual IC?

pierluigividale:
> > not sure what the computational cost is of creating more and more variables, and more and more output streams.
> 
> This is 2D and independent columns, so it scales linearly with the number of horizontal grid cells and is a constant with the vertical resolution, so don't worry about costs and memory. If we're worried about speed we have to invest more into the transforms or maybe write a semi-Lagrangian advection scheme one day.

OK, cool. I went walking for a bit, and I was thinking what to do here. Good to know that I do not need to worry.

pierluigividale:
> You had the `temp[k] < temp_freezing` check outside of the vertical loop over `k`. I moved this in, so it only snows if the temperature in the layer `k` where convection/condensation occurs. This means that snow wouldn't melt when falling through warmer layers or rain doesn't turn into snow when falling through colder layers.

OUCH, I put a note on my yellow pad to ask you, then I completely forgot. Thanks.

milankl:
Let it snow, let it snow, let it snow. But it snows in funny places! 1 Month (March) accumulated after some spinup in Jan+Feb

image

image

milankl:
I reckon in order to resolve this we generally have to think about reevaporation and thawing snow while it falls ...

src/physics/define_column.jl
milankl: Yes, not IC. We currently use `ColumnVariables` to extract one column before the parameterizations for that column are evaluated, this makes sure we have somewhat contiguous memory access as the parameterizations access things in the vertical which is far apart in memory. You see this logic here

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/2085fa1badd20e052f1ac584af99ae27ff32698b/src/physics/tendencies.jl#L20-L36

you reset the column memory, get the data from the diagnostic variables read it into the column, evaluate the parameterzations on it then write back into the tendencies/diagnostics and reuse the same column memory for the next grid point.

pierluigividale:
> Let it snow, let it snow, let it snow. But it snows in funny places! 1 Month (March) accumulated after some spinup in Jan+Feb
> 
> image image

You stole my planned Issues title!!! :)

pierluigividale:
Yes, I think that the problem here is that we are scanning the entire column, which of course is going to cause quite a bit of frozen precipitation to be diagnosed. In reality, as you say, when such precipitation starts falling, it will melt.
For now, maybe, we could limit the levels over which we make the simple threshold-based decision?

pierluigividale:
Do you really want to S&M, or would it be best to re-think the criterion?

pierluigividale:
OK, is this similar to allocate/deallocate in F90?

PL

milankl:
> OK, is this similar to allocate/deallocate in F90? PL

Not sure. A column is only created once and sits in `simulation.diagnostic_variables.column` and is reused for every horizontal grid point.

milankl:
> Do you really want to S&M, or would it be best to re-think the criterion?

I think we should rework how reevaporation + melting + freezing on the way down should work. Like wouldn't it be a better assumption to say snow is created below some freezing threshold but the column to the ground has to be below freezing (0˚C) for snow to reach the ground, otherwise convert it back to precipitation? Relatedly, what happens if convection occurs in a layer with excess humidity. This falls down a rain but if it falls through a layer where the convection would cause drying then does the rain reevaporate?

milankl:
[Frierson et al. 2006](https://journals.ametsoc.org/view/journals/atsc/63/10/jas3753.1.xml) do this

image

I don't want this to be too complicated but there might be something simple yet more sophisticated possible

pierluigividale:
What you describe here is the correct (macro)physics.

Happy to read that paper and try to implement.

Cheers,
PL

milankl:
Converting to draft to revise condensation and convection wrt reevaporation & co as discussed above

src/physics/large_scale_condensation.jl
```diff
@@ -113,6 +113,19 @@ function large_scale_condensation!(
...
             precip, snow = let_it_snow && temp[k] < freezing_threshold ? (snow, precip) : (precip, snow)
+            
+            # --- NEW: latent heat and temperature changes due to freezing ---
+            if snow > 0
+                # equivalent humidity change associated with snow (reverse of precip scaling)
+                δq_freeze = -snow / (Δσ[k] * pₛΔt_gρ)
+
+                # temperature tendency from freezing
+                δT_freeze = -(Lᵢ / cₚ) * δq_freeze
+                temp_tend[k] += δT_freeze
```
milankl: Now we do
```julia
# latent heat release for enthalpy conservation
δT = -Lᵥ_cₚ * δq

# temperature tendency from freezing
δT_freeze = -(Lᵢ / cₚ) * δq_freeze
temp_tend[k] += δT_freeze

temp_tend[k] += δT
```

wouldn't it make more sense to do

```julia
δT = -Lᵥ_cₚ * δq
δT += -(Lᵢ / cₚ) * δq_freeze
temp_tend[k] += δT
```

milankl: But could we not just decide in `δT = -Lᵥ_cₚ * δq` which latent heat to use? First calculating precipitation and then going back to a temperature change feels like a detour?

pierluigividale: So, I have left it that open because you seemed to want to also look into re-evaporation and/or sublimation.

For now, if we are just dealing with going from vapour to liquid and/or ice, we could indeed simplify. 

Make L depend on snow. I am going out on a limb here with my understanding of Julia syntax:

L = snow > 0 ? (Lᵥ + Lᵢ) : Lᵥ
δT = -L / cₚ * δq
temp_tend[k] += δT

pierluigividale: which also requires commenting out the first δT calculation

milankl:
Yes I would like to write a control on reevaporation. Maybe just `reevaporation=1` means everything reevaporates in the cell below and one recalculates the latent heat release from vapourization and sublimation and `=0` means it just falls through. And then we can set it to some value maybe 1/2 to start with?

pierluigividale:
OK, for now I have developed something to allow snow to melt and to add the enthalpy to the layer where it happens. It required setting up a snow_flux to account for what comes from above.

I am sure that I have violated all sorts of good programming rules, please have a look at the commit I made just now, which should have this code in it:


# decide whether to turn precip into snow
            precip, snow = let_it_snow && temp[k] < freezing_threshold ? (snow, precip) : (precip, snow)
            snow_flux += snow
            # latent heat release for enthalpy conservation
            L = snow > 0 ? (Lᵥ + Lᵢ) : Lᵥ
            δT = -L / cₚ * δq

            # only accumulate into humid_tend now to allow humid_tend != 0 before this scheme is called
            humid_tend[k] += δq
            temp_tend[k] += δT

            # now test whether any snow has come into this layer as snow_flux and whether the current layer can melt it
            δT_melt = temp[k] - melting_threshold
            if snow_flux > 0 && δT_melt > 0 
                melting_enthalpy = cₚ * δT_melt 
                δq = melting_enthalpy / Lᵢ
                snow   -= δq
                precip += δq
                temp_tend[k] -= δT_melt

pierluigividale:
Something has happened and the system blocks me from committing, claiming that I have committed already.

But I have found I had forgotten about units, as well as forgotten to check that I have enough snow to melt, so I would like to commit this code:

# now test whether any snow has come into this layer as snow_flux and whether the current layer can melt it
            δT_melt = temp[k] - melting_threshold
            if snow_flux > 0 && δT_melt > 0 
                E_avail = cₚ * δT_melt   # J / kg (of air)
                # maximum snow depth that can be melted with that energy
                melt_depth = (E_avail / Lᵢ) * (Δσ[k] * pₛΔt_gρ)
                # actual melt limited by available snow
                melt_amount = min(snow_flux, melt_depth)
                δq_melt = melt_amount / (Δσ[k] * pₛΔt_gρ)
                snow_flux    -= δq_melt
                precip       += δq_melt
                temp_tend[k] -= δT_melt

pierluigividale:
Finally got the code to commit and it seems to work.

pierluigividale:
Some of the physics must be correct, as it snows where the mountains are.

milankl:
Awesome work already @pierluigividale thanks for writing down the algorithm for the snow melting. I've only moved it up, renamed a few things or changed the order of some computations. I like the idea of first calculating the available energy for melting. I've also implemented some reevaporation now that's proportional to the dryness (saturation minus actual humidity) in that cell. And decided to have condensation + reevaporation + melting all implicitly solved such that also what melts can immediately reevaporate which should be equivalent to sublimation. The implicit time stepping is important so that we can condense excess water vapour quickly without numerical instabilities.

I haven't tested this yet but already wanted to push the code. I hope it's all consistent and well documented, there's one tuning parameter left, `reevaporation` which dictates the efficiency of reevaporation, I guess in reality this depends on things like fall speed and updrafts and what not but thought to keep it simple here. We're already introducing enough new complexity.

pierluigividale:
OK, excellent, let us see whether it all passes the tests. Have you looked at the convection module? I put similar code there, but I think that it needs work. Also, please note, as I have written in a separate e-mail, that I have inserted a statement for updating the heating, but I fear that it is either a duplication or a statement in the wrong place.

pierluigividale:
Forgot to say: I strongly support treating processes implicitly whenever possible, which will help avoid instabilities. I also do like trying to do the full set of processes at the same time, else things can become circular and we can end up with lots of flickering tendencies.

milankl:
20 days accumulated snow (from large scale only, convection back to default) after 20 days of spin up. Only snows in mountaineous areas, check, and at high latitudes (this is NH winter), check. But it also snows a little in Africa which seems to be at night because it cools down quite a lot. Maybe something that will change if run for longer

image

We have some 100mm of snow in the Andes within those 20 days, that's 10cm of melted water and somewhere between 30-300cm of snow depth (density 3 to 30x lower). Sounds good enough for now!

pierluigividale:
Yes, I think that what we have is good enough for now. I will now focus on the surface scheme.
What needs doing for this to end up in Main? I have not seen a Squash and Merge window.

milankl:
We need to see how adapt the same ideas to convection and make it consistent. Or we only have snow from condensation also fine with me. 

Then we need to update the docs with the new maths and additional options for users. And add some tests using the snow and reevaporation options. Testing snow I would just check that snow is zero if option snow is false. 

And I would like to do a quick performance check how expensive condensation now is compared to convection.

milankl:
Snow fall April at T63L8

https://github.com/user-attachments/assets/1fae1066-5ad0-4890-82c1-f567a61df36a

milankl:
And snow fall in Jan year 2, looks stable to me, that's good. More snow in central Asia and North America, looks good to me.

https://github.com/user-attachments/assets/7a4a85c4-39f8-4ec6-aa4f-6a3a84e20c00

What's not clear to me is how we would do snow in convection, it's a little trickier to conserve enthalpy there in the BettsMiller scheme. But maybe it's good enough to start to only have snow from large-scale condensation?

Performance is also okay: Depending on how much there's to condense a L8 column takes 20-100ns and zero allocations so that's all fine

pierluigividale:
There is convective snow in nature, so it is not a bad objective to have. However, if, as you say, it is difficult to figure out the energetics, we should pause and study up before we introduce poor physics.

I can implement the ground snow cover with just the input from the large scale snow.

milankl:
I've started writing some docs but then also realised some bugs in the code and so this pull request stays a draft until this is all sorted, sorry! Feel free to have a look at the docs already though or contribute writing the last bits of the math! I'll get back to this later this evening or tomorrow (long train ride)

pierluigividale:
OK, very busy with a proposal and project meetings today, but I can have a go at the docs tomorrow.

milankl:
Checking the impact of reevaporation, initial 10 days of the simulation, top no reevaporation (`reevaporation=0`) and full evaporation (there's no easy "maximum" one can choose, I believe `30` is "medium" but `reevaporation=1e9` is certainly full reevaporation)

image

reevaporation reduces large-scale precipitation, over mountains but also particularly removes it over (I guess) warm surfaces. Doesn't have a very obvious impact on snow (neither on convection). Not sure how to choose the reevaporation strength to be honest

image

pierluigividale:
The beauty of parametrisation, we end up with parameters, and no idea what to set them to, because such things do not exist in nature. Set it to 30 for now. I don't think we should try tuning anything until we have the new radiation parametrisation.

I have read through the documentation, and I have made a couple of suggestions. I think that it would be better to say re-evaporation, it is more readable, but then the flag is reevaporation. Not sure we can put a hyphen in a flag name?

milankl:
> The beauty of parametrisation, we end up with parameters, and no idea what to set them to, because such things do not exist in nature. Set it to 30 for now. I don't think we should try tuning anything until we have the new radiation parametrisation.

Agreed. 

> I have read through the documentation, and I have made a couple of suggestions. I think that it would be better to say re-evaporation, it is more readable, but then the flag is reevaporation. Not sure we can put a hyphen in a flag name?

No we can't. When you change a heading like `## Reevaporation` to `## Re-evaporation` then also check for links to that section which will look like `[Re-evaporation](@ref)`, I've changed this already following your heading rename but otherwise the docs built will fail. Thanks for having a look!

milankl:
Okay all checks pass, we have docs and tests and will make `snow=true` the default with this. Remaining question I have is on terminology. We now use 

- precipitation for rain, not snow because that's what we called it before we had snow
- rain for liquid precipitation but that is only really used internally or in the docs
- snow for frozen precipitation

Should we use `precipitation` now for rain+snow? And rename everything that's called `precip` in the output netCDFs `rain`? But then retain `total_precipitation` for the sum of rain+snow?

pierluigividale:
OK, all species that fall from the sky are precipitation, so all sorts of drops, ice crystals etc. 
I would say precipitation = liquid _ frozen (rain + snow + ...)
Here we are very liberal with the thing that we call snow, it is actually frozen precipitation, but for the time being, without a microphysics scheme, that is all we can do.

So:
1) rename precipitation as rain in NetCDF output
2) total_precipitation = large_scale_rain + convective_rain + large_scale_snow + convective_snow (does not matter we do not have that one yet).

pierluigividale:
Should I wait to squash and merge? I would do those things now, but I am on family duty.

milankl:
Let me do these renamings first so it's together. I did introduce a `total_precipitation_rate` for coupling purposes. We can add output for it too. But it'll have units of kg/m2/s

EDIT: okay I hope I renamed everything correctly, let's see what CI says.

pierluigividale:
Hmmm, I was about to start, I see you have done it already, but checks are failing. Shall wait a little longer for the end.

pierluigividale:
Ah yes, important about the naming coupling variable.

pierluigividale:
It is finally time to press the button!

PR #822, opened 2025-08-22
Introduce `nothing` as valid model component

milankl:
So that `convection = nothing` etc is possible, removes a lot of code redundancy as `NoSeaIce`, `NoConvection` etc don't have to be defined anymore. Current exceptions are

- `NoOrography` which contains zeros arrays
- `NoVegetation` as this still needs to compute the soil moisture availability and dispatching over `Nothing` here seems ambiguous (maybe there's a way to resolve this but maybe not so important for now?)

### DIFF
--- a/docs/src/examples_3D.md
+++ b/docs/src/examples_3D.md
@@ -62,14 +62,14 @@ model = PrimitiveDryModel(
     boundary_layer_drag = LinearDrag(spectral_grid),
 
     # switch off other physics
-    convection = NoConvection(),
-    shortwave_radiation = NoShortwave(),
-    longwave_radiation = NoLongwave(),
-    vertical_diffusion = NoVerticalDiffusion(),
+    convection = nothing,
+    shortwave_radiation = nothing,
+    longwave_radiation = nothing,
+    vertical_diffusion = nothing,
 
     # switch off surface fluxes (makes ocean/land/land-sea mask redundant)
-    surface_wind = NoSurfaceWind(),
-    surface_heat_flux = NoSurfaceHeatFlux(),
+    surface_wind = nothing,
+    surface_heat_flux = nothing,
 
     # use Earth's orography
     orography = EarthOrography(spectral_grid)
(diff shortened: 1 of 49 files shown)


milankl:
@maximilian-gelbrecht any objections? Oceananigans does that too, makes a good amount of code and types redundant and might be a more intuitive user interface?

src/dynamics/drag.jl
```diff
@@ -11,17 +11,7 @@ function drag!(
...
-NoDrag(SG::SpectralGrid) = NoDrag()
-initialize!(::NoDrag, ::AbstractModel) = nothing
-
-function drag!(     diagn::DiagnosticVariables,
-                    progn::PrognosticVariables,
-                    drag::NoDrag,
-                    args...)
-    return nothing
-end
+drag!(diagn, progn, drag::Nothing, args...) = nothing
```
maximilian-gelbrecht, line 14: oh yeah, I like that

docs/src/examples_3D.md
```diff
@@ -85,12 +85,12 @@ and a linear drag term that is applied near the planetary boundary but switches
...
-to the surface wind (which is zero with `NoSurfaceWind`).
-But to also avoid the calculation being run at all we use `NoSurfaceHeatFlux()`
-for the model constructor.
-Many of the `NoSomething` model components do not require the
-spectral grid to be passed on, but as a convention we allow every model component
-to have it for construction even if not required.
+to the surface wind (which is zero with `nothing`).
+But to also avoid the calculation being run at all we use `nothing` for model components
+passed to the model constructor.
+Some model components use a `NoSomething` but most can just be set to `nothing`.
```
maximilian-gelbrecht, line 91: Might be worth to point out the actual exceptions somewhere? This way it could be a bit confusing to users.

src/models/abstract_models.jl
```diff
@@ -9,6 +9,21 @@ abstract type PrimitiveDry <: PrimitiveEquation end
 abstract type PrimitiveWet <: PrimitiveEquation end
 
 abstract type AbstractModelComponent end
+abstract type AbstractParameterization <: AbstractModelComponent end
+
+# any model component set to nothing needs no initialization or finalize!
+initialize!(::Nothing, ::AbstractModel) = nothing
```
maximilian-gelbrecht, line 15: nice!

maximilian-gelbrecht (approved):
This great. I really like all the code simplification.

PR #823, opened 2025-08-22
ice thermodynamics in Clausius Clapeyron

pierluigividale:
I have added Clausius Clapeyron variables in thermodynamics, and added a small energy correction in large_scale_condensation.

Still thinking about the loop to compute what happens to snow as it falls through the layers.


milankl:
this pull request goes from `main` to `plv/snow` (not into main!) and just includes the four last pull requests that are in main now. Do you want to me resolve the conflicts and merge this so you have merged the current state of main into `plv/snow`?

milankl:
I think this pull request is not needed given #817 ?

pierluigividale:
Thanks, I think that I made a mess of things, good that you could sort it out.

Hopefully I can clean house now.

Cheers,
PL

PR #824, opened 2025-08-23
Reduce GPU overhead

maximilian-gelbrecht:
This PR reduces some GPU overhead. 

* It removes the explicit `synchronize` that are not needed (I still need to fully test that though) 
* It revised `on_architecture`/`adapt` a bit in a way more similar to Oceananigans as well. Basically `adapt` really only adapts the data arrays for `Field` and `LowerTriangularArrays` to save any overhead of adapting `Grid`. I'll test that here, maybe some kernels to be adjusted for that again. For actual data transfer between devices `on_architecture` is used then.


maximilian-gelbrecht:
Note to myself, with less overhead, we could also revisit if splitting some of the spectral gradient kernels in multiple kernels is worthwhile

maximilian-gelbrecht:
All of the recent commits here are just a messy merge of main into this branch. But I try to work on finishing this PR next week as well.

maximilian-gelbrecht:
This is pretty much finished. I have to double check with Nsys but otherwise it's good to go.

maximilian-gelbrecht:
Ok, I had a look at a Barotropic model execution in Nvidia Nsight. There's still a lot of CPU time and gaps between GPU executions that we have to look into, but the PR is already a step in the right direction. Also broadcasting, ffts and KA kernels are all in the same stream which is what we want. So I would say this can be merged.

maximilian-gelbrecht:
@milankl Could you have a short look at this, and then we merge this PR?

milankl (commented):
Looks great to me, just to summarize for future development

- use `on_architecture` if you need to move fields/LTAs between devices, e.g. you create on CPU during initialize and then move to the GPU for time stepping, like in https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/726/commits/070bd60336bd7102d28844d8f7fdd26202e0e4ee ?
- don't `synchronize!` after `launch!` (conceptually a wait-all happens after every kernel launch? or can different kernels now overlap?)
- everything is already correct in our Metal extension?

Feel free to merge!

maximilian-gelbrecht:
Thanks! 

> Looks great to me, just to summarize for future development
> 
> * use `on_architecture` if you need to move fields/LTAs between devices, e.g. you create on CPU during initialize and then move to the GPU for time stepping, like in [070bd60(https://github.com/SpeedyWeather/SpeedyWeather.jl/commit/070bd60336bd7102d28844d8f7fdd26202e0e4ee) ?

Yes! 

> * don't `synchronize!` after `launch!` (conceptually a wait-all happens after every kernel launch? or can different kernels now overlap?)

All kernels are launched in the same stream (I've checked that with Nvidia Nsight), so they won't overlap and are always executed sequentially. It's now just that some other CPU code can be executed in the meanwhile. 

> * everything is already correct in our Metal extension?

Yes, looks good!

PR #830, opened 2025-09-01
Start with soil moisture, temperature, SST, sea ice from climatology

milankl:
@pierluigividale at the moment we start with 0.5 soil moisture also over ocean which doesn't really make sense to me. It's fine to start with 0.5 over land, but I think for ocean we should either have

1. soil moisture being 1, meaning that there's just water in the "soil" (my preferred way of thinking about it)
2. soil moisture being 0, as there's no soil 

I also prefer 1 because often the soil moisture is larger towards the coast, so that feels like a more continuous transition.

After ~1/2 year this would yield a top-layer soil moisture like

image


milankl:
But actually we could also just start from the soil moisture climatology we have, would you prefer that as an option? Similar for soil temperature, we could just load in the netCDF files for that, interpolate in space and to the time of year and use that as initial conditions.

milankl:
We actually do that for the `ConstantOceanClimatology` which reuses `SeasonalOceanClimatology` to set the inital conditions for SST see here

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/a3ba9ac152bd47c40060e928749290a51bb59eeb/src/physics/ocean.jl#L220-L226

We could do this for

- soil temperature (easy, we already have `SeasonalLandTemperature`)
- soil moisture (easy, we already have `SeasonalSoilMoisture`)
- SST in the `SlabOcean` (easy, we already have `SeasonalOceanClimatology`)
- sea ice concentration with `ThermodynamicSeaIce` (we have `input_data/sea_ice.nc` but currently don't use it, so we should write a `SeasonalSeaIce` as with the other variables)

too. What do you think?

pierluigividale:
I have read through, and I think that the final suggestion is the most sensible one.
All seem quite easy, except the seasonal sea ice. What is the origin of the data in the NetCDF file?
Also, did you apply a simple algorithm to turning those data into seasonal, or did you have a seasonal product to begin with?
Happy to ask my HC colleagues for a data source on sea ice if needed.

milankl:
Some ERA5 climatology for all variables.

> I have read through, and I think that the final suggestion is the most sensible one.

Meaning you prefer soil moisture equals 0 over ocean? Note that ocean here is only 100% ocean, a cell with 50% ocean would still start with 0.5 then

milankl:
Also I notice that starting with 0.5 basically all grid cells have a humidity >=0.5, is that because evporation over land is really that weak or do we not dry soil moisture with an upward surface humidity flux?

pierluigividale:
No, I meant making maximum use of the NetCDF files.
But, wherever that is not enough, yes, I agree that aiming to create the smoothest possible transition between any boundaries (land/sea, ice-covered/ice-free) is the best solution. In the case of soul moisture, if needed, let us go with 1 near the coastlines.

pierluigividale:
> Also I notice that starting with 0.5 basically all grid cells have a humidity >=0.5, is that because evporation over land is really that weak or do we not dry soil moisture with an upward surface humidity flux?

In lack of precipitation, the top layer should rapidly dry out. There are some characteristics of our current implementation that do not really allow for that, however:
1) our top layer is quite thick, so it has quite a large water storage capacity
2) drainage to the next layer down is not very efficient
3) the evaporative flux formulation favours places with vegetation and the bottom layer as source, so that a soil with low vegetation will be very inactive in terms of evaporation from the surface. This is unrealistic, but for the time being the thick top soil layer prevents the land surface from going to 80C over summer.

It is a compromise for the sake of keeping temperatures down.

pierluigividale:
Are you going to to S&M? The button is staring at me, but I am not the author of this contribution.

milankl:
I want to load initial conditions from file also in this pull request

pierluigividale:
Meaning you want to make sure that all those variables can either be read from the NetCDF files, or from the history dump?

milankl:
Oh they are read when restarting the model as we addressed in #784 but it would be nice to start from more realistic initial conditions

pierluigividale:
OK, ICs it is.

For the soil moisture memory, once I am done with snow, I can experiment with a few things:

1) increase to N soil layers, making it possible to have a top soil of 2cm
2) increase drainage, to make it easier to empty the soil reservoirs.

BTW, I completely forgot to tell you that I gave the output files to Omar Müller, and he ran the river discharge model, TRIPPy, on them. There are two observations from his work:

A) most runoff happens near coastlines
B) the amount of runoff is very large

For B), it may be that we have a units problem. For A), it could be a combination of soil memory issue and land/sea circulation

Cheers,
PL

milankl:
The netCDF soil moisture climatology has 0 soil moisture over the ocean, so if we set the 100% ocean points to 1 then this looks rather funny

image

pierluigividale:
Sigh, hard to win. Best would be a mask.

pierluigividale:
But then we get into all sorts of nasty behaviours with partial grid boxes.

milankl:
If starting with 0 soil moisture this looks better but with funky boundaries

image

as the soil moisture of the netCDF file is masked

image

I'll leave this now though, this would equilibrate itself over time and just requires a better version of `/input_data/soil_moisture.nc`. Maybe we should just redo that based on some ERA5 climatology at somewhat higher resolution

milankl:
@maximilian-gelbrecht did you also have this before that buildkite just fails for unclear reasons?

```julia
ERROR: failed to parse TOML output from running "/root/.cache/julia-buildkite-plugin/depots/0193d4a2-3184-42e9-964a-e84ff0d6a941/packages/CUDA_Compiler_jll/WkzW2/.pkg/select_artifacts.jl", got:
TOML Parser error:
none:12:6 error: invalid bare key character: ':'
  ERROR: IOError: write: broken pipe (EPIPE)
       ^
Stacktrace:
  [1] pkgerror(msg::String)
    @ Pkg.Types ~/.cache/julia-buildkite-plugin/julia_installs/bin/linux/x64/1.11/julia-1.11-latest-linux-x86_64/share/julia/stdlib/v1.11/Pkg/src/Types.jl:68
```

maximilian-gelbrecht:
Yeah, I also got this error yesterday in #835, then I just rerun it without changing anything and it was gone. I don't understand what the error means.

milankl:
yes, weird, now it passes, I saw you had commit just to trigger buildkite, I somehow also never manage to just hit rerun because it complains about me not being signed in

maximilian-gelbrecht:
Yeah I think that's because the build kite CI is running under the general Julialang account and not under a personal account from one of us a an organisation one from SpeedyWeather.

milankl:
Initial conditions for SST now look like (SST over land can be set, I decided to pick some 283K as it's the role of the land-sea mask to decide what's land whats ocean)

image

with the negative temperatures being immediately converted into sea ice conver (and then set to -1.8˚) which looks after the first timestep like

image

Soil moisture first time step

image

which quickly fills up islands that have 0 soil moisture in the initial conditions, after 10 days

image

and initial conditions for soil temperature

image

where again the soil temperature over ocean is set to 283K but could be set to NaN too if needed

PR #833, opened 2025-09-03
Structure for land snow component

milankl:
EDIT: Summary of changes

- `land.snow` is now by default the `SnowModel`
- `albedo.land` is now by default the `LandSnowAlbedo` which calculates albedo based on vegetation too
- sea ice insulation in surface fluxes
- snow insulation in surface fluxes

To be used like (default)

```julia
land = LandModel(spectral_grid, snow=SnowModel(spectral_grid))
albedo = Albedo(ocean=OceanSeaIceAlbedo(spectral_grid), land=LandSnowAlbedo(spectral_grid))
model = PrimitiveWetModel(spectral_grid; land, albedo)
```

### DIFF
--- a/docs/src/land.md
+++ b/docs/src/land.md
@@ -230,32 +230,58 @@ maximum and add a fraction ``p = 0.5`` of that excess water to the layer below
 ``W_2 = W_2 + p \delta W_1 \tfrac{f_1}{f_2}`` the other half of that excess
 water is put into the river runoff.
 
+## Snow model
+
+SpeedyWeather has a simple snow model that allows for snow to lie on land
+and impact albedo and isolate air-land fluxes. Snow is created from
+the precipitation schemes, can melt on the ground depending on soil temperature
+where it is then passed to the soil moisture. We also include a
+runoff/relaxation term to prevent snow piling up without bounds as
+soil temperatures below freezing would never remove such snow.
+In reality this is where one needs a ice sheet model to convert the
+snow to ice and simulate ice flow like in a glacier or in the
+Greenland and Antarctic ice sheets.
+
 ### LandSnowModel
 
 `SnowModel` stores a single snow bucket with depth ``S`` in units of equivalent liquid water height
-(``prognostic_variables.land.snow_depth``). Snow accumulates from the column-integrated large-scale and
-convective snow precipitation rate ``snow_rate`` (``kg/m²/s``) and melts once the top soil layer
-exceeds the melt threshold ``T_{melt}`` (default ``275~K``). The available melt energy in the top
+(`prognostic_variables.land.snow_depth`) and solves the following equation
+
+```math
+\frac{dS}{dt} = P - M - R
+```
+with precipitation ``P``, melting ``M`` and runoff ``R``.
+Snow accumulates from the column-integrated large-scale
+(currently not from convection) snow precipitation rate ``P``, `snow_rate` (``kg/m²/s``),
+and melts once the top soil layer exceeds the melt threshold ``T_{melt}`` (default ``275~K``)
+through the term ``M`` and the runoff is implemented as a weak relaxation term
+back to 0 with a multi-year time scale.
+
+The available melt energy in the top
 layer of thickness ``z₁`` uses the dry-soil heat capacity ``cₛ``:
 
 ```math
 E_{avail} = cₛ \, \max(T_1 - T_{melt}, 0) \, \frac{z₁}{Δt}.
 ```
 
-This caps the maximum melt rate
+This is the maximum melt rate
 
 ```math
 \text{melt}_{rate_{max}} = \frac{E_{avail}}{ρ_w Lᵢ},
 ```
 
-with ``ρ_w`` water density and ``Lᵢ`` latent heat of fusion. A slow runoff/relaxation term prevents
-perennial snow packs from growing without bound,
+with ``ρ_w`` water density and ``Lᵢ`` latent heat of fusion. But note that
+this formulation allows to melt more snow than there actually is, so we need
+to cap the amount to not end up with negative snow. We implement this constrain
+not for terms individually but for the sum of all terms, see below.
+
+A slow runoff/relaxation term prevents perennial snow packs from growing without bound,
 
 ```math
 \text{runoff}_{rate_{max}} = \frac{S}{\tau_{runoff}},
 ```
 
-controlled by ``runoff_time_scale`` (default ``4`` years). Snowfall, melt and runoff form a raw
+controlled by `runoff_time_scale` (default ``4`` years). Snowfall, melt and runoff form a raw
 tendency ``\text{d}S_{max}`` that is further limited so we never remove more snow than is present
 (including what falls during the current step). The actual removal is reported as
 
@@ -263,13 +289,14 @@ tendency ``\text{d}S_{max}`` that is further limited so we never remove more sno
 \text{snow\_melt\_rate} = \text{melt}_{rate_{max}} + \text{runoff}_{rate_{max}} + \text{excess},
 ```
 
-in ``kg/m²/s``, where the `excess` term only appears when the naive tendency would overdraw the bucket.
+in ``kg/m²/s``, where the `excess` term (negative for melting/runoff trying to remove more snow than there is)
+only appears when the naive tendency would overdraw the bucket.
 `snow_melt_rate` therefore combines true melt with the runoff leak and is zero over ocean points.
 Snow depth is clipped to zero and stored as equivalent liquid water height, not physical snow thickness.
 
 The snow budget links into other surface schemes:
 
-- `snow_melt_rate` provides the latent heat sink ``Lᵢ\,\text{snow\_melt\_rate}`` in the land temperature budget.
+- `snow_melt_rate` provides a latent heat sink to the land temperature budget.
 - The same flux (converted to ``m/s`` by dividing by ``ρ_w``) feeds the soil moisture tendency alongside rain.
 - Snow depth drives the snow-albedo calculation and insulates surface heat/humidity fluxes (see below).
 
@@ -280,15 +307,18 @@ Snow cover over land is diagnosed from snow depth using either a linear ramp
 σₛ = \frac{S}{S + \text{snow\_depth\_scale}},
 ```
 
-set via the `snow_cover` keyword of `LandSnowAlbedo`. The land albedo then blends between bare land
-and snow,
+set via the `snow_cover` keyword of `LandSnowAlbedo`. Snow then adds an albedo to the land albedo
+(diagnosed from vegetation cover)
 
 ```math
-albedo_{land} = albedo_{land} + σₛ \, (albedo_{snow} - albedo_{land}),
+albedo_{land} = albedo_{land} + σₛ albedo_{snow}
 ```
 
-so snow depth from the snow bucket immediately brightens land grid cells. `DefaultAlbedo` already
-uses `LandSnowAlbedo`; there is currently no time-evolving snow albedo.
+so snow depth from the snow bucket immediately brightens land grid cells.
+The total albedo is higher over already brighter areas (low vegetation cover)
+and lower over darker areas. This somewhat reflects that in forests the
+snow cover is broken up and snow lies in between trees.
+`DefaultAlbedo` uses `LandSnowAlbedo`; there is currently no time-evolving snow albedo.
 
 ## Albedo
 


milankl:
Do you want me to comment on the current changes or go start sorting those lines?

pierluigividale:
I think that I am still at a stage where I could use some pointers. I would say that the bulk of what I want to do is there, but there are a few decisions to be made, e.g.
1) do we want to compute the sublimation flux here, or inside the momentum module?
2) am I going in the right direction in terms of imposing tendency updates to ST(1), humid(end), and so on
3) I would leave the thermal shielding as really simple at this stage, and just reduce the soil heat flux by imposing a weighted average of the two conductivities (soil, snow). Again, do it here, or insert the code in the temperature module, with the requirement that we pull snow cover information into there?
4) where to put all the snow parameters: separate file, or thermodynamics?

BTW, I still find it confusing that the surface sensible heat flux is called S now that we have an important variable called S. Are you ok that I go into the relevant modules and change it to H? Otherwise I would go for Snow, not S.

pierluigividale:
Today is packing and moving day, not sure I can get much done.

src/physics/land/snow.jl
```diff
@@ -0,0 +1,121 @@
...
+	## Now compute the change in albedo
+	albedo_snow[ij] = snow_albedo_old + (snow_albedo_fresh - snow_albedo_old) * exp(-Δt_snow / τˢ) # time needs to be expresssed in days. 
+	albedo_land[ij] = (1-σₛ[ij]) * albedo_land + σₛ[ij] * albedo_snow[ij]
+
+	F[ij] = infiltration_flux[ij] + melt_flux[ij]  # Need to pass this back to bucket model
+    R[ij] = + Rₛ[ij]   #??? PLV this needs to be returned to the bucket model, it may be enough to do it here
+
+    launch!(architecture(snow_depth), LinearWorkOrder, size(snow_depth),
+        land_snow_kernel!, snow_depth, mask, p1)
+	Δt_snow += Δt  #Increment the time counter since the time of last significant snowfall
```
milankl: Not sure this works. This requires some memory for each grid cell across time steps which makes this a prognostic variable, and I'm not sure how to formulate this as a PDE. It's also not part of the equations you wrote down above, just flagging this early

src/physics/land/snow.jl
```diff
@@ -0,0 +1,121 @@
...
+	end
+	## Now compute the snow area cover fraction based on snow depth
+	#σₛ = Sₐ / (10. * drag_snow + Sₐ) #JULES, from Betts et al.
+	σₛ[ij] = min(1.0, Sₐ[ij] / snow_depth_scale)   # e.g. snow_depth_scale ≈ 0.1 m
+	## Now compute the change in albedo
+	albedo_snow[ij] = snow_albedo_old + (snow_albedo_fresh - snow_albedo_old) * exp(-Δt_snow / τˢ) # time needs to be expresssed in days. 
+	albedo_land[ij] = (1-σₛ[ij]) * albedo_land + σₛ[ij] * albedo_snow[ij]
+
+	F[ij] = infiltration_flux[ij] + melt_flux[ij]  # Need to pass this back to bucket model
+    R[ij] = + Rₛ[ij]   #??? PLV this needs to be returned to the bucket model, it may be enough to do it here
```
milankl: yeah you can do that generally but probably `R[ij] +=` to update the river runoff not overwrite it. This also relates to the order in which the subcomponents are executed, at the moment (see above) we have

```julia
        timestep!(progn, diagn, land.soil_moisture, model)
        timestep!(progn, diagn, land.snow, model)
        timestep!(progn, diagn, land.rivers, model)
        timestep!(progn, diagn, land.vegetation, model)
    end
    timestep!(progn, diagn, land.temperature, model)
```

which we have as the soil moisture determines the heat capacity in the temperature component

src/physics/land/snow.jl
```diff
@@ -0,0 +1,121 @@
...
+	    # temperature tendency
+	    δT[ij] = -Q_melt[ij] / C_soil
+	    soil_temperature[ij,1] += δT[ij]
+	end
+	## Now compute the snow area cover fraction based on snow depth
+	#σₛ = Sₐ / (10. * drag_snow + Sₐ) #JULES, from Betts et al.
+	σₛ[ij] = min(1.0, Sₐ[ij] / snow_depth_scale)   # e.g. snow_depth_scale ≈ 0.1 m
+	## Now compute the change in albedo
+	albedo_snow[ij] = snow_albedo_old + (snow_albedo_fresh - snow_albedo_old) * exp(-Δt_snow / τˢ) # time needs to be expresssed in days. 
+	albedo_land[ij] = (1-σₛ[ij]) * albedo_land + σₛ[ij] * albedo_snow[ij]
```
milankl: This should go into `albedo.jl` I'm seeing this as a `SnowAlbedo` component that on every time step the checks the snow cover/depth and translates that into an albedo that's added or overwrites the land albedo. That should all work fine, but you also seem to construct here something based on the snow age which mans snow age would need to be a prognostic variable. Alternatively we could let snow depth decrease over time to simulate densification of the snow pack and if snow cover then depends on snow depth we'd also reduce albedo over time that way

src/physics/land/snow.jl
```diff
@@ -0,0 +1,121 @@
...
+
+	    # energy used for melting [J/m²]
+	    Q_melt[ij] = ρ_w * Lᵢ * melt_depth[ij]
+
+	    # soil heat capacity [J/m²/K]
+	    C_soil = ρ_soil * cₛ * z₁
+
+	    # temperature tendency
+	    δT[ij] = -Q_melt[ij] / C_soil
+	    soil_temperature[ij,1] += δT[ij]
```
milankl: this should probably go into `LandBucketTemperature` in a similar way as the soil moisture is used therein with the special case that if soil moisture is zero it's the same heat capacity as for dry soil? So we could compute the soil heat capacity therein as a function of snow depth and if that's zero then we're back to where we were before. Not all snow processes have to live in `snow.jl`, I really see the computation here as the ones that are required to update the snow depth/cover, consequences can absolutely live in other land components and be switched off therein with a flag like `snow=false`

src/physics/land/snow.jl
```diff
@@ -0,0 +1,121 @@
...
+	Δt_snow += Δt  #Increment the time counter since the time of last significant snowfall
+end
+
+@kernel inbounds=true function land_snow_kernel!(
+    snow_depth, mask, @Const(p1),
+)
+    ij = @index(Global, Linear)             # every grid point ij
+
+    if mask[ij] > 0                         # at least partially land
+        snow_depth[ij] = 0                  # dummy operation, replace with real logic
```
milankl: remember that you do want to move you kernel into this line 😉

src/physics/land/thermodynamics.jl
```diff
@@ -2,20 +2,41 @@ abstract type AbstractLandThermodynamics <: AbstractModelComponent end
...
+    snow_temp_melt::NF = 273.15+2.5
+	
+    "[OPTION] Amount of cumulative snowfall required to trigger land snow scheme [m/m²]"
+    snowfall_threshold::NF = 0.01
+	
+    "[OPTION] Fresh Snow albedo []"
+    snow_albedo_fresh::NF = 0.8
+
+    "[OPTION] Old Snow albedo []"
+    snow_albedo_old::NF = 0.4
```
milankl: move to a `SnowAlbedo` type?

src/physics/land/thermodynamics.jl
```diff
@@ -2,20 +2,41 @@ abstract type AbstractLandThermodynamics <: AbstractModelComponent end
...
     wilting_point::NF = 0.17
+	
+    "[OPTION] Freezing temperature for snow on top soil layer [K]"
+    snow_temp_freeze::NF = 273.15
+
+    "[OPTION] Melting temperature for snow on top soil layer [K]"
+    snow_temp_melt::NF = 273.15+2.5
+	
+    "[OPTION] Amount of cumulative snowfall required to trigger land snow scheme [m/m²]"
+    snowfall_threshold::NF = 0.01
```
milankl: I'm not sure these are generally needed across all land components, if not then they should live in `SnowModel`

src/physics/land/snow.jl
```diff
@@ -2,8 +2,10 @@ abstract type AbstractSnow <: AbstractParameterization end
 
 export SnowModel    # maybe change for a more concise name later
 @kwdef mutable struct SnowModel{NF} <: AbstractSnow
-    parameter1::NF = 0
-    parameter2::NF = 0
+    drag_snow::NF = 3.0e-5        # [m]
+    soil_density::NF = 1.3e4      # [kg/m2]
+    snow_depth_scale::NF = 0.1    # [m]                       #Tunable parameter
+	τˢ::NF = 864000               # [s] tunable parameter, currently set to 10 days
```
milankl: time scales are conveniently written as `time_scale::Second = Day(10)`, this way a user can use any units `Day(10), Hour(240), Minute(5)` but it's always converted to `Second` on the fly. One then uses this parameter inside a function with `time_scale.value` to get the seconds in integers, or better `Second(time_scale).value` which guarantees that the unit is `Second` even if you wrote `time_scale::Day = Day(10)` above (but we use `Second` as a convention)

src/physics/land/snow.jl
```diff
@@ -25,25 +27,26 @@ end
 function timestep!(
     progn::PrognosticVariables,
     diagn::DiagnosticVariables,
+	column::ColumnVariables,
```
milankl: The land model doesn't operate on columns, you need to leave the function signature as is.

src/physics/land/snow.jl
```diff
@@ -25,25 +27,26 @@ end
...
+	cᵢ = model.land.thermodynamics.heat_capacity_snow
+	cₛ = model.land.thermodynamics.heat_capacity_dry_soil
+	Lᵢ = model.clausius_clapeyron.latent_heat_fusion                  # latent heat of fusion
     # Some atmospheric variables needed to compute snow sublimation
-    P = diagn.physics.total_precipitation_rate      # precipitation (rain+snow) in [m/s]
-	S = diagn.physics.large_scale_snow_rate +  diagn.physics.convective_snow_rate    # Snowfall rate in [m/s]
-    E = diagn.physics.land.surface_humidity_flux    # [kg/s/m²], divide by density for [m/s]
-    R = diagn.physics.land.river_runoff             # diagnosed [m/s]
+    P = diagn.physics.total_precipitation_rate                        # precipitation (rain+snow) in [m/s]
+	S = column.snow_rate_large_scale +  column.snow_rate_convection   # Snowfall rate in [m/s]
```
milankl: As the land model doesn't operate on columns but with one kernel that loops over all horizontal grid points this should refer to the arrays in `diagn.physics` but we currently only have a `total_precipitation_rate` not a `snow_rate` only, will add this ...

src/physics/land/snow.jl
```diff
@@ -100,10 +102,10 @@ function timestep!(
 	σₛ[ij] = min(1.0, Sₐ[ij] / snow_depth_scale)   # e.g. snow_depth_scale ≈ 0.1 m
 	## Now compute the change in albedo
 	albedo_snow[ij] = snow_albedo_old + (snow_albedo_fresh - snow_albedo_old) * exp(-Δt_snow / τˢ) # time needs to be expresssed in days. 
-	albedo_land[ij] = (1-σₛ[ij]) * albedo_land + σₛ[ij] * albedo_snow[ij]
+	column.albedo_land[ij] = (1-σₛ[ij]) * albedo_land + σₛ[ij] * albedo_snow[ij]
 
-	F[ij] = infiltration_flux[ij] + melt_flux[ij]  # Need to pass this back to bucket model
-    R[ij] = + Rₛ[ij]   #??? PLV this needs to be returned to the bucket model, it may be enough to do it here
+	F[ij] = infiltration_flux[ij] + melt_flux[ij]    # Need to pass this back to bucket model
+    diagn.physics.land.river_runoff[ij] = + Rₛ[ij]   #??? PLV this needs to be returned to the bucket model, it may be enough to do it here
```
milankl: everything that depends on `ij` will need to be moved inside the kernel and all arrays/constants be passed on

milankl:
@pierluigividale, @nviebig is keen to take over this pull request and finish it within the next few weeks. Do you mind?

pierluigividale:
> @pierluigividale, @nviebig is keen to take over this pull request and finish it within the next few weeks. Do you mind?

Please go ahead: I am in Japan for a week, and moving house the week when I return, so I am causing delays. I do want to turn my attention to the turbulence schemes.

src/physics/land/snow.jl
```diff
@@ -56,22 +56,35 @@ function timestep!(
...
+    params = (
+        melting_threshold= melting_threshold,
+        cₛ= cₛ,
+        ρ_soil= ρ_soil,
+        z₁= z₁,
+        Δt= Δt,
+        ρ_water= ρ_water,
+        Lᵢ= Lᵢ,
+        r⁻¹= r⁻¹,
+    )
```
milankl: this can be shortened using the `(; kwargs...)` notation

```julia
julia> a=1
1

julia> b=2
2

julia> params = (; a, b)
(a = 1, b = 2)
```

nviebig:
**EDIT: Changed colormap to Devon**

**Model**: SpeedyWeather T31 resolution
**Simulation period**:  6-month spin-up starting July 1999, followed by a full simulation year 2000 with 6-hourly output.
**SpeedyWeather panels**: Monthly means of snow_depth (variable sd) computed from all 6-hourly outputs in each month of Jan–Dec 2000, values shown as log₁₀(snow depth in meters).
**CESM2 panels**: Monthly means of the CESM2 variable snd (snow depth), same year 2000,  as log₁₀(snow depth in meters).

snowdepth_comparison_month_01
snowdepth_comparison_month_02
snowdepth_comparison_month_03
snowdepth_comparison_month_04
snowdepth_comparison_month_05
snowdepth_comparison_month_06
snowdepth_comparison_month_07
snowdepth_comparison_month_08
snowdepth_comparison_month_09
snowdepth_comparison_month_10
snowdepth_comparison_month_11
snowdepth_comparison_month_12

milankl:
OH YEAH, nice! Looks surprisingly good, okay we have some winter snow in Spain but that's probably fine, I blame that generally SpeedyWeather is too cold (which might be more an issue of longwave radiation and surface fluxes) so don't bother too much but the general patterns are there!! Why is there no snow on Antarctica?

nviebig:
Yeah i was wondering the same! Im currently looking at that and i think it has something to do with that maybe in CESM2 Antarctica is not part of the land model!

milankl:
Or is it simply at the top of that colormap? It's confusing that they (you?) use a cyclic colormap for snow, like a lot of snow suddenly looks like no snow, what? 🤣 

EDIT: Sorry I see it's a diverging colourmap, maybe use a linearly increasing one instead, is that Fabio Crameri's berlin? Maybe use oslo or devon instead?

milankl:
So while this all looks great already clearly there's not enough multi-year snow, several possibilities with a preference for doing all 😉 
- using a timescale of 1 year for the runoff/leakage term would - by definition - remove snow on an annual basis, so maybe 3-5 years is a more realistic time scale here
- without the snow albedo the land can warm up too quickly in spring, before we make assessments of whether this is realistic I'd suggest to use that (if you haven't already)
- also the air-land sensible/latent heat fluxes should be reduced given the insulating factor of snow which will further avoid the land from warming up too quickly
- we also should check whether it makes sense to have the land only warm with warmer ground not also through warmer air ...

nviebig:
- [x] Increase the runoff/leakage timescale from 1 year to something like 3–5 years to avoid annual snow loss.
- [x] Reduce surface sensible/latent heat fluxes when snow is present to account for insulation.
- [x] Make the snow–cover → albedo relationship modular (e.g. linear vs saturating `S/(S+scale)` schemes) in `LandSnowAlbedo`.
- [ ] Enable/verify snow albedo so spring warming isn’t unrealistically fast.
- [ ] Revisit how land temperature responds — should warming come only from soil or also from overlying air?

src/physics/land/snow.jl
```diff
@@ -0,0 +1,114 @@
...
+    ij = @index(Global, Linear)             # every grid point ij
+
+    if mask[ij] > 0                         # at least partially land
+		
+		(;melting_threshold, cₛ, ρ_soil, z₁, Δt, ρ_water, Lᵢ, r⁻¹) = params
+        # check for melting of snow if temperature above melting threshold
+        δT_melt = max(soil_temperature[ij, 1] - melting_threshold, 0)
+	
+        # energy available from soil warming above melting threshold [J/m²/s]
+        E_avail = cₛ * δT_melt * z₁ / Δt  # [J/(m³ K)] * [K] * [kg/m³] * [m] / [s] = [J/m²/s] => we remove ρ_soil?
```
milankl: should be
```
E_avail = cₛ * δT_melt * z₁ / Δt  # [J/(m³ K)] * [K] * [m] / [s] = [J/m²/s]
```

milankl: @pierluigividale you originally had `* ρ_soil` on this line, but this doesn't seem to be correct from the units? I think this is because `land.thermodynamics.heat_capacity_dry_soil` is in units of J/m^3/K, so per volume not per mass.

src/physics/land/snow.jl
```diff
@@ -0,0 +1,114 @@
...
+        runoff_rate = r⁻¹ * snow_depth[ij]
+        # maximum sink rate [m/s]
+        max_sink = max(snow_depth[ij] / Δt + snow_fall_rate[ij] / ρ_water, 0)
+        # limit runoff to not exceed available snow after melt and snowfall
+        runoff_rate = min(runoff_rate, max(max_sink - melt_rate, 0))
+        
+        # store to pass to soil moisture [kg/m²/s]
+        snow_runoff_rate[ij] = runoff_rate * ρ_water
+
+        # change snow depth by falling snow minus melting and runoff [m/s]
```
milankl: We only limit the sum of terms now, has the advantage that falling snow that would melt on the ground is directly passed to soil moisture in one time step

src/physics/land/temperature.jl
```diff
@@ -203,21 +207,29 @@ function timestep!(
...
     return nothing
 end
 
 @kernel inbounds=true function land_bucket_temperature_kernel!(
-    soil_temperature, mask, soil_moisture, Rsd, Rsu, Rlu, Rld, Ev, S,
-    @Const(Lᵥ), @Const(γ), @Const(Cw), @Const(Cs), @Const(z₁), @Const(z₂), @Const(Δ), @Const(Δt),
+    soil_temperature, mask, soil_moisture, Rsd, Rsu, Rlu, Rld, Ev, S, M,
+    @Const(Lᵥ), @Const(Lᵢ), @Const(γ), @Const(Cw), @Const(Cs),
+    @Const(z₁), @Const(z₂), @Const(Δ), @Const(Δt),
 )
```
milankl: wrap in NamedTuple?

src/physics/land/thermodynamics.jl
```diff
@@ -2,18 +2,27 @@ abstract type AbstractLandThermodynamics <: AbstractModelComponent end
 
 export LandThermodynamics
 @kwdef mutable struct LandThermodynamics{NF} <: AbstractLandThermodynamics
-    "[OPTION] Heat conductivity λ of the soil [W/(m K)]" 
-    heat_conductivity::NF = 0.42
+    "[OPTION] Heat conductivity λ_{soil} of the soil [W/(m K)]" 
+    heat_conductivity_dry_soil::NF = 0.42
+
+    "[OPTION] Heat conductivity λ_{snow} of snow [W/(m K)]" 
+    heat_conductivity_snow::NF = 0.08
```
milankl, line 9: is that actually needed?

src/physics/albedo.jl
```diff
@@ -154,4 +156,75 @@ function albedo!(
...
+    diagn::AbstractDiagnosticVariables,     # land only
+    diagn_all::DiagnosticVariables,         # all diagnostics
+    progn::PrognosticVariables,
+    albedo::LandSnowAlbedo,
+    model::PrimitiveEquation,
+)
+    (; snow_depth) = progn.land
+    (; albedo_land, albedo_vegetation, albedo_snow, snow_depth_scale, scheme) = albedo
+
+    snow_cover = diagn_all.dynamics.a_2D_grid       # scratch memory
```
milankl, line 232: @pierlvidale you wrote
```julia
	# ## Now compute the snow area cover fraction based on snow depth
	# #σₛ = Sₐ / (10. * drag_snow + Sₐ) #JULES, from Betts et al.
	# σₛ[ij] = min(1.0, Sₐ[ij] / snow_depth_scale)   # e.g. snow_depth_scale ≈ 0.1 m
	# ## Now compute the change in albedo
	# albedo_snow[ij] = snow_albedo_old + (snow_albedo_fresh - snow_albedo_old) * exp(-Δt_snow / τˢ) # time needs to be expresssed in days. 
	# column.albedo_land[ij] = (1-σₛ[ij]) * albedo_land + σₛ[ij] * albedo_snow[ij]
```
is the adaption here fine?

src/physics/land/snow.jl
```diff
@@ -0,0 +1,118 @@
...
+
+    params = (; melting_threshold, cₛ, z₁, Δt, ρ_water, Lᵢ, r⁻¹)
+
+    # sanitize inputs to avoid propagating NaNs from climatologies or I/O
+    soil_temperature .= ifelse.(isfinite.(soil_temperature), soil_temperature, melting_threshold)
+    snow_depth       .= ifelse.(isfinite.(snow_depth), snow_depth, zero(eltype(snow_depth)))
+    snow_fall_rate   .= ifelse.(isfinite.(snow_fall_rate), snow_fall_rate, zero(eltype(snow_fall_rate)))
+    # clear ocean points explicitly
+    mask!(snow_depth, model.land_sea_mask, :ocean; masked_value=zero(eltype(snow_depth)))
+    mask!(soil_temperature, model.land_sea_mask, :ocean; masked_value=melting_threshold)
```
milankl: Removing those, I believe you put them in to check for instabilities.

src/physics/land/snow.jl
```diff
@@ -0,0 +1,118 @@
...
+        # maximum allowed by existing snow depth
+        runoff_rate_max = r⁻¹ * snow_depth[ij]
+
+        # Adding the terms change snow depth by falling snow minus melting and runoff [m/s]
+        # maximum amount of snow change
+        dsnow_max = snow_fall_rate_max - melt_rate_max - runoff_rate_max
+
+        # don't melt or runoff more than there is snow + snow that is falling down this time step
+        # limited amount of snow change by how much there is
+        dsnow = -min(-snow_depth[ij] / Δt, dsnow_max)
```
milankl: there was a sign error here

milankl:
Last commit contains a vegetation-dependent albedo, annual cycle after a few years of simulation is

https://github.com/user-attachments/assets/fabaa4e0-bb81-4135-8859-89fa4fb8d401

so we now have eternal snow on Greenland and in some areas of Antarctica (good) and that eternal blob of sea ice in the Ross sea is gone (good). There is still no multi-year sea ice (not good but expected because we don't have a sea ice thickness). Snow depth annual cycle is

https://github.com/user-attachments/assets/14a74951-e17b-4fba-89c7-4b7f1d34690c

src/physics/land/snow.jl
```diff
@@ -0,0 +1,112 @@
...
+        # don't melt or runoff more than there is snow + snow that is falling down this time step
+        # limited amount of snow change by how much there is
+        dsnow = -min(snow_depth[ij] / Δt, -dsnow_max)
+
+        # snow that we tried to melt/runoff that isn't available though
+        dsnow_excess = dsnow_max - dsnow
+
+        # store to pass to soil moisture [kg/m²/s], combined runoff with melt rate
+        # limited to what's available to melt/runoff
+        snow_melt_rate[ij] = melt_rate_max + runoff_rate_max + dsnow_excess
```
nviebig, line 110: In [`src/dynamics/diagnostic_variables.jl`](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/833/files#diff-aa195f130389e47ecfe117439e87d5c4d22dee9d28bfa849c849dbeaded46fc4L329) L. 329 we defined Snow melt rate in [kg/m²/s] but all of these melt_rate_max + runoff_rate_max + dsnow_excess are in [m/s]? Do we need to multiply by $\rho_{Water}$?

milankl: Well spotted, yes. Because soil moisture expects kg/m^2/s

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/5f381bf595ffba6f4d2afb43bc3ee3edf03d80e0/src/physics/land/soil_moisture.jl#L256-L258

.... before again converting to m/s. We somehow always end up multiplying/dividing by density, I also find that confusing, maybe one day we can make that more consistent across all parameterizations/ocean/land/sea ice code but for now just document well what's expected in which units in the comments. That's why I'm always writing the units down!

PR #836, opened 2025-09-05
Rename SpeedyInternals to SpeedyWeatherInternals 

maximilian-gelbrecht:


maximilian-gelbrecht:
@milankl I renamed the Internal package to be a bit more unique name for us for the registration. 

The other thing of putting the main module in a submodules, do you know if we have to change things about the registration etc in this case? Probably only when we want to release the next version, right? 

FYI: I didn't know before, the way to install subdirs via URL is e.g. `Pkg.add(url="https://github.com/SpeedyWeather/SpeedyWeather.jl/",subdir="LowerTriangularArrays")` btw

maximilian-gelbrecht:
I split the PRs in two, the changes about the main package are in #837, here it's just the renaming.

milankl (approved):

PR #837, opened 2025-09-05
Put main package in subfolder

maximilian-gelbrecht:
So, I tried to put the main package in a subfolder here and add another environment in the main directory. 

Locally on my laptop this works fine in Julia 1.11, but in Julia 1.10 this doesn't work like this. I can't even `dev` one of the other submodules because it complains directly about the others missing.


milankl:
I know this was suggested but does that actually work then with the dependency of the submodules on main?

maximilian-gelbrecht:
Yeah, I mainly just wanted to try it, if it would work easily. 

So far, this doesn't seem to be the case for Julia 1.10.

maximilian-gelbrecht:
This structure does seem to work for Makie in Julia 1.10 though. But I also don't really now want to spend much time into the module structure again.

maximilian-gelbrecht:
I think I am also fine with just keeping it as it is.

milankl:
I don't see an urgency to do that, another possibility is still to put all subpackages into `/lib` which I'd also be happy with if that's important

maximilian-gelbrecht:
Yeah, I am closing this PR. It was a quick attempt if it might work without some amount of work. It doesn't. So let's keep it as it is for now.

Issue #838, opened 2025-09-05
Simulating Mars


digi-chris:
Hi,

I'm trying to run a simluation using terrain data for Mars.

After bringing the terrain data into a NetCDF file, I was getting an error during the simulation:

`Warning: NaN or Inf detected at time step`

I took a bit of a guess, thinking that something that Mars has that the Earth doesn't is a very, very massive mountain - Olympus Mons. This is some 22km above the Mars Datum line.

I capped the maximum height for the NetCDF off at 5000m, and sure enough it started working. My assumption was that SpeedyWeather might only be simulating the lower atmosphere and a 22km mountain is going to have parts of its surface outside of that atmosphere. Maybe that causes the NaN/Inf?

Thinking I could live with the simulation not really working with Olympus Mons but being kind of OK everywhere else (I'm just interested in getting to an approximation at the moment), I then started plugging some atmosphere data into SpeedyWeather to get things closer to Mars.

heat_capacity = 736
mol_mass_dry = 43.3

Setting these two parameters worked, but then I set temp_ref as follows:

temp_ref = 213.5

Now it doesn't work again - same error as before!

Here is the full code I am using:

```
using SpeedyWeather
spectral_grid = SpectralGrid(trunc=170, nlayers=8)
# default angular frequency of Earth's rotation [1/s] = 7.29e-5
clock_start = DateTime(2025,9,1,0,0,0)
planet_body = Earth(spectral_grid, rotation=7.088218367070196E-05, gravity=3.728, length_of_day=88775.23576284682, seasonal_cycle=true, length_of_year=Day(687), equinox=DateTime(2025,3,20), axial_tilt=0.436332, solar_constant = 587.000352089344)

orography = EarthOrography(spectral_grid, path="xxx", file="orography.nc", file_Grid=FullGaussianGrid)
output = NetCDFOutput(spectral_grid, id="current", path="xxx", output_dt=Hour(1))
add!(output, SpeedyWeather.AllOutputVariables()...)
add!(output, SpeedyWeather.TemperatureOutput())
add!(output, SpeedyWeather.HumidityOutput())
add!(output, SpeedyWeather.SurfacePressureOutput())

albedo = GlobalConstantAlbedo(spectral_grid, albedo=0.17)
planet_atmosphere = EarthAtmosphere(spectral_grid;heat_capacity=736,mol_mass_dry_air=43.3,temp_ref=213.15) #,dry_lapse_rate=0.0045,pres_ref=610.5)
[...]

PR #840, opened 2025-09-07
Move radius to model.planet

milankl:
Inspired from #838


milankl:
@maximilian-gelbrecht tests are passing but just wanted to run this by you whether you see any issues with moving `radius` from spectral grid to planet. Chosing the radius paramter in spectral grid makes sense if you think about the spectral grid as defining the space and its discretization like in Oceanigans the grid has Lx, Ly, and nx, ny etc. But we consider our `grid` i.e. spectral_grid.grid as the discretization of the unit sphere or a sphere with unspecified radius so that wasn't consitent anyway. I therefore found it more intuitive to move the radius inside the planet, where if one defines e.g. Mars, one would alos define its radius. 

Consequence is that `Geometry.radius` is now a RefValue as `Geometry(spectral_grid)` does not know the planet, but it'll be set (if not default which is unchanged) at `initialize!`. So there's some risk someone doing some operations with geometry or also Leapfrog (which also contains `radius`) without it being initialized but that only really applies to exoplanet simulations which I think is acceptable and intuitive ("if you haven't intialized then the planet isn't known to any other model components")

maximilian-gelbrecht:
I understand that it makes more sense to have the radius in the planet, but what makes me a bit cautious is that our equations are scaled by radius as well, so it is very central to the model formulation beyond just the discretisation. 

I am okay with these changes, I think in the long term we could consider reworking this a little by splitting model components that are there to set constants and parameters from those that implement processes in some way.

milankl:
I initially also though this may cause problems but scaling the variables happens here

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/74378abe1a068fa55c73989a2684e6f081b92757/src/models/simulation.jl#L79-L80

inside the `simulation = initialize!(model)` call so there's not really any problem of needing the `radius` where it's not available

maximilian-gelbrecht (approved):

src/dynamics/time_integration.jl
```diff
@@ -106,13 +106,12 @@ be a divisor such that an integer number of time steps matches exactly with the
 time step."""
 function initialize!(L::Leapfrog, model::AbstractModel)
     (; output_dt) = model.output
+    (; radius) = model.planet
 
-    if L.adjust_with_output
```
maximilian-gelbrecht, line 110: Is the check not needed anymore?

milankl: it's incorrectly placed, the actual check should happen in `get_dt_millisec` which it does but even with `adjust_with_output = false` the time steps should be recomputed here!

milankl: here this is important because the radius changes the time step, so another radius needs another timestep whether or not `adjust_with_output` is true or false

milankl:
Used the chance to update `show` for `SpectralGrid`, now

```julia
julia> spectral_grid = SpectralGrid(trunc=31, nlayers=8)
SpectralGrid{Spectrum{...}, OctahedralGaussianGrid{...}}
├ Number format: Float32
├ Spectral:      T31 LowerTriangularMatrix
├ Grid:          48-ring OctahedralGaussianGrid, 3168 grid points
├ Resolution:    3.61°, 401km (at 6371km radius)
├ Vertical:      8-layer atmosphere, 2-layer land
└ Architecture:  CPU using Array

julia> spectral_grid = SpectralGrid(trunc=511, nlayers=8)
SpectralGrid{Spectrum{...}, OctahedralGaussianGrid{...}}
├ Number format: Float32
├ Spectral:      T511 LowerTriangularMatrix
├ Grid:          768-ring OctahedralGaussianGrid, 603648 grid points
├ Resolution:    0.261°, 29.1km (at 6371km radius)
├ Vertical:      8-layer atmosphere, 2-layer land
└ Architecture:  CPU using Array
```

PR #850, opened 2025-09-13
Fix correct array type for LTA init

maximilian-gelbrecht:


maximilian-gelbrecht:
This just fixes that LTA's are really initialised correctly on GPU when `rand` etc is called with a GPU Spectrum: 

```Julia 
spectrum = Spectrum(trunc=31, architecture=GPU())
rand = zeros(spectrum) 
```
returns a GPU LTA now as expected. 

This also needed some adjustment to how SpectralTransforms are initialised because pre-computations have to be done on CPU.

maximilian-gelbrecht:
Yes exactly, in RingGrids this was already correct.

PR #852, opened 2025-09-15
Call callbacks before output

milankl:
Suggest by Davide @DavSab1 at NCAS CMSS so that one can use a callback to define an SST anomaly via a callback that adds to a seasonal ocean climatology, e.g.

```julia
spectral_grid = SpectralGrid(trunc=31, Grid=OctaminimalGaussianGrid)

# use a seasonal ocean from file that overwrites SST on every time step
ocean = SeasonalOceanClimatology(spectral_grid)

# define a callback that adds an SST anomaly
struct SSTAnomaly <: SpeedyWeather.AbstractCallback end

# initialize just calls the callback too
SpeedyWeather.initialize!(callback::SSTAnomaly, args...) = SpeedyWeather.callback!(callback, args...)

function SpeedyWeather.callback!(
    callback::SSTAnomaly,
    progn::PrognosticVariables,
    diagn::DiagnosticVariables,
    model::AbstractModel,
)
    set!(progn.ocean.sea_surface_temperature, (λ, ϕ) -> ϕ < 0 ? -5 : 0, model.geometry, add=true)
end

SpeedyWeather.finalize!(calllback::SSTAnomaly, args...) = nothing

# create that callback and add it to the model
callback = SSTAnomaly()
add!(model, callback)

# adjust output
add!(model, SpeedyWeather.OceanOutput()...)
model.output.output_dt = Day(1)

# run simulation
run!(simulation, period=Day(30), output=true)
```

however, at the moment we call first the output then the callbacks which doesn't really make sense to me as I see the callbacks to e.g. change fields which should then be reflected in the output. What's not possible now would be to let the callback change the netcdf output directly which is fine. E.g. from above

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,10 @@
 
 ## Unreleased
 
+- Call callbacks before output not after [#852](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/852)
 - Split extensions across subpackages [#846](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/846)
 
-## v0.17
+## v0.17
 
 - More documentation on GPU and on_architecture [#845](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/845)
 - The documentation has a logo now as well [#842](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/842)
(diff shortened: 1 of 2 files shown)


milankl:
Another example for a callback ramping up albedo with time https://gist.github.com/milankl/59b0cdca90d59f4b48fe26c746b08b46

PR #853, opened 2025-09-15
Mean sea-level pressure output

milankl:
This works now

```julia
add!(model, SpeedyWeather.MeanSeaLevelPressureOutput())
```

surface pressure is 

image

mean sea-level pressure is

image

outstanding question is what to do with $T_v$, virtual temperature in the mean sea-level extrapolatoin? and todo

- [x] write tests
- [x] and/or documentation

### DIFF
--- a/src/output/netcdf_output.jl
+++ b/src/output/netcdf_output.jl
@@ -201,7 +201,7 @@ function add_default!(
     Model::Type{<:PrimitiveDry},
 )
     add_default!(variables, Barotropic)
-    add!(variables, SurfacePressureOutput(), TemperatureOutput())
+    add!(variables, MeanSeaLevelPressureOutput(), TemperatureOutput())
 end
 
 """$(TYPEDSIGNATURES)
--- a/src/output/variables/dynamics.jl
+++ b/src/output/variables/dynamics.jl
@@ -123,6 +123,58 @@ path(::SurfacePressureOutput, simulation) = simulation.diagnostic_variables.grid
 
 """Defines netCDF output for a specific variables, see [`VorticityOutput`](@ref) for details.
 Fields are: $(TYPEDFIELDS)"""
+@kwdef mutable struct MeanSeaLevelPressureOutput <: AbstractOutputVariable
+    name::String = "mslp"
+    unit::String = "hPa"
+    long_name::String = "mean sea-level pressure"
+    dims_xyzt::NTuple{4, Bool} = (true, true, false, true)
+    missing_value::Float64 = NaN
+    compression_level::Int = 3
+    shuffle::Bool = true
+    keepbits::Int = 12
+end
+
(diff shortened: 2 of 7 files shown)


milankl:
https://speedyweather.github.io/SpeedyWeatherDocumentation/previews/PR853/

milankl:
@stella-bourdin you want to write a test that checks that the MSLP in the default simulation is within a certain bound, say 920 hPa to 1080 hPa?

stella-bourdin:
I could try but I'll need you to explain how to write tests!

stella-bourdin:
image

src/output/variables/dynamics.jl
```diff
@@ -121,6 +121,58 @@ end
...
+    name::String = "mslp"
+    unit::String = "hPa"
+    long_name::String = "mean sea-level pressure"
+    dims_xyzt::NTuple{4, Bool} = (true, true, false, true)
+    missing_value::Float64 = NaN
+    compression_level::Int = 3
+    shuffle::Bool = true
+    keepbits::Int = 12
+end
+
```
milankl, line 137: introducing a `transform` here so that the user can change the unit of the `mslp` output. E.g. `MeanSeaLevelPressure(unit="kPa", transform = (x) -> exp(x)/1000)` would store all mslp output directly in kPa. 

In Julia a function is somewhat equivalent to a type, so `transform::F` is of type `F` and `MeanSeaLevelPressureOutput{F}` is parametric wrt to that function that way to compiler can directly compile the transform into the compilation of `output!` :wink:

src/output/netcdf_output.jl
```diff
@@ -201,7 +201,7 @@ function add_default!(
     Model::Type{<:PrimitiveDry},
 )
     add_default!(variables, Barotropic)
-    add!(variables, SurfacePressureOutput(), TemperatureOutput())
+    add!(variables, MeanSeaLevelPressureOutput(), TemperatureOutput())
```
milankl, line 204: This makes MSLP now default output for primitive models instead of surface pressure which is too dominated by orography to be meaningful dynamically

milankl:
The following lines are currently in #858 but could be moved to here then this can merge independently. In [`/test/output/netcdf_output.jl`](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/858/files#diff-716bbcca7d5f7afd569a12050c656ed19879a3bda882c17878423eedf512a96f)
```julia
    # Test surface variables
    ## MSLP
    p_scale = median(ds["pres"].var[:,:,end]) # Surface pressure order of magnitude
    mslp = ds["mslp"].var[:, :, end]
    @test all(0.8 .< mslp/p_scale .< 1.2)
```

EDIT: Done in [c05366a](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/853/commits/c05366a7f7b3da8f4539a376530ca8160176e41c)

PR #855, opened 2025-09-15
More general forcing of pressure, vorticity and divergence

milankl:
At the moment one can't force the pressure tendency equation because the `.pres_tend_grid` array is reused for `vertical_velocity!` now we recompute this therein (should be cheap because it's 2D only). This allows for

```julia
@. pres_tend_grid += u_mean_grid*∇lnp_x + v_mean_grid*∇lnp_y
```

now with `+=` such that `pres_tend_grid` can contain a custom forcing as an additional term on the right-hand side. Note this is forcing in grid space only.

@jiacheng-atmos suggested this because of [Robinson 1988](https://doi.org/10.1175/1520-0469(1988)045%3C3413:IWFIIA%3E2.0.CO;2) their equation 3

image

contains this forcing term $f$. 

TODO

- [x] test forcing the vorticity equation
- [ ] test forcing the divergence equation
- [ ] test forcing the pressure equation

Also moves the Held Suarez temperature relaxation and the linear drag into forcing and drag respectively and removes them from the parameterizations

### DIFF
--- a/src/dynamics/forcing.jl
+++ b/src/dynamics/forcing.jl
@@ -270,7 +270,7 @@ $(TYPEDFIELDS)"""
     "[DERIVED] log of sigma level per layer"
     logσ::VectorType
 
-    "[DERIVED] (inverse) relaxation time scale per layer and latitude"
+    "[DERIVED] relaxation time scale per layer and latitude (inverse, 1/s)"
     temp_relax_freq::MatrixType
 
     "[DERIVED] Term a to calculate equilibrium temperature function of latitude"


milankl:
note that [f43e217](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/855/commits/f43e2170a74764684854b1f18efe85c05da35bd0) actually fixed a bug: We had a test on StochasticStirring forcing vorticity in the primitive models but this was written into an array that was actually just overwritten, so the strength was now reduced as the earlier setting was unstable when the forcing now actually was accumulated in and not overwrittten

milankl:
The docs currently don't contain any information how to force temperature or humidity and this needs to be tested. I'm thinking of moving `HeldSuarez` and `LinearDrag` (both currently parameterizations) to `<:AbstractForcing` and `<:AbstractDrag` then they can be tests for this too!

milankl:
`HeldSuarez <: AbstractForcing` now forces the temperature equation instead of being implemented as parameterization, looks good

image

jiacheng-atmos:
Hi Milan, thank you for helping add this wonderful feature! : )

PR #857, opened 2025-09-17
Higher default BulkRichardsonDrag via surface gusts + bug fixes

### LINKED ISSUE #856: SpeedyWeather's tropics too cold
December temperature of SpeedyWeather left and Fortran SPEEDY v41 left

Image

milankl:
fixes #856

Currently boundary layer drag coefficients used for surface fluxes via Bulk-Richardson are often zero in stable stratification

image

now introduced a minimum drag of 1e-4 for that scheme

image


milankl:
With [8d8dc4f](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/857/commits/8d8dc4fd662a4738a437795cfb4552b3f7345be1) the defaults for `BulkRichardsonDrag` now produce a drag that I like much more

- introduced a gust speed of 5m/s (similar as used in model.surface_momentum_flux) to prevent too low drags causing negligible fluxes for weak winds
- critical Richardson value is still $Ri_c = 10$ which is 10x too high compared to typical values but otherwise drags are too low (barely above 1e-4)
- changed the roughness length to 0.01 m which Wikipedia says is equivalent to open [terrain/grass/little obstacles](https://en.wikipedia.org/wiki/Roughness_length)
- introduced a minimum drag of 1e-4 that with values above barely kicks in

image

milankl:
Near-surface (k=8) air temperature are also higher (28-30C max) which fixes #856 and there's a light daily cycle now visible in it over the big land masses too

https://github.com/user-attachments/assets/387b981f-e93d-4cdf-ad74-c6c33c055e1b

milankl:
It now doesn't stop convecting + raining in the tropics so maybe the humidity fluxes should be smaller now


https://github.com/user-attachments/assets/dbe9dfba-288b-4719-a9d4-cae7cad2c1a1

milankl:
Convection looks better now


https://github.com/user-attachments/assets/72779391-43d8-4e6f-bd18-2fc61bad0e0e

milankl:
drag coefficients at T63 look more balanced now

image


40-day accumulated convection looks more realistic now

image

but surface temperatures are even colder now, I blame the longwave radiation scheme

image

and total precipitation is more patchy now rather than it raining everywhere in the tropics

image

milankl:
superseded by #873 (all changes have been manually moved over anyway)

PR #858, opened 2025-09-18
Implement surface wind and temperature outputs

stella-bourdin:
Comparison of Ts and T_bottom:
image

Comparison to u10 and u_bottom:
image

Same for v
Spurious points in ratios are attributed to interpolation errors where u_bottom is very close to zero
image

### DIFF
--- a/docs/src/output.md
+++ b/docs/src/output.md
@@ -261,15 +261,21 @@ using SpeedyWeather, GeoMakie, CairoMakie
 spectral_grid = SpectralGrid()
 model = PrimitiveWetModel(spectral_grid)
 simulation = initialize!(model)
+
+# Add mean sea-level pressure for visualisation
+add!(model, SpeedyWeather.MeanSeaLevelPressureOutput())
+
 run!(simulation, period=Day(3), output=false) # some spin-up
-run!(simulation, period=Day(2), output=true)
+run!(simulation, period=Day(5), output=true)
 
-animate(simulation, output_file="test_vor_animation.mp4", variable="vor", level=1) # animate vorticity at the first vertical level
+# animate mean sea-level pressure
+animate(simulation, output_file="test_mslp_animation.mp4", variable="mslp") 
 ```
+![test_mslp_animation](test_mslp_animation.mp4)
 
-![test_vor_animation](test_vor_animation.mp4)
 
-For more options for `animate`, see below: 
+For 3D variables you can provide e.g. `level=1` as keyword argumen to `animate`
+to specify the vertical level to visualise. For more options for `animate`, see below: 
 
 ```@example netcdf 
 @doc SpeedyWeather.animate
--- a/src/output/variables/boundary_layer.jl
+++ b/src/output/variables/boundary_layer.jl
@@ -1,0 +1,86 @@
+"""Defines netCDF output for a specific variables, see [`VorticityOutput`](@ref) for details.
+Fields are: $(TYPEDFIELDS)"""
+@kwdef mutable struct ZonalVelocity10mOutput <: AbstractOutputVariable
+    name::String = "u10"
+    unit::String = "m/s"
+    long_name::String = "10m zonal wind speed"
+    dims_xyzt::NTuple{4, Bool} = (true, true, false, true)
+    missing_value::Float64 = NaN
+    compression_level::Int = 3
+    shuffle::Bool = true
+    keepbits::Int = 12
+end
+
+path(::ZonalVelocity10mOutput, simulation) = simulation.diagnostic_variables.grid.u_grid
+
+"""Defines netCDF output for a specific variables, see [`VorticityOutput`](@ref) for details.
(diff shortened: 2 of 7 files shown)


docs/src/output.md
```diff
@@ -180,6 +180,16 @@ delete!(output, :div)
 If you change the `name` of an output variable, i.e. `SpeedyWeather.DivergenceOutput(name="divergence")`
 the key would change accordingly to `:divergence`.
 
+### Surface variables 
+Mean sea-Level pressure (`mslp`), surface temperature (`tsurf`) and 10m winds (`u10`, `v10`) are computed specifically for the output.
+
+**Mean sea-level pressure** is computed as $p_{surf} \times e^{\frac{g h}{R_d T_v}}$.
```
milankl: Documenter.jl requires equations to be put into ` ``a = 1`` ` quotation marks instead of using `$`

src/output/variables/boundary_layer.jl
```diff
@@ -0,0 +1,139 @@
...
+    # Retrieve u_bottom
+    u_or_v_grid = path(variable, simulation)
+    nlayers = size(u_or_v_grid, 2)
+    u_or_v_bottom = field_view(u_or_v_grid, :, nlayers)
+    
+    z_bottom = simulation.diagnostic_variables.dynamics.a_2D_grid
+    u_or_v10 = simulation.diagnostic_variables.dynamics.b_2D_grid
+    
+    # Compute z_bottom
+    z_surf = max.(simulation.model.orography.orography, 0) #m, set negative values to zero
```
milankl: this is allocating. you could avoid this by using `z_bottom .= max.(...)` and then adding `.+= T_bottom * ...` in line 61 on the fly

src/output/variables/boundary_layer.jl
```diff
@@ -0,0 +1,139 @@
...
+    T_bottom = field_view(simulation.diagnostic_variables.grid.temp_grid, :, nlayers)
+    Δp_geopot = simulation.model.geopotential.Δp_geopot_full[end]
+    @. z_bottom = z_surf + T_bottom * Δp_geopot / simulation.model.planet.gravity
+
+    # Compute u10
+    z₀ = simulation.model.vertical_diffusion.z₀
```
milankl, line 66: I'm wondering whether we always would want to use the roughness length from the vertical diffusion component ... we may want to tweak that without immediately chaning the 10m output, what do you think?

src/output/variables/boundary_layer.jl
```diff
@@ -0,0 +1,139 @@
...
+    T_grid = simulation.diagnostic_variables.grid.temp_grid
+    nlayers = size(T_grid, 2)
+    T_bottom = field_view(T_grid, :, nlayers)
+
+    # Other parameters
+    κ = simulation.model.atmosphere.κ
+    σ_bottom = simulation.model.geometry.σ_levels_full[end]
+
+    # Compute Ts
+    @. Ts = T_bottom * σ_bottom ^ (-κ) - 273.15 # Convert to °C
```
milankl: this could be written as `transform` again that could be user facing to determine the units of the output?

src/output/variables/dynamics.jl
```diff
@@ -145,7 +143,7 @@ function output!(
     # escape immediately after first call if variable doesn't have a time dimension
     ~hastime(variable) && output.output_counter > 1 && return nothing
 
-    lnpₛ = path(variable, simulation)
+    lnpₛ = simulation.diagnostic_variables.grid.pres_grid
```
milankl: I would probably still use `path(::MeanSeaLevelPressureOutput, simulation)` here given that it's the core variable that's read in?

test/output/netcdf_output.jl
```diff
@@ -142,7 +153,22 @@ end
...
+
+    ## test u10, v10 existence
+    @test haskey(ds, "u")
+    @test haskey(ds, "u10")
+    @test haskey(ds, "v")
+    @test haskey(ds, "v10")
+
+    # and 10m values cannot exceed lowermost layer of u,v value
+    @test maximum(abs.(ds["u10"].var[:, :, end])) < maximum(abs.(ds["u"].var[:, :, end, end]))
+    @test maximum(abs.(ds["v10"].var[:, :, end])) < maximum(abs.(ds["u"].var[:, :, end, end]))
```
milankl, line 166: you used `max(array...)` I changed this to `maximum` which is the correct way of doing this in Julia as it's non-allocating. Splatting out arrays is generally not good practice unfortunately

milankl (commented):
Very nice! I think I would like to redirect this pull request to main. Feel free to address the outstanding points otherwise I'll do this in the next days. Many thanks Stella!!!

src/output/variables/boundary_layer.jl
milankl: let's leave this for now, I'll add a TODO into the code

milankl:
Docs are rendered as

image

will update that

PR #869, opened 2025-09-25
Update and rename CITATION.cff to CITATION.bib

### LINKED ISSUE #804: Zenodo archive not created since v0.10
Every release should trigger a Zenodo archive. But since v0.10 it doesn't, the last v0.16 for example triggers this error

Image

There seems to be some issue of how `CITATION.cff` is handled, see https://github.com/zenodo/zenodo/issues/2515 most of which seem to involve the way how `license` is declared in the `CITATION.cff` which we do not use though?

milankl:
hopefully fixes #804


milankl:
@abelsiqueira this removes the `preferred-citation` level, that's what you meant I hope? Many thanks for this input!

abelsiqueira:
I think in this case you want both a root level and a preferred-citation (to link to the JOSS paper). The root level would be the description of the software only: authors, title, type: software, license: MIT, etc. You can add the JOSS DOI as an `identifier`, for instance, to differentiate from the Zenodo DOI. 
The preferred-citation should include some of the current preferred-citation: authors (same or different), title (of the JOSS paper), type: article, and other article-specific things, such as the number, pages, etc.

Zenodo should just ignore preferred-citation, and that is fine, because you're creating the software deposition. (https://help.zenodo.org/docs/github/describe-software/citation-file/)
GitHub, however, should use the preferred-citation for the "Cite this repository" BibTeX generator. (https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files#citing-something-other-than-software)
That's why you need both a root level (Zenodo/DOI/Software info) and a preferred-citation level (GitHub BibTeX, and hopefully humans reading it).

I give an overview of cffinit going over the main CFF fields here, if that helps: https://www.youtube.com/watch?v=zcgLIT5Qd4M

milankl:
Awesome! I think I got it now and understood what all the CFF stuff is about

PR #872, opened 2025-10-09
Enhance README with interactive usage example

milankl:
Added a pluto usage gif and updated the globe videos to not be laggy anymore. EDIT higher quality gif using ffmpeg's palettegen options that removed some previous weird grainy artefacts

speedyweather_pluto

not the highest resolution but colors aren't distorted anymore. I think that's good enough, otherwise the file size increases a lot as it's a long gif but very static overall


milankl:
The mp4 is 1/10th of the size ...

https://github.com/user-attachments/assets/e5a755d2-9a77-4c76-a374-217ca658072e

Issue #735, opened 2025-05-21
Test for UV_from_vordiv!


maximilian-gelbrecht:
We currently don't have a unit test for `UV_from_vordiv!` and I wanted to add one. I am currently struggling to set up one, as I am getting wrong results. Probably I am missing some scaling we do.  

All of the four tests are failing at the moment. 

@milankl Do you have an idea? 

```Julia 
using SpeedyWeather, Test 
NF=Float32 

SG = SpectralGrid(; NF, nlayers=1)
S = SpectralTransform(SG)   

# randomly generate u, v fields in grid space
u_orig_grid = SG.Grid(randn(NF, SG.npoints))
v_orig_grid = SG.Grid(randn(NF, SG.npoints))

# copy u, v fields for later comparison
u_orig_grid_copy = deepcopy(u_orig_grid)
v_orig_grid_copy = deepcopy(v_orig_grid)

# apply 1/coslat-scaling to u, v fields (as they are expected to have this scaling in spectral space)
RingGrids.scale_coslat⁻¹!(u_orig_grid_copy)
RingGrids.scale_coslat⁻¹!(v_orig_grid_copy)

# transform u, v fields to spectral space
u_orig = transform(u_orig_grid_copy, S)
v_orig = transform(v_orig_grid_copy, S)
        
# compute vorticity and divergence in spectral space
vor = curl(u_orig, v_orig, S)
div = divergence(u_orig, v_orig, S)
        
U = similar(u_orig)
V = similar(v_orig)
SpeedyWeather.SpeedyTransforms.UV_from_vordiv!(U, V, vor, div, S)
        
# transform back and unscale u, v fields
u = transform(U, S)
v = transform(V, S)
RingGrids.scale_coslat⁻¹!(u)
RingGrids.scale_coslat⁻¹!(v)

@test u ≈ u_orig_grid
@test v ≈ v_orig_grid
        
# way via inverse laplacian
# streamfunction and velocity potential
Ψ = ∇⁻²(vor, S)
Φ = ∇⁻²(div, S)

∂Ψ_∂λ, ∂Ψ_∂θ = ∇(Ψ, S)
∂Φ_∂λ, ∂Φ_∂θ = ∇(Φ, S)

∂Ψ_∂λ_grid = transform(∂Ψ_∂λ, S)
∂Ψ_∂θ_grid = transform(∂Ψ_∂θ, S)
[...]

PR #873, opened 2025-10-13
GPU Parameterizations via kernel over ij

milankl:
Overview of changes

- rewrite the physics parameterizations Oceananigans-style with `ij` index as first argument for GPU kernel fusion across parameterizations
- remove the `ColumnVariables`, parameterizations would directly read from / write to the arrays in `diagnostic_variables`
- recompute a bunch of variables on the fly (e.g. saturation humidity, moist static energy) none of them are very expensive
- have geopotential once centrally computed as it is a vertical integration that should not be redone on the fly?
- unpack `model` to "model parameters" (components that are generally contain parameters/boundary conditions needed to compute parameterizations) and "parameterizations" that are are evaluated one after another
- make sure that all parameterizations can be passed onto the GPU (only isbits fields!)
- some model components like orography can be adapted to only retain the actually GPU-adaptable fields
- generally this means we have custom types outside of kernels and we unpack them into named tuples inside that behave very similarly (`model.orography.orography` would be the same whether it's a `::AbstractModel` or `::NamedTuple`)
- geopotential is now calculated in grid space only, not in spectral space to avoid an additional transform or to do that calculation twice
- initialize the model on both CPU and GPU

Other changes
- `OneBandLongwave` implemented with `FriersonLongwaveTransmissivity`
- Boundary layer drag revisited
- Snow/sea ice insulation now in flux computation
- disables the OneBandShortwave as those still have to be adapted to the ij-logic

### DIFF
--- a/src/dynamics/geopotential.jl
+++ b/src/dynamics/geopotential.jl
@@ -81,8 +81,8 @@ end
 
 function geopotential_cpu!(geopotential, temp, humid, orography, gravity, Geopotential, atmosphere)
     nlayers = size(temp, 2)
-    @inbounds for lm in eachharmonic(temp)
-        geopotential_compute!(lm, geopotential, temp, humid, orography, gravity, Geopotential.Δp_geopot_half, Geopotential.Δp_geopot_full, nlayers, atmosphere)
+    @inbounds for ij in eachgridpoint(geopotential)
+        geopotential_compute!(ij, geopotential, temp, humid, orography, gravity, Geopotential.Δp_geopot_half, Geopotential.Δp_geopot_full, nlayers, atmosphere)
     end
 end 
 


milankl:
Looking at this I'm almost wondering whether we should do something like

```julia
ij = @index(Global, Linear)     # every horizontal grid point ij

for p in model.parameterizations
    key, parameterization = p
    parameterization!(ij, diagn, progn, parameterization, model)
end
```

instead. This would give more flexibility in terms of adding/removing parameterizations, changing the order etc. But it might also just be confusing and too easily a mess for the user

milankl:
What I'm also not sure about is that we'd easily have many nested parameterization calls, e.g. `albedo!` would call up to 4 methods of `albedo!` before an actual non-dispatch/unpacking operation is hit. Do we `@inline` them all, or should we just leave them as they are?

```julia
# dispatch over model.albedo
albedo!(ij, diagn::DiagnosticVariables, progn::PrognosticVariables, model::PrimitiveEquation) =
    albedo!(ij, diagn, progn, model.albedo, model)

# composite OceanLandAlbedo: call separately for ocean and land with .ocean and .land
function albedo!(ij, diagn::DiagnosticVariables, progn, albedo::OceanLandAlbedo, model)
    albedo!(ij, diagn.physics.ocean, progn, albedo.ocean, model)
    albedo!(ij, diagn.physics.land, progn, albedo.land, model)
end

# function barrier to unpack
albedo!(ij, diagn, progn, albedo::OceanSeaIceAlbedo, model) = albedo!(ij, diagn.albedo, progn.ocean, albedo)

# actual albedo computation
function albedo!(ij, diagn_albedo::AbstractField2D, ocean, albedo::OceanSeaIceAlbedo)
    (; sea_ice_concentration ) = ocean
    (; albedo_ocean, albedo_ice) = albedo

    # set ocean albedo linearly between ocean and ice depending on sea ice concentration
    diagn_albedo[ij] = albedo_ocean + sea_ice_concentration[ij] * (albedo_ice - albedo_ocean)
end
```

maximilian-gelbrecht:
Oh yeah, I also already wanted to start this, I am happy to chime in here as well. 

> Looking at this I'm almost wondering whether we should do something like
> 
> ```julia
> ij = @index(Global, Linear)     # every horizontal grid point ij
> 
> for p in model.parameterizations
>     key, parameterization = p
>     parameterization!(ij, diagn, progn, parameterization, model)
> end
> ```
> 
> instead. This would give more flexibility in terms of adding/removing parameterizations, changing the order etc. But it might also just be confusing and too easily a mess for the user

Yes, something like this would be good. I think we should use this opportunity to have the parametrization really modular and with a standard API so that adding and modifying further parametrization is even easier. 

With exactly the pseudocode as above, I am not 100% sure yet. Basically, a key question for the design here is whether or not we hand the `model` over the kernels or not. That's also something we discussed a lot for Terrarium. The problem with handing over the full `model` is that then also the complete model and every single of its fields has to be adapted for GPU usage. That's realistically not really easily possible and probably not what we want, e.g. the output will be very problematic there, and also generally all `Strings` can't be adapted to GPU usage. So that mean, probably we'll just hand over an array or custom struct with all parametrizations and then have function signatures `parameterization!(ij, diagn, progn, parameterization)`. Do you think we need other components from our `model` types in the parametrizations? Then we'd probably best just hand over exactly those. 

> What I'm also not sure about is that we'd easily have many nested parameterization calls, e.g. `albedo!` would call up to 4 methods of `albedo!` before an actual non-dispatch/unpacking operation is hit. Do we `@inline` them all, or should we just leave them as they are?

I think inlining everything should be fine, right?

milankl:
> The problem with handing over the full model is that then also the complete model and every single of its fields has to be adapted for GPU usage. That's realistically not really easily possible and probably not what we want, e.g. the output will be very problematic there, and also generally all Strings can't be adapted to GPU usage. So that mean, probably we'll just hand over an array or custom struct with all parametrizations and then have function signatures parameterization!(ij, diagn, progn, parameterization). Do you think we need other components from our model types in the parametrizations? Then we'd probably best just hand over exactly those.

There's mostly parameters in `model` that need to be available to execute the parameterization. So sure, `convection!` needs `model.convection` but it's much more complicated than that because we generally need read access to `model.planet`, `model.clausius_clapeyron`, `model.atmosphere` and generally all of them. If someone implements a new parameterization and needs access to a model component that isn't passed on then that's really annoying as many function signatures need to be adapted. At the moment I'm trying to unpack stuff only just before the actual compute kernel is formulated. That way extending/adapting parameterizations is much more self-contained to a few lines of code.
Alternatively, we could have one kernel per parameterization like we do for land, ocean and sea ice. But that doesn't allow us to fuse them. How does Oceananigans do it? is there not a possibility to have read access to everything from `model` that's GPU compatible?

maximilian-gelbrecht:
Oceananigans also doesn't hand over the full model struct, and instead unpacks a lot before launching kernels. 

In Terrarium we mostly solved this by being self-contained in all submodels. I know this isn't quite as easy here, but something I could imagine is that we'll basically just group those important parameters structs together in a new struct that we do actually adapt for GPU usage. 

Let's have a call about this some time this week. I also need to have a think before, but I am sure we can figure something out that works.

maximilian-gelbrecht:
Maybe we'll just have something like 

```Julia 
struct Parameterizations 
   atmosphere # these are mostly constant parameters ..
   clausius
   planet 
   ... 
   processes # this is a vector or dictionary with all parametrization processes 
end 
```

milankl:
We now simply have 

```julia
function get_parameterizations(model::PrimitiveWet)
    return (# diffusion
            vertical_diffusion = model.vertical_diffusion,
            
            # hydrological cycle
            convection = model.convection,
            large_scale_condensation = model.large_scale_condensation,
            
            # radiation
            albedo = model.albedo,
            optical_depth = model.optical_depth,
            shortwave_radiation = model.shortwave_radiation,
            longwave_radiation = model.longwave_radiation,
            
            # surface fluxes
            boundary_layer_drag = model.boundary_layer_drag,
            surface_momentum_flux = model.surface_momentum_flux,
            surface_heat_flux = model.surface_heat_flux,
            surface_humidity_flux = model.surface_humidity_flux,
            
            # stochastic physics
            stochastic_physics = model.stochastic_physics,
    )
end
```
as a method for `PrimitiveWet` and another one for `PrimitiveDry`, i.e. where the model is defined one would just have a method that says what the parameterizations are. Similarly we have
```julia
function get_model_parameters(model::PrimitiveWet)
    return (orography = model.orography.orography,              # use only orography field as strings aren't GPU-compatible
            atmosphere = model.atmosphere,
            planet = model.planet,
            geometry = model.geometry,
            land_sea_mask = model.land_sea_mask.mask,           # similarly for mask
            clausius_clapeyron = model.clausius_clapeyron,
    )
end

```
We could eventually rework this into something with a user interface or done somehow automatically but it just has to return a `NamedTuple` that serves as a subset of `model`. That way it can be used exactly like `model` inside the parameterizations but it doesn't contain the non-GPU compatible model components. I'm unpacking e.g. `orography` into just the orography field, to throw out the strings etc. Does this make sense to you?

We have a problem with the albedo though as it can be both: holding an array or just holding parameters to calculate the albedo on every time step diagnostically. So I reckon we need to split this somehow into `model.prescribed_albedo` and `model.albedo` so that the former behaves like orography or land-sea mask (boundary condition) and the latter dynamically computes it (even if that's just copying array elements around)

maximilian-gelbrecht:
I think for individual components like the albedo in this case, we could also just define a custom `adapt_structure` that circumvents this problem. 

As discussed yesterday, I don't think that's an approach that we should do on the most top-level for our main model type, but I think for a low level component, that's fine.

maximilian-gelbrecht:
So, for all "normal" components/processes it would suffice to just add a 
```Julia 
@adapt_structure ManualAlbedo
```
for example. For the `AlbedoClimatology` we could add 

```Julia 
function Adapt.adapt_structre(to, albedo::AlbedoClimatology) 
       (albedo = adapt(to, albedo.albedo), )
end 
```

maximilian-gelbrecht:
And yeah, let me know if I can do something concrete here for this PR.

milankl:
> And yeah, let me know if I can do something concrete here for this PR. 

Figure out whether we need to `@inline` for all the `parameterization!` calls? 

I have adapted large scale condensation yesterday so we start testing that

maximilian-gelbrecht:
> > And yeah, let me know if I can do something concrete here for this PR.
> 
> Figure out whether we need to `@inline` for all the `parameterization!` calls?
> 
> I have adapted large scale condensation yesterday so we start testing that

Even after reading up a bit more about it, I don't think there's much that we can do then just to do a A/B test what's faster. We should probably also take the Enzyme performance into account as well, as inlining of course changes the IR. 

For comparison, Oceananigans basically inlines just about everything.

milankl:
Okay this at least runs through now on the CPU

```julia
spectral_grid = SpectralGrid(trunc=31)
```

executing
- large-scale condensation
- albedo
- surface condition (the new surface thermodynamics)
- surface momentum flux (renamed from surface wind)
- surface heat flux
- surface humidity flux
- stochastic physics

milankl:
While this does run through now, it allocates massively and is 10x slower already (though now all parameterizations are even included)

Was:
```julia
julia> @btime SpeedyWeather.parameterization_tendencies!($diagn, $progn, $progn.clock.time, $model)
  1.199 ms (2 allocations: 96 bytes)
```

now:
```julia
  13.434 ms (220893 allocations: 146.24 MiB)
```

milankl:
Also interestingly, `model.orography.orography` inside the kernel isn't an `::AbstractField2D` it becomes `Base.Experimental.Const`? It can be used the same way but type constraints are therefore avoided

```julia
julia> SpeedyWeather.parameterization_tendencies!(diagn, progn, model)
ERROR: MethodError: no method matching geopotential!(::Int64, ::Field{…}, ::Field{…}, ::Base.Experimental.Const{…}, ::Float32, ::Geopotential{…})
The function `geopotential!` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  geopotential!(::Any, ::AbstractField3D, ::AbstractField3D, ::AbstractField2D, ::Real, ::Geopotential)
   @ Main.SpeedyWeather ~/git/SpeedyWeather/src/dynamics/geopotential.jl:87
```

milankl:
So it's mostly the (theoretically cheap 2D) surface fluxes being the culprit, if I switch them off I get

```julia
  505.083 μs (25397 allocations: 2.11 MiB)
```

EDIT: no something weird nonlinear is going on here, e.g. convection + large-scale condensation only don't cause 100MB allocations, neither do surface fluxes,  but together they do.

maximilian-gelbrecht:
> Also interestingly, `model.orography.orography` inside the kernel isn't an `::AbstractField2D` it becomes `Base.Experimental.Const`? It can be used the same way but type constraints are therefore avoided
> 
> ```julia
> julia> SpeedyWeather.parameterization_tendencies!(diagn, progn, model)
> ERROR: MethodError: no method matching geopotential!(::Int64, ::Field{…}, ::Field{…}, ::Base.Experimental.Const{…}, ::Float32, ::Geopotential{…})
> The function `geopotential!` exists, but no method is defined for this combination of argument types.
> 
> Closest candidates are:
>   geopotential!(::Any, ::AbstractField3D, ::AbstractField3D, ::AbstractField2D, ::Real, ::Geopotential)
>    @ Main.SpeedyWeather ~/git/SpeedyWeather/src/dynamics/geopotential.jl:87
> ```

This is by our design in #824 as well. When loading into kernels we convert the `Field` into a regular `CuArray`/`Array` (because of our `adapt_structure`) into a `CuDeviceArray`when on GPU and then into a `Base.Experimental.Const` because of `@Const` (all in one step, but this is the logic chain). We might have to reconsider this approach though (that I more or less copied from Oceananigans). I have to check in #878

maximilian-gelbrecht:
Ok, this is now the big main feature branch for the overhaul of the parametrizations. I just copy over here some notes from the other PR: 

## Some little GPU-ify notes

* Error messages are more opaque when something goes wrong -> Claude does a pretty good job at explaining those errors 
* Functors don't work on GPU unless they are inlined (and therefore effectively not functors anymore) 
* Generally, when in doubt, better `@inline` a function
* No Colons `:` 
* We highjack `adapt` to just be responsible for transfer to kernels, for some structs like `SpectralGrid` it explictily adapts to `nothing` -> If you want to transfer objects between devices, use `on_architecture` 
* We currently adapt our `Field` and `LowerTriangularray` to just adapt the `.data` field only. If you need the `Spectrum` or `Grid` it should be in the `parameters`/`model` 
* Structs should ideally don't have extra type parameters. When we have types that can be directly reconstructed just from the arguments alone, it makes our live for kernel adaption a lot easier 
* No mutable structs allowed -> if you really need it think about workarounds with just handing over the important fields
* Iterations over `NamedTuples` don't work, here we e.g. manual unroll a loop with a `@generated` function  
* Think about reducing memory bandwidth by reducing the amount of diagnostic variables we really need 
* Think about performance when branch/code divergence needs to be done

maximilian-gelbrecht:
In principle we could even add a 

```Julia 
add!(model, parametrization) 
```

function similar to how users can add tracers. That looks cool for demonstrations but it has the problem that the order of the parametrizations is important. So it's also not really that practical in a real application, even with potential kwarg like `add!(model, parametrization, i = 1)`.

src/physics/convection.jl
```diff
@@ -168,37 +122,46 @@ function convection!(
...
-	column.rain_convection *= pₛΔt_gρ                                   # convert to [m] of rain during Δt
-    column.rain_rate_convection = column.rain_convection / Δt_sec       # rate: convert to [m/s] of rain
+    pₛΔt_gρ = (pₛ * Δt_sec / gravity / water_density) * deep_convection # enforce no precip for shallow conv
+	rain_convection *= pₛΔt_gρ                                           # convert to [m] of rain during Δt
+    
+    # Store precipitation in diagnostic arrays
+    diagn.physics.rain_convection[ij] += rain_convection
+
+    # TODO: double check, because preivously this was rain_rate_convection, which didn't exist
+    diagn.physics.total_precipitation_rate[ij] = rain_convection / Δt_sec    # rate: convert to [m/s] of rain
```
maximilian-gelbrecht: Just commenting here, to point out that I changed this here just so that I don't get any errors. It still needs to be corrected to the correct rate variable. 

But maybe the rain rates are also one thing were we can actually also accumulate all in one rate instead of separate ones to save some memory.

RingGrids/src/field.jl
```diff
@@ -401,6 +405,12 @@ Base.:(==)(F1::AbstractField, F2::AbstractField) = fields_match(F1, F2) && F1.da
 Base.all(F::AbstractField) = all(F.data)
 Base.any(F::AbstractField) = any(F.data)
 
+# reductions
+Base.minimum(F::AbstractField) = minimum(F.data)
+Base.maximum(F::AbstractField) = maximum(F.data)
+Base.sum(F::AbstractField) = sum(F.data)
+Base.extrema(F::AbstractField) = extrema(F.data)
```
milankl, line 412: @maximilian-gelbrecht wierdly these were need and currently not covered by our broadcasting efforts

RingGrids/src/field.jl
```diff
@@ -517,6 +527,7 @@ end
 Adapt.adapt_structure(to, field::AbstractField) = Adapt.adapt(to, field.data)
   
 Architectures.architecture(field::AbstractField) = architecture(field.grid)
+Architectures.on_architecture(field::AbstractField, x) = on_architecture(architecture(field), x)
```
milankl, line 530: @maximilian-gelbrecht added this as it felt convenient to just get a field on the architecture of another field

milankl:
Also @maximilian-gelbrecht I've also made some change here so that `simulation = initialize!(model)` works for PrimtiveWetModel on the GPU just so we don't duplicate efforts. Sorry this isn't merged yet ... 🙈

RingGrids/src/field.jl
milankl: Relatedly,

```julia
field[field .> a] .= b
```

doesn't currently work but weirdly

```julia
mask = field .> a
fill!(field[mask.data], b)
```

does!

maximilian-gelbrecht: `sum` should probably have `kwargs` for specificing the `dims`. 

Mh, I guess it's good that `field .> a` is also a `Field` but then the indexing doesn't work anymore. Btw: 
```Julia 
field[(field .> 0.5).data] .= 0
```
Also works. It's just the problem that only boolean arrays work for indexing.

maximilian-gelbrecht:
That's great though, one step closer to it running ;)

milankl:
```julia
julia> spectral_grid = SpectralGrid(architecture=GPU());
julia> model = PrimitiveWetModel(spectral_grid);
julia> simulation = initialize!(model);
julia> progn, diagn, model = SpeedyWeather.unpack(simulation);
julia> SpeedyWeather.parameterization_tendencies!(diagn, progn, model)
julia> 
```

all parameterizations run through on GPU and CPU now :partying_face:

maximilian-gelbrecht:
Great! So now, a lot of cleanup, a bit of benchmarking and then we both do a thorough review?

maximilian-gelbrecht:
Here's the script I used in the other PR
```Julia
using SpeedyWeather, BenchmarkTools, CUDA

for ntrunc in [31, 63, 127, 255, 511]
    arch = GPU()
    spectral_grid = SpectralGrid(trunc=ntrunc, architecture=arch)
    @btime CUDA.@sync SpeedyWeather.parameterization_tendencies!($diagn, $progn, $model)
end 
```

milankl:
The last commit computes the geopotential in grid space instead of spectral space as we currently do in main. While geopotential calculation in spectral space is cheaper (because fewer wavenumbers than horizontal grid points) we need the grid geopotential for the parameterizations anyway and it's not a variable that I want to recalculate on the fly due to the vertical integration. So I want to move towards calculating the geopotential once and have it's own field in `diagnostic_variables.grid.geopotential` which can then be reused.

In contrast I'm trying to move towards recalculating the virtual temperature on the fly as it's cheap to do so and so never store that variable explicitly but have `virtual_temperature` as a function that can be called anytime with temperature and specific humidity (and an atmosphere for the $$\mu$$ factor that depends on the gas constants for dry vs vapour)

milankl:
TODO just to keep track of changes
- [ ] check that the new geopotential calculation works (e.g. for a dry core)
- [ ] check that the new geopotential calculation didn't affect the ShallowWaterModel (functions are shared)
- [ ] initialization of ocean, land and initial conditions
- [ ] double check that the precipitation is correctly stored and passed on
- [ ] vertical diffusion threw some errors

TODOs for the future (not this PR)
- [ ] remove central virtual temperature calculation and recompute on the fly

src/dynamics/initial_conditions.jl
```diff
@@ -380,51 +395,56 @@ $(TYPEDFIELDS)"""
...
+        r = $radius * acos(X)
 
         # Eq (13)
-        return -2perturb_uₚ*radius/R^2 * exp(-(r/R)^2) * acos(X)/sqrt(1-X^2) * cosφc * sind(λ-λc)
+        return -2 * $perturb_uₚ * $radius / $R^2 * exp(-(r / $R)^2) * acos(X) / sqrt(1 - X^2) * $cosφc * sind(λ - $λc)
     end
 
     # apply those to set the initial conditions for vor, div
-    set!(progn, model; vor=vor, div=div, lf=1)
+    set!(progn, model; vor = vor, div = div, lf = 1)
```
maximilian-gelbrecht: This isn't directly related to this PR, but just a reminder that we had problems with that version of `set!` for `vor` and `div`. unit tests were always failing. this is also related to #735, #563, #893, #763. So I would be careful with it.

maximilian-gelbrecht:
Ok, full model initializiation now works for me on GPU

milankl:
Yes for me too, on both CPU and GPU, but can't execute the parameterization on the GPU because I introduced the model class 🙈 will fix this later!

milankl:
Quick performance check where we are, all with

```julia
@btime SpeedyWeather.parameterization_tendencies!($diagn, $progn, $model)
```

(or `$t::DateTime` on `main` as additional argument), excluding vertical diffusion as that still throws some errors. GPU is a small L4 GPU 

| Architecture | Resolution | Time | Allocations | Faster |
| - | - | - | - | - |
| M4 CPU main | T31L8 | 1.4ms | (2 allocations: 96 bytes) | 1x |
| Intel CPU this PR | T31L8 | 12.581 ms | (12511 allocations: 16.52 MiB) | 0.1x |
| L4 GPU this PR | T31L8 | 94.955 μs | (557 allocations: 94.39 KiB) | 15x |

and at higher resolution, still OctahedralGaussianGrid, dealiasing of 2

| Architecture | Resolution | Time | Allocations | Faster |
| - | - | - | - | - |
| M4 CPU main | T127L16 | 24.506 ms | (2 allocations: 96 bytes) | 1x |
| Intel CPU this PR | T127L16 | 178.176 ms | (164309 allocations: 209.96 MiB) | 0.1x |
| L4 GPU this PR | T127L16 | 95.631 μs | (557 allocations: 94.39 KiB) | 255x |

So on the GPU we are doing absolutely the right thing but the same code on the CPU just creates allocations, increasingly so with higher resolution which tells me there's actually some arrays allocated or similar.

milankl:
Vertical diffusion is now working on the GPU. I only changed the function signature with https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/873/commits/762b57ce5d2773c7704da1847493f4e50106cc89 @maximilian-gelbrecht I forgot that `var::AbstractField` doesn't work on the GPU (and throws weird errors) as we only pass on `var.data` ... 🙈 

Now the timings are slightly slower with vertical diffusion but only a few percent, again at T31L8 on the L4 GPU

```julia
  101.562 μs (588 allocations: 98.52 KiB)
```

and on the CPU another few allocations now

```julia
  16.968 ms (12603 allocations: 16.52 MiB)
```

I might generally expect the new CPU version to be slower as the memory layout now is not working in our favour anymore. But will need to get to the bottom of this

milankl:
@vchuravy any idea what we could test it there's a large amount of allocations on the CPU but not on the GPU? 👆 

UPDATE: Some RefValue dereferencing we took out, but then also we now just use loop on the CPU and bypass KA, that's seems to work fine and is minimally more code in this case.

milankl:
Just ran the T127L16 comparison CPU (M4) vs GPU (L40) again and I get on the CPU
```julia

  9.870 ms (21 allocations: 704 bytes)
```

and on the GPU
```julia
  58.248 μs (551 allocations: 89.00 KiB)
```

so also green flags here! 170x faster on the GPU while not being slow on the CPU 🎉 So I think performance-wise we solved all problems now

milankl:
Okay all parameterization run stable now but there are still some peculiarities, surface longwave radiation up has really low values at coastlines

image

presumably connected to the soil temperature having a funny mask it shouldn't have

image

Similar for SST

image

and surface sensible heat flux

image

It rains everywhere in the tropics

image

and a lot of snow in the extratropics

image

the last two presumably because of the surface temperature being too cold

image

also likely related is the soil moisture going nuts in the tropics because all that water is dumped down there

image

also related is cloud top (currently as layer index) showing generally too many clouds

image

milankl:
I think I resolved most issues but now the surface air temperature drops over a 1-year simulation to freezing almost everywhere

image

src/dynamics/geopotential.jl
```diff
@@ -23,24 +39,56 @@ Precomputes constants for the vertical integration of the geopotential, defined
...
+    (; orography) = model.orography
+
+    # this triggers a fallback to use virtual temperature = absolute temperature for PrimitiveDry
+    A = model isa PrimitiveWet ? model.atmosphere : nothing
+
+    arch = architecture(temp)
+    launch!(arch, RingGridWorkOrder, (size(temp, 1),), geopotential_kernel!, geopotential, temp, humid, orography, g, G, A)
+end
+
+@kernel function geopotential_kernel!(geopotential, temp, humid, orography, gravity, Geopotential, atmosphere)
```
maximilian-gelbrecht, line 89: I'd do `@kernel inbounds=true function ... ` and then remove the `@inbounds ` below. I think all those array indices are safe here.

src/dynamics/implicit.jl
```diff
@@ -315,25 +315,26 @@ function implicit_correction!(
     # SEMI IMPLICIT CORRECTIONS FOR DIVERGENCE
     # calculate the combined tendency G = G_D + ξRG_T + ξUG_lnps to solve for divergence δD
     (; pres_tend, div_tend) = diagn.tendencies
-    G = diagn.dynamics.a        # reuse work arrays, used for combined tendency G
-    geopot = diagn.dynamics.b   # used for geopotential
+    G = diagn.dynamics.a                # reuse work arrays, used for combined tendency G
+    geopotential = diagn.dynamics.b
+    geopotential .= 0                   # reset geopotential work array for accumulation below
 
     for k in 1:nlayers
```
maximilian-gelbrecht, line 323: Oh, this we still have to put in kernels. I'll do that in another PR. Why isn't this causing errors when initialising the model?

src/dynamics/tendencies.jl
```diff
@@ -873,26 +891,29 @@ function SpeedyTransforms.transform!(
...
-    temp_grid_prev, humid_grid_prev, u_grid_prev, v_grid_prev) = diagn.grid
+    pres_grid_prev, u_grid_prev, v_grid_prev, temp_grid_prev, humid_grid_prev) = diagn.grid
 
     vor   = get_step(progn.vor, lf)     # relative vorticity at leapfrog step lf
     div   = get_step(progn.div, lf)     # divergence at leapfrog step lf
     temp  = get_step(progn.temp, lf)    # temperature at leapfrog step lf
     humid = get_step(progn.humid, lf)   # humidity at leapfrog step lf
     pres  = get_step(progn.pres, lf)    # logarithm of surface pressure at leapfrog step lf
 
+    # @info progn.pres[1, :]
```
maximilian-gelbrecht: leftover from debugging?

src/dynamics/time_integration.jl
```diff
@@ -372,24 +376,20 @@ function timestep!(
...
-        parameterization_tendencies!(diagn, progn, time, model)
+        parameterization_tendencies!(diagn, progn, model)
         ocean_timestep!(progn, diagn, model)    # sea surface temperature and maybe in the future sea ice
         sea_ice_timestep!(progn, diagn, model)  # sea ice
         land_timestep!(progn, diagn, model)     # soil moisture and temperature, vegetation, maybe rivers
     end
 
-    if model.dynamics                                           # switch on/off all dynamics
-        forcing!(diagn, progn, lf2, model)
-        drag!(diagn, progn, lf2, model)
```
maximilian-gelbrecht, line 392: did forcing! and drag! move into the dynamics_tendencies! or why are they not here anymore?

src/physics/define_column.jl
```diff
@@ -1,146 +1,3 @@
+# deprecated define to avoid errors when defining old code, to be removed
 abstract type AbstractColumnVariables end
```
maximilian-gelbrecht, line 2: We can remove the last traces of ColumnVariables, now, right?

src/physics/land_sea_mask.jl
```diff
@@ -45,21 +45,27 @@ function mask!(
...
-        end
-    end
-
+    arch = architecture(field)
+    launch!(arch, RingGridWorkOrder, size(field), mask_kernel!, field, mask, val, masked_val)
     return field
 end
 
+# 2D, 3D or ND variant via Cartesian indexing
+@kernel inbounds=true function mask_kernel!(field, mask, val, masked_val)
```
maximilian-gelbrecht, line 54: This could also be a broadcast `field[mask.data .== val] .= val`, no? 

The `.data` is necessary here beyond some current broadcasting problems with Field, because only Array{Bool} or CuArray{Bool} can be used for indexing like this.

src/physics/radiation/longwave_radiation.jl
```diff
@@ -85,11 +71,14 @@ $(TYPEDFIELDS)"""
     "[OPTION] Radiative forcing constant (W/m²/K²)"
     α::NF = 0.025
 
+    "[OPTION] Eemissivity of the atmosphere [1]"
+    emissivity_atmosphere::NF = 0.98
+
     "[OPTION] Effective emissivity for surface flux over ocean [1]"
-    emissivity_ocean::NF = 0.6
+    emissivity_ocean::NF = 0.98
```
maximilian-gelbrecht: Any reason for the changed values here? Just for debugging?

src/physics/radiation/longwave_radiation.jl
```diff
@@ -98,65 +87,76 @@ $(TYPEDFIELDS)"""
...
+@propagate_inbounds function longwave_radiation!(ij, diagn, progn, longwave::JeevanjeeRadiation, model)
+
+    T = diagn.grid.temp_grid                            # to match Seeley, 2023 notation
+    dTdt = diagn.tendencies.temp_tend_grid
+    pₛ = diagn.grid.pres_grid_prev[ij]                  # surface pressure [Pa]
+    nlayers = size(T, 2)
+    (; temp_tend_grid) = diagn.tendencies
+
+    (; α) = longwave
+    τ⁻¹ = inv(convert(eltype(T), Second(longwave.time_scale).value))
```
maximilian-gelbrecht, line 107: This is correct here, but something I am wondering about in general is when `convert` is okay. 

In some cases I already had problems with it because the compiler was complaining about it being a runtime dispatch, in other words cases the compiler seems to be able to figure out this statically. In theory it should always be able to specialise on this based on the input arguments. But sometimes this doesn't seem to work. I am a bit confused.

src/physics/surface_fluxes/humidity.jl
```diff
@@ -1,174 +1,207 @@
...
 
-    flux_sea *= (1 - land_fraction)             # weight by ocean fraction of land-sea mask
-    column.flux_humid_upward[end] += flux_sea   # accumulate with += into end=lowermost layer total flux
-    column.surface_humidity_flux = flux_sea     # output/diagnose: ocean sets flux (=), land accumulates (+=)
+    # store without weighting by land fraction for coupling
+    diagn.physics.ocean.surface_humidity_flux[ij] = flux_ocean         
+    flux_ocean *= (1 - land_fraction)             # weight by ocean fraction of land-sea mask
+
+    # output/diagnose: ocean sets flux (=), land accumulates (+=)
+    diagn.physics.surface_humidity_flux[ij] = flux_ocean
```
maximilian-gelbrecht, line 89: What actually happens when we have a rock planet without ocean, it still works, right? Because all these variables are allocated anyway even without the ocean model?

src/dynamics/geopotential.jl
milankl: good call, should be our standard anyway to write `@kernel inbounds=true` and I guess if it doesn't blow up immediately, CI will tell us if we do access out of bounds!

src/dynamics/tendencies.jl
milankl: good catch removed

src/dynamics/implicit.jl
milankl: 1) the implicit initialization is happening in the first time steps because the first time steps use different dt (actually also something that's not needed with Lorenz N-cycle scheme proposed in #917 
2) this particular code is executed on every time step?

src/dynamics/time_integration.jl
milankl: yes because that's what we do in the other models too! just for consistency

src/physics/define_column.jl
milankl: Not yet there's some functionality like @nviebig's `OneBandRadiation` I don't want to translate here that have function signatures with dispatch over `ColumnVariables` so I wanted to leave this for now!

src/physics/land_sea_mask.jl
milankl: For `field[mask.data .== val]` field and mask need to same shape no? Because mask is always 2D but field may be 3D+?

src/physics/radiation/longwave_radiation.jl
milankl: introduced an `emissivity_atmosphere`, previously longwave radiation wasn't going down, only up, so extended the JeevanjeeRadiation in that way. It was essentially 0.6 up 0 down so the surface cooled somewhat, now we have 0.98 up and 0.98 down but maybe the latter is actually the problem we currently have, I need to read up on this but it seems 0.96 should be used for ocean, 0.9 maybe for land, and more like 0.7ish for the atmosphere ...

src/physics/radiation/longwave_radiation.jl
milankl: I'm doing this because 
```julia
τ⁻¹ = 1/longwave.time_scale.value
```
would be `::Float64` so converting the integer first to `Float32`. Alternatively we could do
```julia
τ⁻¹ = one(NF) / Second(longwave.time_scale).value
```
that's an implicit conversion?

src/physics/surface_fluxes/humidity.jl
milankl: Yes that should still work, it would run through the ocean heat flux computation but with land_fraction=1 the ocean flux would be zeroed out. But at the moment doing

```julia
surface_humidity_flux = SurfaceHumidityFlux(land=SurfaceLandHumidityFlux(spectral_grid), ocean=nothing)
model = PrimitveWetModel(spectral_grid; surface_humidity_flux)
```
would cause problems as the land would accumulate into
```julia
diagn.physics.surface_humidity_flux[ij] += flux_land
```
without this being reset ...

milankl: On it!

src/dynamics/implicit.jl
maximilian-gelbrecht: I forgot that I already made a PR about that, haha https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/891

src/physics/radiation/longwave_radiation.jl
maximilian-gelbrecht: Both is fine. This comment wasn't about this concrete line of code here, more generally I had some problems with convert in GPU kernels in past. In cases like this it seems to work though.

src/physics/surface_fluxes/humidity.jl
milankl: That's been resolved

milankl:
Near-surface air temperatures reaching absolute zero 🙈 

image

But I think it's the longwave scheme that does something funny because OLR goes up instead of down with surface temperatures

image

milankl:
Made some changes to the `JeevanjeeRadiation` longwave radiation scheme and now everything is stable and not to cold anymore. Woohoo. But not to self: We really need something better than that. But that's not for this PR, seasonal cycle of surface temperatures, note every frame is 0 UTC 


https://github.com/user-attachments/assets/a399737e-95d8-4ed6-9fb1-b894ab128a1d

milankl:
Now the `PrimitiveDryModel` is somewhat unstable as it blows up after 1 year due to some physics. Top with `physics=true` and bottom with `physics=false`. No parameterization should have such a strong influence on near-surface divergence ... Maybe it's the boundary layer drag being too strong over the ocean 

image

milankl:
Not sure it's cause or consequence but we have a lot of cyclones in the tropical oceans, shown here in the boundary layer drag (increases with surface wind speed)

https://github.com/user-attachments/assets/89f87fbe-ee84-4696-bafd-62ce91e0da89

milankl:
Okay yeah a drag that's too strong over the ocean caused this now the boundary lag drag looks more decent. I've introduced a surface roughness length for land (0.5m default) and ocean (1e-4m default) and weight it by the land-sea mask

```julia
z₀ = land_fraction*drag.roughness_length_land + (1 - land_fraction)*drag.roughness_length_ocean
```

now we have

https://github.com/user-attachments/assets/09bdd1ec-6c85-4ea8-b68c-b96fde26cc81

with the daily cycle of wind reversals in some areas (land-sea breeze due to diurnal heating) still nicely visible.

Took the 1e-4m for ocean surface roughness length $z_0$ from [Drennan, Taylor, Yelland, 2005](https://doi.org/10.1175/JPO2704.1) with significant wave height $H_s$. If that's 1m we would have $z_0 = O(1e-4)m$  which looks good enough for now

image

@gregrmunday relevant for your project ☝🏼

milankl:
And now we have temperatures reaching absolute zero again ... 

image

I guess in the `PrimitiveDryModel` with JeevanjeeRadiation that's not too unsurprising that this arises and causes problems. Maybe okay to ignore for now but how hard can it be to just implement a `OneBandLongwave` scheme ...

milankl:
Not so hard! Just did the Frierson scheme

image

https://github.com/ExeClim/Isca/blob/138982081e13157e40a246c9615559a055e142ef/src/atmos_param/two_stream_gray_rad/two_stream_gray_rad.F90#L690-L695

milankl:
Now everything looks stable, surface maybe a tad hot but also no clouds so not surprising

image

SSTs too hot but again no clouds in the tropics

image

too much convection but again no clouds ...

image

large-scale condensation reasonable

image

OLR

image

soil temperatures maybe not too bad given no clouds 😉 (12UTC in December)

image

soil moisture ... high

image

a lot of it available to evaporate

image

milankl:
Had to take the `Adapt.@adapt_structure` out of `@testset` because of https://github.com/JuliaGPU/Adapt.jl/issues/102#issue-3713039805

maximilian-gelbrecht:
Yeah, I also already encountered that issue with adapt. 

Do we actually now want to define a few default variables from the parametrizations? But that's also easily doable in a follow up PR.

milankl:
> Do we actually now want to define a few default variables from the parametrizations? But that's also easily doable in a follow up PR.

With the first version of this new variable system I'd really like to go (almost) all in and define in a follow-up PR

```julia
struct Variables{Pr, G, T, D, Ph, S}
    prognostic::Pr
    grid::G
    tendencies::T
    dynamics::D
    physics::Ph
    scratch::S
end
```
as a single struct that holds all variables with their fields then being managed the `variables(::ModelComponent)` definitions each with possible names spaces. 


Then we just pass `vars::Variables` around and access is as simple as `simulation.vars.tendencies.u` or similar. What do you think?

maximilian-gelbrecht:
Yes, that's good and also very similar to what we do in Terrarium. Compared to all of our GPU revisions, that's also a fairly minor and quick change honestly (adjusting e.g. the tracers to this system just took me maybe 30 minutes). It probably makes sense to think about this together with the timestepping as well though. I haven't looked in detail in it, but whatever we do with the variable system also must the needs of possible time stepping schemes and caches, as you also wrote on Zulip. 

(cc @karinakowalczyk)

milankl:
> Compared to all of our GPU revisions, that's also a fairly minor and quick change

Yes in theory quick but hundreds of function signatures will need to be adapted and work arounds be found for places where we do dispatch over `::PrognosticVariables` particularly as I reckon only `::Variables` would be a struct and everything in it a NamedTuple ... ? Plus it'll be a breaking change 😉

milankl:
EVERYTHING PASSSES 🎉 merge merge merge merge before it's too late, no just kidding, Max I let you review/approve!

milankl:
Regarding the geopotential: In the future I only want the geopotential in grid space. We need that anyway in the parameterizations and so while the dynamical core currently uses geopotential in spectral space because it's a tad faster I'd rather have this unified instead of computing geopotential twice, in spectral and grid. So the version you had to branch now ideally would go soon anyway

src/SpeedyWeather.jl
```diff
@@ -99,10 +100,10 @@ using SpeedyTransforms
 
 export SpeedyTransforms, SpectralTransform
 export transform, transform!
-export spectral_truncation, spectral_truncation!
```
maximilian-gelbrecht, line 102: I think you actually removed that `export` and then I added it again. Do you still want to remove it?

src/SpeedyWeather.jl
```diff
@@ -42,6 +44,7 @@ export DateTime, Millisecond, Second, Minute, Hour, Day, Week, Month, Year, Cent
 export initialize!, finalize!
 
 # import utilities
+export Utils
```
maximilian-gelbrecht, line 47: I don't actually know, when you export the whole module, do you automatically reexport all of its exports? Or why are we doing that here?

src/physics/land/soil_moisture.jl
```diff
@@ -26,14 +30,23 @@ export SeasonalSoilMoisture
     monthly_soil_moisture::GridVariable4D
 end
 
+# TODO to adapt create a ManualSeasonalSoilMoisture component like AlbedoClimatology is adapted to ManualAlbedo
+# Adapt.adapt_structure(to, soil::SeasonalSoilMoisture) = adapt(to, ManualSeasonalSoilMoisture(soil.monthly_soil_moisture))
```
maximilian-gelbrecht, line 34: Could also just do it as a NamedTuple: `adapt_structure(to, (; monthly_soil_moisture = adapt_structure(to, monthly_soil_moisture))`

src/physics/land/vegetation.jl
```diff
@@ -84,6 +90,10 @@ export VegetationClimatology
     low_cover::GridVariable2D
 end
 
+# TODO to adapt create a ManualVegetationClimatology component like AlbedoClimatology is adapted to ManualAlbedo
+# do all vegetations need a low_veg_factor?
+# Adapt.adapt_structure(to, veg::VegetationClimatology) = adapt(to, ManualVegetationClimatology(veg.high_cover, veg.low_cover))
```
maximilian-gelbrecht, line 95: same as above, but if there's another reason for a `ManualVegetationClimatology` then, it's also fine like that. It's just not necessary to define it only for the `adapt`

src/physics/ocean.jl
```diff
@@ -50,12 +50,18 @@ function Base.show(io::IO, O::AbstractOcean)
     print_fields(io, O, keys)
 end
 
+# variable that AbstractOcean requires
+variables(::AbstractOcean) =
```
maximilian-gelbrecht, line 54: Does this now also apply to `ocean = nothing` like this?

src/SpeedyWeather.jl
milankl: Yeah would like to remove it, I don't consider it to be public interface of SpeedyWeather, should be SpeedyTransforms functionality only

src/SpeedyWeather.jl
milankl: No it's just so that to access submodules within submodules one doesn't have to write `SpeedyWeather.SpeedyWeahterInternals.Utils ...`. `export Module` doesn't also exports its exports, only the module name, see here

```julia
julia> module A

       module B
           export f
           f(x) = x
       end

       export B
       using .B
       f(1)

       end
Main.A

julia> using .A

julia> B   # now B is defined because exported
Main.A.B

julia> B.f(1)   # functions within are easily accessible
1

julia> f(1)   # but not exported to main
ERROR: UndefVarError: `f` not defined in `Main`
Suggestion: check for spelling errors or missing imports.
Stacktrace:
 [1] top-level scope
   @ REPL[5]:1

julia> A.f(1)   # only exported within A
1
```

milankl: I think it makes sense to export all modules within SpeedyWeather?

src/physics/land/soil_moisture.jl
milankl: Absolutely, something for later!

src/physics/land/vegetation.jl
milankl: Yes for this actually I thought it would be good to have a ManualVegetationClimatology at some point but not for now, just wanted to put this in so it's clear that at the moment `model.land` isn't adapted

src/physics/ocean.jl
milankl: Even with `ocean = nothing` the `sea_surface_temperature` is being created but I actually don't know why

```julia
julia> model = PrimitiveWetModel(spectral_grid, ocean=nothing)
julia> progn = PrognosticVariables(model)
├ vor:   T31, 8-layer, 2-steps LowerTriangularArray{ComplexF32}
├ div:   T31, 8-layer, 2-steps LowerTriangularArray{ComplexF32}
├ temp:  T31, 8-layer, 2-steps LowerTriangularArray{ComplexF32}
├ humid: T31, 8-layer, 2-steps LowerTriangularArray{ComplexF32}
├ pres:  T31, 1-layer, 2-steps LowerTriangularArray{ComplexF32}
├ random_pattern: T31, 1-layer LowerTriangularArray{ComplexF32}
├┐ocean:
│├ sea_surface_temperature:  Field{Float32} on 48-ring OctahedralGaussianGrid
│├ sea_ice_concentration:    Field{Float32} on 48-ring OctahedralGaussianGrid
│├ sensible_heat_flux:       Field{Float32} on 48-ring OctahedralGaussianGrid
│└ surface_humidity_flux:    Field{Float32} on 48-ring OctahedralGaussianGrid
├┐land:
│├ soil_temperature:         Field{Float32} on 2-layer, 48-ring OctahedralGaussianGrid
│├ soil_moisture:            Field{Float32} on 2-layer, 48-ring OctahedralGaussianGrid
│├ snow_depth:               Field{Float32} on 48-ring OctahedralGaussianGrid
│├ sensible_heat_flux:       Field{Float32} on 48-ring OctahedralGaussianGrid
│└ surface_humidity_flux:    Field{Float32} on 48-ring OctahedralGaussianGrid
├ tracers: 0, ()
├ scale: 1.0
└ clock: 2000-01-01T00:00:00
```

I do want to revisit the variable system anyway and also properly check that we only allocate what's needed and non-existing variables don't cause errors in other components but for now I'd rather have more variables than needed and everything working. In the end we'd want `ocean = nothing` to be possible or maybe even a `ocean = CoupledOcean()` to allocate the respective fields but allow, e.g. Oceananigans to take control of them.

PR #878, opened 2025-10-16
GPU version of parametrizations over ij 

maximilian-gelbrecht:


src/physics/zenith.jl
```diff
@@ -225,13 +225,13 @@ function cos_zenith!(
...
-            cos_zenith, solar_hour_angle_0E, sinδ, cosδ, sinlat, coslat, lons)
+            cos_zenith, solar_hour_angle_0E, sinδ, cosδ, sinlat, coslat, lons, cos_zenith.grid.whichring)
 end
 
 # Kernel for solar zenith calculation with daily cycle
-@kernel inbounds=true function solar_zenith_kernel!(cos_zenith, @Const(solar_hour_angle_0E), @Const(sinδ), @Const(cosδ), @Const(sinlat), @Const(coslat), @Const(lons))
+@kernel inbounds=true function solar_zenith_kernel!(cos_zenith, @Const(solar_hour_angle_0E), @Const(sinδ), @Const(cosδ), @Const(sinlat), @Const(coslat), @Const(lons), @Const(whichring))
     ij = @index(Global, Linear)
-    j = whichring(ij, cos_zenith)
+    j = whichring[ij]
```
milankl, line 234: I've made some changes to RingGrids as `whichring(ij, grid)` was recomputing instead of using `grid.whichring` you'll see this when resolving conflicts. I thought we decided to have `grid.whichring` as array agnostic so why can't one do `whichring(ij, cos_zenith)` here now?

src/physics/albedo.jl
```diff
@@ -111,12 +115,15 @@ function initialize!(albedo::AlbedoClimatology, model::PrimitiveEquation)
...
 
-    a = albedo.file_Grid(ncfile[albedo.varname].var[:, :], input_as=Matrix)
+    a = on_architecture(model.architecture, albedo.file_Grid(ncfile[albedo.varname].var[:, :], input_as=Matrix))
     interpolate!(albedo.albedo, a)
 end
 
 parameterization!(ij, diagn, progn, albedo::AlbedoClimatology, model) = albedo!(ij, diagn.albedo, albedo.albedo)
 
+# For GPU usage just discard the extra information and treat it as a `ManualAlbedo`
+Adapt.adapt_structure(to, albedo::AlbedoClimatology) = adapt(to, ManualAlbedo(albedo.albedo))
```
milankl, line 135: I like!

src/physics/zenith.jl
maximilian-gelbrecht: It's because of the changes in #824.  I have to try out some stuff here, maybe I have to revert some of these anyway. Then I can explain in a bit more detail.

maximilian-gelbrecht:
@milankl  

One annoying problem came up, on which I'd like to ask you for your opinion. 

In very short: Can we get rid of the `NF, ArrayType` type parameters in our data structures? 

Why?

So, previously we only handed over `Fields` and `LowerTriangularArrays` to kernels, that's why our `PrognosticVariables`, `DiagnosticVariables` and their subtypes didn't have any `adapt_structure` defined for them. It wasn't needed. Now, with the parametrizations this seems unavoidable. 

* Unpacking everything before is just not really an option. Right? 

So, for GPU adaption, we need to define the `adapt_structure` that defines how the data structure is adapted to kernel usage. The problem lies in our constructers for these types. Per default `@adapt_structure` would just recursively adapt everything ([see the code](https://github.com/JuliaGPU/Adapt.jl/blob/master/src/macro.jl)) and call the type with all arguments converted. This isn't working here because such constructers don't exist for our types as not all type parameters can be determined automatically from the input arguments only. We typically have `NF, ArrayType` as initial type parameters that can't automatically be determined from the input arguments. 

* Solution 1) 

Manually write those adapt. Even with LLM support that's kind of annoying and quite error prone in the long run. The function would kind of look like this: 

``` 
function Adapt.adapt_structure(to, tendency::Tendencies)
    adapted_fields = [Adapt.adapt_structure(to, tendency.$name) for name in fieldnames(Tendencies)]
    return Tendencies{eltype(tendency), typeof(adapted_fields[1].data), typeof(adapted_fields[1]), 
    typeof(adapted_fields[2]), typeof(adapted_fields[10]), typeof(adapted_fields[9]), 
    typeof(adapted_fields[12]), typeof(adapted_fields[end-1])}(adapted_fields...)
end
```

* Solution 2) 

Follow Solution 1) but turn it into a macro that does it automatically and dynamically. 

* Solution 3) 

Get rid of the `NF,ArrayType` parameters in all our data structs. It's redundant information anyway. My impression is that we don't use it much anyway, or am I missing something? If we get rid of those parameters, all the default constructors and adapts should work. 

I am in favour of Solution 3).

maximilian-gelbrecht:
Okay, I did that change to drop `NF` and `ArrayType` already now,  it needed much less changes and revisions than I thought before.

maximilian-gelbrecht:
@milankl Okay, after a bit of unnecessary back and forward, I came to the realisation that `Dictionary` just won't work on GPU because their mutable. I should have thought about that from the start. That's a bit of a problem for us and might have to rethink some things. Of course we could just replace them with `NamedTuples` but then we loose e.g. the ability to add and remove tracers. Let's have a short call about this or a chat the next days. 

I'll also have a look at the performance problem in the other PR

milankl:
> I came to the realisation that `Dictionary` just won't work on GPU because their mutable.

If that's a requirement then that's fine. It's cool in showcases to add a tracer on the fly but honestly in reality not having a dictionary isn't a massive problem. NamedTuples can be used instead and in this case we can also think about activating or deactivating tracers. While mutability isn't supported as you say one can still conditionally construct NamedTuples before they are passed to the kernel. Just to check does this also mean that `mutable struct` for our parameterizations isn't going to fly and they have to be immutable and so we have to think in terms of adapting a parameterizations via reconstructing the model?

maximilian-gelbrecht:
Yes mutable structs in generally don't fly e.g. a simple MWE: 

```Julia 


using SpeedyWeather, Adapt, CUDA, KernelAbstractions
mutable struct Fooo{A}
    a::A
    b::Int
end

Adapt.@adapt_structure Fooo

foo = Fooo(cu(rand(10)), 1)

@kernel function bar(foo::Fooo)
    @inbounds foo.a .= 1.0
end

SpeedyWeather.launch!(arch, SpeedyWeather.LinearWorkOrder, (10,), bar, foo)
```
This doesn't work
```
ERROR: GPU compilation of MethodInstance for gpu_bar(::KernelAbstractions.CompilerMetadata{…}, ::Fooo{…}) failed
KernelError: passing non-bitstype argument

Argument 3 to your kernel function is of type Fooo{CuDeviceVector{Float32, 1}}, which is not a bitstype:


Only bitstypes, which are "plain data" types that are immutable
and contain no references to other values, can be used in GPU kernels.
For more information, see the `Base.isbitstype` function.
```


> While mutability isn't supported as you say one can still conditionally construct NamedTuples before they are passed to the kernel. Just to check does this also mean that `mutable struct` for our parameterizations isn't going to fly and they have to be immutable and so we have to think in terms of adapting a parameterizations via reconstructing the model?

I am exactly sure what you mean with this. So reconstructing instead of mutating is one option when we would actually like to mutate. The other option is just to hand over more low-level options to the kernels and/or unpack structs. I think with the variables we really don't want to start to unpack the `PrognosticVariables` and `DiagnosticVariables`, it just gets messy. But with the model we already doing that anyway to some degree. So similarly to how we have a custom `adapt_structure` for `ClimatologyAlbedo` now, we could also think about similar stuff in case we really need the mutability for some processes.

maximilian-gelbrecht:
@milankl I am getting some really weird "Saved new PDF of paper" commits here suddenly, without me doing something. I have no idea where this is coming from except that it's somehow related to the JOSS paper. Do you have any idea what's going on here?

maximilian-gelbrecht:
This is working now: 

```Julia 
import Pkg
Pkg.activate("test")

using SpeedyWeather, CUDA

arch = GPU()
spectral_grid = SpectralGrid(trunc=31, architecture=arch)
```

milankl:
Awesome, shall we merge this and then I can continue working on #873 ? You could branch off again if you wanted to work independently of me

milankl:
I'm worried about the changes to `docs/joss/paper.pdf` I kind of don't want to have them in the git history ... Maybe we can replace the file with the old? EDIT: That seems to have worked!

maximilian-gelbrecht:
## Some little GPU-ify notes

* Error messages are more opaque when something goes wrong -> Claude does a pretty good job at explaining those errors 
* Functors don't work on GPU unless they are inlined (and therefore effectively not functors anymore) 
* Generally, when in doubt, better `@inline` a function
* No Colons `:` 
* We highjack `adapt` to just be responsible for transfer to kernels, for some structs like `SpectralGrid` it explictily adapts to `nothing` -> If you want to transfer objects between devices, use `on_architecture` 
* We currently adapt our `Field` and `LowerTriangularray` to just adapt the `.data` field only. If you need the `Spectrum` or `Grid` it should be in the `parameters`/`model` 
* Structs should ideally don't have extra type parameters. When we have types that can be directly reconstructed just from the arguments alone, it makes our live for kernel adaption a lot easier 
* No mutable structs allowed -> if you really need it think about workarounds with just handing over the important fields
* Iterations over `NamedTuples` don't work, here we e.g. manual unroll a loop with a `@generated` function  
* Think about reducing memory bandwidth by reducing the amount of diagnostic variables we really need 
* Think about performance when branch/code divergence needs to be done

PR #880, opened 2025-10-21
Move SpeedyParameters to SpeedyWeatherInternals

### LINKED ISSUE #879: Move SpeedyParameters to SpeedyWeatherInternals
This is a stopgap solution to make the Speedy parameter handling available for external use (namely in Terrarium) while avoiding the creation of another registered package (for now).

bgroenks96:
Also fixes the precompile issues with the `@parameterize` macro.

Resolves #879

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+## v0.17.3
+
 - Move SpeedyParameters to SpeedyWeatherInternals [#880](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/880)
 - Remove JOSS paper draft workflow [#874](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/874)
 - Change particle tracking to every 4 hours [#881](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/881)


Project.toml
```diff
@@ -91,3 +79,15 @@ Statistics = "1.10"
 TOML = "1"
 UnicodePlots = "3.3"
 julia = "1.10"
+
```
bgroenks96, line 82: I'm not sure why this is here? I didn't make this change...

maximilian-gelbrecht: I'd assume it's the automatic formatting that also changed the `[sources]`. 

I am a bit confused, but wouldn't the deps need to become `[weakdeps]` and their `[compats]` kept?

bgroenks96:
@maximilian-gelbrecht @milankl Are we still keeping all of the tests in SpeedyWeather?

Project.toml
bgroenks96: Why? They are just dependencies of a dependency? It's not an extension module.

maximilian-gelbrecht: Ah yes, sorry, I just got confused.

maximilian-gelbrecht:
> @maximilian-gelbrecht @milankl Are we still keeping all of the tests in SpeedyWeather?

For now yes, but it's more due to laziness than due to design.

bgroenks96:
As good a reason as any!

maximilian-gelbrecht:
Could you check if the broken tests for the parameter systems are actually still broken? I think I marked those as broken, when refactoring towards standalone packages.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+## v0.17.3
+- Move SpeedyParameters to SpeedyWeatherInternals
+
```
milankl: You want #882 to be merged first? If then I'd still put this under unreleased and then we can have a separate PR just to bump to v0.17.3 where then introduce `## v0.17.3`

bgroenks96:
> Could you check if the broken tests for the parameter systems are actually still broken? I think I marked those as broken, when refactoring towards standalone packages.

They were indeed broken, fixed now.

> Given you want to use this outside of SpeedyWeather, I'd be in favour of finding a more general name for the package, e.g. as you already call it: ParameterHandling.jl ?

ParameterHandling.jl already exists. But anyway, the point of moving it to SpeedyWeatherInternals is to avoid having to name, create, and maintain a separate package.

CHANGELOG.md
bgroenks96: That would cause precompile errors in the current release of SpeedyWeather.

bgroenks96: I guess we could wait to register the new version of `SpeedyWeatherIntenals` though.

maximilian-gelbrecht: #882 is merged now, so I think you can also just go ahead an bump the version of both after this is merged.

milankl: Given @bgroenks96 already changed the SpeedyWeather version in Project.toml we apprently do this in one and not a separate PR just to bump the version. Happy with this

PR #882, opened 2025-10-21
SpectralGrid constructor revised, e.g. from grid

maximilian-gelbrecht:
@bgroenks96 wanted a constructor for `SpectralGrid` from a `grid` instance for coupling to Terrarium. So why not? 

This also does a small revision to the constructor. 

I removed the `ArrayType` kwarg because it can't be used anymore anyway. At several parts of the model we infer the type directly from the architecture. If you want a different array type you can still do that by sub-typing an existing architecture and defining `array_type(::Architecture)` for it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ## v0.17.2
 
+- SpectralGrid can now be initialized from a grid instance as well [#882](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/882)
 - Expand scratch memory handling in spectral transforms for better differentiability with Enzyme and revise FFT plans [#791](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/791)
 - Increase version of RingGrids, LowerTriangularArrays and SpeedyTransforms to v0.1.1 [#875](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/875)
 - Pluto notebook example in README [#872](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/872/)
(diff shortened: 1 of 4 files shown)


milankl (approved):
Looks good to me! 🎉

milankl:
Hu? The garbage collector complains and there's a Vararg issue?

```julia
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
You may need to write `f(x::Vararg{T})` rather than `f(x::Vararg{<:T})` or `f(x::Vararg{T}) where T` instead of `f(x::Vararg{T} where T)`.
GC error (probable corruption)
Allocations: 420398476 (Pool: 420135881; Big: 262595); GC: 717
```

maximilian-gelbrecht:
I don't know I think it's one of those Enyzme errors that just unfortunately happens occasionally still. I just rerun the job now, and then we'll see. Most likely it'll pass

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ## v0.17.2
 
+- SpectralGrid can now be initialized from a grid instance as well [#882](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/882)
```
milankl, line 7: Just realised that the changelog entry isn't under `## Unreleased` ...

milankl (commented):
fixed in #881 and [a9160ff](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/881/commits/a9160ff12763dba2e7529b3722e4ac9daaace1f2)

PR #884, opened 2025-10-24
Dynamical core kernels

milankl:
Claude writes the kernels in our dynamical core!

And is being very rude

```
Let me try a Python script to handle this properly
```

Note: I had to rename some unicode characters manually it complained about otherwise I didn't do anything, all Claude, let's see what happens!

### DIFF
--- a/src/dynamics/tendencies.jl
+++ b/src/dynamics/tendencies.jl
@@ -129,12 +129,19 @@ function pressure_gradient_flux!(
     (; uv∇lnp ) = diagn.dynamics
 
     # PRESSURE GRADIENT FLUX
-    @inbounds for k in eachlayer(u_grid, v_grid, uv∇lnp)
-        for ij in eachgridpoint(u_grid, v_grid, uv∇lnp)
-            # the (u, v)⋅∇lnp_s term
-            uv∇lnp[ij, k] = u_grid[ij, k]*∇lnp_x[ij] + v_grid[ij, k]*∇lnp_y[ij]
-        end
-    end
+    # Launch the kernel for pressure gradient flux calculation
+    launch!(architecture(uv∇lnp), RingGridWorkOrder, size(uv∇lnp), _pressure_gradient_flux_kernel!,
+            uv∇lnp, u_grid, v_grid, ∇lnp_x, ∇lnp_y)
+end
+
+@kernel inbounds=true function _pressure_gradient_flux_kernel!(
+    uv∇lnp, u_grid, v_grid, @Const(∇lnp_x), @Const(∇lnp_y)
+)
+    # Get indices
+    ij, k = @index(Global, NTuple)
+    
+    # Calculate the (u, v)⋅∇lnp_s term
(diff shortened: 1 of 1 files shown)


src/dynamics/tendencies.jl
```diff
@@ -129,12 +129,19 @@ function pressure_gradient_flux!(
...
+    # Launch the kernel for pressure gradient flux calculation
+    launch!(architecture(uv∇lnp), RingGridWorkOrder, size(uv∇lnp), _pressure_gradient_flux_kernel!,
+            uv∇lnp, u_grid, v_grid, ∇lnp_x, ∇lnp_y)
+end
+
+@kernel inbounds=true function _pressure_gradient_flux_kernel!(
+    uv∇lnp, u_grid, v_grid, @Const(∇lnp_x), @Const(∇lnp_y)
+)
+    # Get indices
+    ij, k = @index(Global, NTuple)
```
maximilian-gelbrecht, line 141: pressure gradient flux can also be 2D, no? then this won't work

src/dynamics/tendencies.jl
```diff
@@ -146,13 +153,23 @@ function temperature_anomaly!(
...
-            temp_grid[ij, k]      -= Tₖ  # absolute temperature -> anomaly
-            temp_virt_grid[ij, k] -= Tₖ  # virtual temperature -> anomaly
-        end
-    end
+    # Launch the kernel for temperature anomaly calculation
+    launch!(architecture(temp_grid), RingGridWorkOrder, size(temp_grid), _temperature_anomaly_kernel!,
+            temp_grid, temp_virt_grid, temp_profile)
+end
+
+@kernel inbounds=true function _temperature_anomaly_kernel!(
```
maximilian-gelbrecht, line 161: Can't we do the whole thing also quite easily with regular broadcasting?

src/dynamics/tendencies.jl
```diff
@@ -192,24 +209,46 @@ function vertical_integration!(
...
+    uv∇lnp_sum_above[ij, k] = u_mean_grid[ij] * ∇lnp_x[ij] + v_mean_grid[ij] * ∇lnp_y[ij]
+    
+    # Add k-th element to the vertical sum
+    u_mean_grid[ij] += u_grid[ij, k] * Δσₖ
+    v_mean_grid[ij] += v_grid[ij, k] * Δσₖ
+    div_mean_grid[ij] += div_grid[ij, k] * Δσₖ
+end
+
+@kernel inbounds=true function _vertical_integration_spectral_kernel!(
+    div_mean, div, @Const(Δσₖ), @Const(k)
```
maximilian-gelbrecht, line 244: probably better to do this column based and iterate over the vertical layers in the kernel I'd say (as we d in the column based parameterisations.

maximilian-gelbrecht (commented):
Just some initial comments to the kernels. 

As you already made friends with Claude, I found it quite helpful for writing some quick benchmark scripts as well. I never committed them in the end though.

maximilian-gelbrecht:
What's actually our plan here? Do we first want to have a version working on GPU with the old tendency logic and then reformulate it later? 

Anyway I am also happy to chime in a bit here in this PR.

maximilian-gelbrecht:
Superseded by #934

Issue #847, opened 2025-09-11
Use Dict or NamedTuple to hold variables


milankl:
Compared to `struct` dictionaries come with the advantage that fields can be added or removed at any time. Considerations

- `Dict`s are slower than `struct` but maybe not important if we typically have O(10) fields in them and just unpack them anyway before passing to a kernel
- get/set access `struct.fieldname` is what we currently have but dictionaries use `dict[key]` but [PropertyDicts.jl](https://github.com/oschulz/propdicts.jl) woud allow to use both where `dict.key` would be much less intrusive to what we currently have
- we could implement this initially only for parameterizations/land/ocean components and see how it goes
- components like `convection` could then add a `rain_convection` field such that it doesn't exist when not needed
- this removes the necessity to add a field to the diagnostic variables when implementing a new physics parameterization

Just some ideas for now

milankl:
I think the idea to have all variables as `NamedTuple` could fly. In the end we don't dispatch over the variable type much and where we do I'm sure we can find another way. Using a NamedTuple for the model or model components doesn't make sense as we wouldn't be able to dispatch over them and a lot of decisions are based on those

if we start with
```julia
progn = (vor=zeros(...),)
```
then the question is how we could in `initialize!(model)` decide how to merge this `NamedTuple` with other variables so that it can be constructed during initialisation? I guess one way to do this would be to let model components implement a `variables` function that returns variables needed as named tuple, e.g.
```julia
function variables(large_scale_condensation, spectral_grid)
    return (rain_large_scale = zeros(GridVariable2D, grid),)
end
```
which could then all be merged. Fallback would be
```julia
variables(::Any, ::SpectralGrid) = NamedTuple()
```
such that variables that aren't needed aren't allocated. We then need to modify the "one" variable tuple by `merge` the named tuples somehow, maybe like
```julia
function variables(model)
     diagn = NamedTuple()
     for component in model
         vars = variables(component, model.spectral_grid)
         diagn = merge(diagn, vars)
     end
     return diagn
end
```

milankl:
Using `NamedTuple` for variables comes with the disadvantage that one can't dispatch on it however one could wrap a `NamedTuple`

```julia
julia> struct Variables{T}
           named_tuple::T
       end

julia> Base.getproperty(V::Variables, s::Symbol) = getproperty(getfield(V, :named_tuple), s)

julia> V = Variables((a=1, b=2))
[...]

PR #885, opened 2025-10-27
Revise Variables towards using NamedTuples and Modularity 

maximilian-gelbrecht:
This addresses #847 

Inspired by @bgroenks96 implementation for Terrarium. 

Currently this is set up as a drop-in replacement. The hierarchy of the variables and how they are accessed is kept exactly the same. We could think about changing some things later though. 

The basic idea is that the `PhysicsVariables` and prognostic and diagnostic land and ocean variables are replaced by NamedTuples. Each of these parameterziation then defines a `variables` function to add to those during initialisation like e.g. so: 
```Julia 
function variables(::SlabOcean)
    return (DiagnosticVariable(name=:surface_shortwave_down, dims=Grid2D()),
    DiagnosticVariable(name=:surface_shortwave_up, dims=Grid2D(), namespace=:ocean),
    DiagnosticVariable(name=:surface_longwave_down, dims=Grid2D()),
    DiagnosticVariable(name=:surface_longwave_up, dims=Grid2D(), namespace=:ocean),
    DiagnosticVariable(name=:surface_humidity_flux, dims=Grid2D(), namespace=:ocean),
    DiagnosticVariable(name=:surface_sensible_heat_flux, dims=Grid2D(), namespace=:ocean),
    PrognosticVariable(name=:sea_ice_concentration, dims=Grid2D(), namespace=:ocean),
    PrognosticVariable(name=:sea_surface_temperature, dims=Grid2D(), namespace=:ocean),
    ) 
end
```

Compared to Brian's Terrarium  implementation I simplified/hard-coded the namespaces, because I don't think we need that much flexibility and also the actual array initialisation is of course different because we don't use Oceananigans fields. Also there are no closures. 

This is a draft but the minimal example is working 

```Julia 
import Pkg 
Pkg.activate(".")

using SpeedyWeather

arch = CPU()
spectral_grid = SpectralGrid(trunc=31, architecture=arch)
simulation =  initialize!(model)
```

This still needs 
- [x] some discussion what exactly we want 
- [x] adjusting ShallowWater and Barotropic 
- [x] documentation 
- [x] more testing 
- [ ] connect to the output system? 
- [ ] a lot of miscellaneous small stuff (I marked a lot of TODOs in the code)


bgroenks96:
I'm not sure we even need the namespaces in Terrarium.... it remains to be seen. They're mostly intended to help support tiling later on.

In general, I like this, but one thing you might also want to consider is whether or not the same variable will show up in multiple places, as it currently does in Terrarium, which would imply the need for a central registry (as disucssed [here](https://github.com/TUM-PIK-ESM/Terrarium.jl/issues/35))

maximilian-gelbrecht:
> In general, I like this, but one thing you might also want to consider is whether or not the same variable will show up in multiple places, as it currently does in Terrarium, which would imply the need for a central registry (as disucssed [here](https://github.com/TUM-PIK-ESM/Terrarium.jl/issues/35))

At the moment we even explicitly want this, e.g. we have `diagn.ocean.albedo` and `ding.land.albedo`. We could of course easily rename those to `albedo_land`. I don't think it's quite as much as a problem here as it is for Terrarium which is just a bit more modular then what we have here. That being said, it probably makes sense to issue at least a warning when a duplicate is detected in the initialisation.

milankl:
Some comments before doing an actual review looking through all of the code

- I like the idea of `variables(::ModelComponent)` returning a `PrognosticVariable`, `DiagnosticVariable` which is thought of as an _instruction_ describing what actually has to be done when allocating the variables. I initially though of passing the `SpectralGrid` also to `variables` so the required variable can directly be allocated within but I prefer your approach that the actual allocation then happens in `initialize_variables`
- Names spaces and paths: IIUC then you think of name spaces as being independent of the actual `simulation.a.b.c.d.var_name` path and more like a tag? We currently have our variables nested over several levels, e.g. `diagnostic_variables.physics.ocean.variable_name` but I am wondering whether that's needed. I could see a setup where we only allow for `variables.path.variable_name` with `path` being a single level and I'm probably suggesting

variables
- prognostic
- grid
- tendencies
- dynamics
- physics/parameterizations
- scratch

but having several levels might be nice to actually then distinguish between ocean/land in various places

variables
- prognostic
1. vorticity
2. ocean
a. sea_surface_temperature
3. land
a. soil_temperature
- diagnostic ...

So what about adding `path::String` to `DiagnosticVariable` (which eventually could be just `Variable`) such that instead of
```julia
(
DiagnosticVariable(name=:albedo, dims=Grid2D(), desc="Albedo", units="1"),
DiagnosticVariable(name=:albedo, dims=Grid2D(), desc="Albedo over the ocean", units="1", namespace=:ocean),
DiagnosticVariable(name=:albedo, dims=Grid2D(), desc="Albedo over the land", units="1", namespace=:land)
)
```
(would albedo 2&3 even exist or be flagged as duplicate?) one would do
```julia
(
DiagnosticVariable(name=:albedo, path="diagnostic.physics", ...),
DiagnosticVariable(name=:albedo, path="diagnostic.physics.ocean", ..., namespace=:ocean),
DiagnosticVariable(name=:albedo, path="diagnostic.physics.land", ..., namespace=:land)
)
```

maximilian-gelbrecht:
I do see the namespace here as a path. It's just limited to one level or a hard-coded path - symbol mapping, as dealing with deeper paths and nested tuples also becomes a bit annoying (albeit it's possible). 

What I currently do is that I initialise based on the namespace in the `DiagnosticVariables` and `PrognosticVariables` constructors. 

I also wouldn't really overthink these paths or namespaces here and just mostly hard-code them because we want a modular atmospheric model, but it'll still always be an atmospheric model with fluid dynamics and parametrizations on top - we need some flexibility, but also not complete flexibility.

maximilian-gelbrecht:
## Some benchmarks 

Related to some comments from @milankl  in #873 

I just benchmarked this PR against `main` on CPU

This PR 
```Julia 
spectral_grid = SpectralGrid(trunc=31)
  907.500 μs (51 allocations: 29.08 KiB)
```


vs 

`main` 
```Julia
spectral_grid = SpectralGrid(trunc=31)
julia> @btime  SpeedyWeather.parameterization_tendencies!($diagn, $progn, $progn.clock.time, $model)
  1.573 ms (2 allocations: 96 bytes)
```

So this looks quite good! Not sure how exactly I improved the performance compared to your PR. We could still track down the few more allocations it makes though. 

The `mg/ijphysics-gpu` branch has almost identical performance to this branch btw. 

### On GPU (Nvidia A40) 

```Julia 
using SpeedyWeather, BenchmarkTools, CUDA

for ntrunc in [31, 63, 127, 255, 511]
    arch = GPU()
    spectral_grid = SpectralGrid(trunc=ntrunc, architecture=arch)
    @btime CUDA.@sync SpeedyWeather.parameterization_tendencies!($diagn, $progn, $model)
end 
```

| trunc | time       |
|-------|------------|
| 31    | 140.601 μs |
| 63    | 138.658 μs |
| 127   | 170.305 μs |
| 255   | 334.762 μs |
| 511   | 936.610 μs |

So, that's also already pretty good news. Even in low resolutions it's faster on GPU, and the T511 on GPU is about the same as the T31 on my M3 ! I still have to do a proper profiling though. 
```
Julia> CUDA.@profile trace=true SpeedyWeather.parameterization_tendencies!(diagn, progn, model)
Profiler ran for 844.96 µs, capturing 1190 events.

Host-side activity: calling CUDA APIs took 188.11 µs (22.26% of the trace)
┌──────┬───────────┬──────────┬────────────────┐
│   ID │     Start │     Time │ Name           │
├──────┼───────────┼──────────┼────────────────┤
│   45 │ 185.49 µs │ 79.39 µs │ cuLaunchKernel │
│ 1186 │ 746.73 µs │ 38.86 µs │ cuLaunchKernel │
└──────┴───────────┴──────────┴────────────────┘

Device-side activity: GPU was busy for 59.37 µs (7.03% of the trace)
┌──────┬───────────┬──────────┬─────────┬────────┬──────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────
│   ID │     Start │     Time │ Threads │ Blocks │ Regs │ Name                                                                                                           ⋯
├──────┼───────────┼──────────┼─────────┼────────┼──────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────
│   45 │ 257.97 µs │  2.38 µs │     256 │    158 │   25 │ gpu_solar_zenith_kernel_(CompilerMetadata<StaticSize<_40320__>, DynamicCheck, void, void, NDRange<1, StaticSiz ⋯
│ 1186 │ 784.87 µs │ 56.98 µs │     256 │    158 │   80 │ gpu_parameterization_tendencies_kernel_(CompilerMetadata<StaticSize<_40320__>, DynamicCheck, void, void, NDRan ⋯
└──────┴───────────┴──────────┴─────────┴────────┴──────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────
                       
```        
So there's still a lot of room for improvement as it seems despite good initial performance. Launching kernels with `diagn` and `progn` definitely has a decent amount of overhead. For higher resolutions this will matter less, but outside of the one big kernel for the parametrizations we should really avoid it if we can. It absolutely isn't worth it for some small operation.

milankl:
Awesome!! Performance here also depends on the atmospheric state as the while loops in convection are conditional on vertical profiles. But as long as it's the same ballpark then this is good! Kernel abstractions always creates some allocations that one doesn't need to worry about right?

maximilian-gelbrecht:
Yes, I agree, and yes as far as I know KA does create a few. And 51 allocations isn't really that much, and they stay the same independently of resolution. Even on GPU T511,  I only got `936.610 μs (1662 allocations: 92.84 KiB)`. See the post above, I edited in some initial GPU benchmarks. They look great as well. 

So how do we proceed best? This PR is mostly about the variables, then we have #878 and #873 as well. I'd welcome a review here and in #878

src/physics/ocean.jl
```diff
@@ -236,6 +239,10 @@ function timestep!(
     return nothing
 end
 
+#TODO: it's not actually prognostic, but used as such so far? 
+variables(::ConstantOceanClimatology) = (PrognosticVariable(name=:sea_surface_temperature, dims=Grid2D(), namespace=:ocean),)
```
milankl, line 243: Yes but it still requires it and actually writes into it so I think this is all good!

src/physics/ocean.jl
```diff
@@ -288,6 +295,9 @@ function timestep!(
     return nothing
 end
 
+#TODO: it's not actually prognostic, but used as such so far? 
+variables(::AquaPlanet) = (PrognosticVariable(name=:sea_surface_temperature, dims=Grid2D(), namespace=:ocean),)
```
milankl, line 299: Same here

src/variables/prognostic_variables.jl
```diff
@@ -117,29 +59,30 @@ export PrognosticVariables
...
+    scale::RefValueNF = Ref(one(typeof(real(vor[1]))))
 
     "Clock that keeps track of time, number of timesteps to integrate for."
     clock::ClockType = Clock()
 end
 
 Adapt.@adapt_structure PrognosticVariables
 
-Base.eltype(::PrognosticVariables{SpectrumType, GridType, SpectralVariable2D, SpectralVariable3D, SpectralVariable4D, GridVariable2D}) where {SpectrumType, GridType, SpectralVariable2D, SpectralVariable3D, SpectralVariable4D, GridVariable2D <: AbstractField{NF}} where NF = NF 
+Base.eltype(progn::PrognosticVariables) = eltype(real(progn.vor[1]))
```
milankl: `real(eltype(progn.vor))` to make it GPU compatible?

src/variables/variables.jl
```diff
@@ -0,0 +1,202 @@
...
+    
+    # Convert to NamedTuple with tuples instead of vectors
+    return NamedTuple{Tuple(namespaces)}(Tuple(Tuple(namespace_dict[ns]) for ns in namespaces))
+end
+
+get_prognostic_variables(model::AbstractModel) = get_prognostic_variables(variables(model)...)
+get_diagnostic_variables(model::AbstractModel) = get_diagnostic_variables(variables(model)...)
+
+# variables of nothing are nothing
+variables(::Nothing) = ()
```
milankl: empty tuple? 
```julia
julia> typeof(())
Tuple{}
```

src/variables/variables.jl
```diff
@@ -0,0 +1,202 @@
...
+
+# TODO: not quite sure yet about the nlayers or where it'll go
+# initialize a NamedTuple from variables 
+function initialize_variables(SG::SpectralGrid, nlayers::Integer, variables...) 
+    (; grid, spectrum, SpectralVariable2D, SpectralVariable3D, GridVariable2D, GridVariable3D) = SG
+
+    zero_init(var::AbstractVariable{Grid2D}) = zeros(GridVariable2D, grid)
+    zero_init(var::AbstractVariable{Grid3D}) = zeros(GridVariable3D, grid, nlayers)
+    zero_init(var::AbstractVariable{Spectral2D}) = zeros(SpectralVariable2D, spectrum)
+    zero_init(var::AbstractVariable{Spectral3D}) = zeros(SpectralVariable3D, spectrum, nlayers)
```
milankl: why not just 
```julia
zero(::AbstractVariable{Grid2D}, SG::SpectralGrid) = zeros(SG.GridVariable2D, SG.grid)
zero(::AbstractVariable{Grid3D}, SG::SpectralGrid, nlayers) = zeros(SG.GridVariable3D, SG.grid, nlayers)
```
defined in global scope?

src/variables/variables.jl
```diff
@@ -0,0 +1,202 @@
...
+Indicator type for spectral 2D variables (l, m).
+"""
+struct Spectral2D <: AbstractVariableDims end
+
+"""
+    $TYPEDEF
+
+Indicator type for spectral 3D variables (l, m, k).
+"""
+struct Spectral3D <: AbstractVariableDims end
```
milankl, line 29: We will also need `Spectral4D` for the atmospheric prognostic variables no?

milankl: But can be added later!

milankl (commented):
I think this is great and I would merge this into the ijphysics branches! 

While there is scope to extend this and modify this to actually cover all allocated variables (likely with nesting) this is a great start and as long as it is non-breaking I'd be happy to merge this even in to the main branch and then we can properly overhaul this and introduce a breaking change revisiting the nesting of the variables? I think at some point I'd like to merge the `prognostic_variables`/`diagnostic_variables` variables to just `variables` with the following structure

simulation.variables
- prognostic
- grid
- tendencies
- dynamics
- parameterizations
- ocean
- land
- scratch

and prognostic variables 

simulation.prognostic
- vorticity
- divergence
- temperature
- pressure
- ocean
- land
- particles
- tracers
- clock

where `ocean` and `land` would be neat to have as nested given the duplication of surface fluxes that otherwise would need `_ocean`, `_land` .... But still, overall this would flatten the structure a little bit and make it easier for people to find things I believe. This would be a very breaking change but I think that'll be good nevertheless! Given we will never have that many prognostic variables

src/variables/prognostic_variables.jl
maximilian-gelbrecht: Good call. Unfortunately `real(eltype(progn.vor))` gives `ComplexF32` though, and it seems you can't do a double `eltype(ComplexF32)` to actually get the precision... 

I'll probably revert back to the very long parametric version...

src/variables/variables.jl
maximilian-gelbrecht: Yes! the variables functions are expected to give tuples. Or technically, they are expected to return a something iterable.

src/variables/variables.jl
maximilian-gelbrecht: Yes, good idea.

maximilian-gelbrecht:
ok, then I'll finish up some documentation and testing and then merge it into the feature branch.

bgroenks96:
> At the moment we even explicitly want this, e.g. we have diagn.ocean.albedo and ding.land.albedo. We could of course easily rename those to albedo_land. I don't think it's quite as much as a problem here as it is for Terrarium which is just a bit more modular then what we have here. That being said, it probably makes sense to issue at least a warning when a duplicate is detected in the initialisation.

That's not quite what I mean. The central registry is not intended to map to a single variable (or path, as you put it) but rather to standardize names, improve discoverability, and avoid things like

```julia
variables(::ComponentA) = (
    DiagnosticVariable(:albedo, namespace=:ocean)
)

variables(::ComponentB) = (
    DiagnosticVariable(:Albedo, namespace=:land)
)

variables(::ComponentC) = (
    DiagnosticVariable(:land_albedo, namespace=:land)
)
```
In Terrarium, we also allow these variables to have other metadata like units and descriptions which we are currently duplicating everywhere which is also a problem...

Of course you can get away without it by just keeping all of the naming conventions in your head, but that makes it harder for new developers to start developing new components. It's really just a user and developer experience feature.

PR #887, opened 2025-10-27
Ring, nested and matrix order for OctaHEALPixGrid

milankl:
This implements functionality to reorder data on an OctaHEALPixGrid from ring order (our default) to nested order (for trivial interpolation)

image

created with

<details>

```julia
using SpeedyWeather, GLMakie, GeoMakie

nlat_half = 8
grid = OctaHEALPixGrid(nlat_half)
N = RingGrids.get_npoints(grid)

field = Field(collect(1:N), grid)    # assumed nested
field_nested = ring_order(field)     # convert to ring order

fig = Figure(size=(800, 800))
ax1 = GeoAxis(fig[1, 1], dest = "+proj=ortho +lon_0=30 +lat_0=45", title="OctaHEALPix, ring order")
ax2 = GeoAxis(fig[1, 2], dest = "+proj=ortho +lon_0=30 +lat_0=45", title="OctaHEALPix, nested order")
meshimage!(ax1, -180..180, -90..90, rotr90(GeoMakie.earth()); npoints = 100)
meshimage!(ax2, -180..180, -90..90, rotr90(GeoMakie.earth()); npoints = 100)
lines!(ax1, GeoMakie.coastlines(), color=:black, linewidth=0.3)
lines!(ax2, GeoMakie.coastlines(), color=:black, linewidth=0.3)

londs, latds = RingGrids.get_londlatds(grid)

for ij in 1:N
    text!(ax1, londs[ij], latds[ij], text=string(field[ij]), color=:black, align=(:center, :center))    
    text!(ax2, londs[ij], latds[ij], text=string(field_nested[ij]), color=:black, align=(:center, :center))    
end

faces = RingGrids.get_gridcell_polygons(typeof(grid), grid.nlat_half, add_nan=true)
lines!(ax1, vec(faces); color=:black)
lines!(ax2, vec(faces); color=:black)

hidedecorations!(ax1)
hidedecorations!(ax2)
fig
```

</details>

In short, the nested order has the following pattern (which is slightly adapted from [Gorski's HEALPix paper](https://iopscience.iop.org/article/10.1086/427976), but honestly their ordering is also ... weird, I believe we are more consistent).  Excalidraw [link](https://excalidraw.com/#room=333279bcbedda36b7ad8,amYZ4QiNgI0MS5uH4hvU4g)

image

employed at the level of the finest cells (colours) and also when taking 2x2 cells together, nicely nesting all the way to the 4 quadrants of the sphere. This allows one to have trivial conservative interpolation, like data in this ordering, e.g.

```
1 2 3 4 ...
```

is remapped to N times higher resolution by repeating every entry 4N times

```
1 1 1 1 2 2 2 2 3 3 3 3 4 4 4 4 ...
```

and in the other direction by averaging 1:4, 5:8, 9:12, ...

@bgroenks96 This is the mathmatical RingGrids foundation I thought about to do higher resolution land simulations in combination with a lower resolution atmosphere. Remapping fluxes would be trivial and trivially conserving!

@maximilian-gelbrecht I have no opinion yet whether or how to store the order of data on an OctaHEALPix grid which may be now any of the following
- ring order (current default)
- nested order (as discussed above)
- matrix order (allows for interpolation-free data storage in a large square matrix)

There's several options

- don't store the order information at all
- store it in the `OctaHEALPixGrid` as `order::O = RingOrder()` but then reordering a field would require a new grid (which might be the best option given we can't do reordering in-place anyway)
- define a new `OctaHEALPixNestedGrid` but then most functionality is identical between ring/nested
- store it in `Field` but then most fields can't be reordered (only HEALPix/OctaHEALPix but this only implements for the latter)

The interface would look something like

```julia
ring_order(field) -> field
nested_order(field) -> field
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## Unreleased
 
 - Ring, nested and matrix order for OctaHEALPixGrid [#887](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/887)
+- Fixes a bug in the allocating spectral transforms on GPU [#906](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/906)
+- ScratchMemory allocation bug [#905](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/905)
+- Coriolis and adiabatic conv GPU-ready [#904](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/904)
 - The barotropic model can now also be initiliazed on GPU [#903](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/903)
 - Set particle sigma to layer they are advected on [#902](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/902)
 - Vertical advection kernelized and GPU-ready [#895](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/895)
--- a/SpeedyTransforms/src/scratch_memory.jl
+++ b/SpeedyTransforms/src/scratch_memory.jl
@@ -63,8 +63,8 @@ function ScratchMemory(
     grid_type = nonparametric_type(grid)   # always use nonparametric concrete type
 
     # RESOLUTION PARAMETERS
-    nlon_max = get_nlon_max(grid_type, grid.nlat_half)    # number of longitudes around the equator
-                                            # number of longitudes per latitude ring (one hemisphere only)
+    nlon_max = get_nlon_max(grid_type, grid.nlat_half)  # number of longitudes around the equator
+                                                        # number of longitudes per latitude ring (one hemisphere only)
     nfreq_max = nlon_max÷2 + 1                      # maximum number of fourier frequencies (real FFTs)
 
     return ScratchMemory(NF, architecture, nfreq_max, nlayers, grid.nlat_half, nlon_max)
--- a/SpeedyTransforms/src/spectral_transform.jl
+++ b/SpeedyTransforms/src/spectral_transform.jl
@@ -309,7 +309,7 @@ function SpectralTransform(
 )
     # get trunc from dealiasing if not provided
     trunc = trunc > 0 ? trunc : get_truncation(grid, dealiasing)
-    spectrum = Spectrum(trunc+1+one_more_degree, trunc+1)
+    spectrum = Spectrum(trunc+1+one_more_degree, trunc+1; architecture=architecture(grid))
     return SpectralTransform(spectrum, grid; kwargs...)
 end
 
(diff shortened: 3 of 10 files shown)


milankl:
Sparsity structure of a potential sparse-matrix reordering operator

image

but I think recomputation of indices is fast enough, on a 10km grid I get

```julia
julia> @btime RingGrids.reorder!($field_nested, $RingGrids.RingOrder(), $field);
  29.927 ms (2 allocations: 96 bytes)
```

maximilian-gelbrecht:
This is great! 

> There's several options
> 
> * don't store the order information at all
> * store it in the `OctaHEALPixGrid` as `order::O = RingOrder()` but then reordering a field would require a new grid (which might be the best option given we can't do reordering in-place anyway)
> * define a new `OctaHEALPixNestedGrid` but then most functionality is identical between ring/nested
> * store it in `Field` but then most fields can't be reordered (only HEALPix/OctaHEALPix but this only implements for the latter)

Our grids are also used for inferring the coordinates etc so I would say a new type is in order, how would you adjust functions like `get_lond`, `get_latd` etc otherwise? But I'd probably introduce a new abstract type as well, so `AbstractOctaHEALPixGrid`, `OctaHEALPixGrid <: `AbstractOctaHEALPixGrid` and `OctaHEALPixNestedGrid <: `AbstractOctaHEALPixGrid`

milankl:
> Our grids are also used for inferring the coordinates etc so I would say a new type is in order, how would you adjust functions like `get_lond`, `get_latd` etc otherwise? But I'd probably introduce a new abstract type as well, so `AbstractOctaHEALPixGrid`, `OctaHEALPixGrid <: `AbstractOctaHEALPixGrid`and`OctaHEALPixNestedGrid <: `AbstractOctaHEALPixGrid`

Yes, I agree with you that this is the right way to do this, but honestly looking through RingGrids there's a lot of functionality that assumes ring order and so implementing `OctaHEALPixNestedGrid`, `OctaHEALPixMatrixGrid` would take more time than I currently want to spend on it. So, for now, I would prefer to implement this feature as "experimental" and when we know we actually use it for a bigger project then we can do it thoroughly?

milankl:
For the `interleave_with_zeros` there's performance wise still potential

Oscar Smith suggested https://graphics.stanford.edu/~seander/bithacks.html#InterleaveBMN
Andy Dienes:

```julia
function interleave_faster(ui::T) where T <: Unsigned
    s = 2 * sizeof(ui)
    one_ = one(T)
    all_ones = one_ << (4 * s) - one_
    r = ui & (one_ << (2 * s) - one_)
    while s > 0
        mask = all_ones ÷ ((one_ << s) + one_)
        r = (r | (r << s)) & mask
        s >>= 1
    end
    return r
end
```

jakni suggsted: https://github.com/BioJulia/BioSequences.jl/blob/f318aec9c12aa145af0ba391c4555c6144738911/src/bit-manipulation/bit-manipulation.jl#L128

And apparently x86 and arm have this as instruction but LLVM doesn't support it?!?

EDIT: The above has not been implemented ☝🏼 (yet) and can be added if needed.

Issue #843, opened 2025-09-09
SnowLand challenges


pierluigividale:
I have cleaned up much of the snow code in mk/snowland, so that at least I have cleared the compiler errors.

However, this happens, telling me that there is still something quite fundamental missing:

The function `timestep!` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  timestep!(::PrognosticVariables, ::DiagnosticVariables, ::NoVegetation, ::PrimitiveEquation)
   @ SpeedyWeather ~/.julia/packages/SpeedyWeather/q31sb/src/physics/land/vegetation.jl:17
  timestep!(::PrognosticVariables, ::DiagnosticVariables, ::ConstantOceanClimatology, ::PrimitiveEquation)
   @ SpeedyWeather ~/.julia/packages/SpeedyWeather/q31sb/src/physics/ocean.jl:230
  timestep!(::PrognosticVariables, ::DiagnosticVariables, ::Real, ::PrimitiveEquation, ::Integer, ::Integer)
   @ SpeedyWeather ~/.julia/packages/SpeedyWeather/q31sb/src/dynamics/time_integration.jl:333
  ...

milankl:
arises in #833

milankl:
you would need to print the actual error message that's in the lines before `Clostest candidates are:` but I think I know why because you changed the function signature of `timestep!(..., ::SnowModel, ....)` to include `ColumnVariables` but that's not how it's called! Remember atmospheric parameterizations are written column-wise, but the land model operates over the entire horizontal and therefore has a kernel that takes care of the looping over `ij`

pierluigividale:
OK, I can see now what the issue is. HOWEVER, when I tried to go at it from prognostics, I was told that the variable did not exist. Going for column was a way to get around the compiler, but indeed it does not work.

Basically I need to pull surface winds, albedo, sat_hum from the rest of the model to do what I need to do at the land surface.
Maybe I can show you the compiler errors when you are here.

milankl:
yes let's hack that next week

pierluigividale:
Snow cover works with either formula, and seems very reasonable.
HOWEVER, there is no impact on albedo at all. Are we updating the right variables? is diagn.albedo the one?

    # set land albedo linearly between bare land and snow depending on snow cover [0, 1]
    diagn.albedo .= albedo_land .+ snow_cover .* (albedo_snow - albedo_land)

[...]

PR #888, opened 2025-10-29
OneBandShortwave radiation scheme with diagnostic cloud reflection

nviebig:
Adds a one-band shortwave radiation scheme with diagnostic cloud-top reflection, following [Molteni & Kucharski (2008), Appendix A (ICTP SPEEDY v4.1, Sec. B.4)](https://users.ictp.it/~kucharsk/speedy_description/km_ver41_appendixA.pdf). Implements the visible-band formulation of Eqs. (26)–(33) for cloud-dependent transmissivity and reflection.

Key additions
- add `OneBandShortwave` parameterisation mirroring SPEEDY’s B.4 constants
- diagnose cloud cover from RH and precipitation thresholds and apply cloud-top reflection per Eq. (33)
- update the column shortwave flux bookkeeping (surface up/down, TOA) accordingly

Next steps
- extend to a two-band (visible + near-IR) treatment
- include ozone absorption and upper-atmosphere attenuation
- implement the stratocumulus enhancement from Eq. (34)

<p align="center">
  
  
</p>

<p align="center">
  <b>TransparentShortwave (left)</b> vs <b>OneBandShortwave (right)</b> — surface shortwave down flux (SSD)
</p>

TODO
- [x] Runs through
- [x] Write tests
- [x] Brief documentation
- [x] Ozone?
- [x] Stratocumulus?
- [x] Background transmittance?


milankl:
One check failed which literally just checks whether `CHANGELOG.md` has been altered. Done this now in [a212ffe](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/888/commits/a212ffe17cc4609689492772d363bbb03cb9ac86) We have this to force ourselves to be more aware of which pull request is in which version and it also reminds you what has been merged into main since you branched off. 

TL;DR: every pull request requires a new line logging the changes made summarised as e.g.

```julia
- OneBandShortwave radiation implemented with diagnostic clouds [#888](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/888)
```

milankl:
Then your code contains a good amount of so-called gremlins, these are unicode characters that look like other unicode characters but aren't

image

In docstrings that may not be too bad but e.g. in markdown it took me ages to find out that I didn't have pressed SPACE but OPTION+SPACE which creates on macbooks a characters that looks like a space but disables the markdown rendering of section titles (see bottom lines). In VSCode there's a Gremlin tracker extension. Not all Gremlins are too bad but just replace them with easier characters if possible

src/physics/shortwave_radiation.jl
```diff
@@ -86,4 +86,132 @@ function shortwave_radiation!(
...
+
+struct OneBandShortwave <: AbstractShortwave
+    # thresholds / weights
+    rh_cl::Float64        # RH_cl,  thresholds for RH and cloud water content (Eqs. 2627).
+    rh_clp::Float64       # RH'_cl,     ---------------------''-----------------------
+    q_cl::Float64         # Q_cl,       ---------------------''-----------------------
+    wp_cl::Float64        # w_pcl   weighting of precipitation term (Eq. 28).
+    pmax_cl::Float64      # p_maxcl (mm/day) 
+    # optics
+    alb_cl::Float64       # A_cl (cloud albedo, visible) (Eq. 33).
```
milankl: This hardcodes the number format to Float64, many GPUs are slow with Float64 or don't even support it (macbooks) we therefore make the number format `NF` parametric as done in many other structs. Use `@kwdef` as a macro to define defaults

milankl: If you write comments as a string in the line above then the documentation can automatically read this out

```julia
    "[OPTION] RH threshold for cloud cover = 0 [1]"
    rh_cl::NF = 0.3
```

src/physics/shortwave_radiation.jl
```diff
@@ -86,4 +86,132 @@ function shortwave_radiation!(
...
+Notes:
+- `abs_wv_vis` expects q in g kg⁻¹. If your model uses kg kg⁻¹, internally scale by 1000 (or rescale the coefficient).
+- `abs_dry` is calibrated per layer mass of Δp = 10⁵ Pa (SPEEDY convention). Match your Beer-Lambert exponent accordingly.
+- If you dont implement Eq. 34 stratocumulus, you can ignore `alb_cls`, `cls_*`, `clf`, `gses*`.
+"""
+OneBandShortwave() = OneBandShortwave(
+    0.30, 1.00, 0.20,   # rh_cl (-), rh_clp (-), q_cl (g kg⁻¹)
+    0.20, 10.0,         # wp_cl (-), pmax_cl (mm day⁻¹)
+    0.43                # alb_cl (-)
+)
```
milankl: We do generator functions based on `::SpectralGrid; kwargs...`

src/physics/shortwave_radiation.jl
```diff
@@ -86,4 +86,132 @@ function shortwave_radiation!(
...
+    r      = clamp((RH[min(k_cl, nlayers)] - RH_cl) / max(RH_clp - RH_cl, T(1e-6)), zero(T), one(T))
+
+    P_mmday = T(86.4) * (rain_rate_convection + rain_rate_large_scale)
+    P_mmday = min(scheme.pmax_cl, P_mmday)
+    CLC     = min(one(T), scheme.wp_cl * sqrt(P_mmday) + r^2)
+
+    # ---- transmissivity per layer: τ_layer = exp(- τk / μ) ----
+    τ = similar(qv)
+    @inbounds @simd for k in 1:nlayers
+        τ[k] = exp(-τk[k] / μ)
```
milankl: where is τk[k] actually being written into? And why do we divide by the cos zenith here? That should be the top flux boundary condition

src/physics/shortwave_radiation.jl
```diff
@@ -86,4 +86,132 @@ function shortwave_radiation!(
...
+    qv  = humid
+    qsat = sat_humid
+    T   = eltype(qv)
+    τk = @view optical_depth_shortwave[1:nlayers, 1]
+
+    # ---- derived diagnostics ----
+    RH = @. clamp(qv / max(qsat, eps(T)), 0, 1) # clamp relative humidity between 0 and 1
+    k_cl = (1 <= cloud_top <= nlayers) ? cloud_top : max(nlayers - 1, 1) # cloud top layer index
+    RH_cl  = scheme.rh_cl 
+    RH_clp = scheme.rh_clp
```
milankl: We usually follow a style whereby inside the struct names are long and telling, e.g. `planet.gravity` but then you can abbreviate locally inside a function body to match the math notation, e.g. `g = planet.gravity`

src/physics/shortwave_radiation.jl
```diff
@@ -86,4 +86,132 @@ function shortwave_radiation!(
...
+       humid, sat_humid, albedo_ocean, albedo_land,
+       optical_depth_shortwave, flux_temp_downward, flux_temp_upward,
+       cloud_top, rain_rate_convection, rain_rate_large_scale) = column
+
+    # ---- aliases / guards ----
+    μ   = max(cos_zenith, 1f-3)                         # avoid divide-by-zero
+    qv  = humid
+    qsat = sat_humid
+    T   = eltype(qv)
+    τk = @view optical_depth_shortwave[1:nlayers, 1]
```
milankl: This is already a matrix with nlayers x 1 so no need for a view

src/physics/shortwave_radiation.jl
```diff
@@ -86,4 +86,132 @@ function shortwave_radiation!(
...
+    RH_cl  = scheme.rh_cl 
+    RH_clp = scheme.rh_clp
+    r      = clamp((RH[min(k_cl, nlayers)] - RH_cl) / max(RH_clp - RH_cl, T(1e-6)), zero(T), one(T))
+
+    P_mmday = T(86.4) * (rain_rate_convection + rain_rate_large_scale)
+    P_mmday = min(scheme.pmax_cl, P_mmday)
+    CLC     = min(one(T), scheme.wp_cl * sqrt(P_mmday) + r^2)
+
+    # ---- transmissivity per layer: τ_layer = exp(- τk / μ) ----
+    τ = similar(qv)
```
milankl: this is allocating, there are scratch arrays you can use to avoid allocations but you don't even need them here because `column.optical_depth_shortwave` already exists

src/physics/shortwave_radiation.jl
```diff
@@ -86,4 +86,132 @@ function shortwave_radiation!(
...
+    τk = @view optical_depth_shortwave[1:nlayers, 1]
+
+    # ---- derived diagnostics ----
+    RH = @. clamp(qv / max(qsat, eps(T)), 0, 1) # clamp relative humidity between 0 and 1
+    k_cl = (1 <= cloud_top <= nlayers) ? cloud_top : max(nlayers - 1, 1) # cloud top layer index
+    RH_cl  = scheme.rh_cl 
+    RH_clp = scheme.rh_clp
+    r      = clamp((RH[min(k_cl, nlayers)] - RH_cl) / max(RH_clp - RH_cl, T(1e-6)), zero(T), one(T))
+
+    P_mmday = T(86.4) * (rain_rate_convection + rain_rate_large_scale)
```
milankl: The rates come in `m/s` so this is correct, but I would probably not hardcode the `T(86.4)` here but move it to the struct as a parameter, that comes with the advantage that this part of the equation has a user interface for tuning, e.g. you could add `precipitation_time_scale::Second = Day(1)` and then use `precipitation_time_scale.value` (an integer) `/1000` for unit conversion

src/physics/shortwave_radiation.jl
```diff
@@ -86,4 +86,132 @@ function shortwave_radiation!(
...
+    D_incident_ct = zero(T)
+    # Clear sky portion until cloud top
+    @inbounds for k in 1:(k_cl - 1)
+        D *= τ[k]
+        flux_temp_downward[k + 1] += D
+    end
+    
+    # Cloudy portion from cloud top downward
+    D_incident_ct = D
+    D *= (one(T) - scheme.alb_cl * CLC)
```
milankl: You can write `1` which will automatically be converted to the number format if everything else is type stable is actually done at compile time

milankl:
```julia
julia> f(x) = 1 - x

julia> @code_llvm f(2.0)
; Function Signature: f(Float64)
;  @ REPL[1]:1 within `f`
define double @julia_f_1005(double %"x::Float64") #0 {
top:
; ┌ @ promotion.jl:431 within `-` @ float.jl:492
   %0 = fsub double 1.000000e+00, %"x::Float64"
   ret double %0
; └
}
```

See the `1.000000e+00` got hardcoded into the compiled code!

src/physics/shortwave_radiation.jl
```diff
@@ -86,4 +86,132 @@ function shortwave_radiation!(
...
+    up_ocean = albedo_ocean * D
+    up_land  = albedo_land  * D
+    column.surface_shortwave_up_ocean = up_ocean
+    column.surface_shortwave_up_land  = up_land
+
+    # Computes grid-cell-average surface albedo and reflected shortwave flux, 
+    albedo_sfc = (one(T) - land_fraction) * albedo_ocean + land_fraction * albedo_land
+    column.surface_shortwave_up = (one(T) - land_fraction) * up_ocean + land_fraction * up_land
+
+    U = albedo_sfc * D
```
milankl: this is just `column.surface_shortwave_up` no?

src/physics/shortwave_radiation.jl
```diff
@@ -86,4 +86,132 @@ function shortwave_radiation!(
...
+        flux_temp_upward[k] += U
+        if k == k_cl
+            dU = scheme.alb_cl * CLC * D_incident_ct
+            U  += dU
+            flux_temp_upward[k] += dU
+        end
+    end
+
+    # ---- 4) TOA outgoing SW ----
+    column.outgoing_shortwave_radiation += U
```
milankl: I don't think this has to be added? There's no other scheme that one would need to accumulate into?

milankl:
Cloud top in meters

image

Surface shortwave radiation down in W/m^2

image

test/physics/shortwave_radiation.jl
```diff
@@ -0,0 +1,16 @@
+@testset "Shortwave radiation" begin
+    spectral_grid = SpectralGrid(trunc=31, nlayers=8)
+    
+    for SW in (TransparentShortwave, OneBandShortwave)
+        @testset "SW: $(SW)" begin
```
milankl: You can just write `@testset for ...`

test/physics/shortwave_radiation.jl
```diff
@@ -0,0 +1,16 @@
...
+            sw = SW(spectral_grid)
+            model = PrimitiveWetModel(spectral_grid; shortwave_radiation=sw)
+            simulation = initialize!(model)
+            run!(simulation, period=Day(5))
+            ssrd = simulation.diagnostic_variables.physics.surface_shortwave_down
+            TRD = model.planet.solar_constant * simulation.diagnostic_variables.physics.cos_zenith
+            @test all((0 .<= ssrd) .& (ssrd .<= TRD))
```
milankl: You can add a similar test for outgoing shortwave radiation, for albedo=1 (whether from clouds or surface) everything is reflected but lower albedos some shortwave radiation is absorbed, so `@test all(0 .<= osr .<= TRD)` should also hold!

milankl: And you can fuse logical operations in Julia
```julia
julia> 0 < 1 <= 2 < 3 == 3 < 4
true
```

milankl:
Because #855 and #889 got merged since you branched off there's now a conflict in CHANGELOG.md (that's normal) it'll tell you nicely what has been implemented since and in most cases you just "Accept both changes" but generally your pull request then should be listed at the top under "# Unreleased" I let do you that

docs/src/radiation.md
```diff
@@ -58,19 +59,140 @@ flux from below at interface ``k+1/2`` (``k`` increases downwards, see
...
+The `OneBandShortwave` scheme provides a single-band (broadband) shortwave radiation parameterization, including diagnostic cloud effects following [^KMB06].
+
+**Cloud diagnosis:**
+Cloud properties are diagnosed from the relative humidity and total precipitation in the atmospheric column. The cloud base is set at the interface between the lowest two model layers, and the cloud top is the highest layer where both
+
+$$
+\mathrm{RH}_k > \mathrm{RH}_{cl} \quad \text{and} \quad Q_k > Q_{cl}
+$$
+
+are satisfied. The cloud cover (CLC) in a layer is then given by
```
milankl, line 87: I know this is written in the Fortran SPEEDY documentation but this is, as it's currently implemented actually not true. `cloud_top` starts at `nlayers+1` then if large-scale condensation occurs it'll be moved up to the highest layer k where condensation occurs `cloud_top = min(k, cloud_top)` and similar for convection. You could see what difference it makes to actually implement the `cloud_top` calculation as outlined here

src/physics/shortwave_radiation.jl
```diff
@@ -35,12 +53,12 @@ function shortwave_radiation!(
     (; cos_zenith, land_fraction, albedo_ocean, albedo_land) = column
     (; solar_constant) = planet
 
-    R = solar_constant * cos_zenith         # top of atmosphere downward radiation
-    column.surface_shortwave_down = R       # transparent atmosphere so same at surface (before albedo)
+    D = solar_constant * cos_zenith         # top of atmosphere downward radiation
+    column.surface_shortwave_down = D       # transparent atmosphere so same at surface (before albedo)
```
milankl: thanks for making this consistent!

docs/src/radiation.md
milankl: This might be important to check because `ImplicitCondensation` actually occurs from 95% humidity onwards, not 100%, the idea here is that a grid cell with 95% average rel humidity will have sub-grid scale variability where 100% should be reached ... this is technically in contrast to the threshold of 100% used here, like you can have clouds without rain but there shouldn't be rain without clouds

src/physics/shortwave_radiation.jl
```diff
@@ -86,4 +107,298 @@ function shortwave_radiation!(
...
+    (; nlayers, cos_zenith, land_fraction,
+       humid, sat_humid, albedo_ocean, albedo_land,
+       transmittance_shortwave, flux_temp_downward, flux_temp_upward, dry_static_energy) = column
+
+     
+    # Weight and max precip for cloud cover calculation
+    wpcl = radiation.precipitation_weight
+    pmcl = radiation.precipitation_max
+    # Contribution to cloud cover from precipitation, convert m/s to mm/day (Speedy multiplies by 86.4 because its rates are already in kg m⁻² s⁻¹ (= mm s⁻¹))
+    P = wpcl*sqrt(min(pmcl, (86400 * (column.rain_rate_large_scale + column.rain_rate_convection) * 1000)))
```
milankl: `/ 1000` ?

src/physics/shortwave_radiation.jl
```diff
@@ -86,4 +107,298 @@ function shortwave_radiation!(
...
+    stratocumulus_cloudcover = zero(P)
+    if radiation.use_stratocumulus
+        # Compute static stability (GSEN)
+        GSEN = dry_static_energy[nlayers] - dry_static_energy[nlayers-1]
+
+        # Use tunable parameters from struct
+        stability_min = radiation.stratocumulus_stability_min
+        stability_max = radiation.stratocumulus_stability_max
+        cover_max = radiation.stratocumulus_cover_max
+        clfact = 1.2  # Added missing factor from SPEEDY
```
milankl: Don't hardcode this but move into the `struct` also this way it's hardcoded as Float64!

src/physics/shortwave_radiation.jl
```diff
@@ -86,4 +107,298 @@ function shortwave_radiation!(
...
+        
+        (; absorptivity_dry_air, absorptivity_aerosol, absorptivity_water_vapor,
+        absorptivity_cloud_base, absorptivity_cloud_limit) = radiation
+
+        sigma_levels = model.geometry.σ_levels_half
+        surface_pressure = column.pres[end]  # This is in Pa
+
+        # Zenith angle correction factor (SPEEDY's zenit)
+        # azen = 1.0, nzen = 2 in SPEEDY
+        zenit_factor = 1.0 + 1.0 * (1.0 - cos_zenith)^2
```
milankl: Hardcoding float64 again! `1 + (1 - cos_zenith)^2` is generic and will always promote to any float format!

You can check this with
```julia
julia> 1 + (1 - 5f-1)^2
1.25f0

julia> 1.0 + (1.0 - 5f-1)^2
1.25
```
in the former float32 stays float32 in the latter it's promoted (=implicit, automatic conversion) to float64. The underlying reason is that this is a generic rule
```julia
julia> Base.promote_type(Float32, Float64)
Float64
```

whereas
```julia
julia> Base.promote_type(Float32, Int)
Float32
```

preserves whether one uses float16, 32 or 64!

src/physics/shortwave_radiation.jl
```diff
@@ -86,4 +107,298 @@ function shortwave_radiation!(
...
+        # Cloud absorption term based on cloud base humidity (SPEEDY logic)
+        q_base = nlayers > 1 ? humid[nlayers-1] : humid[nlayers]
+        cloud_absorptivity_term = min(absorptivity_cloud_base * q_base,
+                                    absorptivity_cloud_limit)
+
+        for k in 1:nlayers
+            q = humid[k]
+
+            # Aerosol factor: use mid-level sigma, squared
+            sigma_mid = 0.5 * (sigma_levels[k] + sigma_levels[k+1])
```
milankl: Don't hardcode Float64!

src/physics/shortwave_radiation.jl
```diff
@@ -86,4 +107,298 @@ function shortwave_radiation!(
...
+
+            # Add cloud absorption below the FINAL cloud top
+            if k >= cloud_top
+                layer_absorptivity += cloud_absorptivity_term * cloud_cover
+            end
+
+            # Compute differential optical depth with zenith correction
+            # CRITICAL: Normalize pressure to 10^5 Pa since absorptivities are per 10^5 Pa
+            delta_sigma = sigma_levels[k+1] - sigma_levels[k]
+            normalized_pressure = surface_pressure / 1e5  # Convert Pa to units of 10^5 Pa
```
milankl: Again don't hardcode float64! Also if you can write as multiplication not as division as its faster, `/ 1e5` vs `*1e-5`

milankl: Here I would write `/ 10000` probably

src/physics/shortwave_radiation.jl
milankl: Also what's the redundant `1.0 *` for?

milankl (commented):
Well done already really good, some points

- let me know if you want me to fix some of my comments this week of you can do this next week
- we need to add some tests for the different flags in `OneBandShortwave`. We don't test various parameter combinations as this would become an overkill but the tests should cover all branches
- Ideally we would create a quick plot showing that there are differences between with/without top cloud reflection and with/without stratocumulus reflection, this could also go into the documentation be produced automatically

milankl:
The last commit should solve the `animate` problem when building the docs. Other docs issues where around the user interfaces we made when creating a `OneBandShortwave` object -- these have to be reflected in the docs

nviebig:
image

My attempt to compare the outgoing shortwave radiation (OSR) for January 2000 between the SpeedyWeather and CESM2 CMIP6 historical data. The code that produced this plot:

milankl:
While the difference is large in some areas there's only some features that actually worry me. Things that _don't_ worry me

- Antarctica: CESM2 clearly uses a higher albedo over Antarctica so we could/should set this higher. Or rely on the snow model for this, see below.
- Sea ice on Weddell and Ross sea: You probably ran this only for a brief amount of time, if you wait a little longer then the sea surface temperatures in these areas drop and produce sea ice which increases albedo, see #811 
- Himalayas and Andes: While we have #817 at the moment the land model doesn't accumulate the falling snow into a snow cover so the albedo doesn't know about snow cover. Work on this is in #843
- Amazon: Looks like there's not enough high-rising convective clouds here, I would first blame the surface fluxes (which trigger convection) or the convection scheme

I believe these "issue" can be easily solved with turning a few knobs / continuing the work in the cited pull requests. However there's also features in these differences that do require some attention

- Southern hemisphere midlatitude storm tracks: There is clearly not enough clouds here. (You did a monthly average like in the CESM data?)
- Intertropical convergence zone (ITCZ): There's a band of low values in CESM2 (for clouds it should be high no? but maybe it's reflected away) on the equator we clearly don't have enough ITCZ in SpeedyWeather but I also blame the vertical resolution (you did 8 layers?, you can see what happens with 16 layers, 24-32 layers can often be unstable at the moment)

nviebig:
Screenshot 2025-11-14 at 17 58 14
Reference simulation: primitive wet atmosphere with no clouds and fully transparent shortwave radiation

nviebig:
Transparent control: TransparentShortwave; pure solar forcing passes straight to the surface, so the OSR map is just land/ocean albedo.
ssrd_transparent

Background (no aerosols, no clouds, humidity-dependent only): BackgroundShortwaveTransmittance without aerosols; depicts clear‑sky absorption in the column even when cloud cover is zero.
ssrd_background

Background + aerosols: Same as above but with the aerosol absorptivity term enabled, highlighting the extra dimming near the surface.
ssrd_aerosols

Stratocumulus only: DiagnosticClouds with impossible RH/precip thresholds so normal clouds vanish, but the stratocumulus branch stays active; isolates surface-based marine cloud reflection. With background transmittance.
ssrd_stratocumulus

Clouds (no ozone): OneBandShortwave with diagnostic clouds (no stratocumulus); bright patches mark where cloud albedo reflects additional shortwave back to space.
ssrd_clouds

Clouds + Ozone absorption: Background transmittance plus 15/5 W m⁻² ozone absorption in the radiative transfer; OSR drops uniformly by ~20 W m⁻² relative to the background case.
ssrd_ozone

Clouds + Stratocumulus: Diagnostic clouds with use_stratocumulus=true; shows the combined effect of humidity-driven cloud cover plus the low‑level stratocumulus reflection.
ssrd_clouds_stratocumulus


Ran a 10‑week transparent shortwave spin‑up once and reused that restart so every diagnostic starts from the same state. Note that these parameters aren’t hand-tuned for realism; especially the stratocumulus albedo, which appears rather small and should be revisited if you want the stratocumulus run to stand out.

milankl:
MERGE 🎉

PR #890, opened 2025-10-31
Bug? Reset geopotential accumulator in implicit correction

milankl:
Found by @maximilian-gelbrecht 

`dynamics.b` is used to hold `v*q` the northward humidity flux before ... it's weird it's been stable the past 2.5 years but on the other hand the original idea for these semi-implicit corrections is that one uses no orography but I never quite understood why because I believe one could just use surface geopotential as a starting condition for this integration. So maybe we just did a semi-implicit correction relative to an "orography" = `v*q` that was even changing with time ...  Maybe that was also the reason why more vertical layers always caused problems ... I will come up with a testcase and check whether `geopot .= 0` or `geopot[:, end] .= geopot_surf` is better!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Reset geopotential accumulator in implicit correction [#890](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/890)
 - Zenith constructors [#889](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/889)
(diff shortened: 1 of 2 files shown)


milankl:
So I've tested this at T31L32 (32 layers at default horizontal resolution) and fixing the bug does not seem to have any impact on the stability here. I would still fix it of course but I don't see any immediate positive (or negative) impacts this has unfortunately ...

milankl:
Could be merged but would be superceded by #891 anyway

PR #891, opened 2025-11-04
Implicit Correction Kernelized 

maximilian-gelbrecht:
Well, that was annoying to debug, but apparently I found a Julia lang bug while doing so. 

Still WIP because I need some of the GPU fixes from #873 in order to test this on GPU. So #873 needs to be merged first. The test file will be deleted, as will the other versions of the correction. 

On CPU currently the 1 kernel and 4 kernel version are roughly equally fast, and x1.5 of the CPU version. So currently this would be within our tolerances.


milankl:
Okay sounds like ball is in my court to get the remaining parameterizations into #873 and make that PR pass!

milankl:
But so ideally we use the single kernel version because it's ... just one kernel? I see that for testing it's nice to split but otherwise they could be all fused no? Otherwise can we just call the 4 kernels from one launch?

maximilian-gelbrecht:
The four kernel variant has probably an advantage for medium resolutions as it's a kernel over (lm, k) instead of lm. The single kernel needs longer to fill all threads of a GPU. But I'll test how much of a difference that makes after #873 is merged.

maximilian-gelbrecht:
I briefly forgot I even made this PR. 

After the issues with KA are fixed now, there's still a correctness issue here, that'll I address in the next days. 

At least the KA versions are actually pretty fast (faster than the original).

maximilian-gelbrecht:
There are some small numerical differences between the KA implementation and the original implmenetion that seem to cause the whole model to become unstable at the moment 

Differences are of the type 
```Julia 
julia> (div_tend_gpu - div_tend_cpu)[:,6]
2144-element, 65x64 LowerTriangularMatrix{ComplexF32, Array{...}}
 0.0+0.0im           0.0+0.0im                 0.0+0.0im                 0.0+0.0im         …          0.0+0.0im         0.0+0.0im  0.0+0.0im         0.0+0.0im
 0.0+0.0im           0.0+0.0im                 0.0+0.0im                 0.0+0.0im                    0.0+0.0im         0.0+0.0im  0.0+0.0im         0.0+0.0im
 0.0+0.0im           0.0+0.0im                 0.0+0.0im                 0.0+0.0im                    0.0+0.0im         0.0+0.0im  0.0+0.0im         0.0+0.0im
 0.0+0.0im           0.0+0.0im         -6.10352f-5+0.0im                 0.0+0.0im                    0.0+0.0im         0.0+0.0im  0.0+0.0im         0.0+0.0im
 0.0+0.0im           0.0+0.0im                 0.0+0.0im                 0.0+0.0im                    0.0+0.0im         0.0+0.0im  0.0+0.0im         0.0+0.0im
 0.0+0.0im           0.0+0.0im                 0.0+0.0im                 0.0+0.0im         …          0.0+0.0im         0.0+0.0im  0.0+0.0im         0.0+0.0im
 0.0+0.0im           0.0+0.0im                 0.0+0.0im                 0.0+0.0im                    0.0+0.0im         0.0+0.0im  0.0+0.0im         0.0+0.0im
 0.0+0.0im           0.0+0.0im                 0.0+0.0im                 0.0+0.0im                    0.0+0.0im         0.0+0.0im  0.0+0.0im         0.0+0.0im
 0.0+0.0im           0.0+0.0im                 0.0+0.0im                 0.0+0.0im                    0.0+0.0im         0.0+0.0im  0.0+0.0im         0.0+0.0im
    ⋮                                                                                      ⋱             ⋮                                           
 0.0+0.0im   -0.00012207+0.0im                 0.0+3.05176f-5im   0.00012207-0.00012207im             0.0+0.0im         0.0+0.0im  0.0+0.0im         0.0+0.0im
 0.0+0.0im   -3.05176f-5+0.0im                 0.0+0.0im                 0.0+0.0im                    0.0+0.0im         0.0+0.0im  0.0+0.0im         0.0+0.0im
 0.0+0.0im           0.0+0.00012207im          0.0+0.0im                 0.0+3.05176f-5im             0.0+0.0im         0.0+0.0im  0.0+0.0im         0.0+0.0im
 0.0+0.0im           0.0+0.0im                 0.0+0.0im         -0.00012207+0.0im                    0.0+0.0im         0.0+0.0im  0.0+0.0im         0.0+0.0im
 0.0+0.0im           0.0+0.0im                 0.0+0.0im         -6.10352f-5+0.0im         …  -6.10352f-5+0.0im         0.0+0.0im  0.0+0.0im         0.0+0.0im
 0.0+0.0im           0.0+0.0im                 0.0+0.0im                 0.0+0.0im                    0.0+0.0im         0.0+0.0im  0.0+0.0im         0.0+0.0im
 0.0+0.0im           0.0+0.0im                 0.0+0.0im                 0.0+0.0im                    0.0+1.52588f-5im  0.0+0.0im  0.0+1.90735f-6im  0.0+0.0im
 0.0+0.0im  -0.000244141+0.0im                 0.0+0.0im                 0.0+0.0im                    0.0+0.0im         0.0+0.0im  0.0+0.0im         0.0+0.0im
 0.0+0.0im           0.0+0.0im                 0.0+0.0im                 0.0+0.0im                    0.0+0.0im         0.0+0.0im  0.0+0.0im         0.0+0.0im
```

maximilian-gelbrecht:
Okay, so it does seem to work now, at least all tests pass which means it is at least somewhat stable. 

@milankl You are the low-precision expert. To me this does seem to be a floating point precision problem here. We have a lot of accumulation of floating point numbers here. By changing the loop order with the kernel version, the floating point operations are carried out in different order. Especially introducing a local accumulator for adding to `temp_tend` in the beginning turned out to be important and this is a situation where you'll potentially add very small numbers to a large number. On the other hand, I even tried a Kahan summation version and this also didn't work better, so I am a bit confused. In general I am surprised that differences on the order of `rtol=1e-6` can make or break the stability of the model here. With local accumulators everywhere it does seem to work now, though, no Kahan needed. 

I'll merge this PR into #934 later.

maximilian-gelbrecht:
I merge this PR now into the other, but just wanted to point again that the floating point precision issue here is something to be aware of.

PR #892, opened 2025-11-04
Seasonal ocean and land models kernelized

maximilian-gelbrecht:


maximilian-gelbrecht:
@milankl I noticed that `monthly_soil_moisture <: GridVariable4D` but `monthly_temperature <: GridVariable3D ` even though both also run over all soil layers. `SeasonalLandTemperature` only speaks of surface temperature, should we then also assert that `nlayers_soil == 1` when it's used, or more generally that the input dimensions fit the number of soil layers.

maximilian-gelbrecht:
Well, or maybe we just keep it as it is for now, merge it and address the inconsistency after #873 is merged. After #873, I would quickly make another PR that takes `nlayers_soil` out of the `SpectralGrid`, makes it a property of the land model only and uses the new `variables` system. Then, these things will always be consistent anyway. So no need for a temporary fix here.

PR #893, opened 2025-11-04
Use uv_from_vordiv kernel version

maximilian-gelbrecht:
We already had a `uv_from_vordiv!` kernel version in #727, but we didn't use it, because there were problems with Enzyme. Let's see if these problems still exist. Either way we have to find a solution. 

Related to that there's also #763

### DIFF
--- a/SpeedyTransforms/src/spectral_gradients.jl
+++ b/SpeedyTransforms/src/spectral_gradients.jl
@@ -264,7 +264,7 @@ end
     l = l_indices[lm]
 
     # Get the coefficients for the current lm index
-    z = vordiv_to_uv_x[lm]
+    z = im*vordiv_to_uv_x[lm]
     vordiv_uv1 = vordiv_to_uv1[lm]
     vordiv_uv2 = vordiv_to_uv2[lm]
     
@@ -273,10 +273,10 @@ end
         # U = -∂/∂lat(Ψ) - no lm-1 term for first element
         U[I] = vordiv_uv2 * vor[lm+1, k]
         # V = ∂/∂λ(Ψ)
-        V[I] = im * (z * vor[I])
+        V[I] = z * vor[I]
     elseif l==(lmax-1)     # extra in case vor[lmax,:] != 0, see comment in UV_from_vordiv!
         U[I] = -vordiv_uv1 * vor[lm-1, k]    # meridional gradient again (but only 2nd term from above)
-        V[I] = im * (z * vor[I])          # zonal gradient again (as above)
+        V[I] = z * vor[I]          # zonal gradient again (as above)
     elseif l==lmax         # extra in case vor[lmax,:] != 0, see comment in UV_from_vordiv!
         U[I] = -vordiv_uv1 * vor[lm-1, k]
         V[I] = 0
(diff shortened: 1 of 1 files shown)


maximilian-gelbrecht:
Well, good news, the Enzyme issue is gone. 

As far as I can see it the test fail because of the issues, we already had in #735 and #763

maximilian-gelbrecht:
@vchuravy it seems Enzyme v0.13.105 broke our differentiability. I am pinging here, but I see our Enzyme CI failing since Friday in every PR (so it can't really be us, and it's not directly related to this PR). The stack trace is in the CI report for the Enzyme CI. The test it runs is this one: https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/main/test/differentiability/full_diff_CI.jl 
Do you see something immediately or should I try to condense this done further / minimal? 

And yeah, I should really find a bit faster set up for your CI in Enzyme.jl to best avoid this situation in the future 🫣

maximilian-gelbrecht:
The regular CI tests actually just failed because of a stupid mistake I overlooked. Once everything passes now, this will be merged.

maximilian-gelbrecht:
Shit now Enzyme continues to fail here..., this was a problem before

vchuravy:
~Locally the Enzyme tests pass on 1.11~ Well forget that the tests are disabled for 1.11

maximilian-gelbrecht:
In someway it has something to do with one of the coefficients being a complex number, because the test was passing before when I mistakingly forgot to multiply by `im`, this was the only change. I look into actually isolating this further, but this is also independent from the other more general problem since Enzyme 0.13.105, that I also want to isolate further. The latter is more important now.

maximilian-gelbrecht:
Huh, now, it's passing.  The only change I made now was to multiply by `im` in a different line, mathematically it's identical. 

So this can be merged now, but I'll try to investigate this as well after the bigger Speedy+Enzyme issue. 

I just re run the CI to be a bit more certain this wasn't just a fluke.

SpeedyTransforms/src/spectral_gradients.jl
```diff
@@ -273,18 +273,18 @@ end
         # U = -∂/∂lat(Ψ) - no lm-1 term for first element
         U[I] = vordiv_uv2 * vor[lm+1, k]
         # V = ∂/∂λ(Ψ)
-        V[I] = z * vor[I]
+        V[I] = im * (z * vor[I])
```
milankl: okay that's interesting that this is an important change for Enzyme. Also the `()` parentheses, did you put them in deliberately to affect the order?

maximilian-gelbrecht: I did put them there deliberately, but purely out of a hunch I can't explain :D

Maybe just to make it different to the version before.

maximilian-gelbrecht: Gosh, now it's failing again

maximilian-gelbrecht: Okay, 2 out of 3 times it didn't work, I guess it was just a fluke :(

maximilian-gelbrecht: The whole issue is honestly quite weird because this kernel is nothing special compared to the other ones. Next week I'll try to properly isolate the issue.

maximilian-gelbrecht:
Ok, so there are still issues here wrt Enzyme. I'll try to isolate the issue over this week. 

In order to go forward with the GPU version, I'll do a version where we'll continue to use the old version on CPU and have differentiability there, and use KA just for GPU to get the GPU online for proper testing and benchmarking.

SpeedyTransforms/src/spectral_gradients.jl
vchuravy: If it is a memory computation then the failure is sadly on-determistic.

PR #895, opened 2025-11-06
Vertical advection kernelized

maximilian-gelbrecht:
This is a very direct GPU adaption. We'll do that first, and later when we really want to optimise the GPU performance we could experiment with fusing the kernels here as well. 

I already checked the correctness of all versions. 

On CPU the kernelled version is roughly x1.1 of the old version, so this is quite good already.


milankl:
You're a rocket, I mean you're a train 🚂 🚂

maximilian-gelbrecht:
Ok, interestingly, this here seems to be another case where the kernel version on ARM is almost identical in performance to the old versions, but on x86 the kernel version is quite a lot worse (x6).

The GPU version works on GPU and is already at T41 faster than on CPU

PR #896, opened 2025-11-06
Test also on Julia v1.12

milankl:


maximilian-gelbrecht:
The error comes from Enzyme which isn't compatible with v1.12 yet. I am also okay with just marking this test as broken in Julia v1.12 for now. 

LLVM changed to untyped pointers which makes the live for Enzyme a bit harder unfortunately and Julia 1.12 is the first version to use a LLVM where that's the case.

maximilian-gelbrecht:
Aside from Enzyme all tests pass 🎉

milankl:
Great, so but marking the test as broken would actually tell us then when it's fixed and then we can mark these tests as unbroken again and remove the if v1.12 branching?

```julia
julia> @test_broken true
Error During Test at REPL[8]:1
 Unexpected Pass
 Expression: true
 Got correct result, please change to @test if no longer broken.

ERROR: There was an error during testing
```

Happy to merge this then!

milankl:
What worries me a bit are CI times, on v1.12 the last one here was 48min ... 😴

maximilian-gelbrecht:
Currently it's just hardcoded to be broken. We could do it with a try catch statement instead to really see when it works. But then, it's Enyzme, we will know when it's compatible and when we can work with it in 1.12

milankl:
But the biggest problem regarding CI time is actually compilation time, e.g.

```julia
Test Summary:  | Pass  Total     Time
Land sea masks |    3      3  1m28.9s
Test Summary:            | Pass  Total     Time
Ocean and sea ice models |   28     28  3m28.7s
Test Summary:            | Pass  Total     Time
Large-scale condensation |   36     36  2m06.3s
Test Summary: | Total     Time
Convection    |     0  2m36.0s
Test Summary:  | Pass  Total     Time
Single Albedos |   18     18  2m49.2s
Test Summary:     | Pass  Total     Time
Albedo composites |  108    108  2m44.8s
```

these are all tests that aren't expensive once compiled but where the `model` changes some O(10) testing different model components -- as far as I understand this once `model` is of another type (which it is when any of its components is different) everything that has `model` in its arguments has to be recompiled. So maybe we need a PR that goes after `model` arguments and puts in function barriers e.g. what I've been trying to do in the parameterizations (but now looking at it horribly failed ...)

```julia
# function barrier, needs to be recompiled for every change to model
parameterization!(ij, diagn, progn, shortwave::TransparentShortwave, model) =
    shortwave_radiation!(ij, diagn, shortwave)

# does not depend on other things in model so no recompilation needed
function shortwave_radiation!(ij, diagn, shortwave)
    ...
end
```

milankl:
Asked this in the Julia slack

Compiler question to avoid unnecessary recompilation: Say I have a type Model with many parametric fields but functions that will only act on some of them e.g.

```julia
struct Model{A, B, C, D, E}
   a::A
   b::B
   c::C
   d::D
   e::E
end

model = Model(1, 1.2, "a", :a, (1,2))

# version 1: don't unpack, compute directly
f(model) = model.a^2

# version 2: function barrier to unpack, compute in h
g(model) = h(model.a)
h(x) = x^2
```

Now assume the `^2` operation is expensive to compile. If I understand things correctly in version 1 without a function barrier any change to `model` (not even `model.a`) would require `f` to be recompiled, e.g. say I create `model = Model(1, 1.2, "a", :a, (1, 2.0))` so the` typeof(model)` changed triggering recompilation of `f` although `f(model)` is exactly the same before and after.

In contrast, if I created a function barrier unpacking model then the expensive-to-compile "core" of `g(model)` (the `^2`) would be in another function that doesn't require recompilation because `h(::A)` has already been compiled. Am I correct here? So is it a reasonable to introduce a bunch of function barriers that unpack in these situations to avoid unnecessary recompilation?

navidcy (approved):

milankl:
superseded by #957

PR #898, opened 2025-11-06
Soil layers defined in land instead of SpectralGrid

maximilian-gelbrecht:
This is based on the `mk/ijphysics` branch and takes the `nlayers_soil` out of the `SpectralGrid` and instead makes it solely a property of the land model. 

@bgroenks96 you might like this.


bgroenks96:
Can confirm, I do in fact like this!

maximilian-gelbrecht:
The only somewhat annoying thing about this is that we have to rethink our output a little bit, because so far we were constructing the work arrays for interpolation already when the outputter was constructed. But now we don't have the amount of soil layer available at that point anymore. But we can probably move that to the `inititialize!`. 

In a separate PR we could also connect the new variable system to the output, i.e. getting descriptions and units of the variables from the `Variable` system and not hard coding them anymore in the outputs.

milankl:
> In a separate PR we could also connect the new variable system to the output, i.e. getting descriptions and units of the variables from the `Variable` system and not hard coding them anymore in the outputs.

Yes one could but I wouldn't actually give this high priority. The reason is that we allow for output variable structs to have a `transform` (not a spectral transform, maybe there could be a better name...) -- a unit conversion function. E.g. the temperature outputs have `transform = x -> x - 273.15` so the units in that case are wrong. If someone wants sea surface temperature in Kelvin instead it would be simply

```julia
sst_output = SpeedyWeather.SeaSurfaceTemperatureOutput(unit="K", transform=x->x)
add!(model, sst_output)
```

The pressure output uses `transform = x -> exp(x)/100` so even the description of "Logarithm of surface pressure" is wrong. So overall I think there's value to keep them separate ?

Are you imagining something like

```julia
shortwave_radiation = MyShortwaveRadiation(spectral_grid, kwargs...)
model = PrimitiveWetModel(spectral_grid; shortwave_radiation)
add!(model.output, variables(shortwave_radiation))
```

so that one can reuse the `variables` call to specify all those for output? This also seems complicated because some variables might be spectral (not for output) some gridded etc ...

maximilian-gelbrecht:
Ok, yeah, I can see that. One could probably still find a nice way to combine both, but that's not for now.

SpeedyWeather/src/output/netcdf_output.jl
```diff
@@ -85,7 +85,7 @@ $(TYPEDFIELDS)"""
     # SCRATCH FIELDS TO INTERPOLATE ONTO
     const field2D::Field2D
     const field3D::Field3D
-    const field3Dland::Field3D
+    field3Dland::Field3D
```
maximilian-gelbrecht: that's a change I am not super happy about, but I am also not quite sure how else we'd do it without a major revision of the output system.

milankl: why is this needed? Do we need 3D land variables at different resolutions/number of vertical layers?

maximilian-gelbrecht: When the number of soil layers is only defined in the land model itself, then it's not known at the time of the creation of the `NetCDFOutput` object, so the `field3Dland` can only be allocated at `initialize!`.

SpeedyWeather/src/physics/land/land.jl
```diff
@@ -30,17 +34,17 @@ end
...
-    @component temperature::T = LandBucketTemperature(spectral_grid)
-    @component soil_moisture::SM = LandBucketMoisture(spectral_grid)
-    @component snow::SN = SnowModel(spectral_grid)
-    @component vegetation::V = VegetationClimatology(spectral_grid)
+    @component geometry::G = LandGeometry(spectral_grid, nlayers = DEFAULT_NLAYERS_SOIL)
+    @component thermodynamics::TD = LandThermodynamics(spectral_grid, geometry)
+    @component temperature::T = LandBucketTemperature(spectral_grid, geometry)
+    @component soil_moisture::SM = LandBucketMoisture(spectral_grid, geometry)
+    @component snow::SN = SnowModel(spectral_grid, geometry)
+    @component vegetation::V = VegetationClimatology(spectral_grid, geometry)
```
maximilian-gelbrecht, line 43: I made `geometry` a part of all the land component signatures because I thought that's the most general thing to do, so that in principle all constructors can see the number of soil layers. Currently most them don't actually need it though, so one could also revert that change.

maximilian-gelbrecht:
Tests seem to pass now. This is ready for review.

SpeedyWeather/src/models/primitive_wet.jl
```diff
@@ -207,6 +207,10 @@ function initialize!(model::PrimitiveWet; time::DateTime = DEFAULT_DATE)
     return Simulation(prognostic_variables, diagnostic_variables, model)
 end
 
+"""$(TYPEDSIGNATURES)
+Extract the number of soil layers from the model."""
+@inline get_soil_layers(model::PrimitiveWetModel) = get_soil_layers(model.land)
```
milankl, line 212: Would we need a `get_soil_layers(land::Nothing) = 0` fallback?

SpeedyWeather/src/output/netcdf_output.jl
milankl: Other option would be to have a `NetCDFOutput(::SpectralGrid, ::LandGeometry)` constructor that's called by default? And a warning being thrown if a manual `NetCDFOutput` is created with `nlayers_soil` being inconsistent with `model.land.geometry.nlayers`?

milankl: Or another option would be to continue using `DEFAULT_NLAYERS_SOIL` and only output those top layers in case someone runs a simulation with more layers, printing a warning once.

SpeedyWeather/src/physics/land/land.jl
milankl: No I like this but maybe this could be optional?

SpeedyWeather/src/physics/land/soil_moisture.jl
```diff
@@ -34,9 +34,10 @@ end
 # Adapt.adapt_structure(to, soil::SeasonalSoilMoisture) = adapt(to, ManualSeasonalSoilMoisture(soil.monthly_soil_moisture))
 
 # generator function
-function SeasonalSoilMoisture(SG::SpectralGrid; kwargs...)
-    (; NF, GridVariable4D, grid, nlayers_soil) = SG
-    monthly_soil_moisture = zeros(GridVariable4D, grid, nlayers_soil, 12)
+function SeasonalSoilMoisture(SG::SpectralGrid, geometry::LandGeometry; kwargs...)
```
milankl: optional by also retaining the previous method with `::LandGeometry` just unpacking `nlayers`?

milankl (commented):
Great! Approve of most things though

- I think we can add a method `NetCDFOutput(::SpectralGrid, ::AbstractLand)` or similar that is called by default using the nlayers in there but then use a default nlayers otherwise and output the top nlayers in case the land model has more?
- Also make the `LandComponent(::SpectralGrid, ::LandGeometry)` optional but default so that the old constructor still works and a user wouldn't need to learn that for `LandComponent` they also have to first create a land geometry and pass that on as well

That way we have fewer breaking changes?

milankl:
You're not making any changes to the submodules here and the tests of those are now independent of SpeedyWeather so saving some CI time by skipping their tests

SpeedyWeather/src/output/netcdf_output.jl
maximilian-gelbrecht: Yeah that's a good point. Previously I also thought that some of our models actually use 1 soil layer, but we use 2 layers everywhere, so it's fair to just assume two as a default and issue a warning when there's a mismatch, as you write. Just got to document it well somewhere.

SpeedyWeather/src/physics/land/land.jl
maximilian-gelbrecht: Yeah fair, we can just declare them with a `nothing` default argument in most cases.

SpeedyWeather/src/models/primitive_wet.jl
maximilian-gelbrecht: good catch

SpeedyWeather/src/physics/land/land.jl
maximilian-gelbrecht: The only component that actually uses the `LandGeometry` currently at construction is `SeasonalSoilMoisture` btw.

maximilian-gelbrecht: Ok done

SpeedyWeather/src/output/netcdf_output.jl
maximilian-gelbrecht: Done

maximilian-gelbrecht:
Ok, all tests passing, everything addressed. If there's nothing else, I'll merge this later.

SpeedyWeather/test/physics/land.jl
```diff
@@ -4,7 +4,8 @@
     for Temperature in (SeasonalLandTemperature, ConstantLandTemperature, LandBucketTemperature)
         for Model in (PrimitiveDryModel, PrimitiveWetModel)
 
-            temperature = Temperature(spectral_grid)
+            geometry = LandGeometry(spectral_grid)
+            temperature = Temperature(spectral_grid, geometry)
```
milankl, line 8: Should we also test whether the `Temperature(spectral_grid)` constructor works here? Just as an additional line? Even though the other one is being used?

SpeedyWeather/test/physics/land.jl
```diff
@@ -25,10 +26,11 @@ end
...
 
-                    temperature = Temperature(spectral_grid)
-                    soil_moisture = SoilMoisture(spectral_grid)
-                    vegetation = Vegetation(spectral_grid)
-                    land = LandModel(spectral_grid; temperature, soil_moisture, vegetation)
+                    geometry = LandGeometry(spectral_grid)
+                    temperature = Temperature(spectral_grid, geometry)
+                    soil_moisture = SoilMoisture(spectral_grid, geometry)
+                    vegetation = Vegetation(spectral_grid, geometry)
+                    land = LandModel(spectral_grid; geometry, temperature, soil_moisture, vegetation)
```
milankl, line 39: Same here, test the constructors without `geometry` provided?

milankl (approved):

SpeedyWeather/test/physics/land.jl
maximilian-gelbrecht: Yeah, fair.

maximilian-gelbrecht:
@bgroenks96 This is merged now. We'll probably still wait with a new release a bit, but for the next revision of the Speedy-Terrarium coupling we can use this.

milankl:
I was actually wondering when we should do the next release. Also for the CPU a good amount of things have been implemented and fixed.

bgroenks96:
Now?

PR #899, opened 2025-11-07
Remove unnecessary double imports of architecture

gregordecristoforo:
Another pedantic PR. I noticed some double imports while working on the AMD implementation. Feel free to ignore it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## Unreleased
 
 - Remove double imports of `architecture` in SpeedyWeatherInternals/ext/ [#899](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/899)
+- Documentation typos [#894](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/894)
+- Seasonal ocean and land climatology models GPU-ready [#892](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/892)
 - Zenith constructors [#889](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/889)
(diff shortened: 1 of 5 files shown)


maximilian-gelbrecht:
Thanks! Let us know how you are doing with the AMD implementation if you have something to report! Neither me nor Milan have a AMD GPU available, so it was a really low priority so far. But with KernelAbstractions, it should be possible. 

As you can see in the PRs, we currently do a lot of work on the GPU version and it shouldn't take that long anymore to have a first version of the primitive model working as well. This also means that a lot of things are currently still subjected to change though.

gregordecristoforo:
Thanks @maximilian-gelbrecht, I am currently working on a [`SpeedyWeatherInternalsROCmExt` extention](https://github.com/gregordecristoforo/SpeedyWeather.jl/blob/AMDGPU/SpeedyWeatherInternals/ext/SpeedyWeatherInternalsROCmExt.jl) since I have access to the LUMI supercomputer with MI250X GPUs.

I am still relatively new to Julia as this is the first established Julia project I would like to contribute to, so I am currently still dealing with some (probably quite trivial) errors. I'll will reach once I have more concrete questions.

maximilian-gelbrecht:
KernelAbstractions takes care of most things luckily, so the crucial point for other architecutres/ devices to work with Speedy is the Fourier transform. Aside from the `SpeedyWeatherInternalsROCmExt` you would need to implement a version of the [`SpeedyTransformsCUDAExt`](https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/main/SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl) that defines the bindings to `CUFFT`.

Issue #870, opened 2025-09-26
GPU performance of land and sea ice model


milankl:
I just took the `ThermodynamicSeaIce` model (it's relatively simple and 2D) and compared its performance on 

- GPU: NVIDIA A100-PCIE-40GB (sm_80, 35.000 GiB / 39.675 GiB available)
- CPU: 64 × AMD EPYC 7282 16-Core Processor (single core only)

at different resolutions each OctahedralGaussianGrid at dealiasing=2 and land sea mask 0 everywhere (all ocean) thought to document this

| Resolution | km | nlat_half | grid points | dt [s] | 1 step CPU | 1 step GPU | SYPD CPU | SYPD GPU | faster |
| - | - | - | - | - | - | - | - | - | - |
| T31     | 400 | 24 | 3168 | 2400 | 13μs | 24μs | 505000 | 274000 | 0.54x |
| T511   | 30    | 384 | 603648 | 150 | 1.8ms | 31.5μs | 230 | 13000 | 59x |
| T1279 | 12    | 960 | 3720960 | 60 | 11.5ms | 78μs | 14.2 | 2100 | 150x |

@maximilian-gelbrecht @bgroenks96

milankl:
And similarly the land model (soil moisture and temperature) setting land sea mask to land everywhere

| Resolution | km | nlat_half | grid points | dt [s] | 1 step CPU | 1 step GPU | SYPD CPU | SYPD GPU | faster |
| - | - | - | - | - | - | - | - | - | - |
| T31     | 400 | 24 | 3168 | 2400 | 27μs | 60μs | 243000 | 110000 | 0.45x |
| T511   | 30    | 384 | 603648 | 150 | 5.5ms | 92μs | 75 | 4513 | 60x |
| T1279 | 12    | 960 | 3720960 | 60 | 34ms | 300μs | 4.8 | 550 | 113x |

maximilian-gelbrecht:
This is great already. Especially because we still have a lot of potential to improve on GPU. I also looked at it in Nvidia Nsight some weeks ago, and we could still get faster then what we do right now 🚀

milankl:
Ran this with

```julia
using SpeedyWeather, CUDA

spectral_grid = SpectralGrid(trunc=31, architecture=GPU())
model = PrimitiveWetModel(spectral_grid)
(; ocean, land, sea_ice) = model

progn = PrognosticVariables(spectral_grid)
[...]

PR #901, opened 2025-11-07
AMDGPU extension

gregordecristoforo:
I would like to provide a short update on my work on porting SpeedyWeather.jl to AMD GPUs. 

This slightly modified version of the [model from the GPU chapter of the documentation](https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/architectures_gpu/) runs in this branch on LUMI's AMD MI250X GPUs using the `AMDGPU` package:
```julia
using SpeedyWeather, AMDGPU 
architecture = SpeedyWeather.GPU()
spectral_grid = SpectralGrid(trunc=41, nlayers=1, architecture=architecture)           

model = BarotropicModel(spectral_grid=spectral_grid)
AMDGPU.@allowscalar simulation = initialize!(model)
```
Please ignore the diffs in the Project.toml file for now, I'll clean that up before sending a normal PR and also add a change-log entry.

So far, this code only utilizes KernelAbstractions. As @maximilian-gelbrecht pointed out in  #899 an AMD specific version of [SpeedyTransformsCUDAExt](https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/main/SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl) might be necessary. I haven't had time to look at this yet, but judging from the time it takes to run this simple model the code might fall back to the CPU for the FFTs (my limited understanding might be wrong here, I still need to get a better understanding of the code base).

I'll continue working on this next week, any comments would be appreciated :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - AMDGPU extension [#901](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/901)
+- Don't adapt rings in RingGrids on GPU [#920](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/920)
 - README updates [#915](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/915#pullrequestreview-3469109183)
 - Fixes some formatting/LaTeX rendering in Docs [#916](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/916)
 
--- a/RingGrids/src/field.jl
+++ b/RingGrids/src/field.jl
@@ -520,11 +520,10 @@ Architectures.architecture(field::AbstractField) = architecture(field.grid)
 
 function Architectures.on_architecture(arch, field::Field{T, N, ArrayType, Grid}) where {T, N, ArrayType, Grid} 
     adapted_data = on_architecture(arch, field.data)
-    if ismatching(field.grid, typeof(adapted_data))
-        return Field(adapted_data, on_architecture(arch, field.grid))
-    else # if not matching, create new grid with other architecture
-        #@warn "Adapting field to new architecture with $(typeof(adapted_data))"
-        return Field(adapted_data, Grid(field.grid, architecture(typeof(adapted_data))))
-    end
+
+    # if not matching, create new grid with other architecture
+    arch = ismatching(field.grid, typeof(adapted_data)) ? arch : architecture(typeof(adapted_data))
+    
+    return Field(adapted_data, on_architecture(arch, field.grid))
 end 
     
--- a/RingGrids/src/grid.jl
+++ b/RingGrids/src/grid.jl
@@ -66,8 +66,7 @@ end
 
 # change the architecture of a grid, keep all other fields 
 function (::Type{Grid})(grid::Grid, architecture::AbstractArchitecture) where {Grid<:AbstractGrid}
-    Grid_ = nonparametric_type(Grid)                # strip away parameters of type, obtain from arguments
-    return Grid_(grid.nlat_half, architecture, adapt(array_type(architecture), grid.rings), adapt(array_type(architecture), grid.whichring))
+    return on_architecture(architecture, grid)
 end
 
 function (::Type{Grid})(
@@ -172,7 +171,6 @@ and then each grid point per ring. To be used like
 
 Accesses precomputed `grid.rings`."""
 @inline eachring(grid::AbstractGrid) = grid.rings
-@inline eachring(grid::AbstractGrid{<:GPU}) = Vector(grid.rings) # on GPU transfer indices back to CPU first 
 
 """$(TYPEDSIGNATURES)
 Computes the ring indices `i0:i1` for start and end of every longitudinal point
@@ -274,7 +272,20 @@ Architectures.ismatching(grid::AbstractGrid, array::AbstractArray) = ismatching(
 
 Architectures.architecture(grid::AbstractGrid) = grid.architecture
 
+# only transfer the whichring on GPU
 function Architectures.on_architecture(arch::AbstractArchitecture, grid::Grid) where Grid<:AbstractGrid 
     Grid_ = nonparametric_type(Grid)
-    return Grid_(grid, arch)
+    return Grid_(grid.nlat_half, 
+                arch, 
+                grid.rings, 
+                on_architecture(arch, grid.whichring))
 end 
+
+# don't adapt the CPU-only rings 
+function Adapt.adapt_structure(to, grid::AbstractGrid)
+    Grid_ = nonparametric_type(typeof(grid))
+    return Grid_(grid.nlat_half, 
+                Adapt.adapt_structure(to, grid.architecture), 
+                nothing, 
+                Adapt.adapt_structure(to, grid.whichring))
+end
--- a/RingGrids/src/grids/full_clenshaw.jl
+++ b/RingGrids/src/grids/full_clenshaw.jl
@@ -50,5 +50,3 @@ get_lond(::Type{<:FullClenshawGrid}, nlat_half::Integer) = get_lond(FullGaussian
 
 # QUADRATURE
 get_quadrature_weights(::Type{<:FullClenshawGrid}, nlat_half::Integer) = clenshaw_curtis_weights(nlat_half)
-
-Adapt.@adapt_structure FullClenshawGrid
--- a/RingGrids/src/grids/full_gaussian.jl
+++ b/RingGrids/src/grids/full_gaussian.jl
@@ -57,5 +57,3 @@ end
 
 # QUADRATURE
 get_quadrature_weights(::Type{<:FullGaussianGrid}, nlat_half::Integer) = gaussian_weights(nlat_half)
-
-Adapt.@adapt_structure FullGaussianGrid
--- a/RingGrids/src/grids/full_healpix.jl
+++ b/RingGrids/src/grids/full_healpix.jl
@@ -41,5 +41,3 @@ get_lond(::Type{<:FullHEALPixGrid}, nlat_half::Integer) = get_lond(FullGaussianG
 # QUADRATURE (use weights from reduced grids though!)
 get_quadrature_weights(::Type{<:FullHEALPixGrid}, nlat_half::Integer) =
     equal_area_weights(HEALPixGrid, nlat_half)
-
-Adapt.@adapt_structure FullHEALPixGrid
--- a/RingGrids/src/grids/full_octahealpix.jl
+++ b/RingGrids/src/grids/full_octahealpix.jl
@@ -41,5 +41,3 @@ get_lond(::Type{<:FullOctaHEALPixGrid}, nlat_half::Integer) = get_lond(FullGauss
 # QUADRATURE (use weights from reduced grids though!)
 get_quadrature_weights(::Type{<:FullOctaHEALPixGrid}, nlat_half::Integer) =
     equal_area_weights(OctaHEALPixGrid, nlat_half)
-
-Adapt.@adapt_structure FullOctaHEALPixGrid
--- a/RingGrids/src/grids/healpix.jl
+++ b/RingGrids/src/grids/healpix.jl
@@ -160,5 +160,3 @@ function each_index_in_ring!(   rings::AbstractVector,
         rings[j] = index_1st:index_end              # turn into UnitRange
     end
 end
-
-Adapt.@adapt_structure HEALPixGrid
--- a/RingGrids/src/grids/octahedral_clenshaw.jl
+++ b/RingGrids/src/grids/octahedral_clenshaw.jl
@@ -113,8 +113,6 @@ function each_index_in_ring!(   rings,
     end
 end
 
-Adapt.@adapt_structure OctahedralClenshawGrid
-
 ## CONVERSION
 # Base.Matrix(G::OctahedralClenshawGrid{T}; kwargs...) where T =
 #     Matrix!(zeros(T, matrix_size(G)...), G; kwargs...)
--- a/RingGrids/src/grids/octahedral_gaussian.jl
+++ b/RingGrids/src/grids/octahedral_gaussian.jl
@@ -108,5 +108,3 @@ function each_index_in_ring!(   rings,
         rings[j] = index_1st:index_end              # turn into UnitRange
     end
 end
-
-Adapt.@adapt_structure OctahedralGaussianGrid
--- a/RingGrids/src/grids/octaminimal_gaussian.jl
+++ b/RingGrids/src/grids/octaminimal_gaussian.jl
@@ -96,5 +96,3 @@ function each_index_in_ring!(   rings,
         rings[j] = index_1st:index_end              # turn into UnitRange
     end
 end
-
-Adapt.@adapt_structure OctaminimalGaussianGrid
--- a/RingGrids/src/interpolation.jl
+++ b/RingGrids/src/interpolation.jl
@@ -6,6 +6,7 @@ struct GridGeometry{
     Grid,
     VectorType,
     VectorIntType,
+    VectorRange,
 } <: AbstractGridGeometry
     grid::Grid                  # grid, e.g. FullGaussianGrid
 
@@ -18,8 +19,8 @@ struct GridGeometry{
     nlons::VectorIntType        # number of longitudinal points per ring
     lon_offsets::VectorType     # longitude offsets of first grid point per ring
 
-    # rings, CPU copy of grid.rings
-    rings::Vector{UnitRange{Int}} 
+    # rings, GPU/architecture copy (if needed) of grid.rings
+    rings::VectorRange
 end
 
 GridGeometry(field::AbstractField; kwargs...) = GridGeometry(field.grid; NF=eltype(field), kwargs...)
@@ -57,9 +58,10 @@ function GridGeometry(
     # vector type
     VectorType = array_type(architecture, NF, 1)
     VectorIntType = array_type(architecture, Int, 1)
+    device_rings = on_architecture(architecture, grid.rings)
 
-    return GridGeometry{typeof(grid), VectorType, VectorIntType}(
-        grid, nlat_half, nlat, npoints, londs, latd_poles, nlons, lon_offsets, Vector(grid.rings))
+    return GridGeometry{typeof(grid), VectorType, VectorIntType, typeof(device_rings)}(
+        grid, nlat_half, nlat, npoints, londs, latd_poles, nlons, lon_offsets, device_rings)
 end
 
 Base.show(io::IO,G::GridGeometry) = print(io,"GridGeometry for $(G.grid)")
@@ -252,8 +254,9 @@ function _interpolate!(
     architecture::AbstractArchitecture      
 )
     (; npoints_output, ij_as, ij_bs, ij_cs, ij_ds, Δabs, Δcds, Δys) = interpolator.locator
-    (; npoints, rings) = interpolator.geometry
-    
+    (; npoints) = interpolator.geometry
+    (; rings) = interpolator.geometry.grid # CPU version even on GPU
+
     # 1) Aout's length must match the interpolator
     # 2) input A must match the interpolator's geometry points (do not check grids for view support)
     @boundscheck size(Aout, 1) == length(ij_as) || throw(DimensionMismatchArray(Aout, ij_as))
@@ -553,7 +556,7 @@ function find_grid_indices!(I::AnvilInterpolator,       # update indices arrays
     (; js, ij_as, ij_bs, ij_cs, ij_ds ) = I.locator
     (; Δabs, Δcds ) = I.locator
     (; nlons, lon_offsets, nlat ) = I.geometry
-    (; rings ) = I.geometry.grid
+    (; rings ) = I.geometry # architecture version (GPU if needed)
     
     # Convert λs to the same type as lon_offsets if needed
     λs_converted = convert.(eltype(lon_offsets), λs)


maximilian-gelbrecht:
Looks great so far. 

I am surprised you don't get an error when running it because the default would be to just use FFTW on CPU for the Fourier transforms. 

In the SpeedyTransformsAMDGPUExt, you'd probably only need to adjust one or two lines of code, e.g.
https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/70a7f295cb26c7bbc59d5c560a959bcd6f6174ba/SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl#L13

and 

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/70a7f295cb26c7bbc59d5c560a959bcd6f6174ba/SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl#L30

to call the correct FFT plan. The rest of the extension should probably work fine if you just copy it. We have these functions defined there that execute the plans because CUFFT doesn't allow plans to be applied to views, but FFTW (CPU) and GenericFFT (CPU) does.

milankl:
Code to just execute the transforms

```julia
using SpeedyWeather, AMDGPU

spectral_grid = SpectralGrid(trunc=42, nlayer=8, architecture=GPU())

# create some data

# transforms in-place using a precomputed SpectralTransform (contains plans, scratch etc)
transform!(field, L, S)   # spectral to grid space
transform!(L, field, S)   # grid to spectral space

# these transforms should work
# allocates output and recomputes spectral transform
transform(field)   
transform(L)
```

gregordecristoforo:
Before running this code on AMD GPUs I tried to run it on an NVidia GPU and found that the code
```Julia
using SpeedyWeather, CUDA

spectral_grid = SpectralGrid(trunc=42, nlayers=8, architecture=GPU())

# create some data

# transforms in-place using a precomputed SpectralTransform (contains plans, scratch etc)
transform!(field, L, S)   # spectral to grid space
transform!(L, field, S)   # grid to spectral space

# these transforms should work
# allocates output and recomputes spectral transform
transform(field)   
```
throws the error:
```Julia
ERROR: MethodError: Cannot `convert` an object of type 
  LowerTriangularArray{Float32,1,CuArray{Float32,1,CUDA.DeviceMemory},Spectrum{GPU{CUDABackend},CuArray{UnitRange{Int64}, 1, CUDA.DeviceMemory},CuArray{Int64, 1, CUDA.DeviceMemory}}} to an object of type 
  LowerTriangularArray{Float32,1,CuArray{Float32,1,CUDA.DeviceMemory},Spectrum{CPU{KernelAbstractions.CPU},Vector{UnitRange{Int64}},Vector{Int64}}}
The function `convert` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  LowerTriangularArray{T, N, ArrayType, S}(::AbstractArray, ::S) where {T, N, ArrayType<:AbstractArray{T, N}, S<:LowerTriangularArrays.AbstractSpectrum}
   @ LowerTriangularArrays ~/git/SpeedyWeather.jl/LowerTriangularArrays/src/lower_triangular_array.jl:14
  convert(::Type{LowerTriangularArray{T1, N, ArrayTypeT1, S}}, ::LowerTriangularArray{T2, N, ArrayTypeT2, S}) where {T1, T2, N, S<:LowerTriangularArrays.AbstractSpectrum, ArrayTypeT1<:(AbstractArray{T1}), ArrayTypeT2<:(AbstractArray{T2})}
   @ LowerTriangularArrays ~/git/SpeedyWeather.jl/LowerTriangularArrays/src/lower_triangular_array.jl:652
  convert(::Type{T}, ::T) where T<:AbstractArray
   @ Base abstractarray.jl:16
  ...

```
is that to be expected or shall I create an Issue?

maximilian-gelbrecht:
Ah okay, I can reproduce it. It's only the last allocating `transform` that causes the error. We probably don't test those allocating ones enough, as they are not used anywhere in the model.

You don't have to open an issue, I am already on it and will merge a fix in the next hours.

maximilian-gelbrecht:
Okay fix is done and working, I just wait until the CI is finished and then merge it.

gregordecristoforo:
After commit [c7003fd](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/901/commits/c7003fd82342c264e40f837d321825ac79fd6b1e)  the following code runs on LUMI's MI250X AMD GPUs:

> Code to just execute the transforms
> 
> ```julia
> using SpeedyWeather, AMDGPU
> 
> spectral_grid = SpectralGrid(trunc=42, nlayers=8, architecture=GPU())
> S = SpectralTransform(spectral_grid)
> 
> # create some data
> field = zeros(Float32, spectral_grid.grid, 8)
> L = randn(ComplexF32, spectral_grid.spectrum, 8)
> 
> # transforms in-place using a precomputed SpectralTransform (contains plans, scratch etc)
> transform!(field, L, S)   # spectral to grid space
> transform!(L, field, S)   # grid to spectral space
> 
> # these transforms should work
> # allocates output and recomputes spectral transform
> transform(field)   
> transform(L)
> ```
Also the land and sea ice models of seem to work (If I understand it correctly). The code examples of https://github.com/SpeedyWeather/SpeedyWeather.jl/issues/870#issuecomment-3512272379 and https://github.com/SpeedyWeather/SpeedyWeather.jl/issues/870#issuecomment-3512331178 don't throw any errors on AMD GPUs. 

It turns out that using AbstractFFTs directly with ROCArrays chooses the correct FFT plans (see this [PR](https://github.com/JuliaGPU/AMDGPU.jl/pull/689) as an example).

maximilian-gelbrecht:
That's great! 

Could you just do a small benchmark running the full barotropic model, e.g. at `trunc=255` and `trunc=511` for a model days compared to a CPU execution? 

Btw with #903 merged, you also don't have to add the `AMDGPU.@allowscalar` anymore at initialization.

SpeedyTransforms/ext/SpeedyTransformsAMDGPUExt.jl
```diff
@@ -0,0 +1,95 @@
...
+    using SpeedyTransforms.RingGrids
+    using SpeedyTransforms.LowerTriangularArrays
+    
+    # Override FFT package deciding function
+    SpeedyTransforms.which_FFT_package(::Type{<:ROCArray{<:AbstractFloat}}) = AbstractFFTs
+
+    """$(TYPEDSIGNATURES)
+    Util function to generate FFT plans based on the array type of the fake Grid 
+    data provided. Uses indexing as we seemingly can't use views for FFT planning 
+    with CUFFT."""
```
maximilian-gelbrecht: Just a minor thing, but all the docstrings should probably not mention CUFFT/CUDA anymore

gregordecristoforo:
Thanks for the hint @maximilian-gelbrecht! The code runs significantly slower than on the CPU or my laptop GPU from Nvidia so I suspect there are still some issues with it. I am starting to profile the code now.

maximilian-gelbrecht:
For the barotropic model the only thing that matters for performance is actually the transforms, so you can restrict any profiling / debugging on that. It's very unlikely the performance regression comes from somewhere else.

gregordecristoforo:
Here are some initial results. For the example code from above (https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/901#issuecomment-3517311920 with `trunc=255`) the CPU version results in:
```Julia
julia> @benchmark transform!(field, L, S)   # spectral to grid space
BenchmarkTools.Trial: 142 samples with 1 evaluation per sample.
 Range (min … max):  34.675 ms …  40.545 ms  ┊ GC (min … max): 0.00% … 13.65%
 Time  (median):     35.003 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   35.446 ms ± 988.335 μs  ┊ GC (mean ± σ):  1.01% ±  2.29%

   ▁█▅▂                                                         
  ▆████▅▃▁▂▁▂▂▁▁▂▁▁▂▁▁▁▃▆▄▆▃▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂ ▂
  34.7 ms         Histogram: frequency by time         40.4 ms <

 Memory estimate: 4.83 MiB, allocs estimate: 2667.
```
Whereas the GPU version shows
```Julia
julia> @benchmark transform!(field, L, S)   # spectral to grid space

BenchmarkTools.Trial: 8 samples with 1 evaluation per sample.
 Range (min … max):   37.777 ms …    1.606 s  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     650.630 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   741.234 ms ± 654.045 ms  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █   ▁           ▁              ▁              ▁           ▁ ▁  
  █▁▁▁█▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁█▁█ ▁
  37.8 ms          Histogram: frequency by time          1.61 s <

 Memory estimate: 3.23 MiB, allocs estimate: 80584.

```
I have no Idea why there is such a huge spread for the GPU version. I'll look further into it tomorrow. Thanks for all the help!

maximilian-gelbrecht:
It has be either the Legendre transforms or the Fourier transform then. You can check that it really hits the right version of the Legendre transform, we have one kernel version for GPU and one "normal" version for CPU in `SpeedyTransforms/src/legendre_ka.jl` and `SpeedyTransforms/src/legendre.jl`. If it's not that, it's probably still something with the FFT plans.

milankl:
You can also benchmark Fourier vs Legendre independently, under the hood `transform!` just calls `_fourier!` then `_legendre!` or vice versa depending on grid -> spectral or spectral -> grid

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/bb3a3469f9111d39b8d3647106faf8ca907d34ec/SpeedyTransforms/src/spectral_transform.jl#L489-L516

gregordecristoforo:
I believe I found the underlying reason for the slow performance. If I am correct it has nothing to do with SpeedyWeather. The benchmarks shown above show a huge range for the time it takes to run this function. The lower end is similar to the benchmarks ran on the Nvidia GPU on my laptop, the upper end is roughly two orders of magnitude higher. 

I was able to reproduce this behavior with this minimal example:
```Julia
julia> using AMDGPU

julia> using AbstractFFTs

julia> x = ROCArray(rand(Float32, 1000, 2000));

julia> p = plan_rfft(x)
rocFFT real forward plan for 1000×2000 ROCArray of Float32

julia> @benchmark p * x
BenchmarkTools.Trial: 5230 samples with 1 evaluation per sample.
 Range (min … max):   14.408 μs … 5.647 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     528.425 μs             ┊ GC (median):    0.00%
 Time  (mean ± σ):   952.150 μs ± 1.003 ms  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▃▂▁▂▂▁▁▂▁▂▁▁ ▁▁▁▁▁▁▁ ▁▁▁▁▁ ▁▁ ▁ ▁ ▁▁ ▁▁▁ ▁▁ ▁           ▄▅ ▁
  █████████████████████████████████████████▇█████████████████ █
  14.4 μs      Histogram: log(frequency) by time      2.67 ms <

 Memory estimate: 320 bytes, allocs estimate: 12.
 
 julia> @which p * x
*(p::AMDGPU.rocFFT.rROCFFTPlan{T, true, false, N}, x::ROCArray{T, N}) where {T<:Union{Float32, Float64}, N}
     @ AMDGPU.rocFFT /projappl/project_462000008/decristoforo/.julia/packages/AMDGPU/TqRG0/src/fft/fft.jl:315

```
As you can see, we have a similar spread here when using the FFT plan provided by `AMDGPU`. 

LUMI uses ROCm 6.0.3 which is almost two years old by now which might be an explanation for this performance. 

Do you have any suggestions? Shall I focus on implementing further tests and benchmarks (even with this performance)?Also, I will reach out to our contact person from AMD who is in contact with the LUMI team.

Thanks for all the feedback so far, even if it looks quite disappointing. 

Edit: This huge range in execution times strangely does not occur in a fresh Julia session.

milankl:
@luraess is this something you can reproduce on your Radeon? (if you have a minute?) Wondering whether there is something about AMDGPU.jl or the ROCm version or the combination of both on LUMI?

luraess:
> @luraess is this something you can reproduce on your Radeon? (if you have a minute?) Wondering whether there is something about AMDGPU.jl or the ROCm version or the combination of both on LUMI?

I can take a look. Are you referring to https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/901#issuecomment-3522781329, right?

gregordecristoforo:
I have to leave work now but I will test this more thoroughly tomorrow. This problem seems to occur after some time, I wonder whether this is due to the GC not working properly.

milankl:
The `p * x` is allocating even though algorithmically it doesn't need to be. Also on CUDA we haven't found a way yet to actually write out the FFT result directly into a preallocated array

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/96854a9e2302e3a551d401c2b9ac2c5a8f0602dc/SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl#L65-L67

A similar issue is there with FFTW on the CPU because our output needs to be strided

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/96854a9e2302e3a551d401c2b9ac2c5a8f0602dc/SpeedyTransforms/src/fourier.jl#L37-L38

luraess:
I gave https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/901#issuecomment-3522781329 a try on a 7900XTX GPU with ROCm 7.1:

```
julia> @bs p * x
Chairmarks.Benchmark: 1323 samples with 1 evaluation.
 Range (min … max):   8.590 μs … 470.926 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     16.210 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   70.397 μs ± 100.864 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▇▄▄                             ▁▂▂▂▂  ▁                     
  ████▆▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▆██████▇██▇▇▇▇▆▆▇▅▅▆▄▅▄▁▅▁▄▅ █
  8.59 μs       Histogram: log(frequency) by time       357 μs <

 Memory estimate: 320.0 bytes, allocs estimate: 12.

julia> AMDGPU.versioninfo()
[ Info: AMDGPU versioninfo
┌───────────┬──────────────────┬───────────┬──────────────────────────────────────────────────────────────────────────
│ Available │ Name             │ Version   │ Path                                                                    ⋯
├───────────┼──────────────────┼───────────┼──────────────────────────────────────────────────────────────────────────
│     +     │ LLD              │ -         │ /opt/rocm-7.1.0/lib/llvm/bin/ld.lld                                     ⋯
│     +     │ Device Libraries │ -         │ /home/lraess/.julia/artifacts/b46ab46ef568406312e5f500efb677511199c2f9/ ⋯
│     +     │ HIP              │ 7.1.25424 │ /opt/rocm-7.1.0/lib/libamdhip64.so                                      ⋯
│     +     │ rocBLAS          │ 5.1.0     │ /opt/rocm-7.1.0/lib/librocblas.so                                       ⋯
│     +     │ rocSOLVER        │ 3.31.0    │ /opt/rocm-7.1.0/lib/librocsolver.so                                     ⋯
│     +     │ rocSPARSE        │ 4.1.0     │ /opt/rocm-7.1.0/lib/librocsparse.so                                     ⋯
│     +     │ rocRAND          │ 2.10.5    │ /opt/rocm-7.1.0/lib/librocrand.so                                       ⋯
│     +     │ rocFFT           │ 1.0.35    │ /opt/rocm-7.1.0/lib/librocfft.so                                        ⋯
│     +     │ MIOpen           │ 3.5.1     │ /opt/rocm-7.1.0/lib/libMIOpen.so                                        ⋯
└───────────┴──────────────────┴───────────┴──────────────────────────────────────────────────────────────────────────
                                                                                                      1 column omitted

[ Info: AMDGPU devices
┌────┬────────────────────┬──────────┬───────────┬────────────┬───────────────┐
│ Id │               Name │ GCN arch │ Wavefront │     Memory │ Shared Memory │
├────┼────────────────────┼──────────┼───────────┼────────────┼───────────────┤
│  1 │ Radeon RX 7900 XTX │  gfx1100 │        32 │ 23.984 GiB │    64.000 KiB │
└────┴────────────────────┴──────────┴───────────┴────────────┴───────────────┘
```
Note that I had to switch to using Chairmarks.jl (actually https://github.com/astrozot/PrettyChairmarks.jl) for benchmarking as for me, BenchmarkTools would hang in the process and just consume CPU time. I can also report that the timing reported by the benchmarks is consistent when repeating the exercise over and over.

gregordecristoforo:
Thanks @luraess! It is promising to see that your upper end is an order of magnitude lower than on LUMI.

May I ask you to run the following code on your 7900XTX GPU with ROCm 7.1:

```Julia
using SpeedyWeather, AMDGPU
architecture = GPU()
spectral_grid = SpectralGrid(trunc=255, nlayers=1, architecture=architecture)
model = BarotropicModel(spectral_grid=spectral_grid)
@time run!(simulation, period=Day(10))
```
On LUMI's MI250X GPU with ROCm 6.0.3 the simulation takes 362 seconds. Do you observe similar numbers?

The equivalent code runs on my GeForce RTX 3050 on my laptop for 79 seconds which is roughly 4.5x faster.

Edit: @maximilian-gelbrecht  For `trunc=511` I receive around 22 min for the MI250X and 13.5 min for the GeForce RTX 3050 (1.63x faster).

luraess:
I am getting
```
julia> architecture = SpeedyWeather.GPU()
ERROR: MethodError: no method matching GPU()
The type `GPU` exists, but no method is defined for this combination of argument types when trying to construct it.

Closest candidates are:
  GPU(::D) where D
   @ SpeedyWeatherInternals ~/.julia/dev/SpeedyWeather/SpeedyWeatherInternals/src/Architectures/Architectures.jl:48

Stacktrace:
 [1] top-level scope
   @ REPL[4]:1
```
while being on https://github.com/gregordecristoforo/SpeedyWeather.jl/tree/AMDGPU

milankl:
It's just `architecture = GPU()`, it's exported, but also doesn't live in SpeedyWeather but in SpeedyWeatherInternals

gregordecristoforo:
Thanks @milankl, I updated the code example above, sorry about that!

luraess:
Are the timings you report in https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/901#issuecomment-3526956501 for `trunc=41` or `trunc255`?

On the Radeon, I am getting 78 seconds for `trunc=41`:
```
julia> @time run!(simulation, period=Day(10))
Weather is speedy: 100%|████████████████████████████████████████████████████████| Time: 0:01:13 (32.00 years/day)
 78.096737 seconds (20.05 M allocations: 723.401 MiB, 0.11% gc time, 1.60% compilation time: <1% of which was recompilation)
```

gregordecristoforo:
Thanks @luraess! The times in [my previous comment](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/901#issuecomment-3526956501) are reported for `trunc=255`. It looks like there is still a lot of room for performance improvements for the AMDGPU port. I will look into potential funding sources so I can spend more time on this part after I move from Oxford back to UiT in two weeks.

Tomorrow I'll have a look at the .buildkite pipeline to run the tests on the AMDGPU queue and investigate why
```Julia
using SpeedyWeather, AMDGPU
architecture = SpeedyWeather.GPU()
```
doesn't work yet (I am still learning the Julia packaging conventions). Once these two things are fixed I hope the PR is ready to be reviewed/merged as an experimental AMD compatible version. Thanks again for all of your help!

docs/src/architectures_gpu.md
```diff
@@ -3,10 +3,10 @@
...
+Some of SpeedyWeather.jl already supports GPU acceleration, e.g. the barotropic model. Our development focuses on CUDA GPUs, but other architectures are thinkable in the future as well, as our approach relies on the device agnostic `KernelAbstractions.jl`. An experimental port to AMD GPUs using the `AMDGPU` package is available but AMD-specific performance optimizations are not implemented yet. The SpeedyWeather.jl submodule `Architectures` encodes all the information of the device we run our models on. In order to initialize a model on a GPU, we need to load the `CUDA` or `AMDGPU` package and pass the architecture to the model constructor. For example, to initialize a barotropic model on a GPU, we can do the following:  
 
 ```julia
-using SpeedyWeather, CUDA 
+using SpeedyWeather, CUDA/AMDGPU
```
maximilian-gelbrecht: The example in the docs should be working code, ready to be copy-paste, please use  something like `# AMDGPU` to indicate that users could also import `AMDGPU` instead and not ` / `

test/GPU/AMDGPU/architecture.jl
maximilian-gelbrecht, line 1: I find it a lot of redundant code to just copy all of the GPU tests here. 

Actually the beauty of how it's written currently is that the code is actually completely identical, the only difference is which library imported (so `using CUDA` vs `using AMDGPU`). The only tests that differ are the `architecture` tests below that check if it's really a `ROCArray`. 

So I'd say, a separate `test/AMDGPU/runtests.jl` and `test/AMDGPU/architecture.jl` is fine, but otherwise the AMDGPU tests should just use the exact same scripts as the CUDA tests and don't duplicate them. Then we don't have to maintain more code. 

Maybe we could reorganise the folder structure a bit for this to make this a bit more clear. So one folder `test/CUDA`, one folder `test/AMDGPU` and one folder `test/GPU` with all the general GPU tests.

luraess:
Thanks for the clarification @gregordecristoforo . Something seems to still be pretty off in AMDGPU (possibly some issues with rocFFT) that the timings are so much different wrt CUDA. Else, both are pretty much on par when it comes to more basic operations or saxpy-like compute kernels.

Wrt testing, only minor adaptions are needed to the Buildkite yaml file in order to also enable AMD GPU runners.

The suggestions raised by @maximilian-gelbrecht about redundancy in tests is a valid one and one could indeed have tests that are just enough agnostic wrt backend in order to avoid duplication but also not loose verbosity which should be there for testing.

For both points above regarding testing, you may want to have a look at https://github.com/PTsolvers/Chmy.jl/tree/main/.buildkite and https://github.com/PTsolvers/Chmy.jl/tree/main/test to get inspiration.

docs/src/architectures_gpu.md
gregordecristoforo: Resolved in [47161cd](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/901/commits/47161cd9686ea8adc44bf78d758f6341f4721e0e)

test/GPU/AMDGPU/architecture.jl
gregordecristoforo: Good point! I refactored the `test/GPU` directory accordingly. The `SpeedyWeather.jl/test/GPU/runtests.jl` script should now identify which GPU backend is available and choose the according one. I also left `runtests.jl` in the `CUDA`, `AMDGPU` and `MetalGPU` directory in case someone wants to run them individually when more tests are added.

I tested it on LUMI's AMD GPUs and on my Nvidia GPU. I don't have access to a Metal GPU atm. What do you think of these changes @maximilian-gelbrecht ?

SpeedyTransforms/src/SpeedyTransforms.jl
```diff
@@ -25,7 +25,9 @@ import LowerTriangularArrays: lm2i, get_lm_range, get_2lm_range
 # TRANSFORM
 export  SpectralTransform,
         transform!,
-        transform
+        transform,
+	_fourier!,
```
maximilian-gelbrecht: please don't export those

test/GPU/CUDA/spectral_transform.jl
maximilian-gelbrecht, line 1: the transforms are the most crucial part of our code in many ways, I think it's fine to use this test for all GPU architectures as well

SpeedyTransforms/src/SpeedyTransforms.jl
gregordecristoforo: Ah, that was just for testing, thanks for spotting it!

test/GPU/CUDA/spectral_transform.jl
gregordecristoforo: `test/GPU/CUDA/spectral_transoform.jl` used `CUDA.cu` so I put it into the CUDA dir. But I found a way of making it agnostic to the GPU backend by using `on_architecture(gpu_arch, f_north_cpu)`. 

It runs now on AMDGPU and CUDA. I can't test Metal unfortunately.

Project.toml
```diff
@@ -23,6 +23,7 @@ KernelAbstractions = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 LowerTriangularArrays = "ed20b33f-a660-45d0-a353-9c8c20bb5814"
 NCDatasets = "85f8d34a-cbdd-5861-8df4-14fed0d494ab"
+PrettyChairmarks = "aafa11c5-44f9-44a1-b829-427e6ce1ffc2"
```
maximilian-gelbrecht: remove that dependency again please

maximilian-gelbrecht (approved):
Great, just remove that one extra dependency again, and then we can merge it.

gregordecristoforo:
Done! I still want to set up the .buildkite tests but that can be done in a separate PR if you prefer.

maximilian-gelbrecht:
Both fine with me. What do you actually mean with buildkite tests though? Could we run the CI on LUMI as well?

gregordecristoforo:
The idea is to run all AMDGPU tests for SpeedyWeather on @luraess' machine which has a 7800XT GPU. Ludovic told me to look at https://github.com/PTsolvers/Chmy.jl/tree/main/.buildkite as an example. I look at it now.

maximilian-gelbrecht:
I don't think we can run our GPU CI on Ludovic's machines. 

GPU CI time is a bit hard to come by, currently we just use the one from JuliaGPU, but that's limited in the amount of time we have (not explicitly but implicitly). Theoretically I think they also have AMDGPU and Metal runners, but I am not sure we really need to use them at the moment, just to save scarce recourses.

maximilian-gelbrecht:
Oh I just heard, he did. So then, nevermind my previous comment :)

test/GPU/CUDA/spectral_transform.jl
milankl: Oceananigans has Metal CI now

https://github.com/CliMA/Oceananigans.jl/blob/d270593c2f28278132b59dcc5fbce76fb8201b7b/.github/workflows/ci.yml#L138-L161

but I don't we should do this now. With Metal there's no FFT library so we would need to rely on the unified memory using FFTW via CPUs but that's not for now ... let's first make sure everything runs on one GPU 😄

gregordecristoforo:
It looks like the buildkite tests are passing. If there is anything left please let me know :)

maximilian-gelbrecht:
But now this is just using JuliaGPU's AMD runners and not Ludovic's, or not?

maximilian-gelbrecht:
Wouldn't we just need a new, separate buildkite workflow for that?

luraess:
My AMD GPU runner is registered in the JuliaGPU Buildkite org, so any tests in Buidlkite that target AMDGPU actually run on "my" server, which is meant (and fine) so.

.buildkite/pipeline.yml
```diff
@@ -19,4 +19,27 @@ steps:
...
+                using Pkg
+                Pkg.develop(; path=pwd())
+                Pkg.add(["AMDGPU","Adapt","KernelAbstractions"])'
+
+      julia -e 'println("+++ :julia: Running GPU tests")
+                include("test/GPU/runtests.jl")'
+    agents:
+      queue: "juliagpu"
+      rocm: "*"
+      rocmgpu: "*"
```
luraess, line 44: This targets the AMD GPU runner. Currently we have 2, the "base" one which is a 7800 XT and most of the time I also enable the second which is a 7900 XTX.

maximilian-gelbrecht:
Ah ok, sorry for all the confusion. 

13min CI time on AMD vs 4 min on Nvidia, isn't great, but a lot of time is spent in setup. 

On AMD we get these infos/warning 

```
<head></head>
┌ Info: Global hostcalls detected!
--
  | │ - Source: MethodInstance for SpeedyWeather.gpu_set_field_3d_kernel!(::KernelAbstractions.CompilerMetadata{KernelAbstractions.NDIteration.StaticSize{(4680, 1)}, KernelAbstractions.NDIteration.DynamicCheck, Nothing, Nothing, KernelAbstractions.NDIteration.NDRange{2, KernelAbstractions.NDIteration.StaticSize{(19, 1)}, KernelAbstractions.NDIteration.StaticSize{(256, 1)}, Nothing, Nothing}}, ::AMDGPU.Device.ROCDeviceMatrix{Float32, 1}, ::AMDGPU.Device.ROCDeviceVector{Float32, 1}, ::AMDGPU.Device.ROCDeviceVector{Float32, 1}, ::AMDGPU.Device.ROCDeviceVector{Float32, 1}, ::SpeedyWeather.var"#143#144"{Float32, Float32}, ::SpeedyWeather.var"#81#83")
  | │ - Hostcalls: [:malloc_hostcall]
  | │
  | │ Use `AMDGPU.synchronize(; stop_hostcalls=true)` to synchronize and stop them.
  | └ Otherwise, performance might degrade if they keep running in the background.

```

Currently we never explicitly synchronise because at least with Nvidia every kernel / operation is executed in the same stream, and it's implicitly synchronised once we output. I am not sure exactly what these infos here mean.

luraess:
> 13min CI time on AMD vs 4 min on Nvidia, isn't great, but a lot of time is spent in setup.

126 sec on Nvidia vs 407 sec on AMD if one ignores the setup time, indeed. I guess it may be interesting to further dig in the perf bottleneck. IIRC from the MWE I was asked to run it seems indeed that there is quite a perf gap between Nvidia and AMD, possibly related to FFT. One should certainly dig in further to better understand where the differences are coming from. Also, if we figure out this is e.g. FFT, then maybe making a small C++ reproducer for both AMD (HIP) and Nvidia (HIP or CUDA) maybe further inform whether there is some issue with ROCm libs or the Julia wrapping.

Regarding hostcalls, like https://buildkite.com/julialang/speedyweather-dot-jl/builds/2390#019a82ce-a71f-416b-a8d6-1b3a78f9890a/668-687, what code triggers them? If one nails this down, that'd be helpful.

gregordecristoforo:
These Warnings about host calls do not appear when I am running the tests on LUMI, even when I set `ENV["JULIA_DEBUG"] = "AMDGPU"`.

@luraess do you use some other flags to trigger the warnings? If not, maybe the different behavior occurs due to the different ROCm versions.

luraess:
Interesting. CI was run on ROCm 6.4.4 and no other flags should have been set.

luraess:
Can you maybe re-trigger CI as to target the other AMD GPU CI machine?

gregordecristoforo:
> Can you maybe re-trigger CI as to target the other AMD GPU CI machine?

Sure, could you remind me how to do that?

Edit: Should I specify to use the other AMD GPU CI machine, or is that done on your side and I should just re-trigger the CI?

luraess:
It has now run again and the perf is on par with CUDA. Something is off with the second CI runner and I removed it from the list for now to investigate. We still have the hostcalls though, but this may be related to a ROCm version. I can try to run it locally on 7.1 to see what would happen.

luraess:
Just checked running locally on ROCm 7.1 also emits hostcalls as for 6.4.4 in CI. It would be great to have a MWE related to those in order to see further (can be done in another PR/issue).

gregordecristoforo:
I am afraid I won't be able to create a MWE since these warnings don't appear on LUMI with ROCM 6.0.2. (assuming you don't want me to spam the CI pipeline with guesses). We will have a system upgrade on LUMI in January which will upgrade the ROCm versions to (hopefully) 6.4. after which the warning might be triggered, but this is of course speculative.

luraess:
If I have some spare time I could try looking into it.

.buildkite/pipeline.yml
```diff
@@ -19,4 +19,27 @@ steps:
       queue: "juliagpu"
       cuda: "*"
     if: build.message !~ /\[skip tests\]/
-    timeout_in_minutes: 60
\ No newline at end of file
+    timeout_in_minutes: 60
```
maximilian-gelbrecht: I am happy to decrease this even more. Buildkite is not our general purpose CI. I think we should ideally aim for not exceeding 10min, and 30 minutes as a timeout should sufficient

luraess: Aiming for 30min could be fine.

gregordecristoforo: I decreased it for both CUDA and AMDGPU jobs to 30 min

docs/src/architectures_gpu.md
```diff
@@ -20,7 +20,7 @@ run!(simulation, period=Day(10))
 In order to easily transfer our structures between CPU (e.g. for plotting and output) and GPU, we have the following utilities that can make use of the `architecture` object defined above and the `on_architecture` function, e.g. as follows: 
 
 ```julia
-using SpeedyWeather, CUDA 
+using SpeedyWeather, CUDA # For AMD GPUs, replace `CUDA` with `AMDGPU`
 nlat_half = 6
 arch_cpu = SpeedyWeather.CPU()
 arch_gpu = SpeedyWeather.GPU()
```
milankl, line 26: we currently also export these but maybe it's good to keep them in the SpeedyWeather scope?

SpeedyTransforms/ext/SpeedyTransformsAMDGPUExt.jl
```diff
@@ -0,0 +1,92 @@
+module SpeedyTransformsAMDGPUExt
```
milankl, line 1: Isn't this extension essentially identical to the SpeedyTransformsCUDAExt? If yes we could move the generic code dispatched over `GPUArrays.AbstractGPUArray` into SpeedyTransforms directly and let the extension only define `which_FFT_package` ?

test/GPU/spectral_transform.jl
```diff
@@ -1,3 +1,21 @@
...
+    field = zeros(Float32, spectral_grid.grid, 8)
+    transform!(field, L, S)
+    transform!(L, field, S)
+
+    # 2nd roundtrip
+    L2 = deepcopy(L)
+    transform!(field, L, S)
+    transform!(L, field, S)
+
+    @test L ≈ L2
```
milankl, line 16: I would add here the 1-layer version too as only that will execute the serial FFT code otherwise you're only the the batched method!

SpeedyTransforms/ext/SpeedyTransformsAMDGPUExt.jl
```diff
@@ -0,0 +1,92 @@
...
+        ilons::UnitRange{Int};
+        not_equator::Bool = true
+    ) where {NF<:AbstractFloat, N}
+        rfft_plan = S.rfft_plans[j]     # FFT planned wrt nlon on ring
+        nlayers = size(field, 2)        # number of vertical layers
+
+        if not_equator
+            view(f_out, 1:nfreq, 1:nlayers, j) .= rfft_plan * field.data[ilons, :]
+        else
+            fill!(f_out[1:nfreq, 1:nlayers, j], 0)
```
milankl: Yo this line contains a bug as it doesn't do anything. Slicing creates an array copy, filling it with zeros just fills a copy with zeros that's garbage collected again. @maximilian-gelbrecht is that the origin of the tests failing on odd grids (clenshaw, healpix)?

```julia
julia> A = cu(rand(Float32, 5))
5-element CuArray{Float32, 1, CUDA.DeviceMemory}:
 0.7814894
 0.81775904
 0.282943
 0.7027015
 0.36905903

julia> fill!(A[1:4], 0)
4-element CuArray{Float32, 1, CUDA.DeviceMemory}:
 0.0
 0.0
 0.0
 0.0

julia> A
5-element CuArray{Float32, 1, CUDA.DeviceMemory}:
 0.7814894
 0.81775904
 0.282943
 0.7027015
 0.36905903
```

originates from the CUDA extension https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/d4220a36be7655d99e402575ec4513f4edd8b494/SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl#L68

maximilian-gelbrecht: Well caught! We definitely have to fix this, but the error originated from somewhere else (and isn't actually present anymore). Just replacing it with a `fill!(view(...))` should suffice, right?

milankl: fixed this in [8dde5d4](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/901/commits/8dde5d4466f44a315f1e79be99dd5af4c2b17ab7)

milankl: did just indexing as this should work on both CPU/GPU like `[:, :, i] .= 0`

milankl:
With the last commit I collapsed the whole CPU/GPU differences into
```julia
# CPU version with view
@inline view_only_on_cpu(A::AbstractArray, I...) = view(A, I...)
# GPU version without view
@inline view_only_on_cpu(A::AbstractGPUArray, I...) = A[I...]
```

all the remaining code is shared!

milankl:
Cool, both GPU tests pass both in ~2min 🎉

milankl:
@maximilian-gelbrecht should we test on more grids now?

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/d4220a36be7655d99e402575ec4513f4edd8b494/test/GPU/spectral_transform.jl#L4-L11

at least one of the odd grids?

milankl:
Actually we don't even have to have the SpeedyTransformsAMDGPUExt or the CUDAExt because

```julia
julia> AbstractFFTs.plan_rfft(rand(8))
FFTW real-to-complex plan for 8-element array of Float64
(rdft2-r2hc-direct-8 "r2cf_8")

julia> AbstractFFTs.plan_rfft(cu(rand(8)))
CUFFT Float32-to-ComplexF32 forward plan for 8-element CuArray of Float32
```

AbstractFFTs is imported and extended by FFTW, CUDA and AMDGPU so `AbstractFFT.plan_rfft(array)` will automatically choose the right plan. Question for me is now, shall we just take these two extensions out or shall we leave them in because we may in the end want to specialize some parts of the spectral transform algorithm to specific hardware. Then we would need to get the extension back in ... I don't see a harm in just leaving them as they are?

maximilian-gelbrecht:
> @maximilian-gelbrecht should we test on more grids now?
> 
> https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/d4220a36be7655d99e402575ec4513f4edd8b494/test/GPU/spectral_transform.jl#L4-L11
> 
> at least one of the odd grids?

I was wondering about that as well. I guess OctahedralGaussianGrid and e.g. HEALPixGrid would also suffice? Basically, we should test one grid that includes the equator and one that doesn't.

maximilian-gelbrecht:
> Actually we don't even have to have the SpeedyTransformsAMDGPUExt or the CUDAExt because
> 
> ```julia
> julia> AbstractFFTs.plan_rfft(rand(8))
> FFTW real-to-complex plan for 8-element array of Float64
> (rdft2-r2hc-direct-8 "r2cf_8")
> 
> julia> AbstractFFTs.plan_rfft(cu(rand(8)))
> CUFFT Float32-to-ComplexF32 forward plan for 8-element CuArray of Float32
> ```
> 
> AbstractFFTs is imported and extended by FFTW, CUDA and AMDGPU so `AbstractFFT.plan_rfft(array)` will automatically choose the right plan. Question for me is now, shall we just take these two extensions out or shall we leave them in because we may in the end want to specialize some parts of the spectral transform algorithm to specific hardware. Then we would need to get the extension back in ... I don't see a harm in just leaving them as they are?

Do AMDGPU and CUDA behave the same wrt to plans on views, strides, etc? This is where the different FFT implementations can differ from each other.

milankl:
> Do AMDGPU and CUDA behave the same wrt to plans on views, strides, etc? This where the different FFT implementations can differ from each other. 

I don't know. Didn't check. At the moment all GPU FFTs take stride-1 as input and output stride-1 due to the non-view slicing and the .= copy into a view. Only on the CPU we avoid these two allocating copies. Did we ever try what happens if we use scratch memory to avoid at least the GPU allocations?

luraess:
In general, some heuristics may indeed differ amongst GPU backends. And the fact that CUDA.jl is more fancy in some ways than currently AMDGPU.jl, it may just be good to have extension to deal with those. What's the issue with the FFT on GPU that requires an explicit buffer for the view - or the allocation?

maximilian-gelbrecht:
Ideally we would like to have FFT plans that can use input and output strides. 

The CUFFT.jl (and AMDGPU.jl FFT?) don't work when applied to a strided input, FFTW.jl does. But FFTW.jl also doesn't work to output those to strided views (just inputs can be strided).

In theory, there's low level support available for these things in FFTW, CUFFT and rocFFT (both input and output strides), but there's no nice Julia interface in FFTW.jl and CUDA.jl and no way currently in AbstractFFT to create a plan with output strides either. There's a PR in CUDA.jl that was abandoned years ago that tried to implement it https://github.com/JuliaGPU/CUDA.jl/pull/912/

I opened a separate issue for this: https://github.com/SpeedyWeather/SpeedyWeather.jl/issues/918

maximilian-gelbrecht:
So, there's still an issue with the non-gaussian grids on GPU, but that's not only limited to the AMD extension. I would say we merge this and work on that issue in a different PR.

maximilian-gelbrecht:
somethings wrong with the central Julia registry that's why all tests are failing, it's the same in other PRs

luraess:
Yes. Let'smwait CloudFlares resumes and then one could re-trigger CI.

PR #902, opened 2025-11-09
Set particle sigma to layer they are advected on

milankl:


milankl:
Given that buildkite currently has a queue but changes here don't affect GPU stuff I'm merging

PR #903, opened 2025-11-10
Barotropic Model initialization (mostly) on GPU

maximilian-gelbrecht:
Makes it so that the `initialize!(::BarotropicModel)` doesn't need an `@allowscalar` on GPU. Some things are explicitly computed on CPU and then transferred (small arrays), other things like e.g. the initial conditions have been kernelized.

### DIFF
--- a/LowerTriangularArrays/src/lower_triangular_array.jl
+++ b/LowerTriangularArrays/src/lower_triangular_array.jl
@@ -100,9 +100,9 @@ Base.sizeof(L::LowerTriangularArray) = sizeof(L.data)
 
 function Base.show(io::IO, ::MIME"text/plain", L::LowerTriangularMatrix)
     Base.array_summary(io, L, axes(L))
-    
+    L_print = architecture(L) <: Architectures.CPU ? L : on_architecture(CPU(), L)
     # the following is copied over from base/arrayshow.jl
-    X = Matrix(L)
+    X = Matrix(L_print)
     # 2) compute new IOContext
     if !haskey(io, :compact) && length(axes(X, 2)) > 1
         io = IOContext(io, :compact => true)
--- a/LowerTriangularArrays/src/lower_triangular_array.jl
+++ b/LowerTriangularArrays/src/lower_triangular_array.jl
@@ -100,7 +100,7 @@ Base.sizeof(L::LowerTriangularArray) = sizeof(L.data)
 
 function Base.show(io::IO, ::MIME"text/plain", L::LowerTriangularMatrix)
     Base.array_summary(io, L, axes(L))
-    L_print = architecture(L) <: Architectures.CPU ? L : on_architecture(CPU(), L)
+    L_print = typeof(architecture(L)) <: Architectures.CPU ? L : on_architecture(CPU(), L)
     # the following is copied over from base/arrayshow.jl
     X = Matrix(L_print)
     # 2) compute new IOContext
--- a/src/dynamics/initial_conditions.jl
+++ b/src/dynamics/initial_conditions.jl
@@ -82,39 +82,7 @@ $(TYPEDFIELDS)"""
 end
 
 """$(TYPEDSIGNATURES)
-Start with random vorticity as initial conditions"""
-function initialize!(   progn::PrognosticVariables{NF},
-                        initial_conditions::RandomVorticity,
-                        model::Barotropic) where NF
-
-    # reseed the random number generator, for seed=0 randomly seed from Julia's global RNG
-    seed = initial_conditions.seed == 0 ? rand(UInt) : initial_conditions.seed
-    RNG = initial_conditions.random_number_generator
-    Random.seed!(RNG, seed)
-
-    lmax = progn.trunc + 1
-    power = initial_conditions.power + 1    # +1 as power is summed of orders m
-
-    (; spectrum, nlayers)
-    A = initial_conditions.amplitude
-    ξ = zeros(LowerTriangularArray{Complex{NF}}, spectrum, nlayers)
-
-    lm = 0
-    for m in 1:lmax
-        for l in m:lmax
-            lm += 1
-            ξ[lm, :] .= A*l^power*(2rand(RNG, Complex{NF}) - (1 + 1im))
-        end
-    end
-
-    spectral_truncation!(ξ, initial_conditions.max_wavenumber)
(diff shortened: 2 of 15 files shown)


LowerTriangularArrays/src/lower_triangular_array.jl
```diff
@@ -100,9 +100,9 @@ Base.sizeof(L::LowerTriangularArray) = sizeof(L.data)
 
 function Base.show(io::IO, ::MIME"text/plain", L::LowerTriangularMatrix)
     Base.array_summary(io, L, axes(L))
-    
+    L_print = architecture(L) <: Architectures.CPU ? L : on_architecture(CPU(), L)
     # the following is copied over from base/arrayshow.jl
-    X = Matrix(L)
+    X = Matrix(L_print)
```
milankl: Over in #901 we just had an error when showing `L::LowerTriangularMatrix` as RocArray, that probably solves this?

src/dynamics/initial_conditions.jl
```diff
@@ -112,6 +112,72 @@ function initialize!(   progn::PrognosticVariables{NF},
...
+
+    lmax = progn.trunc + 1
+    power = initial_conditions.power + 1    # +1 as power is summed of orders m
+
+    (; spectrum, nlayers) = progn
+    A = convert(NF, initial_conditions.amplitude)
+    
+    # Pre-generate random values on CPU, then transfer to device
+    nlm = length(spectrum)
+    random_values_cpu = 2 .* rand(RNG, Complex{NF}, nlm, nlayers) .- (1 + 1im)
```
milankl, line 105: As long as we don't create arrays of random values during model integration we just do it on the CPU and transfer right? Like the GPUs all have some `rand` methods too but not it's worth it ...

src/dynamics/initial_conditions.jl
```diff
@@ -112,6 +112,72 @@ function initialize!(   progn::PrognosticVariables{NF},
...
+    ξ,
+    @Const(random_values),
+    A,
+    power,
+    @Const(l_indices),
+    lmax
+)
+    I = @index(Global, Cartesian)
+    lm = I[1]  # spectral coefficient index
+    k = I[2]   # layer index
```
milankl: I don't think I've understood yet when we use `@index(Global, Tuple)` and when `@index(Global, Cartesian)`

src/dynamics/initial_conditions.jl
```diff
@@ -161,9 +227,7 @@ function initialize!(   progn::PrognosticVariables{NF},
 
     ξ = curl(u_spectral, v_spectral, model.spectral_transform; radius)
     
-    ξ[1:1] .= 0        # remove mean in a ugly hacky way that doesn't use scalar indexing for GPU compat
-    # TODO: do this better, we could also just use the @allowscalar macro in cases like this, 
-    # but currenlty CUDA isn't a dependency of SpeedyWeather, so we can't use it here 
+    fill!(ξ[1:1], 0)  # remove mean in a ugly hacky way that doesn't use scalar indexing for GPU compat
```
milankl: This happens now in several parts of the code, sometimes it's read access sometimes write. Is this now the recommended way of doing this?

milankl (commented):
Fantastic yeah! I already had started with some of this but I think this is a good strategy to just do this it by bit for various parts of the model to remove CPU-only code.

LowerTriangularArrays/src/lower_triangular_array.jl
maximilian-gelbrecht: Yes, that should do it. Everything that's displayed or any form of I/O needs to be transferred to the CPU.

src/dynamics/initial_conditions.jl
maximilian-gelbrecht: I mean, it's the initialization, it's not terribly important, but so far I was roughly setting my personal cut off for GPU computation at one full `Field` / `LTA`. So something like an array over just the layers is def computed on the CPU. In this case I was actually igoroing my own guideline, so I am gonna change that (even though it's not super important)

src/dynamics/initial_conditions.jl
maximilian-gelbrecht: Ha, me neither to be honest

maximilian-gelbrecht: I think NTuple might be preferable, it's the simpler structure. (and oceananigans uses it like this as well ;))

maximilian-gelbrecht: I changed that now across the model.

src/dynamics/initial_conditions.jl
maximilian-gelbrecht: Well, it's a way that works. It's okay during initialization. I think when we want to set the mean during integration, we should ideally find a smarter way than doing it like this (e.g. somehow integrating it in other kernels)

src/dynamics/initial_conditions.jl
maximilian-gelbrecht: Ah, here in this case we needed a `rand` with `RNG` which we don't have defined for LTA. And then, we could distinguish between `rand` and `CUDA.rand` here based on the `architecture`, but it's also almost the same and doesn't really matter.

src/dynamics/initial_conditions.jl
maximilian-gelbrecht: Okay, by changing it across the model, I did find a difference between both. `Cartesian` also works if the array might be 1-dimensional or n-dimensional. `NTuple` not.

src/dynamics/initial_conditions.jl
maximilian-gelbrecht: we could also write a `set_mean!(L::LTA)` function that is extended in a LowerTriangularArraysCUDAExt (which we don't have yet.

src/dynamics/initial_conditions.jl
milankl: I agree!

src/dynamics/initial_conditions.jl
maximilian-gelbrecht: Ok, then let's do it with `GPUArrays.@allowscalar`

maximilian-gelbrecht:
From my side this could be merged. What do you think?

milankl:
Happy for you to merge!!!

Issue #721, opened 2025-04-30
CI time


milankl:
All testsets that are currently above 30s

```julia
Test Summary:     | Pass  Total     Time
Grid broadcasting |  477    477  2m59.4s

Test Summary:                   | Pass  Total     Time
LowerTriangularArray: broadcast |  126    126  1m29.9s

Test Summary:                     | Pass  Total   Time
AbstractGridArray: GPU (JLArrays) |  828    828  40.3s

Test Summary:              | Pass  Total     Time
SpeedyTransforms: AD Rules |  862    862  1m17.1s

Test Summary:           | Pass  Total     Time
Vertical advection runs |    6      6  1m12.7s

Test Summary:     | Pass  Total     Time
Albedo composites |   54     54  1m15.8s

Test Summary:                            | Pass  Total   Time
Single time step at uncommon resolutions |   10     10  37.6s

Test Summary:        | Pass  Total   Time
Forcing and drag: 3D |    2      2  35.3s

Test Summary:   | Total   Time
Wet land models |     0  54.5s
```

milankl:
> The broadcasting tests should be definitely reduced in time.

#731 tests broadcasting over only a few grids, 1 full, 1 octahedral, 1 healpix. I don't think there's a need to test for all of them which takes time as every grid is a new type and so everything has to be recompiled.

>     * AbstractGridArray: GPU (JLArrays):
>     * SpeedyTransforms: AD Rules: 

Left those ones unchanged for now

milankl:
A lot of CI is actually compile time. For convenience we have many components now that take `model` as argument. However, when testing through combinations of model components, e.g.

```julia
    for Temperature in (SeasonalLandTemperature, ConstantLandTemperature, LandBucketTemperature)
        for SoilMoisture in (NoSoilMoisture, SeasonalSoilMoisture, LandBucketMoisture)
            for Vegetation in (NoVegetation, VegetationClimatology)
                for Model in (PrimitiveDryModel, PrimitiveWetModel)
```

This means the type of `model` changes for each iteration even though most components are the same. But given it's parametric wrt to those everything that takes `model` is recompiled as it's a new type. So I tried to test only those functions that actually rely on those components, e.g. `land_timestep!` here. But in general I think we can also introduce more function barriers again, e.g.

```julia
f(a, model) = f(a, model.component)
f(a, component) = ...
```

[...]

PR #914, opened 2025-11-14
No specialize model for initialize!

### LINKED ISSUE #909: Compilation time due to TTSX, time to second X, problem
Following up from #721 I'm testing the compile time vs full time of `initialize!(model)` and `run!(simulation)` when changing from default `PrimitiveWetModel` to `convection=nothing`. Technically this does not require *any* recompilation but because the `typeof(model)` changes it does trigger recompilation. This might be prevented with function barriers but this first assess how bad this problem actually is. Tested with `t = @timed ...` and `t.compile_time` and then doing a 2nd time `t.time` for execute time.

####  `initialize!`

| model | compile time | execute time |
| - | - | - |
| default |  25.8s | 0.2s | 
|  no convection | 7.8s | 0.15s |

#### `run!`

| model | compile time | execute time |
| - | - | - |
| default |  29.4s | 2.8s | 
|  no convection | 16.5s | 2.6s |

So for testing some e.g. 3x3x3 combinations of parameterizations/land/ocean/sea-ice combinations CI would spend 27x 16.5s = 7.5min unnecessarily on recompilation but only 1min to actually run those simulations (and often we don't even need to run for 10 days and could make it as short as a few time steps).

This is an interesting problem. While in Julia world the [TTFX](https://discourse.julialang.org/t/taking-ttfx-seriously-can-we-make-common-packages-faster-to-load-and-use/74949) (time to first X, historically X=plot) problem has gained much attention, here we are dealing with TTSX, time to second X, arising because we want modularity. We therefore pass around a very comprehensive `model` with O(30) parameters that are parametric themself so any change to those triggers (unnecessary) recompilation for maaany methods.

milankl:
fixes #909 maybe?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## Unreleased
 
-## v0.17.4
+- @nospecialize model for initialize! methods [#914](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/914)
+
+## v0.17.4
 
 - Show for LowerTriangularArray/Matrix on CPU/GPU [#907](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/907)
 - Ring, nested and matrix order for OctaHEALPixGrid [#887](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/887)
--- a/src/dynamics/adiabatic_conversion.jl
+++ b/src/dynamics/adiabatic_conversion.jl
@@ -21,6 +21,7 @@ function initialize!(
     adiabatic::AdiabaticConversion,
     model::PrimitiveEquation,
 )
+    @nospecialize model
     (; σ_lnp_A, σ_lnp_B) = adiabatic
 
     # Transfer arrays to CPU for computation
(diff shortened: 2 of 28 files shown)


maximilian-gelbrecht:
Didn't really do anything, no? 

CI time even got up a bit here (even though this also could be normal fluctuations of GitHub Actions).

milankl:
Yeah doesn't seem to solve the problem, should benchmark this locally and look at this more carefully though to examine

milankl:
Talking to @aviks at JuliaEO what's needed here to combine a `@nospecialize` annotation within a function barrier. That way compiling the function barrier with a changing `typeof(model)` is quick and easy but the unpacking guarantees of the function barrier allows to only specialize on components of model which don't need to be recompiled if unrelated components are changed.

milankl:
While technically orthogonal to #1175 that PR was more successful at dropping CI time (earlier issue in #721) So closing this here. But could be picked up again in the future

maximilian-gelbrecht:
At JuliaCon in the session, the advice to get the compile time down for a package was also literally to just ask Claude about it. Maybe we just should just do that.

PR #920, opened 2025-11-18
[RingGrids] Don't adapt rings on GPU

maximilian-gelbrecht:


maximilian-gelbrecht:
@milankl I think this at least solves some issues with `eachring` as it's now always on CPU. I basically continues this logic we use in the parametrization to only adapt the fields that really need to be adapted. Your problem might still be somewhere else, but this PR is good to do nonetheless I think

milankl:
So if we now do

```julia
rings = eachring(grid)
launch!(arch, ..., rings, ...)
```

then this wouldn't work right? Most kernels need, if they do, `whichring` and not `rings` so is this why we have `rings` now living on the CPU?

maximilian-gelbrecht:
Correct, `rings` are ranges of indices, mostly for iteration, we don't need them in kernels, we only need the `whichrings` in the kernel.

milankl:
This is great, thanks Max!

PR #921, opened 2025-11-19
Fix round trip test of transforms

maximilian-gelbrecht:
This continues #901 and looks for a fix for the problems with Clenshaw and HealPixGrids on the GPU.


maximilian-gelbrecht:
@milankl There wasn't actually anything wrong with the transforms on GPU. It was just a mix of us misunderstanding the transforms and the tests being a bit suboptimal. With the dealising set as it was, the transforms were not exact, but this was what the tests expected. 

Now, the GPU tests just compares a round trip to the CPU and in the CPU CI we do actually test round trip accuracy. Are the values for the tolerances as you would expect?

milankl:
Wouldn't GPU and CPU errors be approximate? The transform error should be the same on both just that the orders of arithmetic operations may change or CUFFT may use a different algorithm compared to FFTW. If approximate doesn't work what about relative errors being some multiple of eps, e.g. `rtol = 10*eps(NF)` ?

maximilian-gelbrecht:
Yes, they should. And this is what we test now, but before it wasn't.

test/transforms/spectral_transform.jl
```diff
@@ -325,6 +325,46 @@ end
...
+            dealiasing = 3 
+            NF = Float64
+            nlayers = 8
+
+            SG = SpectralGrid(; NF, trunc, Grid, dealiasing, nlayers)
+            S = SpectralTransform(SG)
+
+            # start in spectral space but compare in grid space to 
+            # avoid inaccuracies due to filtering out higher frequencies
+            spec = rand(Complex{NF}, SG.spectrum, nlayers)
```
milankl: I try not to use `rand` as it doesn't create white noise in grid space but always this funny pattern which might just mean a larger transform error due to this singularity

Image

in contrast `randn` in spectral space is ~white noise in grid space too

Image

test/transforms/spectral_transform.jl
```diff
@@ -325,6 +325,46 @@ end
...
+    # only Gaussian transforms are exact, that's why we need have different tolerenaces
+    # the tolerances below were just emperically found and are kept to ensure 
+    # we don't introduce bugs 
+
+    tolerances = [1e-13,  # FullGaussianGrid
+                  1e-12,  # FullClenshawGrid
+                  1e-13,  # OctahedralGaussianGrid
+                  1e-12,  # OctahedralClenshawGrid
+                  1e-2,   # HEALPixGrid
+                  1e-2]   # OctaHEALPixGrid
```
milankl, line 339: yes these are good, if we're already testing all grids we may want to add `OctaminimalGaussianGrid` the tolerance should be somewhere in between the exact and the inexact grids

test/transforms/spectral_transform.jl
maximilian-gelbrecht: Ok, done!

test/transforms/spectral_transform.jl
maximilian-gelbrecht: Included

maximilian-gelbrecht: It's ~1e-5 rtol, I set the tolerance to 5e-4 to have a small margin

PR #924, opened 2025-11-20
Small fixes for new parametrization on GPU 

maximilian-gelbrecht:


src/dynamics/orography.jl
```diff
@@ -1,7 +1,7 @@
 abstract type AbstractOrography <: AbstractModelComponent end
 
 # adapt to GPU only the fields themselves
-Adapt.adapt_structure(to, orog::AbstractOrography) = (orography=orog.orography, geopot_surf=orog.geopot_surf)
+Adapt.adapt_structure(to, orog::AbstractOrography) = (orography=adapt_structure(to, orog.orography), geopot_surf=adapt_structure(to, orog.geopot_surf))
```
maximilian-gelbrecht, line 4: @milankl we just forgot to actually adapt the `Fields` because that's still needed

src/physics/radiation/shortwave_radiation.jl
```diff
@@ -52,7 +52,7 @@ function shortwave_radiation!(ij, diagn, shortwave::TransparentShortwave, planet
     ssrd_land[ij]  = albedo_land * D
 
     # land-sea mask-weighted
-    albedo = (1 - land_fraction)*albedo_ocean + land_fraction*albedo_land
+    albedo = (one(albedo_ocean) - land_fraction)*albedo_ocean + land_fraction*albedo_land
```
maximilian-gelbrecht, line 55: I have some connection problems to my HPC right now, and couldn't really test this..

src/dynamics/orography.jl
milankl: So `@adapt_structure` adapts every struct field if they are adaptable to GPU essentially doing `adapt_structure(to, field)` and reconstructing the struct with it but then one can also be creative and subset the struct / move to NamedTuple but nevertheless every field needs `adapt_structure(to, field)` when done manually? Just slowly getting my head around the logic here...

src/models/primitive_wet.jl
```diff
@@ -94,7 +94,7 @@ $(TYPEDFIELDS)"""
     surface_heat_flux::SH = SurfaceHeatFlux(spectral_grid)
     surface_humidity_flux::HF = SurfaceHumidityFlux(spectral_grid)
     large_scale_condensation::LSC = ImplicitCondensation(spectral_grid)
-    convection::CV = SimplifiedBettsMiller(spectral_grid)
+    convection::CV = BettsMillerConvection(spectral_grid)
```
milankl, line 97: that's another weird one that didn't get merged properly. I thought of git's merge a bit of black magic anyway and it seems there are indeed cases there it merges incorrectly or at least suggests something in a merge conflict that is actually a bit more complicated

milankl: thanks for fixing this

src/physics/radiation/shortwave_radiation.jl
```diff
@@ -28,10 +28,10 @@ end
 initialize!(::TransparentShortwave, ::PrimitiveEquation) = nothing
 
 # function barrier
-parameterization!(ij, diagn, progn, shortwave::TransparentShortwave, model) =
-    shortwave_radiation!(ij, diagn, shortwave, model.planet, model.land_sea_mask)
+@inline parameterization!(ij, diagn, progn, shortwave::TransparentShortwave, model) =
```
milankl, line 31: I haven't `@inline` anything so far, shall we inline everything as a rule of thumb or are there cases where we shouldn't. I'm thinking about the no specialize / CI issues #909 #914 #721 if that logic applies than we should have

```julia
@inline parameterization!(...) = ...
@noinline shortwave_radiation!(...) = ...
```

because otherwise we would overspecialise on `model` in a function that may not depend on all components of model? But still unsure how much that logic actually applies to overall compile time ...

milankl (approved):

src/dynamics/orography.jl
maximilian-gelbrecht: `@adapt_structure` recursively calls itself that's why it works in most cases. If you do manually you have to think about that yourself. So yeah, every field needs to be adapted. Because on the Kernel you don't want a `CuArray` (host array), you want a `CuDeviceArray`.

PR #925, opened 2025-11-21
set! Field with functions on GPU (but not really) 

maximilian-gelbrecht:


src/dynamics/set.jl
```diff
@@ -181,11 +181,22 @@ function set!(
     geometry::Geometry,
     S::Union{Nothing, SpectralTransform}=nothing;
     add::Bool=false,
+    enforce_static_func = false,
```
maximilian-gelbrecht: This kwarg allows to overwrite the explicit transfer to the CPU for functions that are already known at compile time.

milankl:
Awesome thanks Max! Wouldn't most user-defined functions be known at compile time because they defined them in global scope? But then they may have defined them with references to global variables like we had before the eval. I am wondering whether, given `f::Function` one could somehow check that so that the `enfore_static` flag can be automatically set? I already find it confusing what exactly that means 😅

maximilian-gelbrecht:
Maybe https://github.com/julia-vscode/StaticLint.jl can do that? But as long as we don't use `set!` within an actual simulation, it's also not that important I'd say.

milankl:
We do in KolmogorovFlow I think!

maximilian-gelbrecht:
Good points. After briefly trying it out: The typical user definition in a script does work, because it is compiled before, as you said. So I'll set the default the `static` for now, but update the docstring. The Kolmogorov flow example works as well (to be honest I don't quite understand what's the difference to the situation in the initial conditions..., but maybe the function is simple enough that the compiler just directly compiles it away... )

The Enzyme error has nothing to do with PR, and is unfortunately a new Enzyme version breaking something for us :(.

maximilian-gelbrecht:
I think with this change, initialisation on GPU of the PrimitiveWetModel should also work, we just have to call the `set!` in the initial conditions with `static_func=false` once this is merged.

PR #927, opened 2025-11-24
Initial conditions revised (typed precision and working on GPU? + tests)

maximilian-gelbrecht:
I also noticed we had no unit tests for them

### DIFF
--- a/src/models/primitive_dry.jl
+++ b/src/models/primitive_dry.jl
@@ -88,7 +88,7 @@ $(TYPEDFIELDS)"""
     surface_condition::SC = SurfaceCondition(spectral_grid)
     surface_momentum_flux::SM = SurfaceMomentumFlux(spectral_grid)
     surface_heat_flux::SH = SurfaceHeatFlux(spectral_grid)
-    convection::CV = DryBettsMiller(spectral_grid)
+    convection::CV = BettsMillerDryConvection(spectral_grid)
     shortwave_radiation::SW = TransparentShortwave(spectral_grid)
     longwave_radiation::LW = JeevanjeeRadiation(spectral_grid)
     stochastic_physics::SP = nothing


src/dynamics/initial_conditions.jl
```diff
@@ -83,14 +85,16 @@ $(TYPEDFIELDS)"""
     random_number_generator::Random.Xoshiro = Random.Xoshiro(seed)
```
milankl, line 87: Isn't bits type because mutable I think
```julia
julia> using Random

julia> RNG = Random.Xoshiro()
Xoshiro(0x3e54876a70c0c612, 0xd21272bbe52fae0b, 0x4fe85cae42342250, 0x2b5c731fdd76f9f1, 0x739cd4e47995515a)

julia> isbits(RNG)
false

julia> ismutable(RNG)
true
```

src/dynamics/initial_conditions.jl
```diff
@@ -485,8 +525,8 @@ function initialize!(
 
     η(λ, θ) = R^2 / g * (A(λ, θ) + B(λ, θ) * cosd(m * λ) + C(λ, θ) * cosd(2m * λ))
 
-    set!(progn, geometry, vor = ζ, static_func=false)
-    model isa ShallowWater && set!(progn, geometry, pres = η, static_func=false)
+    set!(progn, geometry, vor = ζ, static_func = false)
```
milankl, line 530: We still need the `static_func` business then if we use functors when `set!` causes world age problems?

src/models/primitive_dry.jl
```diff
@@ -51,7 +52,7 @@ $(TYPEDFIELDS)"""
     TS3,    # <:Tuple{Symbol}
 } <: PrimitiveDry
 
-    spectral_grid::SpectralGrid
+    spectral_grid::SG
```
milankl, line 55: oh yeah well spotted, didn't realise we continued to have this as an abstract field ...

maximilian-gelbrecht:
Some of the unit tests don't pass at the moment. Which isn't because of this PR, the initial conditions were just using outdated code. Good we have unit tests now.

src/dynamics/initial_conditions.jl
```diff
@@ -20,33 +20,35 @@ function initialize!(
...
 struct ZeroInitially <: AbstractInitialConditions end
+ZeroInitially(SG::SpectralGrid) = ZeroInitially()
 initialize!(::PrognosticVariables, ::ZeroInitially, ::AbstractModel) = nothing
 
 # to avoid a breaking change, like ZeroInitially
 export StartFromRest
-@kwdef struct StartFromRest{P, T, H} <: AbstractInitialConditions
-    pres::P = ConstantPressure()
-    temp::T = JablonowskiTemperature()
-    humid::H = ZeroInitially()
```
milankl, line 49: but we have a breaking change for all other initial conditions anyway no? Like `ZonalWind()` was possible now it's not anymore, so I'd rather go all the way and have it conistent?

src/dynamics/initial_conditions.jl
maximilian-gelbrecht: That's not important. Those structs get never transferred to a kernel.

src/dynamics/initial_conditions.jl
maximilian-gelbrecht: With the functors we don't need it, but here it's not a functor.

src/dynamics/initial_conditions.jl
maximilian-gelbrecht: What exactly do you mean? What do you want to be more consistent here with the `StartFromRest`?

PR #928, opened 2025-11-24
More fixes for the initial conditions 

maximilian-gelbrecht:
I fixed all initial conditions in the SBahn ;) 

Except for the ZonalJet, there I am not even sure exactly what it's supposed to do, so maybe you could have a look @milankl

### DIFF
--- a/src/dynamics/coriolis.jl
+++ b/src/dynamics/coriolis.jl
@@ -17,7 +17,7 @@ function initialize!(coriolis::Coriolis, model::PrimitiveEquation)
     initialize!(coriolis, model_parameters)
 end
 
-function initialize!(coriolis::Coriolis, model)
+function initialize!(coriolis::Coriolis, model::AbstractModel)
     (; radius, rotation) = model.planet
     (; sinlat) = model.geometry
 
--- a/src/dynamics/initial_conditions.jl
+++ b/src/dynamics/initial_conditions.jl
@@ -20,33 +20,35 @@ function initialize!(
     return nothing
 end
 
-InitialConditions(::Type{<:Barotropic}) = InitialConditions(; vordiv = RandomVelocity())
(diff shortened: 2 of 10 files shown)


src/dynamics/coriolis.jl
```diff
@@ -17,7 +17,7 @@ function initialize!(coriolis::Coriolis, model::PrimitiveEquation)
     initialize!(coriolis, model_parameters)
 end
 
-function initialize!(coriolis::Coriolis, model)
+function initialize!(coriolis::Coriolis, model::AbstractModel)
```
milankl, line 20: had that ambiguity at some point too 🙈

src/dynamics/initial_conditions.jl
```diff
@@ -434,7 +434,9 @@ function initialize!(
     return nothing
 end
 
-struct JablonowskiVorticity{NF}
+
+# <: Function so that it works with our `set!`
+struct JablonowskiVorticity{NF} <: Function
```
milankl, line 439: Ouh, I was wondering what supertype to use, at first this feels hacky, but I don't think it is?

milankl:
Awesome, thanks for the fixes here, pulled them in already to test them!

src/dynamics/coriolis.jl
milankl: Ah yes but the problem is we have to remove the function barrier ...

src/dynamics/initial_conditions.jl
maximilian-gelbrecht: Yeah it does feel a bit hacky, but i also found some people discussing exactly that on Julia discourse and saying it's basically okay to do it like that.  Function doesn't expect any other things to be implemented, and there seem to be no other consequences of subtyping it.

maximilian-gelbrecht: Because if you remove the explicit `Function` from our `set!` then you also get all kinds of ambiguities, so this hack is preferable in my opinion

PR #930, opened 2025-11-28
Adapt for models for GPU parametrization

maximilian-gelbrecht:
I couldn't figure out how to do it with the NamedTuples. 

I like this approach of actually having a model type in the parametrization just with most field `nothing` conceptually actually better. It's simpler. I saves some of the dummy types etc and some other code. 

The type instability in Julia 1.11 that @code_warntype reported is gone. Performance is similar. With this we are down to ~800 bytes allocation on CPU and ~5% better performance, on GPU it slightly increased to 250KiB, and performance is ~10% worse. The problem on GPU was that apperatenly too complex` @genreated ` functions don't work any more on GPU?  I think we can probably still fine-tune this a bit. 

The model types are now all non-mutable with this change.


maximilian-gelbrecht:
Succeeded by #931

PR #931, opened 2025-11-28
GPU parametrization: adapt but tuples

maximilian-gelbrecht:
That's now a compromise solution between #930 and the current #873 

It uses adapt and NamedTuples for something, but not everywhere and it simplifies them enough that also the Julia 1.11 compiler seems to be able to infer everything. It also still reduces allocations a tiny bit compared to #873.


maximilian-gelbrecht:
Performance on 1.12 is still worse than on 1.11 though. I've tried adding a `@simd` and profiling with `@code_warntype`, Cthulhu and JET. All look okay and I can't see a difference.

src/physics/convection.jl
```diff
@@ -202,16 +202,6 @@ function pseudo_adiabat!(
 
     while buoyant && k > 1                  # calculate moist adiabat while buoyant till top
         k -= 1                              # one level up
-            
-        if !saturated                       # if not saturated yet follow dry adiabat
```
milankl, line 206: Ouh, moving this into a simpler `if saturated ... else ... end` structure is not the same though! Starting with `saturated = false` in the first pass saturated might be immediately set to true in which case for that same layer the `if saturated` is being executed. I need to revisit in what order the algorithm is actually supposed to execute. What was the problem here that made you change this?

src/physics/vertical_diffusion.jl
```diff
@@ -80,7 +74,7 @@ end
 
 # function barrier
 parameterization!(ij, diagn, progn, diffusion::BulkRichardsonDiffusion, model) =
-    vertical_diffusion!(ij, diagn, diffusion, model.atmosphere, model.planet, model.orography, model.geopotential, model.class)
+    vertical_diffusion!(ij, diagn, diffusion, model.atmosphere, model.planet, model.orography, model.geopotential, model.model_class)
```
milankl: just name it `model.class` ?

src/physics/tendencies.jl
```diff
@@ -9,23 +9,23 @@ function parameterization_tendencies!(
     (; time) = progn.clock
     cos_zenith!(diagn, time, model)
 
-    model_parameters = get_model_parameters(model)      # subset of GPU-compatible model components
-    parameterizations = get_parameterizations(model)    # subset of model: parameterizations only
+    #model_parameters = get_model_parameters(model)      # subset of GPU-compatible model components
+    #parameterizations = get_parameterizations(model)    # subset of model: parameterizations only
```
milankl: remove?

src/physics/tendencies.jl
```diff
@@ -50,18 +50,18 @@ function parameterization_tendencies_cpu!(diagn, progn, parameterizations, model
...
     return Expr(:block, calls...)
 end
 
 # Use @generated to unroll NamedTuple iteration at compile time also on CPU for performance
 @generated function _call_parameterizations_cpu!(diagn, progn, parameterizations::NamedTuple{names}, model) where {names}
-    calls = [:(
-        for ij in 1:model.geometry.npoints      # horizontal grid points inner loop
-            parameterization!(ij, diagn, progn, parameterizations.$name, model)
+    calls = [quote
+        @simd for ij in 1:model.geometry.npoints      # horizontal grid points inner loop
```
milankl: Does @simd make a difference? Or was that just experimental?

src/physics/tendencies.jl
```diff
@@ -50,18 +50,18 @@ function parameterization_tendencies_cpu!(diagn, progn, parameterizations, model
...
+            @inbounds parameterization!(ij, diagn, progn, parameterizations.$name, model)
```
milankl: These would need an explicit `@propagate_inbounds` on the callee site no?

src/models/primitive_wet.jl
```diff
@@ -133,6 +137,11 @@ $(TYPEDFIELDS)"""
                                 :stochastic_physics,
                             )
     extra_parameterizations::TS3 = (:solar_zenith, :land, :ocean)
+
+    # DERIVED 
+    # used to infer parameterizations at compile-time 
+    params::PV = Val(parameterizations)
+    model_class::MC = PrimitiveWetDummy()
```
milankl: just `class` for `model.class`?

milankl (commented):
Grand! I like this. Looks simpler too. But see comment on convection below

src/physics/convection.jl
maximilian-gelbrecht: Oh, sorry, this was flagged by JET with a warning that `temp_parcel` might be set, but I should have paid better attention, because that was just wrong.

maximilian-gelbrecht: I reversed that again.

src/physics/tendencies.jl
maximilian-gelbrecht: No, I tried it, but it makes no difference, took it out again.

src/physics/tendencies.jl
maximilian-gelbrecht: Took that out again completely as well, didn't do anything for performance, and it's probably better do explicitly denote `@inbounds` in the parmaetrizaitions where it fits (we mostly do that already).

milankl (approved):
Happy for you to merge it!

PR #933, opened 2025-12-01
GPU parametrizations with propagate inbounds

maximilian-gelbrecht:
This addresses performance issues with the ij-parametrizations and Julia 1.12

### DIFF
--- a/test/physics/all_parametrizations.jl
+++ b/test/physics/all_parametrizations.jl
@@ -1,0 +1,19 @@
+@testset "Test all parametrizations with boundschecks" begin 
+    model_classes = (PrimitiveWetModel, PrimitiveDryModel)
+
+    spectral_grid = SpectralGrid()
+    for model_class in model_classes
+        
+        model = model_class(spectral_grid)
+
+        simulation = initialize!(model)
+        run!(simulation, steps=4)
+        
+        # call parameterization without @inbounds to check for boundserrors 
+        _call_parameterizations_cpu!(diagn, progn, get_parameterizations(model), model)
+        
+        for key in keys(diagn.physics)
+            @test !any(isnan.(diagn.physics[key]))
+        end
+    end
+end
--- a/test/physics/all_parametrizations.jl
+++ b/test/physics/all_parametrizations.jl
@@ -10,7 +10,7 @@
         run!(simulation, steps=4)
         
         # call parameterization without @inbounds to check for boundserrors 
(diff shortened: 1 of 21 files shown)


maximilian-gelbrecht:
Testing the performance of 

(no inbounds) 
```Julia 
@btime SpeedyWeather._call_parameterizations_cpu!(diagn, progn, SpeedyWeather.get_parameterizations(model), model)
```

against 

(inbounds) 
```Julia 
@btime SpeedyWeather.parameterization_tendencies!(diagn, progn, model)
```

shows that even the non-inbounds version is faster than before in Julia 1.12. I assume this is due to `@propagate_inbounds` inlining all functions.

test/physics/all_parametrizations.jl
```diff
@@ -0,0 +1,19 @@
...
+        model = model_class(spectral_grid)
+
+        simulation = initialize!(model)
+        run!(simulation, steps=4)
+        
+        # call parameterization without @inbounds to check for boundserrors 
+        _call_parameterizations_cpu!(diagn, progn, get_parameterizations(model), model)
+        
+        for key in keys(diagn.physics)
+            @test !any(isnan.(diagn.physics[key]))
```
maximilian-gelbrecht, line 16: I just super quickly wrote this test, we do expect `NaN` in ocean and land, right? So they should be excluded, but only exactly those, right?

milankl: With the surface fluxes being affected by NaNs in ocean and land I've currently been following the philosophy that the surface fluxes would be set to 0 in those cases. Ideally the land-sea mask and the dataset being read in for the, e.g. soil temperature would align so the NaN would never actually propagate but that's not the case and difficult to guarantee when loading in datasets from netCDF.

So yes, `prognostic_variables.ocean.sea_surface_temperature` or `prognostic_variables.land.soil_temperature` and `.soil_moisture` can contain NaNs but they shouldn't propagate into the diagnostic variables!

milankl (approved):

test/physics/all_parametrizations.jl
milankl: With the "internal" land/ocean/sea-ice models I want to follow a philosophy whereby each has a reasonable fallback value defined for where it's not defined/missing/NaN. I want the land-sea-mask to do the job of masking not NaNs. So the mask should be the single point defining what's ocean what's land. This also has the nice side effect that we can at any point change the land-sea mask without running into problems and newly exposed land just starts being dynamic with that fallback temperature/soil moisture being the initial condition.

PR #935, opened 2025-12-03
Follow-up to #783: finalize @param/@component markers

nviebig:
This PR builds on and unblocks the work started in #783 by carrying forward the @param and @component marker approach onto the current main branch and resolving conflicts.


milankl:
Just created a new "parameters" label for this

nviebig:
Fix parameter reconstruction to pass on not tunable parameters

I encountered an issue when trying to reconstruct models from parameter vectors using the `reconstruct()` function. The problem was that non-tunable params (like `nlayers::Int` in e.g `BulkRichardsonDiffusion`) weren't passed on during reconstruction, causing `UndefKeywordError` exceptions.

The issue occurred in two places:

1. In the `@parameterized` `setproperties` override: it was only passing the parameter patch to the constructor, losing non-parameter fields. Fixed by merging original properties with the patch using `SpeedyParameters.getproperties(obj)`.

2. In the `reconstruct()` function: it didn't handle edge cases like Nothing fields or empty parameter arrays. I added fallback methods

Now the parameter workflow is:
  - Vectorize parameters: `param_vec = vec(params)`
  - Modify e.g. gravity: `param_vec.planet.gravity = 10.0f0`
  - Reconstruct: `new_model = reconstruct(model, param_vec)`

@bgroenks96 Niklas found this issue hence we're requesting a review from you if you find a moment! thanks!

nviebig:
image

**Cloud albedo parameter sweep showing outgoing shortwave radiation sensitivity across 15 cloud reflectivity values from 0 to 1.**

EDIT :

Added code which uses the spin up state, not `initialize!()` (which would create a blank new simulation)

```
# Minimal cloud-albedo sweep setup

spectral_grid = SpectralGrid(trunc=23, nlayers=3)
shortwave_radiation = OneBandShortwave(spectral_grid)
model = PrimitiveWetModel(; spectral_grid, shortwave_radiation)

run!(simulation, period=Hour(2))  # short burn-in

(; prognostic_variables, diagnostic_variables, model) = simulation
progn_baseline = deepcopy(prognostic_variables)
diagn_baseline = deepcopy(diagnostic_variables)

cloud_albedos = range(0.0, 1.0, length=15)
results = Dict()

# Sweep cloud_albedo values starting from the burn-in baseline
for (idx, albedo) in enumerate(cloud_albedos)
    param_vec = vec(parameters(model))
    param_vec.shortwave_radiation.clouds.cloud_albedo = albedo
    model_modified = SpeedyWeather.reconstruct(model, param_vec)
    sim_temp = SpeedyWeather.Simulation(deepcopy(progn_baseline), deepcopy(diagn_baseline), model_modified)
    run!(sim_temp, period=Day(10))
    results[albedo] = (
        prognostic = deepcopy(sim_temp.prognostic_variables),
        diagnostic = deepcopy(sim_temp.diagnostic_variables),
        model = model_modified,
    )
end
```

SpeedyWeatherInternals/src/SpeedyParameters/parameters.jl
```diff
@@ -323,11 +325,11 @@ macro parameterized(expr)
     if has_kwdef && typesig.head == :curly
         # handle type arguments; first extract argument names (discarding upper type bounds)
         typeargs = map(typedef2sig, typesig.args[2:end])
-        # then build method signature
-        push!(block.args, esc(:(SpeedyParameters.setproperties(obj::$(typename){$(typeargs...)}, patch::NamedTuple) where {$(typesig.args[2:end]...)} = $(typename){$(typeargs...)}(; patch...))))
+        # then build method signature - merge original properties with patch to preserve non-parameter fields
+        push!(block.args, esc(:(SpeedyParameters.setproperties(obj::$(typename){$(typeargs...)}, patch::NamedTuple) where {$(typesig.args[2:end]...)} = $(typename){$(typeargs...)}(; SpeedyParameters.getproperties(obj)..., patch...))))
```
bgroenks96: It's not clear to me why this is necessary. Indeed it actually shouldn't be given how `setproperties` is supposed to work. Could you provide a MWE of what wasn't working before you made this change?

SpeedyWeatherInternals/src/SpeedyParameters/parameters.jl
```diff
@@ -177,6 +177,8 @@ the nested structure must match that of `obj`. This function is used to reconstr
 @inline reconstruct(obj::NamedTuple{keys,V}, values::NamedTuple{keys,V}) where {keys,V<:Tuple} = values
 # @inline reconstruct(obj, values::ComponentArray) = reconstruct(obj, NamedTuple(values))
 @inline reconstruct(obj, values::SpeedyParams) = reconstruct(obj, stripparams(values))
+@inline reconstruct(obj::Nothing, values::AbstractArray) = obj  # Nothing fields stay Nothing regardless of values
```
bgroenks96: Why the type `AbstractArray` here?

src/models/primitive_dry.jl
```diff
@@ -105,6 +105,26 @@ end
 prognostic_variables(::Type{<:PrimitiveDry}) = (:vor, :div, :temp, :pres)
 default_concrete_model(::Type{PrimitiveDry}) = PrimitiveDryModel
 
+parameters(model::PrimitiveDry; kwargs...) = SpeedyParams(
```
bgroenks96: Why can't you use `@parameterized` on the `PrimitiveDry` type?

src/physics/surface_fluxes/humidity.jl
```diff
@@ -11,6 +11,11 @@ function SurfaceHumidityFlux(
     return SurfaceHumidityFlux(; ocean, land)
 end
 
+parameters(flux::SurfaceHumidityFlux; kwargs...) = SpeedyParams(
```
bgroenks96: again here I would favor `@parameterized` with `@component` unless there is a good reason to override `parameters` directly.

bgroenks96 (changes requested):
Thanks @nviebig and @milankl for picking this back up.

As I commented inline, I don't understand the change to the emitted `setproperties` method dispatches, and I am skeptical that it is the correct fix. But I need to see what the original error was to determine that.

SpeedyWeatherInternals/src/SpeedyParameters/parameters.jl
bgroenks96: Is the problem that some types do not have default kwarg constructors for the non-param values?

bgroenks96: Ah, ok, I see the issue. And I think the original intent behind overriding `setproperties` here was that the default implementation just sets all properties on the object, thus bypassing any potential logic in the default keyword constructors.

However, I'm a bit surprised that the current fix shown here doesn't cause errors? `getproperties(obj)` will return all of the properties, including those possibly in `patch`. So I think you need:

`(; merge(SpeedyParameters.getproperties(obj), patch)...)`

nviebig: Okay, I think youre right! To be safe i added the merge!!

SpeedyWeatherInternals/src/SpeedyParameters/parameters.jl
nviebig: The goal was to handle the case where Nothing fields get passed array values during recursive reconstruction. We want them to stay Nothing regardless. AbstractArray covers any array type that might show up (Vector, ComponentArray, etc.), so the field stays unchanged.

src/physics/surface_fluxes/humidity.jl
nviebig: Got i!! I didnt fully understand the `@component `function at the beginning, but I've now removed all the manual parameters() overrides and changed the models and components to use `@parameterized` with `@component` markers

src/models/primitive_dry.jl
nviebig: Just changed it!

src/physics/albedo.jl
```diff
@@ -136,9 +136,9 @@ end
 ## OceanSeaIceAlbedo
 export OceanSeaIceAlbedo
 
-@kwdef struct OceanSeaIceAlbedo{NF} <: AbstractAlbedo
-    albedo_ocean::NF = 0.06
-    albedo_ice::NF = 0.6
+@parameterized @kwdef struct OceanSeaIceAlbedo{NF} <: AbstractAlbedo
+    @param albedo_ocean::NF = 0.06 (bounds = 0..1,)
```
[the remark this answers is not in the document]
milankl: Runic should decide this not copilot

src/physics/boundary_layer.jl
```diff
@@ -27,15 +27,15 @@ export BulkRichardsonDrag
 """Boundary layer drag coefficient from the bulk Richardson number,
 following Frierson, 2006, Journal of the Atmospheric Sciences.
 $(TYPEDFIELDS)"""
-@kwdef struct BulkRichardsonDrag{NF} <: AbstractBoundaryLayer
+@parameterized @kwdef struct BulkRichardsonDrag{NF} <: AbstractBoundaryLayer
     "von Kármán constant [1]"
-    κ::NF = 0.4
+    κ::NF = 0.4
```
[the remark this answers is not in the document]
milankl: I agree with your decision to leave some "true" physical constants out. von Kárman is an interesting one because it's not precisely defined but still happy to leave it out

src/physics/convection.jl
```diff
@@ -294,12 +294,12 @@ The simplified Betts-Miller convection scheme from Frierson, 2007,
...
 Fields and options are
 $(TYPEDFIELDS)"""
-@kwdef struct DryBettsMiller{NF, SP} <: AbstractConvection
+@parameterized @kwdef struct DryBettsMiller{NF, SP} <: AbstractConvection
     "[OPTION] Relaxation time for profile adjustment"
     time_scale::Second = Hour(4)
 
     "[OPTION] Surface perturbation of temp to calculate the dry adiabat"
-    surface_temp::SP    # don't set default here but in generator below
+    @param surface_temp::SP    # don't set default here but in generator below
```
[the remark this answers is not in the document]
milankl: don't worry about these they are flying out anyway. But this shows another aspect of `@param` ... we may regard numerical values as parameters but what about booleans or types (e.g. maybe for an optimization it's better to leave out stratocumulus clouds, or you implemented the `SaturatingSnowCover` and `LinearSnowCover`) how would one optimize which one to use? Beyond the scope of this PR but interesting to keep in mind

src/physics/vertical_diffusion.jl
```diff
@@ -10,20 +10,20 @@ end
 vertical_diffusion!(::ColumnVariables, ::Nothing, ::PrimitiveEquation) = nothing
 
 export BulkRichardsonDiffusion
-@kwdef struct BulkRichardsonDiffusion{NF, VectorType} <: AbstractVerticalDiffusion
+@parameterized @kwdef struct BulkRichardsonDiffusion{NF, VectorType} <: AbstractVerticalDiffusion
     nlayers::Int
 
     "[OPTION] von Kármán constant [1]"
     κ::NF = 0.4
```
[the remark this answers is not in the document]
milankl: see above

milankl (commented):
I'm somewhat confused about the usage of `@component` ... some components are marked as such others aren't ... what's the motivation?

nviebig:
> I'm somewhat confused about the usage of `@component` ... some components are marked as such others aren't ... what's the motivation?

@milankl I tried to mark only components that contain `@param` fields. Components like geometry, coriolis, orography aren't marked because i thought they have no tunable parameters

bgroenks96:
@nviebig That makes sense, but maybe for consistency's sake, and in case someone implements a different one with parameters, they should still be marked with `@component`?

nviebig:
> @nviebig That makes sense, but maybe for consistency's sake, and in case someone implements a different one with parameters, they should still be marked with `@component`?

Okay of course! Then I adjust that!

SpeedyWeatherInternals/src/SpeedyParameters/parameters.jl
bgroenks96: Ok, but why are nothing fields getting passed array values? Can you provide an example of where you ran into this?

src/models/primitive_wet.jl
```diff
@@ -57,43 +57,43 @@ $(TYPEDFIELDS)"""
...
+    @component vertical_diffusion::VD = BulkRichardsonDiffusion(spectral_grid)
+    @component surface_thermodynamics::SUT = SurfaceThermodynamicsConstant(spectral_grid)
+    @component surface_wind::SUW = SurfaceWind(spectral_grid)
+    @component surface_heat_flux::SH = SurfaceHeatFlux(spectral_grid)
+    @component surface_humidity_flux::HF = SurfaceHumidityFlux(spectral_grid)
+    @component large_scale_condensation::LSC = ImplicitCondensation(spectral_grid)
+    @component convection::CV = SimplifiedBettsMiller(spectral_grid)
+    @component shortwave_radiation::SW = TransparentShortwave(spectral_grid)
+    @component longwave_radiation::LW = JeevanjeeRadiation(spectral_grid)
     stochastic_physics::SP = nothing
```
bgroenks96: Why not the `stochastic_physics`?

bgroenks96 (approved):

src/models/primitive_wet.jl
nviebig: Must have slipped through!

nviebig: > Ok, but why are nothing fields getting passed array values? Can you provide an example of where you ran into this?


Youre right, i initially thought this was a problem, but i removed it!

milankl (approved):
@bgroenks96 I approve these changes but would like them to get merged only after #873 because I don't want to fix conflicts there but rather here. Do you approve the changes to SpeedyWeatherInternals given they also affect Terrarium?

milankl:
@nviebig sorry with the GPU pull request merged there's a lot of conflicts now, but I don't think any of them will be hard to resolve. Always take the new struct definitions and put the `@param` in front again ... sorry!

milankl (approved):

PR #937, opened 2025-12-04
Lorenz N-Cycle time stepping

### LINKED ISSUE #917: Lorenz N-cycle time integration
@karinakowalczyk and I have been talking about upgrading the time integration scheme we currently use. Some constraints

- we can store several time steps or tendencies in spectral space, compared to grid space that's small, but ideally we only have one version of the variables in grid space (we currently have two with the `_prev` copy)
- the new time stepping should work with the semi-implicit linear dynamics, ideally in a Crank-Nicolson $\alpha = 1/2$ style
- the new scheme should allows us to continue using a ~40min timestep per tendency evaluation, otherwise the model will substantially slow down
- the new scheme ideally would do something better than solving the parameterizations in Euler forward as we currently do

Candidates

1. SIL3 as used in NeuralGCM: [Whitaker and Kar, 2013](https://journals.ametsoc.org/view/journals/mwre/141/10/mwr-d-13-00132.1.xml)
2. 3 or 4-cycle Lorenz combined with CN in variant A or B, suggested by [Hotta et al. 2016](https://journals.ametsoc.org/view/journals/mwre/144/6/mwr-d-15-0330.1.xml)

I'm in favour of 2 as it requires less storage than SIL3 and otherwise feels like a nice drop-in replacement, I believe we essentially just need to update the `leapfrog!` function (well probably define a `NCycleLorenz <: AbstractTimeStepper`) and make some adjustments to which terms are computed at what "leapfrog" time step.

@hottad Long time no speak! Hope all is well. We've made great progress with SpeedyWeather but still use RAW-filtered leapfrogging. It looks like your ImL4+CN-A (or B, AB, ABBA, chapeau for getting an ABBA acronym into a paper 🎵 🪩) should be easy to implement as an alternative probably with backwards compatibility if we wanted. It's been some years since you wrote this paper, is this still what you would suggest and if we created a pull request would you be happy to review it?

karinakowalczyk:
First implementation of the Lorenz N-cycle semi-implicit implementation, see #917

Questions to be worked on

- [ ] Backwards compatibility to Leapfrog?
- [ ] Where to store the G arrays that hold the previous tendency?
- [ ] Do we let a simulation always run to the end of a 3 or 4-cycle integration?
- [ ] Should the ocean, sea ice and land time stepping be integrated into the Lorenz N-Cycle as well?


src/dynamics/implicit.jl
```diff
@@ -371,4 +371,167 @@ function implicit_correction!(
...
+function lorenz_implicit_correction!( #TODO: avoid code duplications -> dispatch
+    diagn::DiagnosticVariables,
+    progn::PrognosticVariables,
+    implicit::ImplicitShallowWater,
+    model::ShallowWater,
+)
+    # Get G (accumulated tendency) and x (current state)
+    # Handle dimensionality: div is 3D, pres is 2D
+    div_G = progn.div[:, :, 2]
+    div_x = progn.div[:, :, 1]
```
milankl: these are allocating you can use `LowerTriangularArrays.lta_view` like a normal view that retains the `LowerTriangularArray` type though. We thought about extending `Base.view` directly but haven't decided on that yet (I think it's safe). Interface is the same

src/dynamics/time_integration.jl
```diff
@@ -301,6 +278,347 @@ function first_timesteps!(
...
+    "[INTERNAL] Current substep within the N-cycle (0 to cycles-1)"
+    substep::Int = 0
+    
+    "[INTERNAL] Total step counter for variant tracking (used by AB and ABBA)"
+    step_counter::Int = 0
+end
+
+"""$(TYPEDSIGNATURES)
+Generator function for NCycleLorenz struct using `spectral_grid` for resolution."""
+function NCycleLorenz(spectral_grid::SpectralGrid; cycles=4, variant=NCycleLorenzA(), kwargs...)
```
milankl: duplicated defaults?

src/dynamics/time_integration.jl
```diff
@@ -301,6 +278,347 @@ function first_timesteps!(
...
+    α::NF = 0.5
+    
+    "[OPTION] Time step in minutes for T31, scale linearly to `trunc`"
+    Δt_at_T31::Second = Minute(30)
+
+    "[OPTION] Adjust `Δt_at_T31` with the `output_dt` to reach `output_dt` exactly"
+    adjust_with_output::Bool = true
+
+    "[OPTION] No Euler first step needed"
+    first_step_euler::Bool = false           # Not used for Lorenz, but needed for compatibility
```
milankl: `first_step_euler(L::Leapfrog) = L.first_step_euler`, `first_step_euler(L::NCycleLorenz) = ...`?

src/dynamics/time_integration.jl
```diff
@@ -301,6 +278,347 @@ function first_timesteps!(
...
+
+    L.Δt_millisec = get_Δt_millisec(L.Δt_at_T31, L.trunc, radius, L.adjust_with_output, output_dt)
+    L.Δt_sec = L.Δt_millisec.value/1000
+    L.Δt = L.Δt_sec/radius
+
+    n = round(Int, Millisecond(output_dt).value/L.Δt_millisec.value)
+    nΔt = n*L.Δt_millisec
+    if nΔt != output_dt
+        @warn "$n steps of Δt = $(L.Δt_millisec.value)ms yield output every $(nΔt.value)ms, but output_dt = $(output_dt.value)ms"
+    end
```
milankl: this feels duplicated wondering whether we should have a shared `Timestep` type for `Leapfrog` and `NCycleLorenz`? Then `initialize!(L::Leapfrog) = initialize!(L.timestep)` or similar?

src/dynamics/time_integration.jl
```diff
@@ -301,6 +278,347 @@ function first_timesteps!(
...
+# Weight coefficient functions (Hotta et al. 2016, Eqs 2-3, 7-8)
+"""$(TYPEDSIGNATURES)
+Compute weight coefficient w for current substep."""
+function weight_coefficient(L::NCycleLorenz)
+    k = L.substep
+    return weight_coefficient(L.variant, k, L.cycles, L.step_counter)
+end
+
+function weight_coefficient(::NCycleLorenzA, k::Int, N::Int, ::Int)
+    return k == 0 ? 1.0 : Float64(N) / Float64(N - k)
```
milankl: Shall we either use `NF` from spectral grid somehow or `Float32` as default?

milankl (commented):
Some comments for now, more tomorrow!

src/dynamics/implicit.jl
maximilian-gelbrecht: We have `get_step(.... , lf_index)` for that :)  (which calls `lta_view`)

src/dynamics/time_integration.jl
```diff
@@ -301,6 +278,347 @@ function first_timesteps!(
...
+    nsteps::Int = 2
+    
+    "[OPTION] Number of cycles N (3 or 4 recommended, 4 is more stable)"
+    cycles::Int = 4
+    
+    "[OPTION] Variant: NCycleLorenzA() (default), B, AB, or ABBA"
+    variant::V = NCycleLorenzA()
+    
+    "[OPTION] Centering parameter: 0.5=Crank-Nicolson (2nd order), 1.0=Backward Euler (1st order)"
+    α::NF = 0.5
```
milankl: should that not stay in `model.implicit`?

src/dynamics/time_integration.jl
```diff
@@ -301,6 +278,347 @@ function first_timesteps!(
...
+    k = L.substep
+    return weight_coefficient(L.variant, k, L.cycles, L.step_counter)
+end
+
+function weight_coefficient(::NCycleLorenzA, k::Int, N::Int, ::Int)
+    return k == 0 ? 1.0 : Float64(N) / Float64(N - k)
+end
+
+function weight_coefficient(::NCycleLorenzB, k::Int, N::Int, ::Int)
+    return k == 0 ? 1.0 : Float64(N) / Float64(k)
```
milankl: don't hardcode Float64, whether explicit `Float64` or implicit `1.0`. I see that you don't have the number format information available, what about

```julia
function weight_coefficient(::Type{NF}, ::NCycleLorenzB, k, N, step) where NF
    return k == 0 ? one(NF) : convert(NF, N) / convert(NF, k)
end

# float32 fallback for all `weight_coefficient`
weight_coefficient(V::NCycleLorenzVariant, k, N, step) = weight_coefficient(DEFAULT_NF, V, k, N, step)
```

for type stability? You can explicitly `@inline` them but Julia very likely will do that anyway

src/dynamics/time_integration.jl
```diff
@@ -301,6 +278,347 @@ function first_timesteps!(
...
+    if iseven(cycle_num)
+        return weight_coefficient(NCycleLorenzA(), k, N, step_counter)
+    else
+        return weight_coefficient(NCycleLorenzB(), k, N, step_counter)
+    end
+end
+
+function weight_coefficient(::NCycleLorenzABBA, k::Int, N::Int, step_counter::Int)
+    if N != 4
+        @warn "ABBA variant only for 4-cycle, using variant A"
```
milankl: This is dangerous but depends on where `weight_coefficient` is currently called. I see it's called outside a kernel but if it was inside then the `if` in combination with `@warn` will likely kill any vectorization as it may need to inject some warning throwing code here. I'd probably safeguard this at a higher level and silently ignore this here?

src/dynamics/time_integration.jl
```diff
@@ -301,6 +278,347 @@ function first_timesteps!(
...
+    )
+    
+    return nothing
+end
+
+@kernel inbounds=true function accumulate_tendency_kernel!(
+    G, F_explicit, @Const(w)
+)
+    lmk = @index(Global, Linear)
+    G[lmk] = w * F_explicit[lmk] + (1 - w) * G[lmk]
```
milankl: I think you can make your life easy and broadcast this

```julia
G .= w * F_explicit .+ (1-w) * G
```

src/dynamics/time_integration.jl
```diff
@@ -301,6 +278,347 @@ function first_timesteps!(
...
+    )
+    
+    return nothing
+end
+
+@kernel inbounds=true function update_state_kernel!(
+    x, G, @Const(dt)
+)
+    lmk = @index(Global, Linear)
+    x[lmk] = x[lmk] + dt * G[lmk]
```
milankl: Same here
```julia
x .+= dt*G
```

but the advantage of writing this as a kernel I see is that you could do both in one kernel, e.g.

```julia
G[lmk] = w * F_explicit[lmk] + (1 - w) * G[lmk]
x[lmk] = x[lmk] + dt * G[lmk]
```

but question: Wouldn't the `F_explicit` here already need to contain the implicit correction?

src/dynamics/time_integration.jl
```diff
@@ -301,6 +278,347 @@ function first_timesteps!(
...
+    w_current = weight_coefficient(L)
+    
+    # Step 1: Accumulate weighted explicit tendencies (Eq. 19)
+    for (varname, tendname) in zip(prognostic_variables(model), tendency_names(model))
+        var = getfield(progn, varname)
+        
+        # Get G based on dimensionality
+        if ndims(var) == 2
+            # 2D array (e.g., pres in shallow water): direct access
+            G = var[:, 2]
```
milankl: this is allocating, use `lta_view` for non-allocating and I think we have `get_step` for that!

src/dynamics/time_integration.jl
```diff
@@ -301,6 +278,347 @@ function first_timesteps!(
...
+            var_tend = tend.tracers_tend[name]
+            spectral_truncation!(var_tend)
+            
+            accumulate_tendency!(G, var_tend, w_current, L)
+        end
+    end
+    
+    # Step 2: Apply implicit correction (Eq. 20)
+    # This modifies G in-place to become dx = (I - α*Δt*L_I)^(-1) * (G + L_I*x)
+    lorenz_implicit_correction!(diagn, progn, model.implicit, model)
```
milankl: this would normally be called within `timestep!` directly after `dynamics_tendencies!` and actually before the horizontal diffusion or do we need to order this differently here?

src/dynamics/time_integration.jl
```diff
@@ -301,6 +278,347 @@ function first_timesteps!(
...
+            update_state!(x, G, dt, L)
+        end
+    end
+    
+    # Update counters
+    L.substep += 1
+    L.step_counter += 1
+    if L.substep >= L.cycles
+        L.substep = 0
+    end
```
milankl: I'm wondering whether this should be in the clock? `clock.timestep_counter` seems like it already does the job of `step_counter`? And then substep could be computed on the fly?

src/dynamics/time_integration.jl
```diff
@@ -301,6 +278,347 @@ function first_timesteps!(
...
+    
+    # Evolve random pattern
+    random_process!(progn, model.random_process)
+    
+    return nothing
+end
+
+"""$(TYPEDSIGNATURES)
+Initialize G (tendency accumulator) to zero at start of integration."""
```
milankl: Is that not normally the case anyway? And for restart/continuation purposes this can start from non-zero no?

src/dynamics/time_integration.jl
```diff
@@ -316,15 +634,27 @@ function timestep!(
...
-
-    # PARTICLE ADVECTION (always skip 1st step of first_timesteps!)
-    not_first_timestep = lf2 == 2
-    not_first_timestep && particle_advection!(progn, diagn, model)
+    # Check which time stepper we're using and dispatch accordingly
+    if model.time_stepping isa NCycleLorenz
+        # Lorenz N-cycle: always use index 1 for current state
+        dynamics_tendencies!(diagn, progn, 1, model)
+        horizontal_diffusion!(diagn, progn, model.horizontal_diffusion, model)
+        lorenz_ncycle_timestep!(progn, diagn, diagn.tendencies, dt, model)
```
milankl: maybe we should introduce `step!` function that dispatches over `model.time_stepping`?

src/dynamics/time_integration.jl
```diff
@@ -340,23 +670,31 @@ function timestep!(
...
-    transform!(diagn, progn, lf2, model)
-    
-    # PARTICLE ADVECTION (always skip 1st step of first_timesteps!)
-    not_first_timestep = lf2 == 2
-    not_first_timestep && particle_advection!(progn, diagn, model)
+    if model.time_stepping isa NCycleLorenz
+        # Lorenz N-cycle path
+        dynamics_tendencies!(diagn, progn, 1, model)
+        horizontal_diffusion!(diagn, progn, model.horizontal_diffusion, model)
+        # Note: implicit correction is called inside lorenz_ncycle_timestep!
```
milankl: why?

milankl (commented):
Went through the rest! I like what you have started to construct but I'm also generally wondering whether we actually want backwards compatibility with leapfrogging? 

- there seems to be quite some annoying `if model.time_stepping isa NCycleLorenz` business going on and this is only the most obvious parts
- e.g. there's other execution orders that are more subtle yet hardcoded to the leapfrogging
- the `_prev` arrays would only need to be allocated with leapfrogging not with ncycle lorenz and so we'd also need to find a way to handle that elegantly

So thinking about this more, I'm not convinced backwards compatibility should be a goal. What if you just break it, start from the BarotropicModel see whether this works with time stepping, then build up to the ShallowWaterModel with the implicit corrections there and when that works we can for the primitive models?

milankl:
also note that the changelog.md needs to be updated 😉 that's what currently fails!

milankl:
Also CI passes because you're bypassing all the lorenz timestepping stuff right?

src/dynamics/time_integration.jl
karinakowalczyk: Looking at the paper (equations (18) - (21)), F^E contains all explicit terms, which in my understanding also includes the horizontal diffusion and then the implicit correction is applied to the full accumulated tendencies. But I am probably misunderstanding this?

maximilian-gelbrecht:
What's actually the status here?

@karinakowalczyk have you had any chance to look at this again?

karinakowalczyk:
I had a look again this week (after a while, sorry!). The barotropic model/explicit part of the scheme seems to be working. I'm still having issues in the shallow water model which I think should come from the implicit correction part. I wasn't really able to figure out what's wrong there. Will try again next week. If I'll still be stuck then, maybe we could have a meeting to discuss this at some point?

maximilian-gelbrecht:
Yes, happy to help and discuss! 

One thing to just be aware of is that we now have a major revision of the variable system coming up in #969 that will be merged very soon. This won't really directly interfere with the time stepping itself, but we won't have the `PrognosticVariables` and `DiagnosticVariables` separately anymore. Instead we just have one `Variables` and then you'd do `vars.prognostic.vor` instead of `progn.vor` or `vars.grid.u` instead of `diagn.grid.u`. 

With this new variable system we can also allocate variables much more flexible, so the two time steps of the Leapfrog solver aren't hardcoded into model anymore, and we could allocate the prognostic variables differently for the Lorenz N-cycle time stepper. That could be really useful.

maximilian-gelbrecht:
Why was this merged already? Just a mistake?

milankl:
So I merged this accidentally, I don't really know how this happened I didn't even press the "merge" button here but was just doing things in VS Code ... I would normally revert this commit but I actually had set up everything so that the additional functionality Karina wrote wasn't executed by default so everything still passes on main. In that sense I might look at https://github.com/SpeedyWeather/SpeedyWeather.jl/tree/lorenz-ncycle-restructure and open a new pull request from there ?!

maximilian-gelbrecht:
So keeping it merged? I had a quick look over it just now. I don't mind.

Issue #938, opened 2025-12-05
Find Enzyme+Speedy issue in old uv_from_vordiv!


maximilian-gelbrecht:
In past versions of  `uv_from_vordiv!` Enzyme produced an error when multiplying by `im` in another line. Currently we avoid this issue by just doing the operation slightly differently, but mathematically identically. 

This here is primarily a memo to myself to investigate this for a MWE for Enzyme. 

Linked to #893

maximilian-gelbrecht:
Resolved by #971

maximilian-gelbrecht:
I am reopening this issue because my impression is that the unit test failing more often than before since merging #971

maximilian-gelbrecht:
But it's still pretty unclear how to best debug or isolate this issue unfortunately.

maximilian-gelbrecht:
This sees to be solved now, with upstanding 1.12 issues bundled in #1051

PR #940, opened 2025-12-10
Debugging uv_from_vordiv! differentiability 

maximilian-gelbrecht:
#893 is merged but a problem remains (#938) 

What makes this a tough problem is that the kernel version of `uv_from_vordiv!` just on its own is differentiable, we only get an error when differentiating it within the full model. This is already taking into account that we call it with a `view` form within the model, I tested that separately already. 

I'll try a few different things here.


maximilian-gelbrecht:
First result of trial and error: It works in a `ShallowWaterModel` but not in a `PrimitiveWetModel`

milankl:
> First result of trial and error: It works in a `ShallowWaterModel` but not in a `PrimitiveWetModel

Only difference is really 2D vs 3D here?

maximilian-gelbrecht:
Superseded by #971

PR #941, opened 2025-12-11
[LowerTriangularArrays] new copyto!

maximilian-gelbrecht:
In #940 I saw a problem with `LTA.copyto!`. And while I don't think this will do anything about differentiability here, this just fixes our really messy old versions. It's actually also much faster.

### DIFF
--- a/docs/src/gradients.md
+++ b/docs/src/gradients.md
@@ -59,7 +59,7 @@ using SpeedyWeather, CairoMakie
 # create some data with wave numbers 0,1,2,3,4
 trunc = 64                  # 1-based maximum degree of spherical harmonics
 L = randn(LowerTriangularMatrix{ComplexF32}, trunc, trunc)
-SpeedyTransform.spectral_truncation!(L, 5)  # remove higher wave numbers
+SpeedyTransforms.spectral_truncation!(L, 5)  # remove higher wave numbers
 G = transform(L)
 heatmap(G, title="Some fake data G")        # requires `using CairoMakie`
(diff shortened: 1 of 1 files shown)


maximilian-gelbrecht:
I don't get why the tests are currently failing, they pass locally. Guess I'll fix that tomorrow.

LowerTriangularArrays/src/lower_triangular_array.jl
```diff
@@ -538,88 +538,47 @@ function Base.copyto!(
...
-        ind_L2_smaller[:, mmax_1+1:end] .= 0 
-    end
-    if maximum(ms) < mmax_2 # threshold L2 based on ms
-        ind_L2_smaller[:, ms[end]+1:end] .= 0 
-    end 
-    ind_L2 = ind_L2_smaller[ind_L2] # also flattens the indices into indices for the L2.data array
+end 
 
-    L1.data[ind_L1,[Colon() for i=1:(N-1)]...] = T.(L2.data[ind_L2,[Colon() for i=1:(N-1)]...])
+# version for 2D arrays/vectors
```
milankl, line 571: Could be merged into one kernel with `I = @index(Global, Cartesian)` ?

maximilian-gelbrecht: `Cartesian` can't be indexed like above with ranges, so it wouldn't work

milankl:
Also just bumped the JLArrays version to 0.3, that's fine I guess? Otherwise ready to go from my side!

PR #912, opened 2025-11-14
Code formatting with runic

maximilian-gelbrecht:
#910 

This is formatting with [Runic](https://github.com/fredrikekre/Runic.jl), I'll make another draft PR with JuliaFormatter

PR #944, opened 2025-12-12
EarthDryAtmosphere implemented

milankl:
- BarotropicModel, ShallowWaterModel and PrimitiveDryModel have now a EarthDryAtmosphere but PrimitiveWetModel has a EarthAtmosphere.
- Introduces AbstractWetAtmosphere, AbstractDryAtmosphere super types and the virtual temperature calculation should now always compile away for dry atmospheres.
- Removes `model.class` as this dispatch can now be done via `model.atmosphere`

TODO

- [x] merge `ClausiusClapeyron` struct + functionality with the Wet/Dry atmosphere distinction here to remove. Having both an atmosphere and a clausius clapeyron model component always felt redundant
- [x] broacasting with `::AbstractAtmosphere` 
- [x] Americanize "vapour" -> "vapor" for consistency
- [x] rename `pres_ref` to `pressure_reference` and `temp_ref` to `temperature_reference` to generally move to more telling+explicit variable names


milankl:
Regarding runic formatting from #912 this example is wild

```julia
function foo()
    return @inbounds for k in 1:10
        k^2
    end
end
```

because of the `@inbounds` there's a `return` put before the loop not after, see https://github.com/fredrikekre/Runic.jl/pull/48 but one can bypass by explicitly adding `return nothing` to the last line which I like more.

milankl:
Broadcasting with `::AbstractAtmosphere` via `broadcastable` and `Ref` as outlined here: https://discourse.julialang.org/t/how-to-tell-julia-that-my-type-is-not-iterable/112767/3

Doesn't allocate
```julia
julia> @ballocated $Tv .= SpeedyWeather.virtual_temperature.($T, $q, $atmosphere)
0
```
and has some vectorization in the LLVM code. Looks good to me!

maximilian-gelbrecht:
> Regarding runic formatting from #912 this example is wild
> 
> ```julia
> function foo()
>     return @inbounds for k in 1:10
>         k^2
>     end
> end
> ```
> 
> because of the `@inbounds` there's a `return` put before the loop not after, see [fredrikekre/Runic.jl#48](https://github.com/fredrikekre/Runic.jl/pull/48) but one can bypass by explicitly adding `return nothing` to the last line which I like more.

Yes I already noticed that as well, it's also probably the thing I most dislike about their formatting, but adding an explicit `return` does negate it, yes

milankl:
T127L16 on L40 GPU yields

```julia
  137.220 μs (772 allocations: 100.77 KiB)
```

so runs through without errors and timing is fine. Merge?

PR #950, opened 2025-12-18
Runic Formatting

maximilian-gelbrecht:
Follow up from #912 

SUMMARY: We have decided on Runic formatting for SpeedyWeather and submodules/packages.

- style will not be enforced via mandatory check required to pass for merging
- contributors apply runic formatting manually e.g. in their VS Code via shift+option+F
- diverging from runic style in single PRs is tolerated but if possible avoided
- contributors are expected to occasionally reformat all files for conistency


milankl:
Shall we enable the runic format checking here? I'm thinking about a policy whereby external pull requests could be merged without adhering to the format style but every now and then we could just include a reformatting everything into our pull requests and to make runic pass again

maximilian-gelbrecht:
I can see your point. I think that would be good in principle. 

In practice, once an external pull request is merged the runic CI will then trigger on the first commit a contributor makes. So I don't think every now and then would be possible.

milankl:
> I can see your point. I think that would be good in principle.
> 
> In practice, once an external pull request is merged the runic CI will then trigger on the first commit a contributor makes. So I don't think every now and then would be possible.

I just want this to be easy on our side. I don't mind pressing a few keys in VS Code to get runic formatting but maybe others are put off if they need to install runic first. Should the runic CI also go into this PR?

milankl:
Also I wouldn't even necessarily bother with the sometimes weird placements of `return`. Whenever it bothers us then we can just manually add a `return nothing` at the end? Fits the philosophy of "Runic is no one's favourite"

maximilian-gelbrecht:
Yeah, totally. 

Maybe we are fine without the CI? Makes it easy for external people, and our core collaborators we tell to do it and configure it so they do it automatically.

milankl:
I like that approach too. Fewer tests to pass and if we diverge too much from Runic we can just occasionally retrigger a whole repo formatting marathon like this one here!

maximilian-gelbrecht:
Off topic: Our parametrization PR had like two thirds of the line changes of a PR changing the formatting of the whole repo. Really goes to show that we basically revised the completed repo.

PR #952, opened 2025-12-18
MatrixSpectralTransform implemented

### LINKED ISSUE #948: Implement a MatrixSpectralTransform for up to 100km resolution
Fourier and Legendre transforms are linear operations and therefore can be represented as a matrix, i.e.
```math
coeffs = LT(FFT(field)) = A*field
```
with `A` one big matrix of size NxM with N the number of harmonics (560 at T31) and M the number of grid points. The size of this matrix gets quickly huge with higher resolution (all Float32)

| Resolution | size | GB |
| - | - | - |
| T31 | 560x3168 | 0.014 |
| T127 | 8384x40320 | 2.7GB |
| T511 | 131840x603648 | 637 GB |

Note that 2 of these transforms are needed to avoid a matrix inverse (compute it only once).
So up to T127 / 100km it is interesting to look at this more closely. How fast are such matrix-multiply based transforms?

## On the CPU (M4)

At T31L8 (layers make the matrix-vector multiplication `(NxM) * M` from above simply a matrix-matrix multiplication `(NxM) * (MxK)` with `K` vertical layers but same matrix operator)

```julia
julia> @btime LinearAlgebra.mul!($v, $A, $field);
  6.975 μs (0 allocations: 0 bytes)

julia> @btime transform!($v, $field, $S);
  69.292 μs (314 allocations: 124.94 KiB)
```

So the big matrix transform is 10x faster than the "fast" transform of FFT + Legendre. At higher resolution of T127L16 this difference increases to 25x

```julia
julia> @btime LinearAlgebra.mul!($v, $A, $field);
  172.292 μs (0 allocations: 0 bytes)

julia> @btime transform!($v, $field, $S);
  5.084 ms (1338 allocations: 2.59 MiB)
```

## On the GPU (L40)

The low-resolution T31L8 big matrix transform here is with 5μs ~40% faster than on an M4 (7μs)

```julia
julia> @btime LinearAlgebra.mul!($(v.data), $A, $(field.data));
  5.101 μs (54 allocations: 2.02 KiB)
```

but that cost basically doesn't increase with higher resolution T127L16

```julia
julia> @btime LinearAlgebra.mul!($(v.data), $A, $(field.data));
  7.208 μs (54 allocations: 2.02 KiB)
```

So here we are 700x faster than the CPU-"Fast" transform version.

## Extrapolate to overall simulation speed

- We do 14 3D transforms at T127L16 each 7μs that's 0.1ms per time step
- At T127 we have a 10min time step, so 144 time steps in 1 simulated day
- So transforms for 1 simulated day take 14.4ms or 5s for a simulated year
- 1 simulated year per 5s is 17280 SYPD

This clearly shows that a `MatrixSpectralTransform` would be insanely useful for up to 100km resolution simulations.

TODO

- Find algorithm that can efficiently compute the coefficients in the big matrix during initialization
- Implement a `MatrixSpectralTransform` with the same `transform!` interface as `SpectralTransform`
- Use `MatrixSpectralTransform` automatically for lower resolutions, e.g. throw warning after some GB

milankl:
- [x] generalise the spectral gradient arrays so that both SpectralTransform and MatrixSpectralTransform can use them
- [x] Implement MatrixSpectralTransform as type and define interface like for SpectralTransform
- [x] Compute the entries of the transformation matrices
- [x] `::SpectralTransform` -> `AbstractSpectralTransform` in many occasions  
- [x] check correctness of transform
- [x] check efficiency of `pinv` (Moore-Penrose pseudoinverse)


milankl:
Interface for now

```julia
julia> spectral_grid = SpectralGrid(trunc=31, nlayers=8);

julia> M = MatrixSpectralTransform(spectral_grid)
MatrixSpectralTransform{Float32, CPU{...}, Array, ...}:
├ Spectral:     T31, 33x32 LowerTriangularMatrix{Complex{Float32}}
├ Grid:         Field{Float32}, 48-ring OctahedralGaussianGrid
├ Truncation:   dealiasing = 2 (quadratic)
├ Architecture: CPU
└ Memory:       28.39 MB matrices, 202.80 KB scratch (8 layers)

julia> model = PrimitiveWetModel(spectral_grid, spectral_transform=M);

julia> simulation = initialize!(model);

julia> model.feedback.debug = false
false

julia> run!(simulation, period=Day(10))
Weather is speedy: 100%|███████████████████████████████████████████████████████| Time: 0:00:20 (117.79 years/day)
```

Blows up because transformation matrices are being zero so far (bad for temperature/pressure I guess) and so speed might not be much telling when carrying NaNs around ...

maximilian-gelbrecht:
Already thinking a little bit ahead, we would then later add a default heuristic right? Something like 

```Julia 
default_spectral_transform(spectral_grid::SpectralGrid) = default_spectral_transform(spectral_grid.architecture, spectral_grid.trunc) 

default_spectral_transform(::CPU, trunc) = trunc < CUT_OFF_CPU ? MatrixSpectralTransform : SpectralTransform 
default_spectral_transform(::GPU, trunc) = trunc < CUT_OFF_GPU ? MatrixSpectralTransform : SpectralTransform 
```

I think I'll only have time for a rather superficial review later.

SpeedyTransforms/src/gradient_arrays.jl
```diff
@@ -0,0 +1,177 @@
...
+recursion_factor(l, m) = sqrt((l^2 - m^2) / (4 * l^2 - 1))
+
+"""$(TYPEDSIGNATURES)     
+Returns a matrix of recursion factors `ϵ` up to degree `lmax` and order `mmax` (1-based) of the `spectrum` in number format `NF`."""
+function recursion_factors(
+        ::Type{NF},             # number format NF
+        spectrum::Spectrum,
+    ) where {NF}
+
+    ϵ_lm = zeros(NF, spectrum)                              # store in lower triangular matrix
```
maximilian-gelbrecht: So this is currently only working with CPU spectra because of the loop below. 

So far what we did is just to do these pre computations on CPU and then the constructor of the struct does an automatic conversion in the end. 

This is done below in the `spectral_transforms.jl` where we are specifically declaring a `spectrum_cpu` just for initialisation purposes. 

We could also write kernels for all of these functions, but it's only at initialisation and it's all rather small matrices, so I wouldn't see this as very urgent.

SpeedyTransforms/src/matrix_transform.jl
```diff
@@ -0,0 +1,128 @@
...
+    forward::MatrixComplexType          # forward transform matrix
+    backward::MatrixComplexType         # backward transform matrix
+
+    # SCRATCH MEMORY
+    scratch_memory::MatrixComplexType   # state is undetermined, only read after writing to it
+
+    gradients::GradientType             # precomputed gradient and integration matrices
+end
+
+Architectures.nonparametric_type(::Type{<:MatrixSpectralTransform}) = MatrixSpectralTransform
```
maximilian-gelbrecht, line 44: I guess it's good to have this always as a default, but do we use this somewhere? So far I was mostly thinking of this as useful for `AbstractArray` types.

SpeedyTransforms/src/matrix_transform.jl
```diff
@@ -0,0 +1,128 @@
...
+    norm_sphere::NF                 # normalization of the l=0, m=0 mode
+
+    # THE ACTUAL TRANSFORM MATRICES for forward = LT(FFT(input)) and backward = IFFT(ILT(input))
+    forward::MatrixComplexType          # forward transform matrix
+    backward::MatrixComplexType         # backward transform matrix
+
+    # SCRATCH MEMORY
+    scratch_memory::MatrixComplexType   # state is undetermined, only read after writing to it
+
+    gradients::GradientType             # precomputed gradient and integration matrices
```
maximilian-gelbrecht, line 41: I like this, also for the regular transform. Makes the structs a bit cleaner

SpeedyTransforms/src/matrix_transform.jl
```diff
@@ -0,0 +1,128 @@
...
+
+function transform!(                        # SPECTRAL TO GRID
+        field::AbstractField,               # gridded output
+        coeffs::LowerTriangularArray,       # spectral coefficients input
+        scratch_memory,                     # explicit scratch memory to use
+        M::MatrixSpectralTransform;         # precomputed transform
+        unscale_coslat::Bool = false,       # unscale with cos(lat) on the fly?
+    )
+    # use first n layers of scratch memory
+    scratch = view(scratch_memory, :, 1:size(coeffs, 2))
```
maximilian-gelbrecht: Have to check if this is good on GPU as `views` aren't really a thing for `CuArrays` in the low-level implmenation. Optimised implementations only exist for explicit contiguous memory (`CuArray`) or strided memory (`StridedCuArray`) as far as I know.

SpeedyTransforms/src/matrix_transform.jl
```diff
@@ -0,0 +1,128 @@
...
+    
+    # multiply into scratch which is complex typed and then take real part into field
+    # imaginary part should be zero but destination is used to store intermediate results
+    # explicitly convert to real also for NaN + NaN*im results
+    LinearAlgebra.mul!(scratch, M.backward, coeffs.data)
+    field.data .= real.(scratch)
```
maximilian-gelbrecht: I am always a bit confused about broadcasting operations like that. Is this allocating because of executing `real.(scratch)` first or not?

maximilian-gelbrecht (commented):
Great that this already coming along well so quickly! 🚀

SpeedyTransforms/src/spectral_transform.jl
```diff
@@ -191,78 +177,8 @@ function SpectralTransform(
     # Δϕ = 2π/nlon is the azimuth every grid point covers
     solid_angles = get_solid_angles(grid)
 
-    # RECURSION FACTORS
-    ϵlms = get_recursion_factors(Spectrum(lmax + 1, mmax))    # one more degree needed!
-
-    # GRADIENTS (on unit sphere, hence 1/radius-scaling is omitted)
-    # meridional gradient for scalars (coslat scaling included
-    # precomputations always on CPU, then transfered to the actual architecture used
-    spectrum_cpu = Spectrum(lmax, mmax)
```
maximilian-gelbrecht, line 200: Here is actually where I declare the spectrum just for initialisation.

SpeedyTransforms/src/gradient_arrays.jl
maximilian-gelbrecht: Without looking at it, I also assume that's why the GPU CI is failing.

milankl: Yes. Somehow dropped your version with `spectrum_cpu` and then `on_architecture` afterwards. Have it back now just not pushed yet! So initialisation works on GPU now

SpeedyTransforms/src/matrix_transform.jl
milankl: I use it for some error strings ... 😅

SpeedyTransforms/src/matrix_transform.jl
milankl: It's a stride-1 view, just taking off the end so a 8-layer transform can be used for 1...8 layers. Seems to compile away on both cpu and gpu

SpeedyTransforms/src/matrix_transform.jl
milankl: It's non-allocating but on the gpu it's another kernel launch unfortunately. It seems that on the GPU it isn't needed but i need to check for correctness if it writes complex directly back to a real array

milankl:
Okay transforms seem to work, backwards transform matrix is now calculated as the Moore-Penrose pseudo-inverse of the forward matrix via SVD. Not sure it's the most efficient way to compute this but it seems to work:

image

However currently the 2D transform (pressure here) seem to be ... off (model is still stable though)

image

maximilian-gelbrecht:
Some quick benchmarks for the dynamical core with the matrix transform. 

Inverse `transform!` and `surface_pressure_tendency!` is what we have to work on primarily. 

### GPU(CUDABackend(false, true)) | Float32 | T63 L16 | OctahedralGaussianGrid | 96 Rings

| - | - | - | - |
| forcing! | 2.053 μs| 0 bytes| 0 |
| drag! | 2.044 μs| 0 bytes| 0 |
| pressure_gradient_flux! | 720.374 μs| 20.31 KiB| 720 |
| linear_virtual_temperature! | 18.779 μs| 4.03 KiB| 143 |
| geopotential! | 46.661 μs| 9.17 KiB| 336 |
| vertical_integration! | 40.981 μs| 9.78 KiB| 300 |
| surface_pressure_tendency! | 1.300 ms| 11.59 KiB| 361 |
| vertical_velocity! | 211.427 μs| 21.44 KiB| 740 |
| linear_pressure_gradient! | 16.445 μs| 3.31 KiB| 113 |
| vertical_advection! | 41.663 μs| 15.16 KiB| 326 |
| vordiv_tendencies! | 3.013 ms| 23.86 KiB| 580 |
| temperature_tendency! | 4.544 ms| 23.06 KiB| 624 |
| humidity_tendency! | 4.503 ms| 20.12 KiB| 543 |
| bernoulli_potential! | 1.505 ms| 12.48 KiB| 333 |
| tracer_advection! | 2.053 μs| 0 bytes| 0 |
| transform! (forward) | 419.416 μs| 3.83 KiB| 149 |
| transform! (inverse) | 1.488 ms| 2.70 KiB| 82 |



### CPU | Float32 | T63 L16 | OctahedralGaussianGrid | 96 Rings

| - | - | - | - |
| forcing! | 2.049 μs| 0 bytes| 0 |
| drag! | 2.049 μs| 0 bytes| 0 |
| pressure_gradient_flux! | 1.614 ms| 2.50 KiB| 54 |
| linear_virtual_temperature! | 23.155 μs| 0 bytes| 0 |
| geopotential! | 85.368 μs| 17.87 KiB| 23 |
| vertical_integration! | 100.170 μs| 0 bytes| 0 |
| surface_pressure_tendency! | 479.380 μs| 0 bytes| 0 |
| vertical_velocity! | 458.287 μs| 2.51 MiB| 12 |
| linear_pressure_gradient! | 22.053 μs| 0 bytes| 0 |
| vertical_advection! | 1.321 ms| 9.34 KiB| 106 |
| vordiv_tendencies! | 4.933 ms| 9.47 KiB| 96 |
| temperature_tendency! | 6.550 ms| 6.98 KiB| 85 |
| humidity_tendency! | 6.570 ms| 5.62 KiB| 75 |
| bernoulli_potential! | 1.835 ms| 1.02 KiB| 18 |
| tracer_advection! | 2.058 μs| 0 bytes| 0 |
| transform! (forward) | 1.699 ms| 0 bytes| 0 |
| transform! (inverse) | 1.709 ms| 0 bytes| 0 |

SpeedyTransforms/src/matrix_transform.jl
```diff
@@ -0,0 +1,196 @@
...
+        # multiply into scratch which is complex typed and then take real part into field
+        # imaginary part should be zero but destination is used to store intermediate results
+        # explicitly convert to real also for NaN + NaN*im results
+        LinearAlgebra.mul!(scratch, M.backward, coeffs.data)
+        field.data .= real.(scratch)
+
+    else    # don't use view for 3D transforms with all layers
+        # TODO if multiplication yields all real then one could write directly into field.data
+        # which would be much faster but if the imaginary part is non-zero this throws an error
+        # so for now use the scratch memory as intermediate storage and then copy real part
```
maximilian-gelbrecht, line 221: I'm wondering if there's way to do this with `reinterpret`-ing complex numbers as real numbers and vice versa.

maximilian-gelbrecht: I just tried a bit naive version of this and it turned out MUCH slower (x20) than before 

```Julia 
function real_matvec!(y, A::AbstractMatrix{Complex{T}}, x::AbstractMatrix{Complex{T}}) where T
    # Reinterpret complex arrays as interleaved real arrays
    A_flat = reinterpret(T, A)  # 2×size(A,1) × size(A,2)
    x_flat = reinterpret(T, x)
    
    A_r = @view A_flat[1:2:end, :]  # real parts (strided view)
    A_i = @view A_flat[2:2:end, :]  # imag parts
    x_r = @view x_flat[1:2:end, :]
    x_i = @view x_flat[2:2:end, :]
    
    LinearAlgebra.mul!(y, A_r, x_r)
    LinearAlgebra.mul!(y, A_i, x_i, -1, 1)
    return y
end
```

In theory this saves a lot of computations, but as far as I know this isn't able to hit the highly optimised BLAS routines and is much much slower because of that. This is on CPU, but we'd probably hit a similar problem on GPU as well. 

@milankl I vaguely recall that you asked about something like this in Julia slack and some people answered about complex and reinterpret and BLAS as well. Could you point me to that conversation?

maximilian-gelbrecht: Afaik routines for strided arrays exist in BLAS and cuBLAS though. I just don't know at the moment, how to call them correctly because in this example, `x_r <: StridedArray`, so I guess it should work? But it doesn't

maximilian-gelbrecht: Claude is convinced it would never work without copying due to limitations in `gemm` but I am not sure yet it's correct ;)

maximilian-gelbrecht: The alternative is a completely custom kernel, which we could translate to Julia from existing ones in C. It's complex from an algorithm perspective but in the end it's also just about 20-30 lines of code.

maximilian-gelbrecht: Ok, so I just tested the following and already get a 2x speed up on CPU version compared to the old version: 

```Julia 
using SpeedyWeather, BenchmarkTools, LinearAlgebra

spectral_grid = SpectralGrid(trunc=63, nlayers=16);
M = MatrixSpectralTransform(spectral_grid)

spec = randn(Complex{Float32}, spectral_grid.spectrum, 16)
field = zeros(Float32, spectral_grid.grid, 16)

@benchmark transform!(field, spec, M) # 14ms 

function real_matvec2!(y, A_r::AbstractMatrix{T}, A_i::AbstractMatrix{T}, x::AbstractMatrix{Complex{T}}) where T
    LinearAlgebra.mul!(y, A_r, real(x))
    LinearAlgebra.mul!(y, A_i, imag(x), -1, 1)
    return y
end

Mback_r = real(M.backward)
Mback_i = imag(M.backward)

@benchmark real_matvec2!(field.data, Mback_r, Mback_i, spec.data) # 7ms 
```

maximilian-gelbrecht: Ah non-allocating version, isn't faster, but well, it's not allocating 

```julia 
scratch = zero(real(spec.data))

function real_matvec3!(y, A_r::AbstractMatrix{T}, A_i::AbstractMatrix{T}, x::AbstractMatrix{Complex{T}}, scratch) where T
    scratch .= real.(x)
    LinearAlgebra.mul!(y, A_r, scratch)

    scratch .= imag.(x)
    LinearAlgebra.mul!(y, A_i, scratch, -1, 1)
    return y
end

@benchmark real_matvec3!(field.data, Mback_r, Mback_i, spec.data, scratch)
```

maximilian-gelbrecht: A version with fully split `x` (before) isn't even faster on CPU then the previous two. On GPU the difference seems to be much smaller. Mh...  not sure, a bit more to try out.

maximilian-gelbrecht:
I pushed a version of the split transform. I'll post some benchmarks later. So far this seems to make a big difference on CPU, but not so much on GPU.

milankl:
Did you find a way to fix the problems of the 2D transform as reported above ☝🏼 ?

maximilian-gelbrecht:
I wrote a fix so that the 2D transforms run without errors. They are still slow on GPU but that's a fundamental problem of them being comparably low-dim that we can't do much about.

maximilian-gelbrecht:
I made a very quick attempt of generating a custom GPU kernel with tiling and it's 50% as fast as the split cuBLAS version. Which honestly isn't even that bad for like 20 minutes of trying, but actually getting the custom kernel to be faster than cuBLAS would need more advanced optimisations (see e.g https://siboehm.com/articles/22/CUDA-MMM) that I am not sure we really have the capacity to do. 

```Julia 

# ============================================================================
# TILED GEMM KERNEL FOR REAL PART OF COMPLEX MATRIX-VECTOR PRODUCT
# Computes y = Re(A * x) = A_r * x_r - A_i * x_i directly from complex A and x
# Uses shared memory tiling for better memory access patterns on GPU
# ============================================================================

# Tile size for the tiled GEMM kernel - must be known at compile time
const TILE_SIZE = 16

"""
Tiled GEMM kernel computing y = Re(A * x) where A and x are complex.
Each workgroup computes a TILE_SIZE × TILE_SIZE block of the output.
Uses shared memory to reduce global memory accesses.

Arguments:
- y: output matrix (M × K), real
- A: input matrix (M × N), complex  
- x: input matrix (N × K), complex
- M, N, K: dimensions
"""
@kernel function real_complex_gemm_kernel!(
    y::AbstractMatrix{NF},
    A::AbstractMatrix{Complex{NF}},
    x::AbstractMatrix{Complex{NF}},
    M::Int, N::Int, K::Int
) where NF

    # Thread indices within the workgroup
    tx = @index(Local, NTuple)[1]
    ty = @index(Local, NTuple)[2]
    
    # Block indices
    bx = @index(Group, NTuple)[1]
    by = @index(Group, NTuple)[2]
    
    # Global output indices
    row = (bx - 1) * TILE_SIZE + tx
    col = (by - 1) * TILE_SIZE + ty
    
    # Shared memory tiles for A and x (real and imag parts separately)
    tile_A_r = @localmem NF (TILE_SIZE, TILE_SIZE)
    tile_A_i = @localmem NF (TILE_SIZE, TILE_SIZE)
    tile_x_r = @localmem NF (TILE_SIZE, TILE_SIZE)
    tile_x_i = @localmem NF (TILE_SIZE, TILE_SIZE)
    
    # Accumulator for the dot product
    acc = zero(NF)
    
    # Number of tiles needed to cover the K dimension (inner dimension N of A)
    num_tiles = cld(N, TILE_SIZE)
    
    # Loop over tiles
    for tile_idx in 1:num_tiles
        # Column index in A / row index in x for this tile
        tile_col = (tile_idx - 1) * TILE_SIZE + ty
        tile_row = (tile_idx - 1) * TILE_SIZE + tx
        
        # Collaboratively load tile of A into shared memory
        if row <= M && tile_col <= N
            a_val = A[row, tile_col]
            tile_A_r[tx, ty] = real(a_val)
            tile_A_i[tx, ty] = imag(a_val)
        else
            tile_A_r[tx, ty] = zero(NF)
            tile_A_i[tx, ty] = zero(NF)
        end
        
        # Collaboratively load tile of x into shared memory
        if tile_row <= N && col <= K
            x_val = x[tile_row, col]
            tile_x_r[tx, ty] = real(x_val)
            tile_x_i[tx, ty] = imag(x_val)
        else
            tile_x_r[tx, ty] = zero(NF)
            tile_x_i[tx, ty] = zero(NF)
        end
        
        # Synchronize to ensure all threads have loaded their data
        @synchronize
        
        # Compute partial dot product for this tile
        # y = A_r * x_r - A_i * x_i
        @unroll for k in 1:TILE_SIZE
            acc += tile_A_r[tx, k] * tile_x_r[k, ty] - tile_A_i[tx, k] * tile_x_i[k, ty]
        end
        
        # Synchronize before loading next tile
        @synchronize
    end
    
    # Write result to global memory
    if row <= M && col <= K
        y[row, col] = acc
    end
end

"""
$(TYPEDSIGNATURES)
Launch the tiled GEMM kernel for computing y = Re(A * x) where A and x are complex.
This is an optimized GPU kernel that uses shared memory tiling.
"""
function real_complex_gemm!(
    y::AbstractMatrix{NF},
    A::AbstractMatrix{Complex{NF}},
    x::AbstractMatrix{Complex{NF}},
    arch::AbstractArchitecture
) where NF
    M, N = size(A)
    N2, K = size(x)
    @assert N == N2 "Inner dimensions must match: A is $M×$N, x is $N2×$K"
    @assert size(y) == (M, K) "Output size must be ($M, $K), got $(size(y))"
    
    # Workgroup size is TILE_SIZE × TILE_SIZE
    workgroup = (TILE_SIZE, TILE_SIZE)
    
    # Number of workgroups needed to cover the output
    ndrange = (cld(M, TILE_SIZE) * TILE_SIZE, cld(K, TILE_SIZE) * TILE_SIZE)
    
    # Launch kernel
    kernel! = real_complex_gemm_kernel!(device(arch), workgroup)
    kernel!(y, A, x, M, N, K; ndrange)
    
    return y
end

"""
$(TYPEDSIGNATURES)
Spectral transform (spectral to grid space) using the optimized tiled GEMM kernel.
Computes y = Re(backward * coeffs) directly without intermediate complex storage.
"""
function transform_tiled!(
    field::AbstractField,
    coeffs::LowerTriangularArray,
    M::MatrixSpectralTransform;
    unscale_coslat::Bool = false,
)
    @boundscheck ismatching(M, field) || throw(DimensionMismatch(M, field))
    @boundscheck ismatching(M, coeffs) || throw(DimensionMismatch(M, coeffs))
    
    # Use the tiled kernel to compute Re(backward * coeffs)
    real_complex_gemm!(field.data, M.backward, coeffs.data, M.architecture)
    
    unscale_coslat && RingGrids._scale_lat!(field, M.coslat⁻¹)
    return field
end
```

milankl:
Nice! This is really cool though because while being 50% as fast it'll save all the memory we'd need to carry around otherwise? @oscardssmith do we have in CUDA.jl a way to accelerate such a kernel e.g. calling the tensor core units?

oscardssmith:
I think https://cuda.juliagpu.org/stable/api/kernel/#WMMA shows how to do it

maximilian-gelbrecht:
I started some benchmarks late Friday. Not even going to post the full results. 

The new inverse transform doesn't seem to so well in practice, but on CPU there's something really wrong, I got very bad performance for the whole dyn core, not only the inverse transforms. On GPU it's fine but not great. 

Starting benchmarks with the old inverse transform now.

maximilian-gelbrecht:
I only had half an eye on this the last week. Even the old version of the inverse transform has some massive performance problem. I am looking into this now.

maximilian-gelbrecht:
Ok, good news and bad news: 

* I added unit tests, but they are failing currently: both round trip and comparison to the standard transform. We have to look into that. It's the inverse transform that's the problem, the forward transform works. 
* For the inverse transform I tested in conjunction with #970 a compilation by Reactant. This one is really, really fast. I mean that is also exactly what Reactant is made for: hitting those highly optimised XLA routines. This might solve the performance problems for the inverse transform.

maximilian-gelbrecht:
So I committed a version here that fixed the inverse transform by actually fully computing the backward matrix as well. This works 

So either we keep this version, or we think about another shortcut that avoids computing it, but `pinv`doesn't work.

maximilian-gelbrecht:
Claude came up with this way to compute the backward matrix from the forward matrix adjusting quadrature weights and FFT normalisation. It does yield correct results, but it's not actually faster than the brute force version. 

```
"""$(TYPEDSIGNATURES)
Derive the backward transform matrix `B` (spectral to grid) directly from the forward matrix `F`
(grid to spectral) without probing. Uses the relationship `B[i,lm] = s_m * conj(F[lm,i]) / w_j`
where `w_j` is the solid angle for ring `j` containing point `i`, and the scale factor `s_m` accounts
for the real FFT convention: `s_m = 1` for `m = 0` or `m ≥ nlon_j ÷ 2` (Nyquist and above),
`s_m = 2` for `0 < m < nlon_j ÷ 2` (positive frequencies that fold with their negative counterpart)."""
function backward_from_forward_matrix(
        forward::AbstractMatrix{Complex{NF}},
        spectrum::AbstractSpectrum,
        grid::AbstractGrid,
    ) where {NF}

    nharmonics, npoints = size(forward)
    backward = zeros(Complex{NF}, npoints, nharmonics)

    # solid angles per ring (quadrature weight × azimuthal spacing)
    solid_angles = get_solid_angles(grid)
    rings = eachring(grid)

    # build the m-value (0-based order) for each linear index lm
    (; lmax, mmax) = spectrum
    m_per_lm = zeros(Int, nharmonics)
    idx = 0
    for m in 1:mmax           # 1-based column index in the lower-triangular layout
        for l in m:lmax       # 1-based row index
            idx += 1
            m_per_lm[idx] = m - 1   # 0-based spherical harmonic order
        end
    end

    # scale factor: real FFT folds negative frequencies, so 0 < m < nlon÷2 carry factor 2
    # but m = 0 and m ≥ nlon÷2 (Nyquist and beyond for that ring) carry factor 1
    for (j, ring) in enumerate(rings)
        w_inv = inv(solid_angles[j])
        nlon_half = length(ring) ÷ 2
        for i in ring
            for lm in 1:nharmonics
                m = m_per_lm[lm]
                s_m = (m > 0 && m < nlon_half) ? NF(2) : one(NF)
                backward[i, lm] = s_m * conj(forward[lm, i]) * w_inv
            end
        end
    end

    return backward
end
```

maximilian-gelbrecht:
What's actually still missing is documentation. But I am willing to merge it without it, and do a bigger documentation update once Reactant works.

milankl:
@gregordecristoforo FYI we merged the matrix spectral transform already as its working but not efficient yet (on neither CPU or GPU I believe) for which we need Reactant to work, see #985. But initial tests that Max did showed potentially so we're keen to try this for low resolution

PR #955, opened 2025-12-23
Changed OneBandShortwave to ij logic

nviebig:
## Goal
- Switch OneBandShortwave to the ij-logic from #873.
- Updated shortwave clouds/transmissivity to match ij logic.
- Docs: enable OneBandShortwave examples again.
- Tests: re-enable shortwave/clouds/transmissivity checks.

### DIFF
--- a/src/physics/radiation/clouds.jl
+++ b/src/physics/radiation/clouds.jl
@@ -61,7 +61,7 @@ export DiagnosticClouds
     use_stratocumulus::Bool = true
 
     "[OPTION] Stratocumulus cloud factor (SPEEDY clfact) [1]"
-    @param stratocumulus_clfact::NF = 1.2 (bounds = Nonnegative,)
+    @param stratocumulus_cloud_factor::NF = 1.2 (bounds = Nonnegative,)
 end
 
 Adapt.@adapt_structure DiagnosticClouds
@@ -111,7 +111,7 @@ Returns (cloud_cover, cloud_top, stratocumulus_cover) tuple."""
     stab_min = clouds.stratocumulus_stability_min
     stab_max = clouds.stratocumulus_stability_max
     cover_max = clouds.stratocumulus_cover_max
-    clfact = clouds.stratocumulus_clfact
+    cloud_factor = clouds.stratocumulus_cloud_factor
 
     # Precipitation contribution (rain rate is in m/s)
     rain_rate = diagn.physics.rain_rate[ij]
@@ -154,9 +154,9 @@ Returns (cloud_cover, cloud_top, stratocumulus_cover) tuple."""
         # normalized static stability factor
         static_stability = clamp((G - stab_min) / (stab_max - stab_min), 0, 1)
 
-        stratocumulus_cover_ocean = static_stability * max(cover_max - clfact * cloud_cover, 0)
+        stratocumulus_cover_ocean = static_stability * max(cover_max - cloud_factor * cloud_cover, 0)
         qsat_surface = saturation_humidity(temp[ij, surface], sigma_levels[surface] * pₛ, model.atmosphere)
-        rh_surface::NF = qsat_surface > 0 ? humid[ij, surface] / qsat_surface : 0   # relative humidity at surface
+        rh_surface = humid[ij, surface] / qsat_surface      # relative humidity at surface
         stratocumulus_cover_land = stratocumulus_cover_ocean * rh_surface
 
         stratocumulus_cover =
--- a/src/dynamics/atmosphere.jl
+++ b/src/dynamics/atmosphere.jl
@@ -141,6 +141,7 @@ function saturation_vapor_pressure(T, A::AbstractWetAtmosphere)
     Lᵥ = A.latent_heat_condensation
     (; R_vapor) = A
     T₀ = A.temperature_freezing
+    T = convert(eltype(T₀), T)      # ensure type of A
     return e₀ * exp(Lᵥ / R_vapor * (inv(T₀) - inv(T)))
 end
 
@@ -151,9 +152,9 @@ Saturation specific humidity as a function of temperature [K] and pressure [Pa],
     qₛ = ε * eₛ(T) / p,
 with saturation vapor pressure eₛ(T), pressure p [Pa] and ratio of gas constants `ε = R_dry/R_vapor`."""
 function saturation_humidity(T, p, A::AbstractWetAtmosphere)
-    e_s = saturation_vapor_pressure(T, A)
+    eₛ = saturation_vapor_pressure(T, A)
     ϵ = A.mol_ratio
-    return ϵ * e_s / p
+    return ϵ * eₛ / p
 end
 
 export TetensEquation


src/physics/radiation/shortwave_radiation.jl
```diff
@@ -141,86 +151,83 @@ export OneBandShortwaveRadiativeTransfer
...
+        if k == cloud_top
+            R = cloud_albedo * cloud_cover
+            U_reflected = D * R
+            D *= (1 - R)
         end
+
+        D_out = D * t[ij, k]
+        # Update temperature tendency due to absorbed shortwave radiation
+        dTdt[ij, k] += flux_to_tendency((D - D_out) / c_p, p_s, k, model)
+        D = D_out
```
milankl, line 203: oh yeah that's smart then you don't have to call `flux_to_tendency` twice with the flux in and flux out. Feel free to change that in the longwave too

src/physics/radiation/clouds.jl
```diff
@@ -125,17 +142,19 @@ function diagnose_cloud_properties(column::ColumnVariables, clouds::DiagnosticCl
...
-        F_ST = max(0, min(1, (GSEN - stab_min) / (stab_max - stab_min)))
+        GSEN = (c_p * temp[ij, nlayers] + geopotential[ij, nlayers]) -
+            (c_p * temp[ij, nlayers - 1] + geopotential[ij, nlayers - 1])
+        F_ST = clamp((GSEN - stab_min) / (stab_max - stab_min), zero(NF), one(NF))
 
         stratocumulus_cover_ocean = F_ST * max(cover_max - clfact * cloud_cover, 0)
-        RH_N = sat_humid[nlayers] > 0 ? humid[nlayers] / sat_humid[nlayers] : 0
-        stratocumulus_cover_land = stratocumulus_cover_ocean * RH_N
+        qsat_surface = saturation_humidity(temp[ij, nlayers], sigma_levels[nlayers] * p_s, model.atmosphere)
+        RH_N = qsat_surface > zero(NF) ? humid[ij, nlayers] / qsat_surface : 0
```
milankl: I think this isn't type stable. Use `zero(NF)` or define `RH_N::NF = 0` then you lock it's type in and it can't be reassigned to another type

src/physics/radiation/shortwave_radiation.jl
```diff
@@ -141,86 +151,83 @@ export OneBandShortwaveRadiativeTransfer
...
-
-    # Clear sky portion until cloud top
-    @inbounds for k in 1:(cloud_top - 1)
-        D *= t[k]
-        flux_temp_downward[k + 1] += D
+    dTdt = diagn.tendencies.temp_tend_grid
+    p_s = diagn.grid.pres_grid_prev[ij]
+    nlayers = size(dTdt, 2)
+
+    @inbounds begin
```
milankl: with the `@propagate_inbounds` macro we use now we automatically inline and inbound everything from the callside (where we have a `@inbounds`), so this is redundant here

src/physics/radiation/shortwave_transmissivity.jl
```diff
@@ -97,19 +108,19 @@ function transmissivity!(
...
             layer_absorptivity += cloud_absorptivity_term * cloud_cover
         end
 
         # Compute differential optical depth with zenith correction
-        # CRITICAL: Normalize pressure to 10^5 Pa since absorptivities are per 10^5 Pa
+        # Normalize pressure to 1e5 Pa since absorptivities are per 1e5 Pa
         delta_sigma = sigma_levels[k + 1] - sigma_levels[k]
-        normalized_pressure = surface_pressure / 100000   # Convert Pa to units of 10^5 Pa
+        normalized_pressure = surface_pressure / 100000
         optical_depth = layer_absorptivity * delta_sigma * normalized_pressure * zenit_factor
```
milankl: now you have the chance to rename `zenit_factor` to `zenith_factor` 😉

milankl:
You get this error now
```julia
SeaIceModel = Nothing: Error During Test at /home/runner/work/SpeedyWeather.jl/SpeedyWeather.jl/test/physics/ocean_sea_ice.jl:16
  type NamedTuple has no field surface_shortwave_up
```

because of our new variable system where individual components are now responsible to declare variables as needed (information that is then used to allocate them). Hence the `variables(::AbstractShortwave)` function. I don't see it immediately what's wrong but somehow we must have ended up  trying to access a variable that's wasn't declared through `variables(...)` before.

src/physics/radiation/shortwave_radiation.jl
nviebig: Great okay! :)

src/physics/radiation/shortwave_transmissivity.jl
nviebig: Oh perfect haha :)

milankl:
I think the documentation fails because there's a `using CairoMakie` coming after `heatmap` is called

src/physics/radiation/clouds.jl
```diff
@@ -81,64 +88,78 @@ end
...
+    rain_rate = diagn.physics.rain_rate[ij]
+    precip_term = min(precip_max, (86400 * rain_rate) / 1000)   # convert to mm/day
+    P = precip_weight * sqrt(precip_term)
 
-    kprtop = column.cloud_top
-    humidity_term_kcltop = zero(P)
-    kcltop = nlayers + 1
+    cloud_top_precipitation = diagn.physics.cloud_top[ij]       # from convection or large-scale condensation
+    humidity_term_cloud_top::NF = 0
+    cloud_top_humidity = nlayers + 1
```
milankl, line 123: I found the previous naming here cryptic (probably from Fortran SPEEDY) and therefore applied some renaming

src/physics/radiation/clouds.jl
```diff
@@ -81,64 +88,78 @@ end
...
         if humidity_k > q_min && qsat > 0
             relative_humidity_k = humidity_k / qsat
 
             if relative_humidity_k >= rh_min
                 rh_norm = max(0, (relative_humidity_k - rh_min) / (rh_max - rh_min))
-                humidity_term_kcltop = min(1, rh_norm)^2
-                kcltop = k
-                break
+                humidity_term_cloud_top = min(1, rh_norm)^2
+                cloud_top_humidity = min(k, cloud_top_humidity)
```
milankl, line 135: removed the break here, not sure whether it's better to use `min` or `break` or write it as a while loop. Do all of these work on the GPU?

src/physics/radiation/shortwave_radiation.jl
```diff
@@ -117,111 +122,114 @@ end
 function initialize!(radiation::OneBandShortwave, model::PrimitiveEquation)
     initialize!(radiation.clouds, model)
     initialize!(radiation.transmissivity, model)
-    return initialize!(radiation.radiative_transfer, model)
```
milankl, line 120: runic while always put a `return` in front of the last line. As we don't actually return anything from initialize! here be explicit and write `return nothing`

src/physics/radiation/shortwave_transmissivity.jl
```diff
@@ -1,92 +1,109 @@
...
-    (; humid, cos_zenith, nlayers) = column
+
+    humid = diagn.grid.humid_grid_prev
+    cos_zenith = diagn.physics.cos_zenith[ij]
+    nlayers = size(t, 2)
 
     sigma_levels = model.geometry.σ_levels_half
     sigma_levels_full = model.geometry.σ_levels_full
-    surface_pressure = column.pres[end]  # This is in Pa
+    normalized_surface_pressure = diagn.grid.pres_grid_prev[ij] / 100000
```
milankl, line 88: Oceananigans started to use rationals in many places, i.e. `* 1//100000` here. Rationals are being compiled away and being replaced by the corresponding value in float32. That way you can actually get a multiplication without writing `* 0.00001f0` ...

src/physics/radiation/shortwave_transmissivity.jl
```diff
@@ -1,92 +1,109 @@
...
         absorptivity_cloud_limit
     )
 
     for k in 1:nlayers
-        q = humid[k]
+        q = humid[ij, k]
 
         # Aerosol factor: use mid-level sigma, squared
-        aerosol_factor = transmissivity.aerosols ? sigma_levels_full[k]^2 : 0
+        aerosol_factor = transmissivity.aerosols ? sigma_levels_full[k]^2 : zero(NF)
```
milankl, line 106: this wasn't type stable

src/physics/radiation/shortwave_transmissivity.jl
```diff
@@ -96,19 +113,18 @@ function transmissivity!(
...
             layer_absorptivity += cloud_absorptivity_term * cloud_cover
         end
 
         # Compute differential optical depth with zenith correction
-        # CRITICAL: Normalize pressure to 10^5 Pa since absorptivities are per 10^5 Pa
+        # Normalize pressure to 1e5 Pa since absorptivities are per 1e5 Pa
         delta_sigma = sigma_levels[k + 1] - sigma_levels[k]
-        normalized_pressure = surface_pressure / 100000   # Convert Pa to units of 10^5 Pa
-        optical_depth = layer_absorptivity * delta_sigma * normalized_pressure * zenit_factor
+        optical_depth = layer_absorptivity * delta_sigma * normalized_surface_pressure * zenith_factor
```
milankl, line 124: Instead of recomputing the normalized pressure on every layer I moved this up so it's only computed once

src/models/primitive_wet.jl
```diff
@@ -96,7 +96,7 @@ $(TYPEDFIELDS)"""
     @component surface_humidity_flux::HF = SurfaceHumidityFlux(spectral_grid)
     @component large_scale_condensation::LSC = ImplicitCondensation(spectral_grid)
     @component convection::CV = BettsMillerConvection(spectral_grid)
-    @component shortwave_radiation::SW = TransparentShortwave(spectral_grid)
+    @component shortwave_radiation::SW = OneBandShortwave(spectral_grid)
```
milankl, line 99: Making the `OneBandShortwave` the default for primitive wet 🎉

milankl (commented):
On my M4 I get

- no short or longwave radiation ~2.4k SYPD
- only longwave ~2k SYPD
- short and longwave ~1.8k SYPD

so somewhat more expensive but within reason. Have you checked that this all runs on the GPU too? Otherwise approved!

milankl:
T127L16 on the L4 GPU without OneBandShortwave
```julia
  220.880 μs (2065 allocations: 119.20 KiB)

  220.592 μs (2065 allocations: 119.20 KiB)
```
and with OneBandShortwave
```julia
  228.669 μs (2105 allocations: 123.27 KiB)

  229.985 μs (2105 allocations: 123.27 KiB)
```
so only 4% slower. On the CPU on my M4 I have ~2.4k SYPD without radiation, 2k SYPD with longwave only, 1.8k SYPD with short (+clouds) and longwave.

milankl:
AMDGPU buildkite is stalling, don't know why but just bypassing this. As long as CUDA works that's fine.

PR #957, opened 2025-12-30
Monorepo structure, separate tests and skip mechanism

milankl:
- puts SpeedyWeather into its own folder 
- separates tests for every submodule
- adds a code of conduct

Added a new skip checks logic based on labels, we support

- `skip-ci`, `skip-docs`, `skip-submodules-ci`, `skip-speedyweather-ci`, `skip-gpu-ci`, `skip-enzyme-ci`, `skip-ringgrids-ci`, `skip-lowertriangulararrays-ci`, `skip-speedyweatherinternals-ci`, `skip-speedytransforms-ci`


milankl:
So the `skip-ci` label disables the enforce changelog workflow that works great, need to find a way to do the same for the other workflows ...

milankl:
When this PR passes and is merged into main then we need to do https://github.com/JuliaRegistries/General/pull/145437 to update SpeedyWeather's registry

milankl:
I forgot that on Julia v1.10 (pre proper monorepo support) Pkg will unfortunately install SpeedyWeather dependencies as registered not as in the repo main. Hence SpeedyTransforms currently fails on v1.10 because it installs the latest release of LowerTriangularArrays instead of main.

maximilian-gelbrecht:
Oh this is great @milankl ! Thanks!

PR #958, opened 2025-12-30
Fix `SpeedyWeather.animate` projection issue

gregrmunday:
Currently the `SpeedyWeather.animate` function doesn't transform the underlying data before supplying it GeoMakie - as a result the output map is upside down and needs to be translated by 180°.

You can see this in the [documentation animation](https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/output#Visualizing-output) where the continents are misaligned with the drawn coastlines and are upside down 🙃

This PR fixes this by:

- [x] Supplying latitude points in ascending order to `GeoMakie`
- [x] Transforming the underlying data from the NetCDF file following the lat/lon coordinate transforms.

Basically, as above, this just means flipping the grid vertically and rolling by 180 degrees horizontally. 

The extra stuff here is just to take care of potential NetCDF input from other sources with dimensions in potentially different orders. 

An issue has been opened in GeoMakie to fix this itself so manual intervention isn't required from its users, therefore we can replace this code when it has been resolved there. https://github.com/MakieOrg/GeoMakie.jl/issues/363


milankl:
Very cool, once the Documentation CI finished building there'll be another check popping up saynig "deploy docs" we can then verify that this actually changed the orientation of `animate` in the docs correctly

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@
 - Don't adapt rings in RingGrids on GPU [#920](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/920)
 - README updates [#915](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/915#pullrequestreview-3469109183)
 - Fixes some formatting/LaTeX rendering in Docs [#916](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/916)
+- Fixes a `SpeedyWeather.animate` projection issue [#958](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/958)
```
milankl: our changelog is organised with newest changes at the top 😉

ext/SpeedyWeatherGeoMakieExt.jl
```diff
@@ -105,13 +105,30 @@ function SpeedyWeather.animate(
...
+        else
+             ds[variable].var[:, :, $tsteps]
+        end
+
+        # Make sure the array is shaped correctly
+        standardized = permutedims(raw_slice, (lon_idx, lat_idx))
+
+        # Apply same transformations as the lat/lons
+        rolled = circshift(standardized, (shift_amt, 0))
+        transf_data = reverse(rolled, dims=lat_idx)
```
milankl: I feel like this should be dealt with in GeoMakie but once till it's fixed there and released we should do this manually

ext/SpeedyWeatherGeoMakieExt.jl
```diff
@@ -74,7 +74,7 @@ function SpeedyWeather.animate(
 
     # Get dimensions
     lon = longitude_shift_180(ds["lon"][:]) # longitudes -180 to 180
-    lat = ds["lat"][:]
+    lat = sort(ds["lat"][:]) # lats -90 to 90
```
milankl: if we can produce an example that `90..-90` doesn't work correctly with GeoMakie then we should flag this as an issue there ...

CHANGELOG.md
gregrmunday: Ahh! Updated in the latest commit :)

ext/SpeedyWeatherGeoMakieExt.jl
gregrmunday: Agreed, that'd be quite smooth!

ext/SpeedyWeatherGeoMakieExt.jl
gregrmunday: Ok in fairness 90...-90 does actually work, so I guess we can leave this as-is if we want? It makes sense to me at least from a readability standpoint to force the lat/lons to ascend (then the vertical flip is better justified) but I suppose not totally necessary! 

Screenshot 2025-12-30 at 17 07 57

gregrmunday:
> Very cool, once the Documentation CI finished building there'll be another check popping up saynig "deploy docs" we can then verify that this actually changed the orientation of `animate` in the docs correctly

Seems fixed in the [docs](https://speedyweather.github.io/SpeedyWeatherDocumentation/previews/PR958/output/) now :)

milankl:
Okay this only failed because of Enzyme (generally meaning that something isn't differentiable) but the actual error is
```
 GC error (probable corruption)
```
so this is not your fault (also you didn't affect any code that we'd need to differentiate through). We sometimes have problems with Enzyme, hence tests currently only run on Julia v1.10. I just reran the failed test let's see what happens

ext/SpeedyWeatherGeoMakieExt.jl
milankl: In these situations the most elegant way to solve this is to open an issue in GeoMakie, fix the problem independently here (as you did) but then add a comment like "following lines currently needed because of [link to that issue] remove/replace with X when resolved"

milankl:
And then while we're at it, we use Runic formatting, see https://github.com/fredrikekre/Runic.jl  and #950 

We've decided not to make this an actual mandatory check required to pass but to occasionally just runic entire files/everything when diverging from a consistent runic style. We've set it personally set it up in our VS codes so a simple shift+option+F applies runic formatting to the entire jl file. Here it was only a `a=b` -> `a = b` spaces added but in #950 it was thousands of lines. As Runic philosophy goes no ones like the runic style but everyone loves the runic formatter, it's a "don't ask questions just do it" approach and I'm happy to spend as little time as possible on whether style detail shuold be A or B 😉

gregrmunday:
> And then while we're at it, we use Runic formatting, see https://github.com/fredrikekre/Runic.jl and #950
> 
> We've decided not to make this an actual mandatory check required to pass but to occasionally just runic entire files/everything when diverging from a consistent runic style. We've set it personally set it up in our VS codes so a simple shift+option+F applies runic formatting to the entire jl file. Here it was only a `a=b` -> `a = b` spaces added but in #950 it was thousands of lines. As Runic philosophy goes no ones like the runic style but everyone loves the runic formatter, it's a "don't ask questions just do it" approach and I'm happy to spend as little time as possible on whether style detail shuold be A or B 😉

Sounds great, I've got it all set up on my end now :)

ext/SpeedyWeatherGeoMakieExt.jl
gregrmunday: I've taken your advice and opened an issue there, and updated the description of this PR as you suggest! Issue is here: https://github.com/MakieOrg/GeoMakie.jl/issues/363

ext/SpeedyWeatherGeoMakieExt.jl
```diff
@@ -105,13 +105,30 @@ function SpeedyWeather.animate(
...
+        end
+
+        # Make sure the array is shaped correctly
+        standardized = permutedims(raw_slice, (lon_idx, lat_idx))
+
+        # Apply same transformations as the lat/lons
+        rolled = circshift(standardized, (shift_amt, 0))
+        transf_data = reverse(rolled, dims = lat_idx)
+
+        transf_data
```
asinghvi17:
```suggestion
    data = lift(fig.scene, tsteps) do tstep
        raw_slice = if is_3d
            ds[variable].var[:, :, level, tstep]
        else
            ds[variable].var[:, :, tstep]
        end

        # Make sure the array is shaped correctly
        standardized = permutedims(raw_slice, (lon_idx, lat_idx))

        # Apply same transformations as the lat/lons
        rolled = circshift(standardized, (shift_amt, 0))
        transf_data = reverse(rolled, dims = lat_idx)

        transf_data
```

These days I tend to prefer explicit `lift(...)` since it makes it extremely clear what is being lifted on - but that's a more minor thing.

SpeedyWeather/ext/SpeedyWeatherGeoMakieExt.jl
```diff
@@ -105,13 +105,30 @@ function SpeedyWeather.animate(
...
+    # Translation length depends on number of gridboxes
+    n_lon = length(lon)
+    shift_amt = div(n_lon, 2)  # Num lons is usually even
+
+    data = @lift begin
+        raw_slice = if is_3d
+            ds[variable].var[:, :, level, $tsteps]
+        else
+            ds[variable].var[:, :, $tsteps]
+        end
```
asinghvi17, line 127: Should this be a view, or is that already lazy?  No need to materialize more than we have to

ext/SpeedyWeatherGeoMakieExt.jl
```diff
@@ -74,7 +74,7 @@ function SpeedyWeather.animate(
 
     # Get dimensions
     lon = longitude_shift_180(ds["lon"][:]) # longitudes -180 to 180
-    lat = ds["lat"][:]
+    lat = sort(ds["lat"][:])                # lats -90 to 90
```
asinghvi17: I'm a bit confused here - this doesn't seem to affect the actual data, only the latitude values?

asinghvi17 (commented):
Seems reasonable, just a couple comments

ext/SpeedyWeatherGeoMakieExt.jl
milankl: our grids are sorted north to south, so I'd say 'reverse' is correct here to get it in ascending order

asinghvi17: My point here is more that we're using `sort` here but should either use the same logic as below (`reverse` effectively) or take `sortperm` and re-order the rows of the data array using the right order determined by `sortperm`.

gregrmunday: Yes fair point! I kind of like `sort` for the latitudes then using the `sortperm` result to reorder the data array. Feels more robust to whatever order the lats come in than assuming they'll always be in reverse order

SpeedyWeather/ext/SpeedyWeatherGeoMakieExt.jl
gregrmunday: I think it should be a view, I've updated this in the latest commit

ext/SpeedyWeatherGeoMakieExt.jl
gregrmunday: I've done what you've suggested here in the latest commit!

milankl:
@gregrmunday could this be updated and merged -- assuming the projection issue persists?

milankl:
Docs preview: https://speedyweather.github.io/SpeedyWeatherDocumentation/previews/PR958/output#Visualizing-output

gregrmunday:
Hi @maximilian-gelbrecht, would you mind having a quick go-over this PR? Milan says you wrote the original 'animate' function so would be good check that you're happy with this fix!

SpeedyWeather/ext/SpeedyWeatherGeoMakieExt.jl
```diff
@@ -74,9 +74,13 @@ function SpeedyWeather.animate(
 
     # Get dimensions
     lon = longitude_shift_180(ds["lon"][:]) # longitudes -180 to 180
-    lat = ds["lat"][:]
+    raw_lat = ds["lat"][:]
     time = ds["time"][:]
 
+    # Sort latitudes
+    lat_sort_idx = sortperm(raw_lat)
```
milankl: Can we not `sort` but `reverse` if descending? I feel like it's a fair assumption that dimensions should be sorted but that the order can change. You can use `issorted`. I don't like the implications that sorting a dimension would mean we generally support unsorted dimensions ...

SpeedyWeather/ext/SpeedyWeatherGeoMakieExt.jl
```diff
@@ -74,9 +74,13 @@ function SpeedyWeather.animate(
...
+    lat = raw_lat[lat_sort_idx]
```
milankl:
```suggestion
    lat_ascending = issorted(raw_lat)
    lat = lat_ascending ? raw_lat : reverse(raw_lat)
    @assert issorted(lat) "Latitudes need to be provided in either ascending or descending order."
```

SpeedyWeather/ext/SpeedyWeatherGeoMakieExt.jl
```diff
@@ -105,13 +109,30 @@ function SpeedyWeather.animate(
...
+        else
+            view(ds[variable], :, :, tstep)
+        end
+
+        # Make sure the array is shaped correctly
+        standardized = permutedims(raw_slice, (lon_idx, lat_idx))
+
+        # Apply same transformations as the lat/lons
+        rolled = circshift(standardized, (shift_amt, 0))
+        transf_data = rolled[:, lat_sort_idx]
```
milankl: Latitudes need to be reversed depending on `lat_ascending`

milankl (commented):
Looks great, I would just suggest to restrict ourselves here to latitudes ascending/descending only

maximilian-gelbrecht (approved):
I am fine with those changes. 

Just while we are already on this function, it would be great to generalise this to allow for Zarr datasets as well. They can be accessed in the same way, so the changes are really minimal (https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/other_output#Zarr-Output) 
~~The only change I think that we'd have to make is that in Zarr output the order of the dimensions is different, so the `view(ds[variable], ... ` would have to be in a different order with Zarr output. Everything else should work as is.~~

I was wrong about the order of dimension, I was just confused between Python and Julia. Even the order is identical. So in the best case all that is needed is just allowing zarr datasets as input arguments.

SpeedyWeather/ext/SpeedyWeatherGeoMakieExt.jl
gregrmunday: Committed your suggestion below...

SpeedyWeather/ext/SpeedyWeatherGeoMakieExt.jl
gregrmunday: Done this in the latest commit!

gregrmunday:
> I am fine with those changes.
> 
> Just while we are already on this function, it would be great to generalise this to allow for Zarr datasets as well. They can be accessed in the same way, so the changes are really minimal (https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/other_output#Zarr-Output) ~The only change I think that we'd have to make is that in Zarr output the order of the dimensions is different, so the `view(ds[variable], ... ` would have to be in a different order with Zarr output. Everything else should work as is.~
> 
> I was wrong about the order of dimension, I was just confused between Python and Julia. Even the order is identical. So in the best case all that is needed is just allowing zarr datasets as input arguments.

Thanks @maximilian-gelbrecht! I've made the suggested changes now. To generalise for `zarr`, would the best path be to update the input argument type to `AbstractDataset` and then write some logic in the dispatch functions to detect whether it's a `Zarr` file or `NetCDF`? Maybe there's something simpler I'm missing. @milankl

SpeedyWeather/ext/SpeedyWeatherGeoMakieExt.jl
maximilian-gelbrecht: @gregrmunday The most pragmatic change would be just to make this `Union{NCDataset, Zarr.ZGroup}` and add Zarr as a dependency to the extension.

maximilian-gelbrecht: And then see if it actually works without any other changes, it might because the access patterns are identical.

asinghvi17: Hm are you not using ZarrDatasets.jl here?

gregrmunday: I see ok! Just to be explicit, I assume it'll need to be loaded using Zarr instead of NetCDF at some point. So some logic required to either load a Zarr or NetCDF file down in the other `animate` multiple dispatches?

```julia
"""
$(TYPEDSIGNATURES)
Create an animation from a NetCDF file. Needs a backend like 
CairoMakie or GtkMakie to be loaded at the time of calling this function.
Takes the same keyword arguments as [`SpeedyWeather.animate`](@ref).
"""
function SpeedyWeather.animate(
        file_path::String;
        kwargs...
    )

    file = NCDataset(file_path)
    output_file = animate(file; kwargs...)
    close(file)

    return output_file
end

"""
$(TYPEDSIGNATURES)
Create an animation from a SpeedyWeather Simulation object. 
Needs the output of a simulation with NetCDF output enabled and a backend like 
CairoMakie or GtkMakie to be loaded at the time of calling this function.
Takes the same keyword arguments as [`SpeedyWeather.animate`](@ref).
"""
function SpeedyWeather.animate(
        simulation::Simulation;
        kwargs...
    )
    # Extract the NetCDF file path from the simulation object
    if simulation.model.output.active
        nc_file = joinpath(simulation.model.output.run_path, simulation.model.output.filename)
    else
        error("NetCDF output is not active. Make sure the simulation has been run with NetCDF output enabled.")
    end

    # Check if the NetCDF file exists
    if !isfile(nc_file)
        error("NetCDF file $(nc_file) not found. Make sure the simulation has been run with NetCDF output enabled.")
    end

    # Call animate with the extracted file path
    return animate(
        nc_file;
        kwargs...
    )
end
```

maximilian-gelbrecht: > Hm are you not using ZarrDatasets.jl here?

No, we use Zarr.jl directly at least for the writing.

maximilian-gelbrecht: > I see ok! Just to be explicit, I assume it'll need to be loaded using Zarr instead of NetCDF at some point. So some logic required to either load a Zarr or NetCDF file down in the other `animate` multiple dispatches?

Yeah, true. Maybe just add an additional optional input argument to that function that defaults to netCDF for now, but also allows for Zarr.

milankl: Had a go at implementing this in https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/958/commits/c976544063deb980bec298dbae85de97540251f3

- introduced a `dataset_type` for `::NetCDFOutput)` and `::ZarrOutput` to dispatch over the dataset type which unfortunately don't have a common super type other than `Any`.
- `animate(::Simulation)` now gets the output path and the dataset type dispatches over those and then calls the generic
- `animate(::Any)` which expects a `NCDataset` or `ZGroup` for now, once there's a common data model among those then we could change that but at the moment it seems `NCDataset` uses CommonDataModel for super types and Zarr.jl uses DiskArrays?

milankl:
There's another issue @maximilian-gelbrecht just figured out which is

```julia
time_units = ds["time"].attrib["units"] # netCDF
time_units = ds["time"].attrs["units"]  # Zarr
```

(nice example why unabbreviated naming "attributes" is less ambiguous) so we'll need to distinguish between them unfortunately ... @lazarusA @alexander-barth how did this come about? NetCDF vs Zarr standards? So annoying 🤣

Alexander-Barth:
If you use `ZarrDatasets`, it is `attrib` too (as in `NCDatasets`).

asinghvi17:
NCDatasets and ZarrDatasets are both backed by CommonDataModel.jl which does use DiskArrays, Zarr is just lower in the stack and more intimately tied to DiskArrays.  ZarrDatasets.jl is a layer on top of Zarr.jl as NCDatasets is a layer on top of raw netcdf.

For this level of thing I'd say using ZarrDatasets is probably the right move.  Zarr.jl will also get some performance improvements fairly soon thanks to @glwagner

maximilian-gelbrecht:
I wrote the extension for writing Zarr output directly with Zarr.jl because there's so little activity and stars on ZarrDatasets.jl. I took that as a red flag. But if you assure me it won't get abandoned then we can also switch.

milankl:
@maximilian-gelbrecht you think this PR should only be merged after a switch from Zarr.jl to ZarrDatasets.jl ? I"m wondering whether we can leave the SpeedyWeatherZarrExt unchanged (because loading ZarrDatasets would also load Zarr which would then load that extension??) but change the GeoMakie extension to also automatically load ZarrDatasets? That way we use the same API inside `animate` but don't need to change the output writer ...

maximilian-gelbrecht:
> @maximilian-gelbrecht you think this PR should only be merged after a switch from Zarr.jl to ZarrDatasets.jl ? I"m wondering whether we can leave the SpeedyWeatherZarrExt unchanged (because loading ZarrDatasets would also load Zarr which would then load that extension??) but change the GeoMakie extension to also automatically load ZarrDatasets? That way we use the same API inside `animate` but don't need to change the output writer ...

Loading with ZarrDatasets, while saving directly with Zarr is also totally possible. We could just do that separately. 
The difference in how the attribute is called to me really seems to be in the basic specifications, ZarrDatasets just defines the netCDF-compatible one on top of that: https://github.com/JuliaGeo/ZarrDatasets.jl/blob/a0dfef986f33eea7ccb5281aeee96191f43fb2c2/src/dataset.jl#L19

bgroenks96 (approved):

milankl:
@gregrmunday what's the remaining issue here? Can this be merged, Brian needs it, and then we just open an issue for any issue that remains?

gregrmunday:
> @gregrmunday what's the remaining issue here? Can this be merged, Brian needs it, and then we just open an issue for any issue that remains?

Yeah I think there was just a discussions about how to generalise to Zarr output, not sure if it's settled or not but I think it could be sorted in another PR. If all tests pass I think it's good to go!

bgroenks96:
I would indeed defer the ZarrDatasets.jl discussion to another issue/PR. Seems orthogonal to me.

gregrmunday:
All tests pass, just the AMD GPU test sort of hanging right now. Am I ok to merge @milankl ?

bgroenks96:
@milankl already told me he was happy to merge it, so I think you're good to go.

maximilian-gelbrecht:
> All tests pass, just the AMD GPU test sort of hanging right now. Am I ok to merge @milankl ?

GPU CI just has long queues, Nvidia and AMD also separate ones. This is why this often needs a few hours.

PR #959, opened 2026-01-02
Data-driven surface roughness scheme

gregrmunday:
PR to implement a machine-learned surface roughness scheme. 

**Land**: a light-weight feedforward neural network
**Ocean**: a symbolic expression distillation of a MLP
**Sea-ice**: an expression used by IFS for $z_0$ over sea-ice

Both have been trained on ERA5 reanalysis.

tldr; more realistic spatially heterogeneous boundary layer drag (bld) over land, and more realistic bld over the ocean + sea-ice.

### DIFF
--- a/SpeedyWeather/ext/SpeedyWeatherLuxExt.jl
+++ b/SpeedyWeather/ext/SpeedyWeatherLuxExt.jl
@@ -5,7 +5,7 @@ import Random, Adapt
 
 const z₀_SATURATION_LIMIT = 3.92f-3 # observed roughness saturation, from Curcic (2020)
 
-function SpeedyWeather._get_asset(path::String, name::String, ArrayType::Type{<:Array}, FileFormat::Type{<:Dict})
+function RingGrids.load_asset(path::String, name::String, ArrayType::Type{<:Array}, FileFormat::Type{<:Dict}, FillValue::Float64)
     data = NPZ.npzread(path)
     return data
 end
@@ -60,7 +60,7 @@ end
 
 Adapt.@adapt_structure SpeedyWeather.LearnedSurfaceRoughness
 function SpeedyWeather.initialize!(surface_roughness::LearnedSurfaceRoughness, ::PrimitiveEquation)
-    weights = SpeedyWeather.get_asset(
+    weights = RingGrids.get_asset(
         surface_roughness.path,
         from_assets = surface_roughness.from_assets,
         name = "land_weights",


src/physics/surface_fluxes/boundary_layer.jl
```diff
@@ -93,6 +95,83 @@ initialize!(::BulkRichardsonDrag, ::PrimitiveEquation) = nothing
     return nothing
 end
 
+const OCEAN_ROUGHNESS_MODEL_REF = Ref{ONNXRunTime.InferenceSession}()
+const LAND_ROUGHNESS_MODEL_REF = Ref{ONNXRunTime.InferenceSession}()
```
milankl: so these are actually the same?

src/physics/surface_fluxes/boundary_layer.jl
```diff
@@ -93,6 +95,83 @@ initialize!(::BulkRichardsonDrag, ::PrimitiveEquation) = nothing
...
+
+function _load_model_safe!(ref_container, filename)
```
milankl: what about doing this during `initialize!(::BulkRichardsonDrag, ::PrimitiveEquation)` (currently just `= nothing`) that's what it's for. Every component decides for themselves what has to be done during initialisation.

src/physics/surface_fluxes/boundary_layer.jl
```diff
@@ -93,6 +95,83 @@ initialize!(::BulkRichardsonDrag, ::PrimitiveEquation) = nothing
...
+const O_INPUT_MEANS = Float32[0.14675693, 0.24450141, 0.17968568, 7.6465526]
+const O_INPUT_STDS = Float32[0.3357911, 6.3831706, 5.53958, 3.6144474]
+const O_OUTPUT_MEAN = Float32(-8.76918)
+const O_OUTPUT_STD = Float32(1.2418048)
+
+# Land normalisation parameters
+const L_INPUT_MEANS = Float32[0.100255094, 0.154690117, 16791.6582, 6.33934355]
+const L_INPUT_STDS = Float32[0.27687377, 0.32951584, 11649.544, 4.8004246]
+const L_OUTPUT_MEAN = Float32(-5.0261893)
+const L_OUTPUT_STD = Float32(2.4474483)
```
milankl: these could be arrays in an `LearnedSurfaceRoughness <: AbstractSurfaceRoughness` type no? Whether you set these as defaults or load them from file during `initialize!` is both possible then

milankl (commented):
So I'm thinking we should have

```julia
abstract type AbstractSurfaceRoughness <: AbstractParameterization end

export ConstantSurfaceRoughness
@kwdef struct ConstantSurfaceRoughness{NF} <: AbstractSurfaceRoughness
    "[OPTION] roughness length over land [m]"
    roughness_length_land::NF = 0.5

    "[OPTION] roughness length over ocean [m]"
    roughness_length_ocean::NF = 1.0e-4
end

Adapt.@adapt_structure ConstantSurfaceRoughness
ConstantSurfaceRoughness(SG::SpectralGrid, kwargs...) = ConstantSurfaceRoughness{SG.NF}(; kwargs...)
initialize!(::ConstantSurfaceRoughness, ::PrimitiveEquation) = nothing

export BulkRichardsonDrag

"""Boundary layer drag coefficient from the bulk Richardson number,
following Frierson, 2006, Journal of the Atmospheric Sciences.
$(TYPEDFIELDS)"""
@kwdef struct BulkRichardsonDrag{NF, SR} <: AbstractBoundaryLayer
    "[OPTION] von Kármán constant [1]"
    von_Karman::NF = 0.4

    "[OPTION] Critical Richardson number for stable mixing cutoff [1]"
    critical_Richardson::NF = 10

    "[OPTION] Drag minimum to avoid zero surface fluxes in stable conditions [1]"
    drag_min::NF = 1.0e-5

    surface_roughness::SR = ConstantSurfaceRoughness{NF}()
end

Adapt.@adapt_structure BulkRichardsonDrag
BulkRichardsonDrag(SG::SpectralGrid, kwargs...) = BulkRichardsonDrag{SG.NF}(; kwargs...)
function initialize!(drag::BulkRichardsonDrag, model::PrimitiveEquation)
    # nothing to initialize for the boundary drag itself
    initialize!(drag.surface_roughness, model)
    return nothing
end
```

that way you can define your own `LearnedSurfaceRoughness <: AbstractSurfaceRoughness` with default still being what's currently done and just create 

```julia
surface_roughness = LearnedSurfaceRoughness(spectral_grid)
boundary_layer_drag = BulkRichardsonDrag(spectral_grid; surface_roughness)
model = PrimitiveWetModel(spectral_grid; boundary_layer_drag)
```

and build up your model Russian doll-style!

src/physics/surface_fluxes/boundary_layer.jl
```diff
@@ -93,6 +95,83 @@ initialize!(::BulkRichardsonDrag, ::PrimitiveEquation) = nothing
...
+        @warn "ONNX Model not found at $path"
+    else
+        # Load directly into the container passed as argument
+        ref_container[] = ONNXRunTime.load_inference(path)
+    end
+end
+
+function __init__()
+    _load_model_safe!(OCEAN_ROUGHNESS_MODEL_REF, "ocean_model/model.onnx")
+    return _load_model_safe!(LAND_ROUGHNESS_MODEL_REF, "land_model/rf_z0_land_model.onnx")
```
milankl: Ah sorry I see you load in the two different models in each container respectively!

src/physics/surface_fluxes/boundary_layer.jl
```diff
@@ -70,7 +71,8 @@ initialize!(::BulkRichardsonDrag, ::PrimitiveEquation) = nothing
     # maximum drag Cmax from that height, stable conditions would decrease Cmax towards 0
     # Frierson 2006, eq (12)
     κ = drag.von_Karman
-    z₀ = land_fraction * drag.roughness_length_land + (1 - land_fraction) * drag.roughness_length_ocean
+    # z₀ = land_fraction * drag.roughness_length_land + (1 - land_fraction) * drag.roughness_length_ocean
+    z₀ = land_fraction * surface_roughness_land(ij, diagn, progn) + (1 - land_fraction) * surface_roughness_ocean(ij, diagn, progn)
```
milankl: Here then generally `surface_roughness_land(ij, diagn, progn, drag)` I guess and `ConstantSurfaceRoughness` can implement it's own super-simple version of that function and `LearnedSurfaceRoughness` can implement another 😉

src/physics/surface_fluxes/boundary_layer.jl
```diff
@@ -21,39 +21,166 @@ initialize!(::ConstantDrag, ::PrimitiveEquation) = nothing
...
+initialize!(::ConstantSurfaceRoughness, ::PrimitiveEquation) = nothing
+
+export LearnedSurfaceRoughness
+@kwdef struct LearnedSurfaceRoughness{NF, M} <: AbstractSurfaceRoughness
+    "ONNX Inference Session for Ocean"
+    ocean_model::M
+
+    # Ocean normalisation parameters
+    ocean_input_means::Vector{NF} = Float32[0.14675693, 0.24450141, 0.17968568, 7.6465526]
+    ocean_input_stds::Vector{NF} = Float32[0.3357911, 6.3831706, 5.53958, 3.6144474]
```
milankl: To execute code on the GPU the array type has to be flexible. `Vector{T}` is a Julia (CPU) array and couldn't be a e.g. CuArray. You can write `::V` instead and add this to the parameters of the type like `{NF, V, M}`. See src/dynamics/geometry.jl or the definition of HyperDiffusion

src/physics/surface_fluxes/boundary_layer.jl
```diff
@@ -21,39 +21,166 @@ initialize!(::ConstantDrag, ::PrimitiveEquation) = nothing
...
+    full_land_path = joinpath(@__DIR__, "../../../input_data", land_path)
+
+    if !isfile(full_ocean_path) || !isfile(full_land_path)
+        @warn "ONNX models not found. Ensure paths are correct."
+    end
+
+    ocean_model = ONNXRunTime.load_inference(full_ocean_path)
+    land_model = ONNXRunTime.load_inference(full_land_path)
+
+    return LearnedSurfaceRoughness{SG.NF, typeof(ocean_model)}(;
```
milankl: and then the new `V` or `VectorType` has to be added here from `(; VectorType) = SG` and conversion will be done automatically

src/physics/surface_fluxes/boundary_layer.jl
```diff
@@ -21,39 +21,166 @@ initialize!(::ConstantDrag, ::PrimitiveEquation) = nothing
...
+end
+
+@propagate_inbounds function surface_roughness_ocean(ij, scheme::ConstantSurfaceRoughness, diagn, progn)
+    return scheme.roughness_length_ocean
+end
+
+@propagate_inbounds function surface_roughness_land(ij, scheme::LearnedSurfaceRoughness, diagn, progn)
+    vₕ = diagn.physics.land.vegetation_high[ij]
+    vₗ = diagn.physics.land.vegetation_low[ij]
+    g = diagn.dynamics.geopotential[ij]
```
milankl: I see that this may look straight forward but doesn't make sense: `diagn.dynamics.geopotential` is a `SpectralVariable3D`. If you index this with `[ij]` (leaving out the vertical index `k`) then you're actually running through the spherical harmonics at the uppermost layer furthest away from the ground. Also the element type is complex that's why you probably got an error and introduced the `real` call below. I don't know what you trained the model on but I assume it was orography then use `model.orography.orography[ij]`

I want to retire the spectral geopotential in `diagn.dynamics.geopotential` as we also have `diagn.grid.geopotential` which is the geopotential on the grid -- but also 3D. If you want to use that at the lowermost layer then it would be

```julia
g = diagn.grid.geopotential[ij, end]
```

which you also could divide by `planet.gravity` again to get height of the lowermost layer above sea level

src/physics/surface_fluxes/boundary_layer.jl
```diff
@@ -21,39 +21,166 @@ initialize!(::ConstantDrag, ::PrimitiveEquation) = nothing
...
+    sd = progn.land.snow_depth[ij]
+
+    raw_inputs = [vₕ, vₗ, g, sd]
```
milankl: this is allocating and will create a (small) vector over and over again. If you need to wrap a few values you can use a tuple `(a, b, c)` but this code here looks to me like you could just have

```julia
normalize(a, m, s) = (a - m) / s
vₕ = normalize(vₕ, scheme.land_input_means[1])
vₗ = ...
g = ...
sd = ...
```

given we'll only have O(10) inputs, I'd just give them names instead of wrapping them into a vector.

src/physics/surface_fluxes/boundary_layer.jl
```diff
@@ -21,39 +21,166 @@ initialize!(::ConstantDrag, ::PrimitiveEquation) = nothing
...
+    # input = Dict("float_input" => predictors)
+
+    # # Run Inference using the model stored in the struct
+    # prediction = scheme.land_model(input)["variable"][1]
+
+    # log_surface_roughness = (prediction * L_OUTPUT_STD) + L_OUTPUT_MEAN
+    # surface_roughness = exp(log_surface_roughness)
+
+    # Symbolic regression replacement expression
+    surface_roughness = ((cos(inputs_norm[4]) / 0.1970652) ^ -7.5047336) * ((inputs_norm[1] + (2.305061 ^ sin(inputs_norm[1]))) ^ (1.3223927 ^ (0.3254184 ^ inputs_norm[2])))
```
milankl: probably preliminary that you're hardcoding the symbolic regression coefficients in here but regarding type stability you could also write them as `0.1970652f0` to hardcode them into Float32. I guess for symbolic regression the right approach would be to have the coefficients in `LearnedSurfaceRoughness` as defaults. That way it's easier to control their number format upfront and it's user facing

src/physics/surface_fluxes/boundary_layer.jl
```diff
@@ -21,39 +21,166 @@ initialize!(::ConstantDrag, ::PrimitiveEquation) = nothing
...
+    ℵ = progn.ocean.sea_ice_concentration[ij]
+    Uₛ = diagn.grid.u_grid[ij, surface]
+    Vₛ = diagn.grid.v_grid[ij, surface]
+    UVₛ = diagn.physics.surface_wind_speed[ij]
+
+    raw_inputs = [ℵ, Uₛ, Vₛ, UVₛ]
+    inputs_norm = Float32.((raw_inputs .- scheme.ocean_input_means) ./ scheme.ocean_input_stds)
+
+    predictors = reshape(inputs_norm, 1, 4)
+    input = Dict("x_input" => predictors)
```
milankl: in addition to the comments from above creating small dictionaries on the fly will also have a (possibly nonegligible) impact on performance, e.g. this is also allocating

```julia
julia> f(x) = Dict("a" => x)

julia> using BenchmarkTools

julia> v = rand(100);

julia> @btime f($v)
  37.344 ns (4 allocations: 448 bytes)
Dict{String, Vector{Float64}} with 1 entry:
  "a" => [0.234329, 0.435124, 0.0358745, 0.37487, 0.751913, 0.77716, 0.470596, …
```

if you want to create some collection on the fly then it's recommended to use either tuples or NamedTuples both of which can be compiled away, e.g.

```julia
julia> f(a,b,c,d) = (a,b,c,d)
f (generic function with 2 methods)

julia> @btime f($a, $b, $c, $d)
  1.541 ns (0 allocations: 0 bytes)
(1, 2, 3, 4)
```

or often helpful if you want to preserve the names of the entries a NamedTuple via `(; ...)` 

```julia
julia> f(a,b,c,d) = (; a,b,c,d)
f (generic function with 2 methods)

julia> @btime f($a, $b, $c, $d)
  1.541 ns (0 allocations: 0 bytes)
(a = 1, b = 2, c = 3, d = 4)
```

see `?NamedTuple` for more information. Both tuple and NamedTuple are immutable hence they don't cause allocations but it obviously depends on what the ONNX runtime accepts. We could also use some views on scratch arrays and write data in those that also wouldn't cause allocations as it would reuse existing memory. That's why we have a handful of scratch memory

src/physics/surface_fluxes/boundary_layer.jl
```diff
@@ -21,39 +21,166 @@ initialize!(::ConstantDrag, ::PrimitiveEquation) = nothing
...
+
+    # Run Inference using the model stored in the struct
+    prediction = scheme.ocean_model(input)["linear_5"][1]
+
+    # Post-process
+    log_surface_roughness = (prediction * scheme.ocean_output_std) + scheme.ocean_output_mean
```
milankl: would this not return a Float32? Then just write `2.43...f-5` below that'll be more type stable then doing computation in Float64 and converting back

src/physics/surface_fluxes/boundary_layer.jl
```diff
@@ -21,39 +21,166 @@ initialize!(::ConstantDrag, ::PrimitiveEquation) = nothing
...
-    "[OPTION] roughness length over ocean [m]"
-    roughness_length_ocean::NF = 1.0e-4
-
     "[OPTION] Critical Richardson number for stable mixing cutoff [1]"
     critical_Richardson::NF = 10
 
     "[OPTION] Drag minimum to avoid zero surface fluxes in stable conditions [1]"
     drag_min::NF = 1.0e-5
+
+    surface_roughness::SR = ConstantSurfaceRoughness{NF}()
```
milankl: either provide default here or 👇🏼

src/physics/surface_fluxes/boundary_layer.jl
```diff
@@ -21,39 +21,166 @@ initialize!(::ConstantDrag, ::PrimitiveEquation) = nothing
...
 end
 
 Adapt.@adapt_structure BulkRichardsonDrag
-BulkRichardsonDrag(SG::SpectralGrid, kwargs...) = BulkRichardsonDrag{SG.NF}(; kwargs...)
-initialize!(::BulkRichardsonDrag, ::PrimitiveEquation) = nothing
+function BulkRichardsonDrag(
+        SG::SpectralGrid;
+        surface_roughness = ConstantSurfaceRoughness(SG), # Default if not provided
```
milankl: or here but I wouldn't do both otherwise it's difficult to see which has precedence

src/physics/surface_fluxes/boundary_layer.jl
```diff
@@ -70,7 +197,12 @@ initialize!(::BulkRichardsonDrag, ::PrimitiveEquation) = nothing
     # maximum drag Cmax from that height, stable conditions would decrease Cmax towards 0
     # Frierson 2006, eq (12)
     κ = drag.von_Karman
-    z₀ = land_fraction * drag.roughness_length_land + (1 - land_fraction) * drag.roughness_length_ocean
+
+    # Calculate land and ocean roughness lengths
+    z₀_land = surface_roughness_land(ij, drag.surface_roughness, diagn, progn)
```
milankl: feel free to pass on orography if you need it

milankl (commented):
Great work already just some comments on Julia specific stuff

src/physics/surface_fluxes/boundary_layer.jl
gregrmunday: I think I've effectively passed on orography using the bottom layer of grid.geopotential now!

src/physics/surface_fluxes/boundary_layer.jl
gregrmunday: Fixed this in the latest commit

src/physics/surface_fluxes/boundary_layer.jl
gregrmunday: As below!

src/physics/surface_fluxes/boundary_layer.jl
gregrmunday: Fixed this in the latest commit!

src/physics/surface_fluxes/boundary_layer.jl
gregrmunday: I *think* I've solved this in the latest commit, by including a memory buffer on initialisation and then overwriting that allocation during each inference step... Is this the scratch array you're talking about?

src/physics/surface_fluxes/boundary_layer.jl
gregrmunday: Do you mean inside the `LearnedSurfaceRoughness` struct? So along with the other normalisation coefficients?

src/physics/surface_fluxes/boundary_layer.jl
gregrmunday: Done this in the latest commit

src/physics/surface_fluxes/boundary_layer.jl
gregrmunday: Yup, done this in the latest commit!

src/physics/surface_fluxes/boundary_layer.jl
gregrmunday: Ok this one I haven't addressed, I couldn't really see inside `HyperDiffusion` or `geometry.jl` but might have missed it...

gregrmunday:
This latest version is pretty heavy memory-wise!

Baseline:
```julia
@btime SpeedyWeather.parameterization_tendencies!($diagn, $progn, $model)
  1.026 ms (21 allocations: 704 bytes)
```

with LearnedSurfaceRoughness
```julia
@btime SpeedyWeather.parameterization_tendencies!($diagn, $progn, $model)
  9.198 ms (158421 allocations: 11.22 MiB)
```

I think this is entirely due to the Lux NN for land surface roughness, since the ocean component is now just an expression. I had trouble getting `@compile` to work with Reactant, where it would always produce the result from the initial 'dummy' data once compiled, even with new input data passed to the model function. Might be user-error though :) 

If we can get `@compile` working I believe it'll look more like the benchmark in #973, but I reckon ultimately this would be solved with global inference for NN-based parameterisations.

src/physics/surface_fluxes/boundary_layer.jl
milankl: I can fix that later too.

gregrmunday:
> This latest version is pretty heavy memory-wise!
> 
> Baseline:
> 
> ```julia
> @btime SpeedyWeather.parameterization_tendencies!($diagn, $progn, $model)
>   1.026 ms (21 allocations: 704 bytes)
> ```
> 
> with LearnedSurfaceRoughness
> 
> ```julia
> @btime SpeedyWeather.parameterization_tendencies!($diagn, $progn, $model)
>   9.198 ms (158421 allocations: 11.22 MiB)
> ```
> 
> I think this is entirely due to the Lux NN for land surface roughness, since the ocean component is now just an expression. I had trouble getting `@compile` to work with Reactant, where it would always produce the result from the initial 'dummy' data once compiled, even with new input data passed to the model function. Might be user-error though :)
> 
> If we can get `@compile` working I believe it'll look more like the benchmark in #973, but I reckon ultimately this would be solved with global inference for NN-based parameterisations.

I actually can get this down to zero new allocations just using raw matrix multiplication, could be worth it in this case since the architecture is so straightforward!

@milankl

gregrmunday:
```julia
  4.818 ms (68171 allocations: 4.83 MiB)

julia> run!(sim_learned, period=Month(1))
Weather is speedy: 100%|████████████████████████████████████████████████████| Time: 0:00:09 (776.12 years/day)
```

It's running a lot faster and with less allocations than before. I'm not sure how we can decrease them any more while running over `ij` with Lux...

gregrmunday:
In the latest commit I've added `surface_roughness` as a parameterisation. This makes sense to me and I think it makes it really nice and flexible to use within SpeedyWeather. 

I've added the land, ocean and overall surface roughness as diagnostic variables, mainly so that I could plot them separately out the box but maybe this isn't the best way of doing it, and maybe takes up extra memory? 

Using `ConstantSurfaceRoughness`:

Screenshot 2026-02-17 at 13 54 32

gregrmunday:
...and now with `LearnedSurfaceRoughness`

Screenshot 2026-02-18 at 11 15 36

gregrmunday:
As part of my work on this it would be very useful to merge `gm/remote-input-data` (#983 ) into this branch so that I can produce some generalisability examples. 

e.g. I've created some 'Pangaea' boundary condition files that I'd like to load in to test the learned roughness scheme, which is now much easier to do using that system.  

My thinking here is that #983 will probably be merged before this PR, so it shouldn't cause any crazy conflicts when it comes time to merge this in. Just wanted to check your thoughts on this @milankl ?

Some fun runs with 'Pangaea' using the ConstantSurfaceRoughness, all custom conditions loaded using the new input system:

Screenshot 2026-02-20 at 09 40 12

Screenshot 2026-02-20 at 09 41 34

gregrmunday:
Snapshot of SR and BLD with deserts (including ice-sheets) and forests flipped. This setup seems to only be stable for about 4.5 months and then it blows up for some reason.

Screenshot 2026-02-25 at 15 02 05
Screenshot 2026-02-25 at 15 00 37

SpeedyWeather/ext/SpeedyWeatherLuxExt.jl
```diff
@@ -0,0 +1,161 @@
...
+
+    return LearnedSurfaceRoughness{SG.NF, typeof(land_nn), typeof(land_params), typeof(land_states)}(;
+        land_nn = land_nn,
+        land_params = land_params,
+        land_states = land_states,
+        kwargs...
+    )
+end
+
+function load_land_parameters!(params, weights::Dict{String, Array{Float32}})
```
maximilian-gelbrecht, line 40: what's the reason for this renaming actually? To me it just seems that this makes it quite inflexible, when retraining / setting up the NN differently

PR #960, opened 2026-01-05
Fix bug in parameter subsetting

bgroenks96:
This PR fixes a bug in `getindex` for `SpeedyParams` where incorrect parameter subsets were selected when multiple parameters with identical values are present.

Example:
```julia
spectral_grid = SpectralGrid(trunc = 31, nlayers = 1)   # define resolution
model_ps = parameters(model)
model_ps_zero = parameters(SpeedyParameters.reconstruct(model, zero(vec(model_ps))))
model_ps_zero_subset = model_ps_zero[["planet"]]
```
on `main` results in the full parameter set being returned due to an erroneous use of ∈ (my bad) that neglects the customized behavior of `Param` types, i.e. only the numeric value is compared.

The updated code uses `===` instead to select only the `SpeeedyParam`s that are exactly identical to the ones identified in the queried subset.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Fix bug in parameter subsetting [#960](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/960)
 - Follow-up to #783: finalize @param/@component markers [#935](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/935)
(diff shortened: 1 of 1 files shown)


bgroenks96:
@milankl any way we could do a release after this is merged?

milankl:
> @milankl any way we could do a release after this is merged?

Yes happy for you to do patch releases of SpeedyWeatherInternals anytime. The only conflict I see is that in #957 I'm restructuring the repo to separate tests such that SpeedyWeatherInternals also has its own tests (which still have to be moved in there) but if this is a bug fix it definitely should go in first.

PR #962, opened 2026-01-10
Fix typo in how_to_run_speedy.md

jveigel:
Typo? als = also?


maximilian-gelbrecht:
Sorry for the late reply, I was on vacation. Thanks for the fix and also your interest in our project!

PR #946, opened 2025-12-15
Update Benchmarks for full GPU version 

maximilian-gelbrecht:

PR #967, opened 2026-01-21
More benchmarks, now for the dynamical core

maximilian-gelbrecht:
Continues #946 and adds detailed benchmarks for all parts of the dynamical core to diagnose performance, will apply that later to #952


milankl:
Btw it looks for github as if buildkite CI runs but it doesn't (5s)

PR #969, opened 2026-01-23
One Variables 💍 to hold them all

milankl:
Major changes

### [BREAKING] `Variables` instead of `PrognosticVariables` and `DiagnosticVariables`

- changes _many_ function signatures including those that a user is expected to extend for customization
- `Variables` is dynamically allocated: `variables(::MyComponent)` has to be defined to declare required variables that are all allocated when `initialize!(model)`, if `model` is needed define `variables(::MyComponent, model::AbstractModel)` (optional `model` argument). Duplicated variables (same name, same path) are removed.
- variable paths have been changed, we now talk about `variables.group.namespace.name` whereby `namespace` is optional, `group` is one of `prognostic` (declared with `PrognosticVariable` etc), `grid`, `tendencies`, `parameterizations`, `particles`, `dynamics`, or `scratch`
- to find variables best to look at the pretty printing of `simulation.variables` which visualises a tree.
- to define a new type of variable one has to define a new variable dimension, see `src/variables/dimensions.jl`

### [BREAKING] Initial conditions modularised

- `model.initial_conditions` is now expected to be a `NamedTuple` over which is iterated to set the variables
- initial conditions order is: ocean, sea ice, land, particles, atmosphere
- `model.ocean` is expected to implement `initialize!(::Variables, ::MyOcean, model)` and set the associated variables, similar for `model.sea_ice`, `model.land`, `model.particle_advection`
- a new (atmosphere) initial condition can therefore be defined as `MyIC <: AbstractInitialCondition`, extending `initialize!(::Variables, ::MyIC, model)` and be passed on as `initial_conditions = (; my_ic = MyIC(...))`
- the order of these can therefore be determined, or individual ones easily be left out or new ones added

### [BREAKING] `set!` adapted

- to accommodate changes to the variable system `set!` now requires the variable `namespace` and `group` to be specified (default `group = :prognostic` and no namespace), e.g. `set!(simulation, sea_surface_temperature=300, namespace = :ocean)`
- `set!` can set any (custom) variable in `simulation.variables` as long as `namespace` and `group` are correctly passed on as keyword arguments

### [FEATURE] Global parameterizations

- Parameterizations now either define `parameterization!(ij, ::Variables, ...)` (column parameterization as before) or `parameterization(::Variables, ...)` (new global parameterization). These are expected to launch their own kernels. See `src/parameterizations/zenith.jl` for example
- global parameterizations are called _before_ column parameterizations
- other callsites e.g. _after_ column parameterizations may be implemented in future PRs

### [BREAKING] Particle advection determines `nparticles`

- `nparticles` removed from `SpectralGrid` and now via `model.particle_advection`, e.g. `ParticleAdvection2D(spectral_grid, nparticles = 30)`

### [FEATURE] Pretty printing with color

- all model components have pretty printing with color now, implemented using `StyledStrings`

### [BREAKING] Kernel launching functionality now in SpeedyWeatherInternals.KernelLaunching

- moved from `Utils` to `KernelLaunching`
- will be released as SpeedyWeatherInternals v0.2

### [BREAKING] Parameterization instead of physics

- the term "physics" meaning parameterizations only is replaced by "parameterizations" for file names and folders. Consequently the model constructors have a kwarg `dynamics_only = true` (to have no parameterizations) instead of previously `physics = false` (`parametrizations` was already taken for the order of parameterizations)

### [FEATURE] Constant short and longwave transmissivity implemented

- `ConstantShortwaveTransmissivity` and `ConstantLongwaveTransmissivity` are implemented and used by default in the primitive dry model for a < 1 atmospheric transmissivity (therefore non-zero emissivity)


maximilian-gelbrecht:
Yeah. Also dropping a reference here to what @bgroenks96 implemented in Terrarium: https://github.com/TUM-PIK-ESM/Terrarium.jl/blob/main/src/state_variables.jl

And an incomplete list of things of we have to address that diverge a bit from the standard design: 

- [ ] There are also work arrays that are neither a `RingGrid` nor `LowerTriangualArray`, of different sizes. 
- [ ] There's also the interpolation interpolator in the diagnostic variables currently 
- [ ] Do we introduce an automatic logic that declaring a `PrognosticVariable` declares also a tendency, grid, grid_prev, ...?

milankl:
Yeah, thanks for the summary of decisions we have to make! I kind of realised them too while designing this. 

1. Custom work arrays: My plan was to define `struct CustomDim <: AbstractVariableDims end` in each case. That's just a name and `Base.zero` can then be extended for `::AbstractVariable{CustomDim}`?
2. Interpolator should also work like 1
3. I thought to start explicit, so 

```julia
variables(::SomeComponent) = 
(
    PrognosticVariable(:abc),
    TendencyVariable(:abc),
    GridVariable(:abc),
)
```

or similar. Depending on #937 that'll change then a bit anyway!

milankl:
Okay this works now

```julia
julia> SpeedyWeather.Variables(BarotropicModel(spectral_grid))
Variables{@NamedTuple{...}, ...}
├ prognostic
│ ├ clock: Clock
│ └ vor: 560×1 LowerTriangularArray{ComplexF32, 2, Array{...}}
├ grid
│ ├ vor: 3168×1, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
│ ├ u: 3168×1, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
│ └ v: 3168×1, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
├ tendencies
│ ├ vor: 560×1 LowerTriangularArray{ComplexF32, 2, Array{...}}
│ └ grid
│   ├ vor: 3168×1, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
│   ├ u: 3168×1, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
│   └ v: 3168×1, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
├ dynamics
├ parameterizations
└ scratch
  ├ a: 560×1 LowerTriangularArray{ComplexF32, 2, Array{...}}
  └ b: 560×1 LowerTriangularArray{ComplexF32, 2, Array{...}}

julia> SpeedyWeather.Variables(ShallowWaterModel(spectral_grid))
Variables{@NamedTuple{...}, ...}
├ prognostic
│ ├ clock: Clock
│ ├ vor: 560×1 LowerTriangularArray{ComplexF32, 2, Array{...}}
│ ├ div: 560×1 LowerTriangularArray{ComplexF32, 2, Array{...}}
│ └ pres: 560-element, 33x32 LowerTriangularMatrix{ComplexF32, Array{...}}
├ grid
│ ├ vor: 3168×1, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
│ ├ u: 3168×1, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
│ ├ v: 3168×1, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
│ ├ div: 3168×1, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
│ └ pres: 3168-element, 48-ring OctahedralGaussianField{Float32, 1} as Array on CPU
├ tendencies
│ ├ vor: 560×1 LowerTriangularArray{ComplexF32, 2, Array{...}}
│ ├ div: 560×1 LowerTriangularArray{ComplexF32, 2, Array{...}}
│ ├ pres: 560-element, 33x32 LowerTriangularMatrix{ComplexF32, Array{...}}
│ └ grid
│   ├ vor: 3168×1, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
│   ├ u: 3168×1, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
│   ├ v: 3168×1, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
│   ├ div: 3168×1, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
│   └ pres: 3168-element, 48-ring OctahedralGaussianField{Float32, 1} as Array on CPU
├ dynamics
├ parameterizations
└ scratch
  ├ a: 560×1 LowerTriangularArray{ComplexF32, 2, Array{...}}
  └ b: 560×1 LowerTriangularArray{ComplexF32, 2, Array{...}}
```

And tracers too which get their own namespace

```julia
julia> add!(model, Tracer(:co2))
Dict{Symbol, Tracer} with 1 entry:
  :co2 => Tracer <: Main.SpeedyWeather.AbstractTracer…

julia> vars = SpeedyWeather.Variables(model)
Variables{@NamedTuple{...}, ...}
├ prognostic
│ ├ vor: 560×1 LowerTriangularArray{ComplexF32, 2, Array{...}}
│ ├ ocean
│   ├ vor: 560×1 LowerTriangularArray{ComplexF32, 2, Array{...}}
│   └ div: 560×1 LowerTriangularArray{ComplexF32, 2, Array{...}}
│ └ tracers
│   └ co2: 560×1 LowerTriangularArray{ComplexF32, 2, Array{...}}
├ grid
│ ├ vor: 3168×1, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
│ ├ u: 3168×1, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
│ └ v: 3168×1, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
├ tendencies
│ ├ vor: 560×1 LowerTriangularArray{ComplexF32, 2, Array{...}}
│ └ vor_grid: 3168×1, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
├ dynamics
│ └ f: 3168-element, 48-ring OctahedralGaussianField{Float32, 1} as Array on CPU
├ parameterizations
└ scratch
```

milankl:
>     * [ ]  There are also work arrays that are neither a `RingGrid` nor `LowerTriangualArray`, of different sizes.

For this one now needs (illustrating with the clock)

```julia
struct ClockVar <: AbstractVariableDims end
Base.zero(::AbstractVariable{ClockVar},     model::AbstractModel) = Clock()
```

and one then can let a component define them as

```julia
PrognosticVariable(:clock, ClockVar(), desc = "Clock", units = "s"),
```

So any new variables types/dimensions just need to be defined as `<: AbstractVariableDims` and need to extend `Base.zero(::AbstractVariable{ClockVar}, model::AbstractModel)`.

>     * [ ]  There's also the interpolation interpolator in the diagnostic variables currently

This should follow from above.

>     * [ ]  Do we introduce an automatic logic that declaring a `PrognosticVariable` declares also a tendency, grid, grid_prev, ...?

No, this has to be done manually for now. But it's automatically done for tracers.

SpeedyWeather/src/variables/variables.jl
```diff
@@ -1,199 +1,201 @@
...
-    namespace::Symbol = :atmosphere
+"""$(TYPEDSIGNATURES)
+Symbol representing a unique identifier for a variable, combining its type, namespace and name and namespace. 
+Used to remove duplicates when extracting variables from the model. E.g. `:Variable_ocean_sea_surface_temperature`."""
+identifier(v::AbstractVariable) =
+    Symbol(nonparametric_type(v), "_", v.namespace, v.namespace == Symbol() ? "" : "_", v.name)
+
+Base.unique(variables::AbstractVariable...) = unique(v -> identifier(v), variables)
+
+@kwdef struct Variables{P, G, T, D, Pa, S} <: AbstractVariables
```
maximilian-gelbrecht: `NamedTuples` are always `NamedTuple{names}{arrays...}`. It might be useful to have the names as parameters of the struct for some functions that loop over the variables. That's what we do in Terrarium in some cases.

maximilian-gelbrecht: See https://github.com/TUM-PIK-ESM/Terrarium.jl/blob/main/src/state_variables.jl

milankl: What's wrong with `keys`?

```julia
julia> keys(simulation.variables.tendencies)
(:vor, :div, :η, :grid)
```

this will return also the namespaces (here `grid`) but I've simply used `if !(key isa NamedTuple)` to filter those out ...

milankl:
@maximilian-gelbrecht I'm unsure about what to do with

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/780fe747ea5f9d111bcb2172191fd74ed929660b/SpeedyWeather/src/variables/prognostic_variables.jl#L212-L213

and

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/780fe747ea5f9d111bcb2172191fd74ed929660b/SpeedyWeather/src/variables/prognostic_variables.jl#L243

in the new variable system given we don't dispatch anymore over them. Could this be turned into `zero_prognostic_variables!(::Variables)` and `fill_prognostic_variables!(::Variables)` ?

SpeedyWeather/src/variables/variables.jl
maximilian-gelbrecht: Nothing inherently wrong with `keys`, but if you take them from the type parameters, it's guaranteed  a compile-time action, is it with `keys` as well? In most cases this won't matter so much though. Anyway of course we can also do this as e.g. 
`progn::Variables{NamedTuple{prog_names, prog_vars}, ...}` even with the current choice.

maximilian-gelbrecht:
Just delete them for now. 

I used those a bit in some of the differentiability scripts, but nowhere else iirc. With NamedTuples it's also way easier to just re-write those quickly later, with the old types that wasn't so easy.

maximilian-gelbrecht:
It doesn't have to be now, but at a later point I'd also revise the particles to just use a `NamedTuple`. The current version isn't compatible with Reactant at all, but making it a `NamedTuple` with vectors `active`, `lon`, `lat`, `\sigma` (or just `location` and `active`) would work.

bgroenks96:
> The current version isn't compatible with Reactant at all, but making it a `NamedTuple` with vectors `active`, `lon`, `lat`, `\sigma` (or just `location` and `active`) would work.

Can you clarify what you mean here? What does Reactant require?

maximilian-gelbrecht:
Reactant doesn't like arrays of custom types (I think, certainly not out-of-the-box), they should be arrays of regular number types. Our particle advection is based on `Vector{Particle{NF}}` where `Particle{NF}` is a struct 

```Julia
struct Particle{NF}
             active::Bool
             lon::NF
             lat::NF
             sigma::NF
end
```

milankl:
> Reactant doesn't like arrays of custom types (I think, certainly not out-of-the-box), they should be arrays of regular number types. Our particle advection is based on `Vector{Particle{NF}}` where `Particle{NF}` is a struct

What about defining eventually

```julia
struct Particles{NF, V1 <: AbstractVector{Bool}, V2 <: AbstractVector{NF}} <: AbstractVector{NF}
    active::V1
    lon::V2
    lat::V2
    sigma::V2
end
```

So instead of having a vector of particles with locations we define `Particles` through vectors of locations

maximilian-gelbrecht:
Yeah sure, that's kind of what I meant, just that now as we have most of these things already as `NamedTuple`s, so you could also just make it a named tuple with these four vectors as well.

bgroenks96:
Just FYI guys, I have had some trouble in Terrarium with long compile times due to the somewhat complicated procedure of building up and merging these `Variable` named tuples. I have had some preliminary success in mitigating this by using `OrderedDict` instead of `NamedTuple` during the initialization process and then converting everything to `NamedTuple`s only at the end when constructing the actual model state object.

milankl:
I haven't faced long compile times yet. But I also rewrote some of the functions although the concept is still the same. E.g. removes_duplicates with push! calls has been replaced by a uniqueness identifier.

bgroenks96:
> removes_duplicates with push! calls has been replaced by a uniqueness identifier

Was this something in SpeedyWeather? I don't think Terrarium has such a function...

Also can you point me to this "uniqueness identifier"?

milankl:
Added more styled strings 🥳 

image

maximilian-gelbrecht:
Adjusting all the tests will be annoying. I think Claude is good for a first try at adjusting those, but ping me if you need any help adjusting the tests that I wrote.

milankl:
> Adjusting all the tests will be annoying. I think Claude is good for a first try at adjusting those, but ping me if you need any help adjusting the tests that I wrote.

Thanks yeah, I believe once everything is up and running Claude should be good at iterating over the tests till they all pass. However, would you mind having a look at the `JLD2Output` already? The netCDF output in the barotropic model is already working but the JLD2Output has some options around prognostic and diagnostic variables that I'm not sure what you need here

```julia
    if output_diagnostic & output_prognostic
        jld2_file["$i"] = (simulation.prognostic_variables, simulation.diagnostic_variables)
    elseif output_prognostic
        jld2_file["$i"] = simulation.prognostic_variables
    elseif output_diagnostic
        jld2_file["$i"] = simulation.diagnostic_variables
    end
```

maximilian-gelbrecht:
I would say, for this PR, just remove those options completely, and then in a follow up PR we can slightly revise JLDOutput again (possible together with the array/to-RAM output we spoke about yesterday), so that you can choose exactly which variables to output.

milankl:
> > removes_duplicates with push! calls has been replaced by a uniqueness identifier
> 
> Was this something in SpeedyWeather? I don't think Terrarium has such a function...
> 
> Also can you point me to this "uniqueness identifier"?

@bgroenks96 Sorry I hadn't responded to this. The idea is to use Julia Base's `unique(f, itr)` function that applies `f` to every element in `itr` and decides the uniqueness on the return value. So `unique(x -> 1, itr)` would only return the first value as all of them are mapped to `1`. If you have a long tuple of all variables each with their dimension and namespace and desc etc then you may have two variables `a` that are different because the `desc` is different or the unit but their path would be the same, so the idea is to write a function that returns that path in some form, e.g.

```julia
identifier(v::AbstractVariable) = Symbol(v.namespace, v.name)
```

the exact syntax doesn't matter as long as it's unique in the sense of there can't be two fields in the same NamedTuple. Then I simply do now

```julia
    vars = filter(v -> v isa VariableType, vars)    # filter by variable type
    vars = unique(v -> identifier(v), vars)         # remove duplicates by identifier (group+namespace+name)
```

and filter that long tuple of all `vars` by their type (`PrognosticVariable`, `GridVariable`, ...) that corresponds to their group (we now do variables -> group -> namespace -> name, with the namespace being optional) and then within one group we remove duplicates by their identifier so that the creation of the NamedTuple doesn't throw this error

```julia
julia> (; a=1, a=1)
ERROR: syntax: field name "a" repeated in named tuple around REPL[42]:1
```

milankl:
@bgroenks96 I realised that the compile time of the whole new variable allocation system is now much more than it used to be

```julia
julia> @time SpeedyWeather.Variables(model)
  8.451465 seconds (12.66 M allocations: 626.808 MiB, 0.97% gc time, 99.38% compilation time)

julia> @time SpeedyWeather.Variables(model)
  0.003962 seconds (5.76 k allocations: 12.699 MiB)
```

maximilian-gelbrecht:
> @bgroenks96 I realised that the compile time of the whole new variable allocation system is now much more than it used to be
> 
> ```julia
> julia> @time SpeedyWeather.Variables(model)
>   8.451465 seconds (12.66 M allocations: 626.808 MiB, 0.97% gc time, 99.38% compilation time)
> 
> julia> @time SpeedyWeather.Variables(model)
>   0.003962 seconds (5.76 k allocations: 12.699 MiB)
> ```

Which two versions are you comparing there? The old one that was just the parametrizations to now everything?

maximilian-gelbrecht:
Over at https://github.com/NumericalEarth/NumericalEarth.jl/pull/133, I noticed that with the new variable system we don't have prescribed ocean/land/sea-ice allocated nicely as before. 

We should probably just address this here in the PR. Either we just introduce `PrescribedOcean`, ... etc like I did temporary in that script, or we make just make e.g. `sea_surface_temperature` just part of the standard variables of `PrimitiveWetModel`. It is needed for the parametrizations, so it should always be allocated. Basically it boils down to whether we want to write `PrimitiveWetModel(.... , ocean=nothing, ...)` or  `PrimitiveWetModel(.... , ocean=PrescribedOcean(), ...)`.

milankl:
Still need to work myself through CI failures but otherwise ready to review I reckon!!

docs/src/forcing_drag.md
```diff
@@ -194,26 +194,23 @@ then this will be called automatically with multiple dispatch.
...
-The first argument has to be of type `DiagnosticVariables` as the diagnostic variables,
-are the ones you want to change (likely the tendencies within) to apply a forcing.
-But technically you can change anything else too, although the results may be unexpected.
-The diagnostic variables contain the current model state in grid-point space and the
-tendencies (in grid and spectral space). The second argument has to be of type
-`PrognosticVariables` because, in general, the forcing may use (information from)
-the prognostic variables in spectral space, which includes in `progn.clock.time` the current
-time for time-dependent forcing. But all prognostic variables should be considered read-only.
+The first argument has to be of type `Variables` as it contains the tendencies you will
+want to change as well as the current model state. But all state fields should be
```
maximilian-gelbrecht, line 209: What language do we use now in general with this? Here you write "all state fields" now, but mean the prognostic variables, right? I think just continuing to name them prognostic is more clear.

maximilian-gelbrecht, line 209: What language do we use now in general with this? Here you write "all state fields" now, but mean the prognostic variables, right? I think just continuing to name them prognostic is more clear.

LowerTriangularArrays/src/spectrum.jl
```diff
@@ -126,10 +126,11 @@ Base.:(==)(s1::Spectrum, s2::Spectrum) =
...
 
 function Base.show(io::IO, S::Spectrum)
-    println(io, "T$(S.mmax - 1) Spectrum")
-    println(io, "├ lmax=$(S.lmax) (degrees)")
-    println(io, "├ mmax=$(S.mmax) (orders)")
-    return print(io, "└ architecture: $(typeof(S.architecture))")
+    println(io, styled"T$(S.mmax - 1) {warning:Spectrum}\{...\}")
+    println(io, styled"├ {info:lmax} = $(S.lmax) {note:(degrees)}")
+    println(io, styled"├ {info:mmax} = $(S.mmax) {note:(orders)}")
+    print(io, styled"└ {info:architecture} = $(typeof(S.architecture))")
```
maximilian-gelbrecht, line 131: what's the difference with StyledStrings?

SpeedyWeather/src/dynamics/clock.jl
```diff
@@ -1,6 +1,6 @@
 const DEFAULT_DATE = DateTime(2000, 1, 1)
 
-abstract type AbstractClock end
+abstract type AbstractClock <: AbstractModelComponent end
```
maximilian-gelbrecht, line 3: Shouldn't Clock then also get a `variables`? But I assume that's somewhere, I haven't seen yet.

SpeedyWeather/src/dynamics/forcing.jl
```diff
@@ -149,6 +147,12 @@ function StochasticStirring(SG::SpectralGrid; kwargs...)
     return StochasticStirring{SG.NF, SG.VectorType}(; lat_mask, kwargs...)
 end
 
+function variables(::StochasticStirring)
+    return (
+        ScratchVariable(:a_2D, Spectral2D(), desc = "2D spectral scratch space", units = "?"),
```
maximilian-gelbrecht: What's our convention on scratch memory now? Always explicitly allocate it, when needed? 

Above in the documentation it says `a` , `b` ,`c`, `d` are always there.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Revised variable system based on NamedTuples [#969](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/969)
```
maximilian-gelbrecht: Probably best to mark this as a very breaking change.

SpeedyWeather/src/SpeedyWeather.jl
maximilian-gelbrecht, line 1: Yes! Agreed with the physics to parametrization renaming

SpeedyWeather/src/dynamics/initial_conditions.jl
```diff
@@ -1,69 +1,59 @@
+""" Initial conditions for SpeedyWeather.jl models, i.e. the state of the atmosphere,
```
maximilian-gelbrecht, line 1: We already discussed this, I like this change!

SpeedyWeather/src/dynamics/initial_conditions.jl
```diff
@@ -1,69 +1,59 @@
...
-    temp::T = JablonowskiTemperature(spectral_grid)
-    humid::H = ConstantRelativeHumidity(spectral_grid)
+"""$(TYPEDSIGNATURES)
+Loops over all initial conditions in `model.initial_conditions` and applies one after another."""
+function initialize!(vars::Variables, model::AbstractModel)
+
+    # first initialize ocean, sea ice and land
+    hasproperty(model, :ocean) && initialize!(vars, model.ocean, model)
+    hasproperty(model, :sea_ice) && initialize!(vars, model.sea_ice, model)
+    hasproperty(model, :land) && initialize!(vars, model.land, model)
```
maximilian-gelbrecht, line 40: What does this initialization do now here? Is it the normal one we had in the model constructors before?

SpeedyWeather/src/dynamics/scaling.jl
```diff
@@ -1,85 +1,46 @@
...
-        diagn::DiagnosticVariables,
-        scale::Real
-    )
-
-    new_scale = scale / progn.scale[]     # undo previous scale and new scale in one go
-    scale!(progn, :vor, new_scale)
-    scale!(progn, :div, new_scale)
-    progn.scale[] = scale               # store scaling information
-    return diagn.scale[] = scale               # store scaling information
+@propagate_inbounds function scale!(vars::Variables, scale::Real)
```
maximilian-gelbrecht: This function could be easily misunderstood now, as it only scales one part of the input `var`. 

I would say one could be more explicit. For example defining it so that it s `scale!(vars.prognostic)` or by calling it `scale_prognostic!`. That would be more transparent.

SpeedyWeather/src/dynamics/scaling.jl
```diff
@@ -1,85 +1,46 @@
...
-"""
-$(TYPEDSIGNATURES)
-Undo the radius-scaling for any variable. Method used for netcdf output."""
-function unscale!(
-        variable::AbstractArray,
-        scale::Real
-    )
-    return variable ./= scale
+Undo the radius-scaling of vorticity and divergence from `scale!(vars, scale::Real)`."""
+function unscale!(vars::Variables)
```
maximilian-gelbrecht, line 55: Same as above

SpeedyWeather/src/dynamics/spectral_grid.jl
```diff
@@ -196,7 +184,6 @@ function SpectralGrid(
         spectrum::Spectrum,
         grid::AbstractGrid,
         dealiasing,
-        nparticles,
```
maximilian-gelbrecht, line 201: That's also a nice side benefit, that this isn't in `SpectralGrid` anymore

SpeedyWeather/src/dynamics/transform.jl
```diff
@@ -0,0 +1,208 @@
...
+    U = vars.scratch.a                              # reuse work arrays
+    V = vars.scratch.b                              # U = u*coslat, V=v*coslat
+    S = model.spectral_transform
+
+    # retain previous time step for vertical advection and parameterizations
+    if initialize == false                          # only store prev after initial step
+        @. u_grid_prev = u_grid
+        @. v_grid_prev = v_grid
+        @. temp_grid_prev = temp_grid
+        @. pres_grid_prev = exp(pres_grid)
```
maximilian-gelbrecht, line 144: Could we rename this to `exp_pres_grid_prev`?

SpeedyWeather/src/dynamics/vertical_advection.jl
```diff
@@ -108,55 +103,55 @@ end
...
-    u > 0, (2ξ[ij, k[1]] - 13ξ[ij, k[2]] + 47ξ[ij, k[3]] + 27ξ[ij, k[4]] - 3ξ[ij, k[5]]) / 60,
-    (2ξ[ij, k[6]] - 13ξ[ij, k[5]] + 47ξ[ij, k[4]] + 27ξ[ij, k[3]] - 3ξ[ij, k[2]]) / 60
+    u > 0, (2ξ[ij, k[1]] - 13ξ[ij, k[2]] + 47ξ[ij, k[3]] + 27ξ[ij, k[4]] - 3ξ[ij, k[5]]) * 1 // 60,
+    (2ξ[ij, k[6]] - 13ξ[ij, k[5]] + 47ξ[ij, k[4]] + 27ξ[ij, k[3]] - 3ξ[ij, k[2]]) * 1 // 60
 )
 
 # 2nd order centered
 @inline reconstructed_at_face(ξ, ij, k, u, ::CenteredVerticalAdvection{NF, 1}) where {NF} =
-    (ξ[ij, k[1]] + ξ[ij, k[2]]) / 2
+    (ξ[ij, k[1]] + ξ[ij, k[2]]) * 1 // 2
```
maximilian-gelbrecht, line 119: I don't know, but just to be sure: the // and fractions are supported in gpu kernels?

SpeedyWeather/src/models/primitive_dry.jl
```diff
@@ -126,15 +130,55 @@ $(TYPEDFIELDS)"""
...
-    extra_parameterizations::TS3 = (:solar_zenith, :land, :ocean, :sea_ice)
 
     # DERIVED
     # used to infer parameterizations at compile-time
     params::PV = Val(parameterizations)
 end
 
-prognostic_variables(::Type{<:PrimitiveDry}) = (:vor, :div, :temp, :pres)
-default_concrete_model(::Type{PrimitiveDry}) = PrimitiveDryModel
+function variables(::Type{<:PrimitiveDry})
```
maximilian-gelbrecht: See my comment from last week. 

Some paramtrizations require e.g a SST, but when the model is initialized with `ocean=nothing`, it previously wasn't allocated. Should we just add this here?

SpeedyWeather/src/dynamics/initial_conditions.jl
```diff
@@ -1,69 +1,59 @@
...
+
+    # then atmosphere as this may include initial conditoins like StartFromFile
```
maximilian-gelbrecht:
```suggestion
    # then atmosphere as this may include initial conditions like StartFromFile
```

SpeedyWeather/src/dynamics/spectral_grid.jl
```diff
@@ -281,6 +267,12 @@ function (::Type{S})(
     return S(spectrum, grid; NF, ArrayType, nlayers, kwargs...)
 end
 
+function variables(::SpectralTransform)
```
maximilian-gelbrecht: The two different spectral transforms need different scratch memory, so the one for the matrix transform is just missing, right?

SpeedyWeather/src/models/barotropic.jl
```diff
@@ -59,15 +59,22 @@ $(TYPEDFIELDS)"""
     feedback::FB = Feedback()
 end
 
-prognostic_variables(::Type{<:Barotropic}) = (:vor,)
-default_concrete_model(::Type{Barotropic}) = BarotropicModel
-
-parameters(model::Barotropic; kwargs...) = SpeedyParams(
```
maximilian-gelbrecht, line 65: replaced or just moved somewhere else?

SpeedyWeather/src/output/feedback.jl
```diff
@@ -17,7 +17,7 @@ $(TYPEDFIELDS)"""
     description::String = ""
 
     "[OPTION] Progress bar length, nothing = full window width"
-    progress_bar_length::Int = 20
+    progress_bar_length::Int = 0
```
maximilian-gelbrecht, line 20: 0 as default?

SpeedyWeather/src/models/barotropic.jl
```diff
@@ -59,15 +59,22 @@ $(TYPEDFIELDS)"""
...
-    planet = parameters(model.planet; component = :planet, kwargs...),
-    atmosphere = parameters(model.atmosphere; component = :atmosphere, kwargs...),
-    forcing = parameters(model.forcing; component = :forcing, kwargs...),
-    drag = parameters(model.drag; component = :drag, kwargs...),
-)
+function variables(::Type{<:Barotropic})
+    return (
+        PrognosticVariable(:clock, ClockDim(), desc = "Clock", units = "s"),
+        PrognosticVariable(:scale, ScalarDim(1), desc = "Scaling of vor and div in the dynamical core", units = "m"),
+        PrognosticVariable(:vor, Spectral4D(2), desc = "Relative vorticity", units = "1/s"),        # 2 for 2 leapfrog steps
```
maximilian-gelbrecht: So this isn't really for this PR, but in the medium term we want to be flexible with the time stepper (e.g. #937). So in principle it would be good to separate this, maybe the time stepper should actually be the component that allocates the prognostic variables. In this case we'd probably need to make the timestepper parametric on the model type, so that this can be easily dispatched at compile time.

SpeedyWeather/src/output/jld2_output.jl
maximilian-gelbrecht, line 1: Once this is merged, I would make a PR with bot help to revise this, so that one can select the output variables. This is much easier now. But yes, let us do this in a separate PR.

SpeedyWeather/src/parameterizations/radiation/longwave_radiation.jl
```diff
@@ -108,51 +108,51 @@ initialize!(::JeevanjeeRadiation, ::PrimitiveEquation) = nothing
...
+    vars.parameterizations.land.surface_longwave_up[ij] = Fₖ_land       # for land model forcing
 
     Fₖ_down = ϵ * σ * T[ij, nlayers]^4
-    diagn.physics.surface_longwave_down[ij] = Fₖ_down          # for surface energy budget
+    vars.parameterizations.surface_longwave_down[ij] = Fₖ_down          # for surface energy budget
     dTdt[ij, nlayers] -= surface_flux_to_tendency(Fₖ_down / cₚ, pₛ, model)   # out of layer k
 
     # land-sea mask weighted combined flux from land and ocean
-    local Fₖ::eltype(T)
+    Fₖ::eltype(T)
```
maximilian-gelbrecht: why was the `local` there, and why isn't it needed anymore?

SpeedyWeather/src/parameterizations/radiation/shortwave_radiation.jl
```diff
@@ -225,29 +207,29 @@ One-band shortwave radiative transfer with cloud reflection and ozone absorption
...
-    diagn.physics.albedo[ij] = albedo
+    vars.parameterizations.surface_shortwave_up[ij] = U_surface_albedo
+    vars.parameterizations.albedo[ij] = albedo
 
     U = U_surface_albedo + U_stratocumulus
     for k in nlayers:-1:1
         U_out = U * t[ij, k]
         dTdt[ij, k] += flux_to_tendency((U - U_out) / cₚ, pₛ, k, model)
-        U_out += k == cloud_top ? U_reflected : zero(U)
+        U_out += ifelse(k == cloud_top, U_reflected, zero(U))
```
maximilian-gelbrecht, line 229: thanks for these changes! The `?` also doesn't work in Reactant, but I haven't done that in the parametrizations yet.

SpeedyWeather/src/parameterizations/ocean.jl
```diff
@@ -44,44 +33,19 @@ but not masked by the land-sea mask, its value is always ignored.
...
-end
-
-# function barrier for all oceans
-function ocean_timestep!(
-        progn::PrognosticVariables,
-        diagn::DiagnosticVariables,
-        model::PrimitiveEquation
+function variables(::AbstractOcean)
+    return (
+        PrognosticVariable(:sea_surface_temperature, Grid2D(), namespace = :ocean, units = "K", desc = "Sea surface temperature"),
```
maximilian-gelbrecht, line 39: See comment about Ocean/Land allocations when `ocean=nothing`

SpeedyWeather/src/parameterizations/parameterizations.jl
```diff
@@ -0,0 +1,43 @@
...
+parameterization!(vars::AbstractVariables, parameterization::Any, model) = nothing
+
+"""$(TYPEDSIGNATURES)
+Initialize an `AbstractParameterization`. This is called once at when calling initialize!(model). 
+The default behaviour is to return `nothing`."""
+initialize!(parameterization::AbstractParameterization, model::AbstractModel) = nothing
+
+"""$(TYPEDSIGNATURES)
+Extract the parameterizations from the model as NamedTuple.
+These are the GPU-compatible components of the model."""
```
maximilian-gelbrecht, line 32: Speaking of GPU: Did you already test the whole PR on GPU? The CI is currently failing earlier than running it. I am just commenting here, because things like these `@generated` functions were a bit hard to get exactly right for GPU. I don't see an immediate reason though, why this function below shouldn't be okay.

SpeedyWeather/src/parameterizations/tendencies.jl
```diff
@@ -0,0 +1,126 @@
...
+    # parameterizations with their own kernel
+    global_parameterizations!(vars, model)
+
+    # parameterizations fused into a single kernel over ij
+    column_parameterizations!(vars, model)
+
+    return nothing
+end
+
+function global_parameterizations!(vars::Variables, model::PrimitiveEquation)
```
maximilian-gelbrecht, line 18: So global parametrization is now the Zenith and what else? 

But yeah, as in #973, I like this, we could slightly extend it in another PR, but here that's really great already!

SpeedyWeather/src/variables/dimensions.jl
```diff
@@ -0,0 +1,85 @@
...
+struct Spectral2D <: AbstractVariableDim end
+Base.zero(::AbstractVariable{Spectral2D}, model::AbstractModel) = zeros(model.spectral_grid.SpectralVariable2D, model.spectral_grid.spectrum)
+
+"""Dimension for 3D spectral variables with vertical levels."""
+@kwdef struct Spectral3D <: AbstractVariableDim
+    n::Int = 0
+end
+Base.zero(v::AbstractVariable{Spectral3D}, model::AbstractModel) = zeros(model.spectral_grid.SpectralVariable3D, model.spectral_grid.spectrum, v.dims.n == 0 ? get_nlayers(model) : v.dims.n)
+
+"""Dimension for 1D latitude-oriented variables."""
```
maximilian-gelbrecht: "latitude-oriented" means one meridian, right?

SpeedyWeather/src/variables/variables.jl
```diff
@@ -1,199 +1,230 @@
...
-# TODO: not quite sure yet about the nlayers or where it'll go
-# initialize a NamedTuple from variables
-function initialize_variables(SG::SpectralGrid, nlayers::Integer, variables...)
-    return NamedTuple{Tuple(map(v -> v.name, variables))}(Tuple(map(var -> zero(var, SG, nlayers), variables)))
+function get_steps(coeffs::LowerTriangularArray{T, 3}) where {T}
+    nsteps = size(coeffs, 3)
+    return ntuple(i -> lta_view(coeffs, :, :, i), nsteps)
 end
+
+export get_step
```
maximilian-gelbrecht, line 292: Again thinking about a more modular time stepping, maybe we could move these somewhere else in a future PR.

SpeedyWeather/src/variables/variables.jl
```diff
@@ -1,199 +1,230 @@
...
-# intialize variables
-Base.zero(::AbstractVariable{Grid2D}, SG::SpectralGrid, nlayers::Int) = zeros(SG.GridVariable2D, SG.grid)
-Base.zero(::AbstractVariable{Grid3D}, SG::SpectralGrid, nlayers::Int) = zeros(SG.GridVariable3D, SG.grid, nlayers)
-Base.zero(::AbstractVariable{Spectral2D}, SG::SpectralGrid, nlayers::Int) = zeros(SG.SpectralVariable2D, SG.spectrum)
-Base.zero(::AbstractVariable{Spectral3D}, SG::SpectralGrid, nlayers::Int) = zeros(SG.SpectralVariable3D, SG.spectrum, nlayers)
+"""Struct that holds all variables of a simulation. Variables are split into 
+groups corresponding in the fields here $(TYPEDFIELDS) Each group can have
+their own namespaces to distinguish between e.g. ocean, land, or tracer variables.
+All non-prognostic groups are considered to be diagnostic with no memory between time steps."""
+@kwdef struct Variables{Po, G, T, D, Pm, Pt, S} <: AbstractVariables
```
maximilian-gelbrecht, line 63: Oh yeah, I like the new `Variables` :)

maximilian-gelbrecht (commented):
This looks really great. I went through everything except for tests fairly quickly. I'll do another run through everything once it's really ready to be merged

bgroenks96:
> Basically it boils down to whether we want to write PrimitiveWetModel(.... , ocean=nothing, ...) or PrimitiveWetModel(.... , ocean=PrescribedOcean(), ...).

Chiming in a bit late on this, but I would vote for being explicit about defaults wherever possible. I understand the appeal of using `= nothing` as a somewhat elegant way to "disable" a particular component, but the problem with it is that it's ambiguous what `nothing` actually means. If it means that the component is literally not there, i.e. nothing gets called and no additional variables are allocated, then I think `nothing` can be OK. But if `nothing` corresponds to some kind of other "prescribed" default implementation, it would be better to create a type that documents this more clearly.

Let's not recreate in Julia the horrible Python standard practice of abusing `None` to mean "whatever default the developer decided to bury somewhere in this callstack".

SpeedyWeather/test/differentiability/full_diff_CI.jl
maximilian-gelbrecht, line 1: That's the only script that's actually executed in the CI. 

The others aren't, just leave them like they are and I'll fix them in a follow up

maximilian-gelbrecht, line 1: I think you need to merge `main` into this, I simplified this script in #995

docs/src/forcing_drag.md
milankl: Yes, good call, we should name things in the docs as they are named inside `simulation`.

LowerTriangularArrays/src/spectrum.jl
milankl: Everything is now nice in color!

image

SpeedyWeather/src/dynamics/clock.jl
milankl: Yes, clock is a funny one because I consider it here to be `<: AbstractModelComponent` (only actually used for pretty printing) but it's also in `simulation.variables`. Conceptually that might be confusing, but it's not actually a problem. To create a clock variable in `simulation.variables` we now have a clock dimension

```julia
struct ClockDim <: AbstractVariableDim end
Base.zero(::AbstractVariable{ClockDim}, model::AbstractModel) = Clock()
```

and the models then define a `PrognosticVariable` called `clock` of that dimension

```julia
function variables(::Type{<:Barotropic})
    return (
        PrognosticVariable(:clock, ClockDim(), desc = "Clock", units = "s"),
...
```

SpeedyWeather/src/dynamics/forcing.jl
milankl: The primitive models by default would have

```julia
julia> simulation.variables
...
└ scratch (521.30 KB)
  ├ a: 560×8 LowerTriangularArray{ComplexF32, 2, Array{...}}
  ├ b: 560×8 LowerTriangularArray{ComplexF32, 2, Array{...}}
  ├ a_2D: 560-element, 33x32 LowerTriangularMatrix{ComplexF32, Array{...}}
  ├ b_2D: 560-element, 33x32 LowerTriangularMatrix{ComplexF32, Array{...}}
  ├ transform_memory: 57×8×24×2 ScratchMemory{Array{ComplexF32,...}, ...}
  └ grid
    ├ a: 3168×8, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
    ├ b: 3168×8, 48-ring OctahedralGaussianField{Float32, 2} as Array on CPU
    ├ a_2D: 3168-element, 48-ring OctahedralGaussianField{Float32, 1} as Array on CPU
    └ b_2D: 3168-element, 48-ring OctahedralGaussianField{Float32, 1} as Array on CPU
```

the 2D models don't need all of them. The documentation should be updated ofc but otherwise not really a problem to let a component require `ScratchVariable(:my_scratch, ...)` it would just save memory to reuse an existing one.

SpeedyWeather/src/parameterizations/parameterizations.jl
milankl: Will check!

LowerTriangularArrays/src/spectrum.jl
maximilian-gelbrecht: Nice, looks good!

CHANGELOG.md
milankl: Added `[BREAKING]` which I think is a good policy to have makes it much easier to decide how to bump the version and add comments on users how to adapt

SpeedyWeather/src/SpeedyWeather.jl
milankl: the whole ECMWF dynamics+physics terminology is really confusing tbh.

SpeedyWeather/src/dynamics/initial_conditions.jl
milankl: The model components `ocean`, `sea_ice` and `land` now can implement their own `initialize!(::Variables, ::Component, ::AbstractModel)` method. I thought to leave `model.initial_conditions` to define the atmosphere and then when you define e.g. a new ocean model it'll also define its required `variables(::NewOcean)` as well as their initial conditions.

SpeedyWeather/src/dynamics/scaling.jl
milankl: Yeah, good call, the problem is that `scale!(vars.prognostic)` can't use dispatch as it's `::NamedTuple`. Maybe we should really just go for `scale_prognostic!` that's most explicit I agree

SpeedyWeather/src/dynamics/spectral_grid.jl
milankl: Yeah this is possible now because we pass on the component that requires a variable to `variables` like
```julia
function variables(P::ParticleAdvection2D)
    (; nparticles) = P
    return (
        PrognosticVariable(:particles, ParticleVectorDim(nparticles), desc = "Particle locations", units = "˚/1"),
```

So `nparticles` can just be defined therein, `nlayers_soil` can be defined where it belongs etc. What we currently can't do is using information across model components like allocating an array of size `nparticles x nlayers_soil` ...

SpeedyWeather/src/dynamics/transform.jl
milankl: Or should we rename `pres` to `log_pres` ?

SpeedyWeather/src/dynamics/vertical_advection.jl
milankl: Yes they should compile away, but I'll double check!!! Reason I'm trying is that on the CPU this compiles to * 0.5 without hardcoding the number format for 0.5

SpeedyWeather/src/variables/dimensions.jl
milankl: Yes, maybe that was Claude ...

SpeedyWeather/src/parameterizations/tendencies.jl
milankl: nothing else so far!! I'm also wondering about the order of the global parameterizations. At the moment it's one order and the parameterizations can implement a column parameterizations or a global or both. but global fallback is nothing

SpeedyWeather/src/SpeedyWeather.jl
maximilian-gelbrecht: As a physicist I also find the dynamical core much more physics related 👀

SpeedyWeather/src/dynamics/transform.jl
maximilian-gelbrecht: Both fine with me. I guess the `log_pres` is actually more used, so maybe go with that

milankl:
> > Basically it boils down to whether we want to write PrimitiveWetModel(.... , ocean=nothing, ...) or PrimitiveWetModel(.... , ocean=PrescribedOcean(), ...).
> 
> Chiming in a bit late on this, but I would vote for being explicit about defaults wherever possible. I understand the appeal of using `= nothing` as a somewhat elegant way to "disable" a particular component, but the problem with it is that it's ambiguous what `nothing` actually means. If it means that the component is literally not there, i.e. nothing gets called and no additional variables are allocated, then I think `nothing` can be OK. But if `nothing` corresponds to some kind of other "prescribed" default implementation, it would be better to create a type that documents this more clearly.
> 
> Let's not recreate in Julia the horrible Python standard practice of abusing `None` to mean "whatever default the developer decided to bury somewhere in this callstack".

I generally agree with this philosophy. E.g. I'm very happy to have a `PrescribedOcean` which requires `PrognosticVariable`s for the surface fluxes that the coupling would write into. But I also think that there are situations where `= nothing` is the most intuitive way to think about it. For example, if one was to set up a simulation for Mars or any other rocky planet, `ocean = nothing` should be possible and yield the correct results. Previously we did have `NoOcean`, `NoSomething` types in many places that we replaced with `nothing`. So I believe it's good to make `= nothing` possible and intuitive even though I acknowledge there are situations where it isn't. For those I'd introduce more telling `PrescribedOcean` or `NoVegetation` --  which is actually a good example. `NoVegetation` survived the move to `nothing` because the role of the vegetation at the moment is to calculate the soil moisture availability. With `nothing` nothing would happen, so that's not correct. Hence using `NoVegetation` remains the first choice here. 

Hence my suggestion is

- make `nothing` generally possible wherever intuitive
- define `NoSomething` whereever it's unclear what `nothing` would do, or where skipping a calculation wouldn't be correct

How does that sound?

maximilian-gelbrecht:
Yes, I agree. So in the concrete case then, we do define `PrescribedOcean` and `PrescribedLand` here, that allocate e.g. SST.

milankl:
> Yes, I agree. So in the concrete case then, we do define `PrescribedOcean` and `PrescribedLand` here, that allocate e.g. SST.

Yes. In the case of coupling to oceananigans and letting oceanigans compute the fluxes `ocean = nothing` is actually more appropriate and then having `surface_heat_flux = SurfaceHeatFlux(ocean=PrescribedOceanHeatFlux(spectral_grid))` etc. Because in that case we don't have control or need SST -- so allocating it would be misleading? However in case coupling is done such that Oceananigans gives us SST but we compute fluxes a `PrescribedOcean` should be defined.

I generally feel like there's so many possible combinations of components that it's really hard to anticipate all possible combinations. But it sounds like we're having a good ground to just add those things later (without anything breaking).

SpeedyWeather/ext/SpeedyWeatherFiniteDifferencesExt.jl
maximilian-gelbrecht, line 1: This massively simplified with the new `Variables`. Nice!

maximilian-gelbrecht:
I started to adjust all of the different (extended) differentiability tests. Not 100% finished with that, but I don't consider those to be absolutely essential, we can also continue with this in another PR.

milankl:
The `set!` function got a new `namespace` keyword argument so that one now has to do

```julia
set!(simulation, sea_surface_temperature = 1, namespace = :ocean)
```

otherwise we'd need to hardcode in where to find `sea_surface_temperature` and given our variables are now super dynamically allocated, I want to avoid that.

If `simulation` is passed on then it defaults to `simulation.variables.prognostic` but otherwise other variables can be set similarly, e.g.
```julia
julia> set!(simulation.variables.grid, model.geometry, geopotential=1)
```

SpeedyWeather/src/dynamics/spectral_grid.jl
milankl: Yes good point, it needs its own dimension then too

SpeedyWeather/src/output/feedback.jl
milankl: yeah = not shown. Otherwise the progress is getting too wide

```julia
 100% Time: 0:00:00 (2000-01-02, 1500.39 years/day, 45 m/s, [-83, 35] ˚C)
```

I've opened an issue in ProgressMeter.jl to modularize the bar, but until then this is all a bit hacky.

SpeedyWeather/src/output/jld2_output.jl
milankl: top!

SpeedyWeather/src/dynamics/spectral_grid.jl
maximilian-gelbrecht: I just generalised this definition to `AbstractSpectralTransform` now, and I think this already did it.

maximilian-gelbrecht:
I continued to look into the extended differentiability tests. They are not all passing, but they are on a similar level as before, and I actually used the opportunity to clean up the code a little bit. Essentially it's just really hard to get reliable estimates with finite differences here. That's visible and provable for processes like `horizontal_diffusion` or the leapfrog step for which I know the analytical solution of the derivative and can see how far of the FD is. All of that isn't really a topic for this PR though. For this PR, I adjusted the tests to the new syntax and cleaned them up a little while I was already on it. 

Aside from that I also updated the benchmark scripts.

maximilian-gelbrecht:
The extended tests workflow is wrong. That's an odd mix between the Enzyme differentiability CI and the Reactant CI now. We didn't previously run this on every commit. I still wouldn't do this. Before we had this run once per week. But I think it's also still run really running well this test, so I would just deactivate this workflow for now.

milankl:
You can just put an `if: false` into the file.

milankl:
To compile the model initialization 

1. initialize components
2. allocate variables
3. set the initial conditions

now takes

| branch | components | variables | initial conditions |
| - | - | - | - |
| `main` | 1.6s | 1.8s | 6.7s |
| `mk/newvars` | 1.7s | 3s | 14.6s |

more details now

```julia
  1.686025 seconds (8.05 M allocations: 523.823 MiB, 8.43% gc time, 94.37% compilation time)
  2.993641 seconds (7.82 M allocations: 393.368 MiB, 0.50% gc time, 99.82% compilation time)
 14.649184 seconds (21.74 M allocations: 1.053 GiB, 1.33% gc time, 99.78% compilation time)
Simulation{...} (4.84 MB)
├ variables::Variables{...} (4.00 MB)
└ model::PrimitiveWetModel{...} (1.05 MB)
```

vs main

```julia
  1.610157 seconds (8.04 M allocations: 523.474 MiB, 2.85% gc time, 94.05% compilation time)
  1.802391 seconds (7.30 M allocations: 366.475 MiB, 2.08% gc time, 99.81% compilation time)
  6.699610 seconds (18.26 M allocations: 907.748 MiB, 2.44% gc time, 99.51% compilation time)
Simulation{PrimitiveWetModel}
├ prognostic_variables::PrognosticVariables{...}
├ diagnostic_variables::DiagnosticVariables{...}
└ model::PrimitiveWetModel{...}
```

SpeedyWeather/src/parameterizations/radiation/zenith.jl
```diff
@@ -115,7 +115,7 @@ function parameterization!(vars::Variables, zenith::AbstractZenith, model::Primi
 end
 
 # as a global parameterization define the column parameterization as doing nothing
-parameterization!(ij, vars::Variables, zenith::AbstractZenith, model::PrimitiveEquation) = nothing
+parameterization!(ij, vars::Variables, zenith::AbstractZenith, model) = nothing
```
maximilian-gelbrecht, line 118: On GPU `model` is just the `NamedTuple` with the adapted entries. So the `parameterization!` has to be able to take in both

milankl: Damn overlooked this

SpeedyWeather/src/dynamics/scaling.jl
```diff
@@ -28,9 +28,9 @@ end
 
 @propagate_inbounds function scale_tendencies!(ij, vars::NamedTuple, scale::Real)
```
maximilian-gelbrecht, line 29: Is the Julia compiler smart enough to figure out these `haskey`s on compile time? Because in principle with `NamedTuples` we can do these things at compile time because the names of the variables are part of the type of `vars`

maximilian-gelbrecht: Could just iterate over all names in `NamedTuples{names}`

SpeedyWeather/src/parameterizations/parameterizations.jl
maximilian-gelbrecht: Done now

SpeedyWeather/src/dynamics/scaling.jl
milankl: In small examples it just compiles this away

```julia
julia> nt = (a=1, b=2)
(a = 1, b = 2)

julia> f(nt) = haskey(nt, :a) ? nt.a : nothing

julia> @code_llvm f(nt)
; Function Signature: f(NamedTuple{(:a, :b), Tuple{Int64, Int64}})
;  @ REPL[8]:1 within `f`
define i64 @julia_f_63937(ptr nocapture noundef nonnull readonly align 8 dereferenceable(16) %"nt::NamedTuple") #0 {
top:
; ┌ @ Base_compiler.jl:54 within `getproperty`
   %"nt::NamedTuple.unbox" = load i64, ptr %"nt::NamedTuple", align 8
   ret i64 %"nt::NamedTuple.unbox"
; └
}

julia> nt = (b=2,)
(b = 2,)

julia> @code_llvm f(nt)
; Function Signature: f(NamedTuple{(:b,), Tuple{Int64}})
;  @ REPL[8]:1 within `f`
define void @julia_f_64000(ptr nocapture noundef nonnull readonly align 8 dereferenceable(8) %"nt::NamedTuple") #0 {
top:
;  @ REPL[8] within `f`
  ret void
}
```

In a simplified form it does it here too

```julia
@propagate_inbounds function scale_tendencies!(ij, vars::NamedTuple, scale::Real)
    return haskey(vars, :humid2) ? 1 : nothing
end
```

for example compiles to `1` or `nothing` ... doing something with `vars` inside the function just spits out more LLVM code I don't understand ...

SpeedyWeather/src/dynamics/transform.jl
milankl: Let's do this in a follow-up PR on renaming variables

SpeedyWeather/src/dynamics/tendencies.jl
```diff
@@ -1002,15 +1002,20 @@ function temperature_average!(
 end
 
 function reset_tendencies!(vars::Variables; value = 0)
-    for tendency in vars.tendencies
```
maximilian-gelbrecht: These forms of nested loops into heterogeneous `NamedTuples` are not fully type stable and therefore break Enzyme differentiability. Let's see if the approach below works (not tested within Enzyme yet)

milankl: Ah shoot I keep forgetting

maximilian-gelbrecht:
The GPU CI passes for Nvidia, but shows an error for AMD that points to the rational number `//` division. 
I don't have any way to test this because I have no AMD GPU available. 

@gregordecristoforo Do you maybe know something about this? 

If not could you still quickly test on one of your AMD GPUs the following MWE:
```Julia 
using AMDGPU
using KernelAbstractions

n = 1024
u = ROCArray(rand(Float32, n))
v = ROCArray(rand(Float32, n))
out = similar(u)

# Test 1 
@. out = 1 // 2 * (u^2 + v^2)

# --- Example 2: KernelAbstractions kernel ---
@kernel function bernoulli_kernel!(out, u, v)
    I = @index(Global, Linear)
    @inbounds out[I] = 1 // 2 * (u[I]^2 + v[I]^2)
end

backend = get_backend(u)
kernel! = bernoulli_kernel!(backend, 256)
kernel!(out, u, v, ndrange = n)
KernelAbstractions.synchronize(backend)
```

milankl:
> The GPU CI passes for Nvidia, but shows an error for AMD that points to the rational number `//` division. 

The reason I started writing some kernels with rationals is that on the CPU they nicely get compiled to their floating point-agnostic equivalent. While `x/2` is compiled to `x * 0.5` (with 0.5 in the format of `x`), in contrast `x/3` is compiled to `x / 3.0` is using division (which is more expensive) not multiplication. But `a // b * x` is always compiled by precomputing the rational to the type of `x`. In contrast `a / b * x` would spit out `Float64`.

Example

```julia
julia> f(x) = 1//3 * x

julia> @code_llvm f(Float16(2))
; Function Signature: f(Float16)
;  @ REPL[11]:1 within `f`
define half @julia_f_3365(half %"x::Float16") #0 {
top:
; ┌ @ promotion.jl:434 within `*` @ float.jl:497
   %0 = fmul half %"x::Float16", 0xH3555
   ret half %0
; └
}
```

So with rationals we have ✅ precomputation of the fraction ✅ type stable ✅ avoid division.

milankl:
The docs preview is here https://speedyweather.github.io/SpeedyWeatherDocumentation/previews/PR969/ the `documenter/deploy` workflow has the wrong link in it

luraess:
Testing on CI machine, Test 2 passes.

The issue seems to be in Test 1 as your writing 
```
@. out = 1 // 2 * (u^2 + v^2)
```
will most likely add a `.` to `//` which fails. Manually writing
```
out .= 1 // 2 .* (u.^2 .+ v.^2)
```
works fine.

maximilian-gelbrecht:
Oh shit, the Enzyme diff fails in a bad way now with zero gradients. That's not good. I need to investigate this.

gregordecristoforo:
> The GPU CI passes for Nvidia, but shows an error for AMD that points to the rational number `//` division. 
> I don't have any way to test this because I have no AMD GPU available. 
> 
> @gregordecristoforo Do you maybe know something about this? 
> 
> If not could you still quickly test on one of your AMD GPUs the following MWE:
> ```Julia 
> using AMDGPU
> using KernelAbstractions
> 
> n = 1024
> u = ROCArray(rand(Float32, n))
> v = ROCArray(rand(Float32, n))
> out = similar(u)
> 
> # Test 1 
> @. out = 1 // 2 * (u^2 + v^2)
> 
> # --- Example 2: KernelAbstractions kernel ---
> @kernel function bernoulli_kernel!(out, u, v)
>     I = @index(Global, Linear)
>     @inbounds out[I] = 1 // 2 * (u[I]^2 + v[I]^2)
> end
> 
> backend = get_backend(u)
> kernel! = bernoulli_kernel!(backend, 256)
> kernel!(out, u, v, ndrange = n)
> KernelAbstractions.synchronize(backend)
> ```

I am currently on an Easter vacation but I can have a look in Tuesday when I am back in office.

maximilian-gelbrecht:
Thanks Gregor, I think we might have already resolved it with Ludovic's help. Let's just hope no other AMD exclusive issues appear.

maximilian-gelbrecht:
All green  ✅ 🎉 

Enzyme issue was just forgetting to set a seed. 

AMD issue seems to be really only that one line in the Bernoulli potential with the `@.` macro.

SpeedyWeather/src/dynamics/tendencies.jl
maximilian-gelbrecht: Ok, it worked

docs/src/forcing_drag.md
maximilian-gelbrecht: @milankl if you are still on the documentation anyway, can you incorporate that change?

SpeedyWeather/src/models/primitive_dry.jl
maximilian-gelbrecht: @milankl So what did we do with this, nothing so far, right? 

Let's still add `PrescribedOcean` and `PrescribedLand` then?

SpeedyWeather/src/models/simulation.jl
```diff
@@ -106,6 +106,7 @@ function finalize!(simulation::AbstractSimulation)
     (; variables, model) = simulation
     finalize!(model.feedback)                       # finish the progress meter, do first for benchmark accuracy
     unscale_prognostic!(variables)                  # undo radius-scaling for vor, div from the dynamical core
+    unscale_grid_variables!(variables)              # undo radius-scaling for grid variables (vor, div)
```
maximilian-gelbrecht: I think this doesn't work as unintended, or am I mistaking? 

Because `unscale_prognostic!` already sets the `scale` back to `1`

maximilian-gelbrecht: Could also add unit tests for functions like that (or let Claude do that).

maximilian-gelbrecht: Or maybe rather for checking the correct scaling in a simulation run.

milankl: Yes my bad! fix is comgin

milankl:
@gregrmunday just got this when running the documentation build. Turns out GitHub can have problems accessing its own servers

```julia
┌ Error: failed to run `@example` block in docs/src/callbacks.md:217-222
│ ```@example callbacks
│ simulation = initialize!(model)
│ run!(simulation, period=Day(3))
│ v = model.callbacks[:storm_chaser].maximum_surface_wind_speed
│ maximum(v)      # highest surface wind speeds in simulation [m/s]
│ ```
│   exception = ("Could not download asset 'land_surface_temperature.nc'. Check your internet connection or if the URL asset path exists."
```

PR #970, opened 2026-01-27
SpeedyWeather + Reactant: Basics and BarotropicModel

maximilian-gelbrecht:
Initial preparations to run Speedy with Reactant. 

This is based on `mk/matrix_transform` as I am thinking that it's easier to avoid FFT plans for the first attempt.


maximilian-gelbrecht:
Test script 

```Julia
import Pkg 
Pkg.activate("SpeedyWeather")
using SpeedyWeather, Reactant, LinearAlgebra
#using CUDA 

arch = SpeedyWeather.ReactantDevice()
nlayers = 1 
spectral_grid = SpectralGrid(; architecture = arch, nlayers)
M = MatrixSpectralTransform(spectral_grid)

spec = randn(Complex{Float32}, spectral_grid.spectrum, nlayers)
field = zeros(Float32, spectral_grid.grid, nlayers)
   
tr = @compile transform!(spec, field, M)

# model construction works 
#model = PrimitiveWetModel(spectral_grid; spectral_transform=M)

# model initialization currently fails (because of the transforms used)
#simulation = initialize!(model)
```

maximilian-gelbrecht:
At the moment, the model construction of all models works. 
We can also run e.g. the spectral transform in matrix form with Reactant thanks to @mofeing

Next, we have to think about the full model initialisation. That's more a question for our own code organisation. When we construct the model with Reactant Arrays, we should probably do all the pre-compute with it as well, but how to do this best? The alternative would be doing the pre-compute without Reactant and then transferring and converting later, but that's a bit annoying in practice. I have to have a think about the best approach.

milankl:
Btw calling NN parameterizations that are compiled with Reactant already works on the CPU, see #973

SpeedyTransforms/src/matrix_transform.jl
```diff
@@ -4,14 +4,13 @@ $(TYPEDFIELDS)"""
 struct MatrixSpectralTransform{
         NF,
         AR,                         # <: AbstractArchitecture
-        ArrayType,                  # non-parametric array type
```
maximilian-gelbrecht, line 7: FYI: Removing this redundant type parameter was actually crucial to make it work with Reactant. This is probably a bug in Reactant (they are looking into it), but it's related to Reactant having to reconstruct the structs it tracks.

SpeedyWeather/src/dynamics/horizontal_diffusion.jl
maximilian-gelbrecht, line 82: Better to do really do all the initialization in kernels, with Reactant you also get warnings about scalar indexing here

We just have one or two more of those that transfer back and forth, I'll do the others later.

SpeedyTransforms/src/matrix_transform.jl
mofeing: `traced_type` is already fixed in main, but maybe `make_tracer` still raises an error. mind giving a quick test?

maximilian-gelbrecht: Thank you! Without this type parameter the transform was already running nicely. 
I quickly checked out the old commit before that and tested it, and this one also works now. But we'll keep this change anyway, because the parameter was a relic basically, and not needed anymore.

maximilian-gelbrecht:
Good news before the weekend: 

Both initialization and time stepping of the `BarotropicModel` (our simplest model) with Reactant compiles and runs without errors 🎉 

I haven't done any form of correctness checks, and there are issues with the initial conditions, at the moment I tested it with just with zero ICs. 

```Julia
import Pkg 
Pkg.activate("SpeedyWeather")
using SpeedyWeather, Reactant, LinearAlgebra, CUDA # cuda is loaded for reactant but this is still on CPU

arch = SpeedyWeather.ReactantDevice()
nlayers = 1 
spectral_grid = SpectralGrid(; architecture = arch, nlayers)
M = MatrixSpectralTransform(spectral_grid)

spec = zeros(Complex{Float32}, spectral_grid.spectrum, nlayers)
field = rand(Float32, spectral_grid.grid, nlayers)
    
# model construction works 
#model = PrimitiveWetModel(spectral_grid; spectral_transform=M)
initial_conditions = InitialConditions(; vordiv=ZeroInitially())
# feedback == nothing is needed currently due to issues with Reactant 
model = BarotropicModel(spectral_grid; spectral_transform=M, feedback=nothing, initial_conditions=initial_conditions)

@jit SpeedyWeather.timestep!(simulation)
```

maximilian-gelbrecht:
@milankl For proper correctness of our models with Reactant we need be to aware of the functions that do have significant branch divergences / control flow those limits/conditions change while running the model. Those need to be explicitly traced by Reactant. I don't have the best overview of that currently. I can see those e.g. in convection and condensation scheme, can you think about more like these?

mofeing:
> For proper correctness of our models with Reactant we need be to aware of the functions that do have significant branch divergences / control flow those limits/conditions change while running the model. Those need to be explicitly traced by Reactant.

for this, you can add the ReactantCore dependency to SpeedyWeather (which just adds the `@trace` macro) and you can add it wherever you need in the main code. but IMO I would just  add it in the following cases:
- control-flow where the predicate is traced (e.g. a `if pred` where `pred::TracedRNumber{Bool}`)
- loops where the partial evaluation unrolls the loop and the resulting code is tooo large

maximilian-gelbrecht:
Thanks! 

> * loops where the partial evaluation unrolls the loop and the resulting code is tooo large

Just to be sure, what do you mean with the "the resulting code is too large", when do we have to watch out there? The typical situation in our model is that loop limits depend only on the model configuration that isn't changed at runtime, so I thought we don't need to trace them, with typical loop lengths of ~8-80 (for vertical layers). There are just some very few exceptions in our parametrizations.

mofeing:
> Just to be sure, what do you mean with the "the resulting code is too large", when do we have to watch out there?

IMO when compilation time is too large due to tracing.

> The typical situation in our model is that loop limits depend only on the model configuration that isn't changed at runtime, so I thought we don't need to trace them, with typical loop lengths of ~8-80 (for vertical layers). There are just some very few exceptions in our parametrizations.

AFAIK this is actually the best case for `@trace for` because that way a `for i in 1:n` gets translated to:

```mlir
%0 = stablehlo.constant 80 : i64 // `n`, in this case n=80
%1 = stablehlo.while(...) ({
  // here the comparison code
 }, {
  // here the body code
})
```

maybe @glou-nes can indicate you better

maximilian-gelbrecht:
Ok, the model isn't actually working yet. ~~Not sure why the `@jit` didn't cause an error. But currently our custom broadcasting for our array wrappers isn't working with the Reactant `TracedRArray` arrays. Not sure yet how to fix that.~~ Almost working now.

glou-nes:
> > Just to be sure, what do you mean with the "the resulting code is too large", when do we have to watch out there?
> 
> IMO when compilation time is too large due to tracing.
> 
> > The typical situation in our model is that loop limits depend only on the model configuration that isn't changed at runtime, so I thought we don't need to trace them, with typical loop lengths of ~8-80 (for vertical layers). There are just some very few exceptions in our parametrizations.
> 
> AFAIK this is actually the best case for `@trace for` because that way a `for i in 1:n` gets translated to:
> 
> ```mlir
> %0 = stablehlo.constant 80 : i64 // `n`, in this case n=80
> %1 = stablehlo.while(...) ({
>   // here the comparison code
>  }, {
>   // here the body code
> })
> ```
> 
> maybe @glou-nes can indicate you better

I can confirm this! Because of unrolling, without tracing loops, we can get pathological tracing / optimizer performance even with simple kernels.

maximilian-gelbrecht:
Ok, now the `BarotropicModel` does actually run with Reactant. I've written it so that users only need to choose the `ReactantDevice` as an architecture and then everything else is handled automatically. 

```Julia 
using SpeedyWeather, Reactant, CUDA # CUDA needs to be loaded even if it's run on CPU

arch = SpeedyWeather.ReactantDevice()
nlayers = 1
spectral_grid = SpectralGrid(; architecture = arch, nlayers)
M = MatrixSpectralTransform(spectral_grid)

# feedback doesn't work with Reactant, but they are currently fixing the issue :) 
model = BarotropicModel(spectral_grid; spectral_transform = M, feedback = nothing) 
run!(simulation, steps=10)
```

I've already written a correctness check, but I still need to fine tune it a bit. Results look quite similar for both models, but not identical.

maximilian-gelbrecht:
To-Dos 

- [x] Some code clean up
- [x] Finish correctness test (e.g. figure out good tolerances, which functions / rollouts to test) 
- [x] Test on GPU/TPU
- [x] ~~Not all initial conditions are working~~ ICs for barotropic are working
- [ ] Benchmark 
- [ ] Particle advection not working (that's harder to do, but we might just delay that) 
- [ ] Look into `@trace` for for loop even if it's not strictly necessary 
- [ ] Continue with the `PrimitiveDryModel` and `PrimitiveWetModel`

maximilian-gelbrecht:
@milankl I've had to do a lot of minor adjustments across the model code, but I think most of them make sense outside of Reactant as well: cleaning up some old implementations, adding a few more tests etc. So ideally I think it would be good to merge this before we finish the complete model on Reactant. 

I'll also look again into #952 next week as both have to merged together. But I think we can merge them while we still regard both the `MatrixSpectralTransform` and the Reactant support as experimental and limited as we done with the GPU support as well.

maximilian-gelbrecht:
Works on GPU as well, expect I get a repeated warning 

```
'86' is not a recognized feature for this target (ignoring feature)
```

Not sure what that means.

maximilian-gelbrecht:
While, the forward model is working, the reverse model / autodiff isn't. 
It already needed a few adjustments of making types proper parametric, but the correct stopper is `Dates`, see: https://github.com/EnzymeAD/Reactant.jl/issues/2046

maximilian-gelbrecht:
After talking to @mofeing , a custom `Dates` seems to be only real way to make the differentiation of the time stepping Reactant compatible. Just ditching using date types at all is the alternative, but we use them so much throughout the model that this might be even more annoying. With some help from Claude, I just set up a custom `TracableDates` submodule now that we load in with `import TracableDates as Dates` as a drop-in replacement for `Dates`.

milankl:
> After talking to @mofeing , a custom `Dates` seems to be only real way to make the differentiation of the time stepping Reactant compatible. Just ditching using date types at all is the alternative, but we use them so much throughout the model that this might be even more annoying. With some help from Claude, I just set up a custom `TracableDates` submodule now that we load in with `import TracableDates as Dates` as a drop-in replacement for `Dates`.

Sorry only trying to catch up with the updates here now. Ditching `Dates` would be a bummer though I might be happy to reinvent `Dates` to make it Reactant-compatible. However there is a quite a lot of functionality we use, including in `physics/radiation/zenith.jl` ... I hope this is not too much work. I guess this also removes our type piracy from #746 as a positive side effect ...

maximilian-gelbrecht:
With regards to Dates I just made a PR for Reactant https://github.com/EnzymeAD/Reactant.jl/pull/2540 that implements the custom dates format there to make `Dates` Reactant-compatible, so that we don't need to have ithe custom date format here anymore. I am not totally finished yet with the PR, I need to test if it actually works here with what we do in Speedy as well.

maximilian-gelbrecht:
Differentiating with Reactant won't work for the next few months at least unfortunately https://github.com/EnzymeAD/Reactant.jl/issues/2406 

I'll finish what I started in the Reactant related PRs, but it will take a bit lower priority because of that.

maximilian-gelbrecht:
I simplified to approach to Reactant a bit here by defining a 
```
@maybe_jit architecture func
```
macro. What it does it that it both checks
a) whether `architecture < :ReactantDevice` 
b) whether it's already within in a Reactant compile context (calling `@jit` or `@compile` inside of another `@jit` or `@compile` results in an error)

So with this we can largely do `initialize!` without any reactant compilation and little changes to the code, but the `timestepping!` gets compiled. 

Reactant CI is failing here because the Dates extension I wrote for Reactant isn't ~~merged~~ released yet.

maximilian-gelbrecht:
The Enzyme CI error is a bit odd. This is unrelated directly to Reactant, but in some way probably related to the changes I've made to the zenith parametrization to accommodate for Reactant. I already checked type stability. That's not the problem 

Edit: Ok, found the culprit: `Base.RefValue{<:DateTime}` causes the error, `Base.RefValue{DateTime}` works.

maximilian-gelbrecht:
I think this could actually be merged very soon. Locally most tests are passing and I just need to fix a few minor issues. The only thing we really have to wait for is a new Reactant release and that should come in a matter of days. 

This PR introduces the basics for Reactant compat and while doing that also fixes a few other issues here and there. It’s non-breaking. 

With the PR the BarotropicModel runs with Reactant. The primitive models will follow in the other PR. 

@milankl what do you think, do you want to review this and should we merge this rather quickly? I think it might be easier to merge this first and then the variables one. But I am also not completely set on that. 

What this PR does: 

- Reactant Architecture basics 
- The `@maybe_jit` macro for conditional Reactant execution (see comment above) 
- Make most types more parametric as many basic types need to be replaced by Reactant types by Reactant 
- In a few ocasions replaces "zeros" with "similar" because this works better with the different Reactant array types
- Fix/simplify broadcasting issues 
- Add a few unit tests 
- GPU-ify a few initialisations that haven't been properly GPU-ified before
- Correctness tests for the BarotropicModel with Reactant

RingGrids/src/field.jl
```diff
@@ -457,11 +457,11 @@ import Base.Broadcast: BroadcastStyle, Broadcasted, DefaultArrayStyle
 # important to not have parameter T (eltype/number format) here to broadcast
 # automatically across the same field type but with different T
 # e.g. FullGaussianField{Float32} and FullGaussianField{Float64}
-struct FieldStyle{N, Grid, Column} <: Broadcast.AbstractArrayStyle{N} end
+struct FieldStyle{N, Grid} <: Broadcast.AbstractArrayStyle{N} end
```
milankl, line 460: Did we have `Column` here for ColumnFields? Happy to deprecate that ...

RingGrids/src/field.jl
```diff
@@ -474,58 +474,53 @@ find_field(::Any, rest) = find_field(rest)
...
+FieldGPUStyle{3, Grid}(::Val{4}) where {Grid} = FieldGPUStyle{4, Grid}()
 
 function KernelAbstractions.get_backend(
         field::F
     ) where {F <: AbstractField{T, N, ArrayType}} where {T, N, ArrayType <: GPUArrays.AbstractGPUArray}
     return KernelAbstractions.get_backend(field.data)
 end
 
+Adapt.parent_type(::Type{<:AbstractField{T, N, ArrayType}}) where {T, N, ArrayType} = ArrayType
+Base.parent(field::AbstractField) = field.data
```
milankl, line 522: What is `parent` for?

SpeedyTransforms/src/spectral_transform.jl
```diff
@@ -416,8 +413,7 @@ function transform(                         # GRID TO SPECTRAL
         field::AbstractField,               # input field
         S::AbstractSpectralTransform,       # precomputed spectral transform
     )
-    ks = size(field)[2:end]                 # the non-horizontal dimensions
-    coeffs = on_architecture(architecture(field), zeros(Complex{eltype(S)}, S.spectrum, ks...))
+    coeffs = similar(field, S.spectrum, Complex{eltype(S)})
```
milankl, line 416: Ouh? How does this version know about the size of the non-horizontal dimensions?

SpeedyWeather/src/dynamics/forcing.jl
```diff
@@ -240,12 +240,25 @@ function forcing!(
...
+    )
+end
+
+@kernel inbounds = true function kolmogorov_flow_kernel!(Fu, s, k, latds)
+    I = @index(Global, NTuple)
+    ij = I[1]
+    layer = I[2]
+    lat = latds[ij]
+
+    Fu[ij, layer] += s * sin(deg2rad(k * lat))
```
milankl, line 261: `sind` doesn't work?

SpeedyWeather/src/dynamics/initial_conditions.jl
```diff
@@ -161,34 +163,36 @@ export RandomVelocity
...
+        initial_conditions::RandomVelocity{NF},
         model::Barotropic
-    )
-
-    NF = eltype(progn)
+    ) where {NF}
 
     # reseed the random number generator, for seed=0 randomly seed from Julia's global RNG
-    seed = initial_conditions.seed == 0 ? rand(UInt) : initial_conditions.seed
+    seed = initial_conditions.seed # == 0 ? rand(UInt) : initial_conditions.seed
```
milankl: Leftover comment?

SpeedyWeather/src/dynamics/random_process.jl
```diff
@@ -72,7 +72,11 @@ that is reseeded on every `initialize!`. Fields are: $(TYPEDFIELDS)"""
 end
 
 # generator function
-SpectralAR1Process(SG::SpectralGrid; kwargs...) = SpectralAR1Process{SG.NF, SG.VectorType}(trunc = SG.trunc; kwargs...)
+function SpectralAR1Process(SG::SpectralGrid; kwargs...)
+    RNG = haskey(kwargs, :random_number_generator) ? typeof(kwargs[:random_number_generator]) : typeof(Random.Xoshiro())
+    SeedType = haskey(kwargs, :seed) ? typeof(kwargs[:seed]) : Int
```
milankl, line 77: Ouh why is that needed?

SpeedyWeather/src/dynamics/spectral_grid.jl
```diff
@@ -97,14 +99,14 @@ struct SpectralGrid{
 
     # PARTICLES
     "[OPTION] number of particles for particle advection [1]"
-    nparticles::Int
+    nparticles::IntType
 
     "[DERIVED] ArrayType of particle vector"
-    ParticleVector::Type{<:AbstractArray}
+    ParticleVector::Type
```
milankl, line 105: Why is that not allowed but `SpectralVariable2D::Type{<:AbstractArray}` is fine?

SpeedyWeather/src/dynamics/spectral_grid.jl
```diff
@@ -232,10 +234,15 @@ function SpectralGrid(
...
+    # TODO: For Reactant we need something else in the long run
+    if typeof(architecture) <: ReactantDevice
+        ParticleVector = typeof(nothing) # this is just a dummy type, it's actually nonworking
+    else
+        ParticleVector = array_type(architecture, Particle{NF}, 1)
+    end
 
     # Create the SpectralGrid with all fields
-    return SpectralGrid{typeof(architecture), typeof(spectrum), typeof(grid)}(
+    return SpectralGrid{typeof(architecture), typeof(spectrum), typeof(grid), typeof(nlat_half), Float64}(
```
milankl: We can just make the dealiasing type Float32 ... feels inconsistent to carry around a float64 here whereas everything else is float32 by default. It just needs to be a float but the precision doesn't matter

SpeedyWeather/src/dynamics/time_integration.jl
```diff
@@ -240,11 +240,10 @@ function first_timesteps!(simulation::AbstractSimulation)
     (; Δt) = time_stepping
 
     # decide whether to start with 1x Euler then 1x Leapfrog at Δt
-    if time_stepping.first_step_euler
+    @trace if time_stepping.first_step_euler
```
milankl, line 243: when do we use `@trace` ?

SpeedyWeather/src/dynamics/time_integration.jl
```diff
@@ -315,7 +312,7 @@ function timestep!(
         lf1::Integer = 2,               # leapfrog index 1 (dis/enables Robert+Williams filter)
         lf2::Integer = 2,               # leapfrog index 2 (time step used for tendencies)
     )
-    model.feedback.nans_detected && return nothing  # exit immediately if NaNs/Infs already present
+    (!isnothing(model.feedback) && model.feedback.nans_detected) && return nothing  # exit immediately if NaNs/Infs already present
```
milankl, line 315: I feel like `feedback = nothing` should just not be allowed. Maybe at the `initialize!(::AbstractModel)` we should check that `@assert !isnothing(model.feedback) "model.feedback should not be nothing"`?

milankl: but maybe if that was the only change needed to make `feedback = nothing` work then I'd be happy with it too!

SpeedyWeather/src/dynamics/tracers.jl
```diff
@@ -6,13 +6,13 @@ export Tracer
 and define it as `active` (=true default). `active=false` will (temporarily)
 disable the time evolution of the tracer. Fields are
 $(TYPEDFIELDS)"""
-@kwdef mutable struct Tracer <: AbstractTracer
+@kwdef mutable struct Tracer{B} <: AbstractTracer
     name::Symbol
```
milankl, line 10: `Symbol`s are never tracked by Reactant?

SpeedyWeather/src/variables/diagnostic_variables.jl
```diff
@@ -312,11 +318,20 @@ end
...
+    # TODO: Particle advection for Reactant nonworking
+    if typeof(architecture) <: ReactantDevice
+        locator = nothing 
+        return ParticleVariables{ParticleVector, Nothing, Nothing, typeof(nparticles)}(;
+            nparticles, locator = nothing, u = nothing, v = nothing, σ_tend = nothing
+        )
+    else
+        VectorNF = array_type(architecture, NF, 1)
+        locator = RingGrids.AnvilLocator(NF, nparticles; architecture)
+    end
```
milankl, line 332: With #969 this could just become 
```julia
function variables(P::ParticleAdvection2D, model::AbstractModel)
    if typeof(model.spectral.architecture) == ReactantDevice)
         return ()
    else
         # return the variables needed
    end
end
```

neat.

milankl (commented):
Looks fantastic already!!

- happy for dropping particle support on reactant for now, that's because of the `Particle` type being custom right?
- did you make it possible to use `output = nothing` and `feedback = nothing` too?

RingGrids/src/field.jl
maximilian-gelbrecht: Yeah, we needed separate Broadcasting styles, but I simplified that now.

RingGrids/src/field.jl
maximilian-gelbrecht: It's used by Reactant, apparently it's also generally expected by array wrapper types. E.g. `parent(::Transpose)` gives you the original array.

SpeedyTransforms/src/spectral_transform.jl
maximilian-gelbrecht: It takes the non-horizontal dimensions from `field`

SpeedyWeather/src/dynamics/forcing.jl
maximilian-gelbrecht: Nope https://github.com/EnzymeAD/Reactant.jl/issues/2349

SpeedyWeather/src/dynamics/random_process.jl
maximilian-gelbrecht: Reactant has its own `RNG`, but maybe we could make this here a bit more elegant.

SpeedyWeather/src/dynamics/time_integration.jl
maximilian-gelbrecht: `@trace` is used when the condition changes at run time. 

So everything related to model setup doesn't need trace. The truncation, or the number of layers doesn't change and can be statically compiled. 

Everything that changes during runtime needs to be traced. In this case here, I am still not 100% settled on this, because it's really only the first time step. Maybe we just refactor it in a way that just doesn't use this `if` here anymore at all.

SpeedyWeather/src/dynamics/time_integration.jl
maximilian-gelbrecht: Yeah, unfortunately we currently need `feedback = nothing` for Reactant. This is because deep down the type hierarchy it uses types like `Memory` that currently aren't compatible with Reactant yet, but there's a PR already that could fix that https://github.com/EnzymeAD/Reactant.jl/pull/2427

SpeedyWeather/src/dynamics/tracers.jl
maximilian-gelbrecht: Good call. I don't know actually, maybe I just never actually tested a tracer....

SpeedyWeather/src/variables/diagnostic_variables.jl
maximilian-gelbrecht: Yeah, there are a few other things e.g. in the intial conditions that are also more simple with #969

SpeedyWeather/src/dynamics/spectral_grid.jl
maximilian-gelbrecht: Because the problem here is the `Particle`. Any type that involves `Particle` isn't allowed. `Type{<:AbstractArray{Float}}` is fine, `Type{<:AbstractArray{Particle}}` isn't

maximilian-gelbrecht:
> Looks fantastic already!!
> 
> * happy for dropping particle support on reactant for now, that's because of the `Particle` type being custom right?

Yep, exactly 

> * did you make it possible to use `output = nothing` and `feedback = nothing` too?

Yes, both.

PR #971, opened 2026-01-27
Relax Enzyme compat again

maximilian-gelbrecht:
At least the MWE of our problem with Enzyme > 0.13.104 (https://github.com/EnzymeAD/Enzyme.jl/pull/2784) isn't causing an error anymore with the most recent Enzyme version (v0.13.121). Let's test if that's also true for the complete model

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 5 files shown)


maximilian-gelbrecht:
Yeah, that works.

I'll do a brief test of the problems reported in #943 and #938 here as well

maximilian-gelbrecht:
Okay, seemingly #938 is resolved, but #943 not.

PR #974, opened 2026-02-02
Fix absorptivity_water_vapor and absorptivity_cloud_base default values

nviebig:
# Fix shortwave absorption default values

## Problem

After implementing clouds and the updated radiation scheme, we observed **excessive convection over the ocean** (~80% of ocean gridpoints triggering convection). Investigation revealed that the atmosphere was absorbing almost no shortwave radiation:

| Metric | Old Value | Expected (Earth) |
|--------|-----------|------------------|
| SW absorbed by atmosphere | **-0.8 W/m²** | 70-80 W/m² |
| Ocean convection frequency | 79% | excessive |
| Convective precipitation | 4.33 mm/day | ~3 mm/day |

**Simulation details**: These diagnostics were computed from a 1-day spinup simulation at T31 resolution with 8 vertical levels. The 79% ocean convection frequency represents the fraction of ocean gridpoints with convective precipitation > 0.1 mm/day at the end of the simulation.

**Before:** 

Screenshot 2026-02-02 at 17 53 39

**After:**

Screenshot 2026-02-02 at 17 53 53

## Root Cause

The `absorptivity_water_vapor` and `absorptivity_cloud_base` parameters were set far too low:
- `absorptivity_water_vapor`: 0.00017 (should be ~75)
- `absorptivity_cloud_base`: 0.000014 (should be ~15)

This caused the atmosphere to not absorb sunlight → net radiative cooling → atmospheric instability → excessive convection triggering.

### The mechanism

```
LOW SW ABSORPTION
       ↓
Atmosphere doesn't absorb sunlight
       ↓
Net radiative cooling (LW emission >> SW absorption)
       ↓
Atmosphere cools → becomes unstable (PT > 0)
       ↓
Over ocean: moisture available (Pq > 0) → CONVECTION
Over land: no moisture (α ≈ 0.08) → NO CONVECTION
```

## Where do these values come from?

The `OneBandShortwave` radiation scheme in SpeedyWeather is adapted from SPEEDY's original **two-band** shortwave scheme. The original SPEEDY Fortran code splits shortwave radiation into visible and near-IR bands with separate absorptivities:

```fortran
! Original SPEEDY (two-band scheme)
abswv1 = 0.022   ! visible band, per g/kg per 10^5 Pa
abswv2 = 15.0    ! near-IR band, per g/kg per 10^5 Pa
fband  = [0.95, 0.05]  ! fraction in visible, near-IR
```

When adapting to a **one-band** scheme, we computed a weighted combination:
```
0.95 × 0.022 + 0.05 × 15.0 × 0.2 = 0.171 per g/kg
```

Which we need to convert to SpeedyWeather's kg/kg convention:
- 0.171 per g/kg → 171.0 per kg/kg

### Why 75 instead of 171?

Screenshot 2026-02-02 at 16 12 10

**Note**: The original SPEEDY Fortran value (171.0) caused **145 W/m²** absorption - too high.

After fixing the unit conversion, we found that the SPEEDY value of 171.0 causes **too much** atmospheric absorption (~145 W/m² vs Earth's ~75 W/m²). This is likely because:

1. SPEEDY's original scheme was tuned for its specific model configuration
2. The one-band approximation doesn't perfectly capture the two-band behavior
3. SpeedyWeather's cloud and humidity distributions differ from SPEEDY

We performed a sensitivity analysis and found that **absorptivity_water_vapor = 75** produces atmospheric absorption close to Earth's observed ~75 W/m².

**Sensitivity analysis details**: We ran PrimitiveWetModel simulations at T31 resolution with 8 vertical levels for 5 days, varying `absorptivity_water_vapor` from 0 to 200. The atmospheric SW absorption was computed as: incoming TOA (340 W/m²) - outgoing SW - surface SW. Linear interpolation identified 75 as the optimal value to match Earth's ~75 W/m² atmospheric absorption.

(Earth's atmospheric SW absorption ~78 W/m²: Trenberth et al. 2009, Wild et al. 2015)

### Note on absorptivity_cloud_base

The `absorptivity_cloud_base` parameter was similarly converted from SPEEDY's cloud absorption coefficients. The value of 15.0 is a placeholder - a proper sensitivity analysis (similar to what was done for water vapor) is still needed to optimize this parameter.

## The Fix

```julia
# shortwave_transmissivity.jl
absorptivity_water_vapor::NF = 75      # was 0.00017 (tuned for ~75 W/m² absorption)
absorptivity_cloud_base::NF = 15.0     # was 0.000014 (needs further tuning)
```

## Results After Fix

Based on sensitivity analysis:

| Metric | Old (0.00017) | New (75) | Target |
|--------|---------------|----------|--------|
| SW absorption (W/m²) | -0.8 | ~76 | 70-80 |
| Ocean convection freq | 79% | ~44% | 30-40% |
| Convective precip (mm/day) | 4.33 | ~3.3 | ~3 |

**Note**: The original SPEEDY Fortran value (171.0) caused **145 W/m²** absorption - too high. The optimal value of **75** was determined by interpolating to match Earth's ~75 W/m² atmospheric absorption.

## Why ocean but not land?

The ocean/land asymmetry is physical: over ocean, soil moisture availability α = 1.0 allows unlimited evaporation, while over land α ≈ 0.08 severely limits evaporation. Convection requires both temperature instability (PT > 0) AND moisture (Pq > 0) - land lacks the moisture.

## Problem remaining: Cloud top too high

Entrainment profiles for cloud depth control will be addressed in a separate PR.

### DIFF
--- a/SpeedyWeather/src/physics/radiation/shortwave_transmissivity.jl
+++ b/SpeedyWeather/src/physics/radiation/shortwave_transmissivity.jl
@@ -35,27 +35,27 @@ $(TYPEDFIELDS)."""
 
     "[OPTION] Absorptivity of dry air [per 10^5 Pa]"
     # Weighted visible + near-IR: 0.95*0.033 + 0.05*0.0 = 0.03135 (SPEEDY absdry, fband weights)
-    @param absorptivity_dry_air::NF = 0.03135 (bounds = 0 .. 1,)
+    @param absorptivity_dry_air::NF = 0.03135 (bounds = Nonnegative,)
 
     "[OPTION] Constant aerosol concentration?"
     aerosols::Bool = true
 
     "[OPTION] Absorptivity of aerosols [per 10^5 Pa]"
     # Weighted visible + near-IR: 0.95*0.033 + 0.05*0.0 = 0.03135 (SPEEDY absaer, fband weights)
-    @param absorptivity_aerosol::NF = 0.03135 (bounds = 0 .. 1,)
+    @param absorptivity_aerosol::NF = 0.03135 (bounds = Nonnegative,)
 
     # Weighted visible + near-IR: 0.95*0.022 + 0.05*15.0*0.2 = 0.171 per g/kg → 1.7e-4 per kg/kg (SPEEDY abswv1, abswv2)
     # Value chosen following PR #974 for a ~75W/m^2 shortwave absorption target
     "[OPTION] Absorptivity of water vapor [per kg/kg per 10^5 Pa]"
-    @param absorptivity_water_vapor::NF = 75 (bounds = 0 .. 1,)
+    @param absorptivity_water_vapor::NF = 75 (bounds = Nonnegative,)
     
     # Weighted visible band: 0.95*0.015 = 0.014 per g/kg → 1.4e-5 per kg/kg (SPEEDY abscl1)
     "[OPTION] Base cloud absorptivity [per kg/kg per 10^5 Pa]"
-    @param absorptivity_cloud_base::NF = 15 (bounds = 0 .. 1,)
+    @param absorptivity_cloud_base::NF = 15 (bounds = Nonnegative,)
 
     "[OPTION] Maximum cloud absorptivity [per 10^5 Pa]"
     # Weighted one-band scaling: 0.95*0.15 = 0.1425 → rounded to 0.14 (SPEEDY abscl2)
-    @param absorptivity_cloud_limit::NF = 0.14 (bounds = 0 .. 1,)
+    @param absorptivity_cloud_limit::NF = 0.14 (bounds = Nonnegative,)
 end
 
 Adapt.@adapt_structure BackgroundShortwaveTransmissivity


milankl:
Skipping submodule CI and enzyme CI as we're only changing parameters. Please do this do save CI time where not needed!

SpeedyWeather/src/physics/radiation/shortwave_transmissivity.jl
```diff
@@ -46,10 +46,10 @@ $(TYPEDFIELDS)."""
 
     "[OPTION] Absorptivity of water vapor [per kg/kg per 10^5 Pa]"
     # Weighted visible + near-IR: 0.95*0.022 + 0.05*15.0*0.2 = 0.171 per g/kg → 1.7e-4 per kg/kg (SPEEDY abswv1, abswv2)
```
milankl, line 47: I think the docstring need to be directly above the field, so move the `# ...` comment up! Otherwise the docstring will probably not render correctly or the wrong line

milankl: Don't worry just did that in [e764a1e](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/974/commits/e764a1e7efff30c4cf65456227401fd6dd4e67c2)

milankl:
Would be good to check the effect on atmospheric temperatures. E.g. layer 8 air temperature before and after similar to above. I just want to make sure this warms the atmosphere accordingly

nviebig:
Global means were computed by averaging the temp_grid field across all horizontal gridpoints at the final timestep for each model level. Each run used the same T31, 8‑level configuration and was integrated for 10 days:

## Temperature Profile: Old vs New

| Level | σ    | T_old (K) | T_new (K) | ΔT (K) |
|------:|:----:|----------:|----------:|-------:|
| 1 | 0.06 | 202.31 | 202.88 | +0.57 |
| 2 | 0.19 | 219.03 | 221.63 | +2.60 |
| 3 | 0.31 | 234.35 | 238.89 | +4.53 |
| 4 | 0.44 | 247.53 | 251.12 | +3.60 |
| 5 | 0.56 | 255.83 | 258.97 | +3.13 |
| 6 | 0.69 | 264.02 | 266.78 | +2.76 |
| 7 | 0.81 | 271.97 | 274.57 | +2.60 |
| 8 | 0.94 | 280.23 | 282.49 | +2.27 |

**Near-surface layer (Level 8):**  
T_old = 280.23 K, T_new = 282.49 K → ΔT = **+2.27 K**

milankl:
Just compared lowermost layer air temperature (top before, bottom after) which are now 6K warmer, that's great. Sure the air over ocean is still somewhat too cold and surface air temperatures (typically 3K warmer than lowermost layer) reach 35˚C so slightly too cold but that's another tuning question

image

Ran this on `#main` with

```julia
julia> using SpeedyWeather
julia> spectral_grid = SpectralGrid()
julia> B_old = BackgroundShortwaveTransmissivity(spectral_grid)
julia> B_new = BackgroundShortwaveTransmissivity(spectral_grid,  absorptivity_water_vapor=75, absorptivity_cloud_base=15)
julia> model = PrimitiveWetModel(spectral_grid, shortwave_radiation = OneBandShortwave(spectral_grid, transmissivity=B_old))
```

and `transmissivity=B_new` afterwards

milankl (approved):

milankl:
@hannahw0od @nviebig happy for you to merge this

hannahw0od:
A comparison with ERA5 convective rain rate for reference

- I added 2 different colormaps (one where rainfall is easier to see in this dataset, the other matching the colormap @nviebig used)
- In the top colormap white indicates undefined areas log(0)

Screenshot (48)


Screenshot (49)

PR #975, opened 2026-02-03
[LowerTriangularArrays] broadcast gpu similar fix?

maximilian-gelbrecht:
I noticed the custom broadcasting uses a constructor with a non-concrete type. We had some issues with GPU broadcasts. Let's see if this changes something.


maximilian-gelbrecht:
Superseded by #977

PR #977, opened 2026-02-03
Simplify broadcasting of RingGrids and LTA

maximilian-gelbrecht:
Not yet finished, but hopefully with a bit simplified logic, it also works with Reactant. 

We also hat some other issues with broadcasting, let's see if it also solves some of them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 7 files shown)


milankl (approved):
Nice yeah, I don't even remember why we had all of these. Probably because we didn't know better?

PR #978, opened 2026-02-04
Shortwave Absorption Bug Fix & Stability Investigation

nviebig:
# Shortwave Absorption Bug Fix & Stability Investigation

## Summary

This PR fixes a cloud reflection bug in the upward shortwave beam that was incorrectly reducing atmospheric absorption at cloud top.

### What Was Wrong

In the upward shortwave loop, cloud reflection was being added to the outgoing radiation **before** computing the atmospheric absorption tendency:

```julia
# BUGGY (before):
U_out = U * t[ij, k]
U_out += k == cloud_top ? U_reflected : zero(U)  # Add reflection first
dTdt[ij, k] += flux_to_tendency((U - U_out) / cₚ, ...)  # Then compute absorption
```

This meant the absorption calculation was:
$$\text{absorption} = (U - (U \cdot t + U_{\text{reflected}})) / c_p$$

The cloud reflection was **subtracted from the atmosphere's absorption**, creating an energy budget error.

### The Fix

Swap the order so absorption is computed first, then the reflection is carried upward:

```julia
# FIXED (after):
U_out = U * t[ij, k]
dTdt[ij, k] += flux_to_tendency((U - U_out) / cₚ, ...)  # Compute absorption first
U_out += k == cloud_top ? U_reflected : zero(U)  # Then add reflection
```

Now the absorption is correct:
$$\text{absorption} = (U - U \cdot t) / c_p$$

## Investigation: Parameter Recalibration Required

PR #974 originally tuned `absorptivity_water_vapor=75` and `absorptivity_cloud_base=15` using 5-day tests against the buggy physics.  (75, 15) crashes at day 139 with fixed physics

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## Unreleased
+
+- Bug fix: Cloud reflectivity on half level above not below [#978](https://github.dev/SpeedyWeather/SpeedyWeather.jl/pull/978)
 - Fix absorptivity_water_vapor and absorptivity_cloud_base default values [#974](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/974)
 - Particle advection GPU ready [#897](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/897)
 - Enzyme compat relaxed as Enzyme bug was fixed and uv_from_vordiv! kernel version used on CPU as well [#971](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/971)


milankl:
This should go in before #979 to release v0.18

milankl:
As we don't currently do long time integration tests, @nviebig did you test that this solves your problem locally?

milankl:
If yes, can you merge?

nviebig:
So the fix is actually a seperate problem from the instability, i just thought maybe i adress both in the same PR.

nviebig:
But im still looking as to why OneBandShortWave triggers this instability (running with TransparentShortwave is stable)

PR #981, opened 2026-02-05
Add soil type-dependent field capacity

### LINKED ISSUE #980: Add diverse soil types
After chatting with @nviebig, it seems like the current global soil field capacity might be too low, and doesn't represent the diverse range of soil types in the real world. Adding representation of more soil types and their respective field capacities might be part of a solution to the convection issues #978 and #974 over land, where soil moisture availability seems too low.

IFS uses 7 soil textural types and kindly provides a nice map through the CDS interface. Initially we could just use this to dynamically set soil types on `initialisation` but also opens the door for inclusion of more complex processes down the line.

Texture types, $\theta_{cap}$ is the field capacity (ours is currently set globally to 0.3):

Image

Thoughts @milankl ?

gregrmunday:
Add a soil type map used by IFS to determine soil-specific field capacities, instead of a global constant.

Closes #980.

Seems the result of adding this is: more rain! I think this is what we expect since the soil-specific field capacities are generally significantly higher than the current global constant, hence higher soil moisture availability &rarr; more evapotranspiration &rarr; more moisture in the atmosphere.    

**Soil moisture availability**

Current
Screenshot 2026-02-05 at 17 03 24

New
Screenshot 2026-02-05 at 17 03 19

**Convective rain**

Current
Screenshot 2026-02-05 at 17 03 54

New 
Screenshot 2026-02-05 at 17 03 50

**Large-scale rain**

Current
Screenshot 2026-02-05 at 17 02 39

New
Screenshot 2026-02-05 at 17 02 33


gregrmunday:
I may not have implemented this in the perfect way 😁

gregrmunday:
Also the `soil_type.nc` is `379KB` - I haven't added this to the PR yet.

milankl:
So my question is, can you achieve the same by simply increasing field capacity from 0.3 ... to 0.5 or similar? Also why does it have so little impact regarding precipitation over the ocean?

milankl:
This PR depends on #983. We should first find a place for another boundary condition file before considering adding a new one. I like the idea to make the land model more realistic but there's also bigger fish to fry like a realistic surface albedo given vegetation, soil moisture, orography, sea ice concentration ...

hannahw0od:
this would be super useful for me! Let me know if there's anything I can do to help :)

gregrmunday:
Here's the soil type grid for reference


Screenshot 2026-03-23 at 11 37 02

gregrmunday:
After the merge with main and #983, this seems to work nicely! Ready for review @milankl ? The `soil_type.nc` already sits in SpeedyWeatherAssets :)

milankl:
I'm reluctant to merge this. The reason is that while it has an impact on precipitation, it doesn't actually address the main problem that our convection scheme needs an overhaul and shouldn't trigger all the time. If we fix that the impact that the soil type introduced here may be different. So essentially this introduces an increasing complexity without a strong justification for it. Furthermore it hardcodes two global fields that are used as boundary condition without a plan how to adapt those to changing climate or other planets.

milankl:
Closing as considered too complex for SpeedyWeather's land model. Similar efforts should go into Terrarium instead

Issue #982, opened 2026-02-05
Enzyme CI failing more often than before


maximilian-gelbrecht:
Since merging #971 (*) the Enzyme workflow that checks differentiability of a single timestep of the `PrimitiveWetModel` fails more regularly than before. That's not great. It's still kind of random and only fails some times, but this makes it even harder to test what the issue really. Individual functions are also fine and it only seems to appear in complex models. 

Examples of failing runs and there stack trace are here: 
https://github.com/SpeedyWeather/SpeedyWeather.jl/actions/runs/21724327082/job/62662246268
https://github.com/SpeedyWeather/SpeedyWeather.jl/actions/runs/21716502143/job/62633847833

(*) #971 updated the Enzyme version and introduced one more function as a kernel version on CPU that was previously problematic for Enzyme.

maximilian-gelbrecht:
@vchuravy Do you have maybe any advice how we can best hunt down the cause of these issues? Especially with those that are only happening sometimes it's unfortunately really difficult.

milankl:
@vchuravy it's typically a ` GC error (probable corruption)` error ... any advice appreciated!

maximilian-gelbrecht:
My impression is that it now fails really almost every time, this is even worse than a few weeks ago, but I am not sure when it changed. My small attempt to slightly change the script in #995 were not successful

milankl-claude:
## MWE found: `Base.copy!(::PrognosticVariables, ...)` via dispatch

I've bisected the issue to a minimal reproducer. The error is **not** from the GC at runtime — it is an Enzyme compiler error (`nodecayed_phis!`) triggered when Enzyme differentiates `Base.copy!(::PrognosticVariables, ::PrognosticVariables)` via dynamic dispatch.

### Reproducer

With `trunc=5, nlayers=1, PrimitiveWetModel` on Julia 1.10 + Enzyme:

```julia
using SpeedyWeather, Enzyme, Dates

spectral_grid = SpectralGrid(trunc = 5, nlayers = 1)
model = PrimitiveWetModel(; spectral_grid)
simulation = initialize!(model)
initialize!(simulation)
run!(simulation, period = Hour(6))

progn = simulation.prognostic_variables

# Test A: call copy! via dispatch — FAILS
function just_copy!(pn, po)
    copy!(pn, po)   # dispatches to SpeedyWeather's Base.copy!
    return nothing
end

dpn = one(progn)
dpo = zero(progn)
[...]

PR #983, opened 2026-02-05
Get SpeedyWeatherAssets on-the-fly

gregrmunday:
This PR removes the `input_data` folder from the SpeedyWeather repository and instead downloads (or finds) the required assets when a particular parameterisation is used.

Seems to work nicely, though I couldn't get around adding another couple of dependencies, sorry @milankl !

### DIFF
--- a/docs/src/input_data.md
+++ b/docs/src/input_data.md
@@ -28,7 +28,7 @@ The interface to load data from
 is via the `get_asset` function
 
 ```@docs; canonical=false
-get_asset
+SpeedyWeather.get_asset
 ```
 
 with `from_assets = true` will look into that repository to find the


maximilian-gelbrecht:
Cool! That's something we wanted to do since a while. 

I never really worked with the artifact system from Julia before, it does download them also for later use, right? Because  we don't SpeedyWeather to always require an internet connection ;)

gregrmunday:
> Cool! That's something we wanted to do since a while. 
> 
> 
> 
> I never really worked with the artifact system from Julia before, it does download them also for later use, right? Because  we don't SpeedyWeather to always require an internet connection ;) 

Yes it does! It creates an Artifacts.toml file locally and adds little addresses to where it downloads the files to, which are all stored under `.julia/artifacts`, so I guess you need internet to run for the first time but not after that :)

SpeedyWeather/src/dynamics/orography.jl
```diff
@@ -189,12 +186,7 @@ function initialize!(
     (; orography, surface_geopotential, scale) = orog
     (; gravity) = P
 
-    # LOAD NETCDF FILE
-    if orog.path == "SpeedyWeather.jl/input_data"
-        path = joinpath(@__DIR__, "../../input_data", orog.file)
-    else
-        path = joinpath(orog.path, orog.file)
-    end
+    path = get_speedy_asset("data", orog.file)
```
milankl: I'd just call it `get_asset` no need to have the Speedy branding everywhere 😉

SpeedyWeather/src/dynamics/orography.jl
milankl: I"m wondering whether we should move this to `get_asset` see suggested function signature below

SpeedyWeather/src/input/get_asset.jl
```diff
@@ -0,0 +1,29 @@
+"""
+    get_speedy_asset(subfolder, filename)
+
+Downloads a file from the SpeedyWeatherAssets repo, adds it to 
+Artifacts.toml in the project root, and returns the file path.
+"""
+function get_speedy_asset(subfolder::String, filename::String)
```
milankl: Suggest to do

```julia
function get_asset(path::String...; array_type::AbstractArray = Array)
```

as a function signature. With the splatting you don't hardcode how many folders we can have

```julia
julia> g(x::String...) = joinpath(x...)
g (generic function with 1 method)

julia> g("weights", "land", "surface_roughness.nc")
"weights/land/surface_roughness.nc"
```

adding an `array_type` as keyword argument here but I think in general we want to support a range of input data formats (not necessarily netcdf!) and how what they are being wrapped into, so actually we should just call

```julia
function get_asset(path::String...; type=Array, format=NCDataset)
    return _get_asset(type, format, joinpath(path...))
end
```

and then have several methods implemented to dispatch over `type`s and `format`s (only needed to implement fields and netcdf for now) which will make less breaking changes later!

SpeedyWeather/src/input/get_asset.jl
```diff
@@ -0,0 +1,29 @@
...
+    # SpeedyAssets base url
+    repo_base = "https://github.com/SpeedyWeather/SpeedyWeatherAssets/raw/refs/heads/main"
+    url = "$repo_base/$subfolder/$filename"
```
milankl: make `const assets_url = "https://github.com/SpeedyWeather/SpeedyWeatherAssets/raw/refs/heads/main"` and define outside of function?

SpeedyWeather/src/input/get_asset.jl
```diff
@@ -0,0 +1,29 @@
...
+
+    project_root = pkgdir(SpeedyWeather)
+    artifact_toml = joinpath(project_root, "Artifacts.toml")
```
milankl, line 65: So this serves as a registry for what has been downloaded to not download again? I like! Maybe save it in the SpeedyWeather folder though ... ?

SpeedyWeather/src/input/get_asset.jl
```diff
@@ -0,0 +1,29 @@
...
+
+    # Check if the artifact already exists in the TOML
+    hash = Artifacts.artifact_hash(filename, artifact_toml)
+
+    # If not found or file not installed create and bind it
+    if hash === nothing || !Artifacts.artifact_exists(hash)
+        hash = PkgA.create_artifact() do artifact_dir
```
milankl: As far as I understand from https://docs.julialang.org/en/v1/stdlib/Artifacts/#Artifacts `Pkg.Artifacts` can be used for older Julia versions but has technically been deprecated as moved to Julia Base, can we just use that?

SpeedyWeather/src/dynamics/orography.jl
```diff
@@ -125,9 +125,6 @@ $(TYPEDFIELDS)"""
 @kwdef struct EarthOrography{NF, GridVariable2D, SpectralVariable2D} <: AbstractOrography
 
     # OPTIONS
-    "path to the folder containing the orography file, pkg path default"
-    path::String = "SpeedyWeather.jl/input_data"
```
milankl, line 129: I'd probably keep this with default to `SpeedyWeatherAssets` because otherwise you don't allow for a user to load in a local file? We'd just need a `if path == "SpeedyWeatherAssets"` to call your new `get_asset` otherwise look locally?

SpeedyWeather/Project.toml
```diff
@@ -6,6 +6,7 @@ version = "0.18"
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
+Artifacts = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
```
milankl, line 9: This is part of Julia Base so no problem adding it as another dependency. I think you need to add a compat entry though.

SpeedyWeather/src/dynamics/orography.jl
gregrmunday: Haha that's fair, Speedy is just very fun to prepend to stuff

SpeedyWeather/src/input/get_asset.jl
gregrmunday: I think that's where it currently saves, inside `SpeedyWeather.jl/SpeedyWeather`

gregrmunday: But yes, it's actually p cool!

SpeedyWeather/src/input/get_asset.jl
gregrmunday: Yeah this I got a little confused about as well. On the Julia docs it does say it's been moved but then the `create_artifact()` and `download()` methods weren't available in the Base version and were in the `Pkg.Artifacts` version. Not sure what's going on there or if I was doing something wrong or not...

SpeedyWeather/src/dynamics/orography.jl
gregrmunday: Yeah that's a good point, I'll add this logic back in

SpeedyWeather/Project.toml
gregrmunday: Ah yes I think I do!

SpeedyWeather/src/dynamics/orography.jl
gregrmunday: I think your comment cut off the line above, so assume you're talking about loading the file inside `get_asset` as well, depending on its input type?

SpeedyWeather/src/input/get_asset.jl
gregrmunday: I can see a potential issue with this approach where the required types for each input format will also require new dependencies in main SpeedyWeather, maybe, if I understand what you're saying correctly.

So for `NCDataset` this is fine but for loading the NN weights in, we'll want an NPZ type I guess which means we need NPZ dep in main. However if we leave it type agnostic and just return a `path` it leaves it to the particular parameterisation to handle it, even if it's in an extension.

I might be missing something though!

SpeedyWeather/src/dynamics/orography.jl
milankl: I think `get_asset` should return data wrapped into an array type as defined, so

```julia
orography = get_asset("boundaries", "orography.nc", type=FullGaussianGrid)
```

Would

- identify the file by path
- load it in given a (here default) format `NCDataset`
- determine resolution from array size
- wrap it into `type` as provided ready to be used

I would probably return CPU arrays for now as the data is loaded by the CPU anyway so the GPU conversion would explicitly follow in the line below

SpeedyWeather/src/input/get_asset.jl
milankl: No you are right, but if you want to load data from `NPZ` you'd clearly need to do `import NPZ` to do

```julia
import NPZ
get_asset("folder", "filename.npz", format=NPZ)   # otherwise NPZ here not available
```

so `_get_asset` for `NPZ` could be extended in the NPZExt!!

SpeedyWeather/src/input/get_asset.jl
milankl: Okay we might ask around to see how to resolve that ...

gregrmunday:
Here's what I've been trying out:

The constructor checks that if the user provided the file path, it checks if the filepath exists, warn if it doesn't, will try to find it in the SWA repo instead
```julia
function EarthOrography(spectral_grid::SpectralGrid; kwargs...)
    (; architecture, NF, GridVariable2D, SpectralVariable2D, grid, spectrum) = spectral_grid
    orography = on_architecture(architecture, zeros(GridVariable2D, grid))
    surface_geopotential = on_architecture(architecture, zeros(SpectralVariable2D, spectrum))

    if haskey(kwargs, :path)
        if !isfile(kwargs[:path])
            @warn "Could not find the asset path provided locally, will attempt to load asset from SpeedyWeatherAssets"
        end
    end

    return EarthOrography{NF, GridVariable2D, SpectralVariable2D}(;
        orography, surface_geopotential, kwargs...
    )
end
```

then at the top of get_asset  we check again if the path exists locally, and if it doesn't try to load from SWA. The default path is the SWA path, and this check is ok since the default path will never exist locally

```julia
function get_asset(path::String; name::String = "", type = NCDataset, format = NCDataset)
    if isfile(path) # check if path is local (custom input)
        try
            asset = _get_asset(path, name, type, format)
            return asset
        catch e
            @warn "Custom asset loading failed with: $e"
            @warn "Attempting to load default asset"
        end
    end
    
    .......
    
end
```  

This approach would require adding the little file path check in each constructor

gregrmunday:
As you suggested @milankl we try a boolean flag approach to all this. I suppose the reason why I hadn't done this was to try to do as much as possible automatically - so if a user supplies the path it will either automatically find it locally, online, or throw an error if it doesn't exists in either.

I'm imagining that we could have several types of each asset file at some point in the future, so the user being able to specify both the SpeedyWeatherAssets path OR a local path could be useful. Just a matter of how you construct the logic I suppose

gregrmunday:
OK I've propagated the logic through all the `structs` and `initialise!` methods for each of the assets we have currently, and updated the `get_asset` method with more comprehensive error handling :)

SpeedyWeather/src/input/get_asset.jl
```diff
@@ -0,0 +1,96 @@
...
+    else
+        return name
+    end
+end
+
+"""
+Downloads a file from the SpeedyWeatherAssets repo, adds it to 
+Artifacts.toml in the project root, and returns the file path.
+"""
+function get_asset(path::String; from_assets::Bool = true, name::String = "", type = NCDataset, format = NCDataset)
```
milankl: add `version::String = "1.0"` here, check for `"main"` other convert to `VersionNumber` or somehow get the right github path. A component like `LearnedSurfaceRoughness` could then have a `version::String` field that is passed to get `get_asset` function

gregrmunday:
The latest on this:

`get_assets.jl` is designed to take in a version number when it's called, as passed by the user during a boundary condition's construction.

```julia
const ASSETS_URL = "https://github.com/SpeedyWeather/SpeedyWeatherAssets/raw/refs"
const DEFAULT_ASSETS_VERSION = v"1"
const ASSETS_VERSIONS_AVAILABLE = (v"1")

"""
Downloads a file from the SpeedyWeatherAssets repo, adds it to 
Artifacts.toml in the project root, and returns the file path.
"""
function get_asset(
        path::String;
        from_assets::Bool = true,
        name::String = "",
        type::Any = NCDataset,
        format::Any = NCDataset,
        version::VersionNumber = DEFAULT_ASSETS_VERSION
    )

    if !from_assets
        if isfile(path) # check if path is local (custom input)
            try
                return _get_asset(path, name, type, format)
            catch e
                throw("Local asset loading failed with: $e")
            end
        else
            throw("Local $name file doesn't exist on this path")
        end
    end

    if !isempty(version.build)
        branch = version.build[1]
        target_url = joinpath(ASSETS_URL, "heads/$branch")
    else
        if version in ASSETS_VERSIONS_AVAILABLE
            target_url = joinpath(ASSETS_URL, "tags/$version")
        else
            throw("SpeedyWeatherAssets version $version not available. Please select from $ASSETS_VERSIONS_AVAILABLE")
        end
    end
...
```

Then using `soil_moisture.jl` as an example:

```julia
@kwdef struct SeasonalSoilMoisture{NF, GridVariable4D} <: AbstractSoilMoisture
    # READ CLIMATOLOGY FROM FILE
    "[OPTION] filename of soil moisture"
    file::String = "soil_moisture.nc"

    "[OPTION] path to the folder containing the soil moisture"
    path::String = joinpath("data", file)

    "[OPTION] flag to check for soil moisture in SWA or locally"
    from_assets::Bool = true

    "[OPTION] SpeedyWeatherAssets version number"
    version::Union{VersionNumber, Nothing} = nothing
...


function initialize!(soil::SeasonalSoilMoisture, model::PrimitiveEquation)
    (; monthly_soil_moisture) = soil

    version_kwarg = soil.version === nothing ? (;) : (; version = soil.version)

    # LOAD NETCDF FILE
    ncfile = get_asset(
        soil.path;
        from_assets = soil.from_assets,
        name = "sm",
        type = NCDataset,
        format = NCDataset,
        version_kwarg...
    )
...
```

So here if the user doesn't pass in anything, the kwarg isn't passed and the default defined in `get_assets.jl` sticks. If they do pass, then it goes through the checks in `get_assets` as shown above. 

What are thoughts @milankl ?

milankl:
> ```
> "[OPTION] SpeedyWeatherAssets version number"
> version::Union{VersionNumber, Nothing} = nothing
> ```

just do `version::VersionNumber = DEFAULT_ASSET_VERSION` ?

milankl:
Otherwise awesome!!! Sounds like we basically solved this problem 🎉 Thanks for putting it into code!

milankl:
> @milankl could be a silly question, is DEFAULT_ASSET_VERSION completely global in SpeedyWeather after it's defined as a `const` in `get_asset.jl`? If so that's v useful!

A `const` is a global constant so it exists in the scope of whole `module`. You can't define a `const` in a local scope because any local scope is temporary (the loop is allowed to end, the function returns and exits ...)

```julia
julia> function local_scope()
           const X = 1
       end
ERROR: syntax: unsupported `const` declaration on local variable around REPL[23]:2
```

gregrmunday:
Alright I think ready for review @milankl ! Would be interested to see if/where it breaks for you...

milankl:
We're getting a bunch of "could not download orography.nc" error from buildkite and github actions CI

gregrmunday:
> We're getting a bunch of "could not download orography.nc" error from buildkite and github actions CI

Think I've solved the build kite one, just waiting on the AMDGPU test

gregrmunday:
All tests passing now but AMDGPU test still hanging...

gregrmunday:
In the latest commit I've added `surface_roughness` as a parameterisation. This makes sense to me and I think it makes it really nice and flexible to use within SpeedyWeather. 

I've added the land, ocean and overall surface roughness as diagnostic variables, mainly so that I could plot them separately out the box but maybe this isn't the best way of doing it, and maybe takes up extra memory? 

Using `ConstantSurfaceRoughness`:

Screenshot 2026-02-17 at 13 54 32

maximilian-gelbrecht:
As we are already discussing open source licenses in other PR. We haven't done this so far, but from what I understand we should actually add an attribution to ECWMF to the files. ERA5 is under CC BY. https://creativecommons.org/licenses/by/4.0/

I am not an expert on this, but I would think that putting a readme with that information into the `input` folder with the download code, should count as a good enough effort? 

It should say something like: 

> The data available for download through these scripts are climatological averages and other data derived from [ECMWF's ERA5 project](https://www.ecmwf.int/en/forecasts/dataset/ecmwf-reanalysis-v5) licensed under [CC BY 4.0](https://creativecommons.org/licenses/by/4.0/)

milankl:
I hope this information is still in the meta data of the netCDF files. But definitely good to write this into a readme

gregrmunday:
> As we are already discussing open source licenses in other PR. We haven't done this so far, but from what I understand we should actually add an attribution to ECWMF to the files. ERA5 is under CC BY. https://creativecommons.org/licenses/by/4.0/
> 
> I am not an expert on this, but I would think that putting a readme with that information into the `input` folder with the download code, should count as a good enough effort?
> 
> It should say something like:
> 
> > The data available for download through these scripts are climatological averages and other data derived from [ECMWF's ERA5 project](https://www.ecmwf.int/en/forecasts/dataset/ecmwf-reanalysis-v5) licensed under [CC BY 4.0](https://creativecommons.org/licenses/by/4.0/)

Happy to do this!

SpeedyWeather/src/input/README.md
```diff
@@ -0,0 +1 @@
+The data available for download through these scripts are climatological averages and other data derived from ECMWF's ERA5 project licensed under CC BY 4.0
```
maximilian-gelbrecht: Could you please add the links to ERA5 and CC BY 4.0 like in my quote? The license literally says we must provide links https://creativecommons.org/licenses/by/4.0/

gregrmunday: Oh shoot yes, I thought I did...

gregrmunday: They're there now :)

milankl:
Quick summary of the discussion today: We have decided against the intended direct use of Artifacts.jl and ArtifactUtils.jl as

- they require one to write manually (or with `add_artifact!`) and so we can't just version them all together easily within SpeedyWeatherAssets
- artifacts have to be tar balls or `.zst` and can't be `.nc`
- now we have `version = v"1"` to specify SpeedyWeatherAssets version per asset that can also be `version = v"1+main"` for experimental main branch of that repo but with a more manual Artifacts.toml one would need to distinguish them somehow by name?

milankl:
But there is the general scope to revisit these decision at some point, what I liked is that this may allow for a nice interface whereby

```julia
SpeedyWeather.add_artifact!("label", "path/to/file")
component = Component(..., path = artifact"label")
```

first line would just define a new artifact hooking it into the new system and the second would allow us to use it. However, we can't do local files though with this.

gregrmunday:
Seems stuck on the Enzyme differentiability again, I assume this isn't anything to do with this PR?

milankl:
I'm wondering whether we actually need the `const AVAILABLE_ASSETS_VERSIONS` -- we also don't for the branches. It'll just fail if you try to load a version that doesn't exist? 

Otherwise wrote some docs and if everything passes then happy for you to merge this

milankl:
> Seems stuck on the Enzyme differentiability again, I assume this isn't anything to do with this PR?

Yes, if it says `GC error (probable corruption)` (just checked it is) then it's already flagged as in issue in #982 and should be ignored here

PR #984, opened 2026-02-06
Introduce orbit and rotation time

milankl:
We want to be able to change `planet.length_of_day` and `planet.length_of_year` so that the solar zenith angle is correctly computed. This should be independent of the `planet.rotation` rate which is used for coriolis to be able to change those independently from another (which is essentially an Earth-centric solar system). Important to isolate effects from coriolis and solar heating.

TODO

- [x] introduce `rotation_time` and `orbit_time` in the `Clock` with their own `orbit_dilation` and `rotation_dilation`
- [x] adapt the `cos_zenith!` functions and calls to use those instead of `clock.time`
- [x] Initialize the `clock` called from `initialize!(simulation)` correctly
- [x] manual testing (videos!)
- [x] CI testing
- [x] documentation


milankl:
@hannahw0od to advance the clock given a $\Delta t$ I found it easier to think of a [time dilation](https://en.wikipedia.org/wiki/Time_dilation), so a factor that compresses (for <1) or stretches time (>1) for orbit/rotation time, i.e. we have now

```julia
function timestep!(clock::Clock, Δt; increase_counter::Bool = true)
    clock.time += Δt

    # step the rotation and orbit time separately for solar zenith calculations
    # using the dilation factors to speed up or slow down the rotation and orbit time relative to the model time
    clock.rotation_time += Millisecond(round(Int, Δt.value * clock.rotation_dilation))
    clock.orbit_time += Millisecond(round(Int, Δt.value * clock.orbit_dilation))
```

In contrast to time dilation in relativistic physics which seems to be additive the _difference_ in elapsed time between two clocks *dilation* here is multiplicative, so tells you how much faster one clock is running compared to another. That's better here because `Δt` is passed on (see above) as it's actually half of the normal value in the first time steps see https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/time_integration/

milankl:
To test this do

```julia
planet = Earth(spectral_grid, length_of_day=Hour(12))
model = PrimitiveWetModel(spectral_grid; planet)

add!(model, SpeedyWeather.SurfaceShortwaveDownOutput())
model.output.output_dt = Hour(1)
run!(simulation, output=true)
```

After 12h sun is slightly east of the dateline although it should be on the dateline?

image

this shift only occurs on the first time steps, 24h is same location as after 12h

milankl:
This runs the sun backwards, so `length_of_day=Hour(-24)` but somehow that also flips the hemispheres ... need to fix!

https://github.com/user-attachments/assets/21b08dbd-21d5-4180-87f8-dffd25777f5f

hannahw0od:
I was looking at the solar declination calculation and plotted some simple demos functions to see if this could be the issue

```Julia
"""
$(TYPEDSIGNATURES)
SinSolarDeclination functor, computing the solar declination angle of
angular fraction of year g [radians] using the coefficients of the
SinSolarDeclination struct."""
function (S::SinSolarDeclination)(g::NF) where {NF}
    planet = S.planet
    axial_tilt = deg2rad(planet.axial_tilt)
    equinox = planet.length_of_day.value * Dates.dayofyear(planet.equinox) / planet.length_of_year.value
    return axial_tilt * sin(g - 2 * (π * convert(NF, equinox)))
end
```

The red line shows a positive g (angle fraction) and equinox
The blue line shows a negative g and equinox

Screenshot 2026-02-06 at 18 39 39

The blue line is flipped like the video shown above. My guess is that in the simulation, g and equinox become negative which is why the sun 'flips'. This may be because ```length_of_day ``` is negative and is used in both g and equinox calculations.

milankl:
What about

```julia
equinox = abs(Second(planet.length_of_day).value) * Dates.dayofyear(planet.equinox) / abs(Second(planet.length_of_year).value)
```

I reckon the equinox here converted from a date to a fraction 0..1 should always be nonnegative.

hannahw0od:
https://github.com/user-attachments/assets/e95d3375-a9b5-4a51-b755-6a5b19b50c99

The Sun now flips hemispheres when I change the code to use absolute values as suggested above. Need to think about why

hannahw0od:
https://github.com/user-attachments/assets/29aee1f4-bb6a-4f44-bd1c-33b37ab61071

I tried running the same code again today and it seems to work. I also checked the specific dates and times for which the Sun was flipping hemispheres from above, and this does not happen when I run the code today. I haven't changed anything from last time, but perhaps before I hadn't saved all the files and reloaded the kernel so there was a bug from that?

milankl:
@hannahw0od I think there might be still issues with `Dates.dayofyear` which may not work as expected? Might be good to check, I added some comments here when I wrote this code regarding generalizing to different length of day/year

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/eeb5784dd59e2f32069406ed13a6b001ecd87737/SpeedyWeather/src/physics/radiation/zenith.jl#L20

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/eeb5784dd59e2f32069406ed13a6b001ecd87737/SpeedyWeather/src/physics/radiation/zenith.jl#L181

milankl:
Okay I think I got it, there was a lot of things wrong because in the `year_angle` and `solar_hour_angle` computations length of day and length of year had to be held constant to use the Dates functionality correctly (which is relative to Earth's calendar, so it indeed treats these as constants). We now entirely rely on `time` being either orbit or rotation time. 

E.g. this is a simulation with days being -10days long, so the planet rotates the other way around wrt to the sun (coriolis unchanged) and 10x slower, and the years being also 10 days long, so much shorter than normal. Given that both periods equal, we have the Bering Strait (and its antipole) being forever winter, the sun is always at the Horizon. Quite interesting

```julia
planet = Earth(spectral_grid, length_of_day=-Day(10), length_of_year=Day(10))
```

https://github.com/user-attachments/assets/3e73bad6-93da-41d4-ac8f-5434fea0cf11

SpeedyWeather/src/time_stepping/clock.jl
```diff
@@ -46,19 +58,34 @@ on_architecture(::AbstractArchitecture, clock::Clock) = clock
...
     clock.time = convert(DT, clock_old.time)
     clock.start = convert(DT, clock_old.start)
+    clock.rotation_time = convert(DT, clock_old.rotation_time)
+    clock.orbit_time = convert(DT, clock_old.orbit_time)
     clock.period = convert(S, clock_old.period)
+    clock.rotation_dilation = convert(NF, clock_old.rotation_dilation)
+    clock.orbit_dilation = convert(NF, clock_old.orbit_dilation)
     clock.step_counter = convert(I, clock_old.step_counter)
     clock.time_step_counter = convert(I, clock_old.time_step_counter)
     clock.n_steps = convert(I, clock_old.n_steps)
```
milankl, line 92: Reactant errors here with `Int64(::TracedRNumber{Int64})`

milankl:
@maximilian-gelbrecht Why is this?

```julia
 ERROR: LoadError: MethodError: no method matching Int64(::Reactant.TracedRNumber{Int64})
The type `Int64` exists, but no method is defined for this combination of argument types when trying to construct it.
```

This PR extends the Clock copy, see above

maximilian-gelbrecht:
In the Reactant extension we define 

```
SpeedyWeather.Clock(architecture::ReactantDevice) = Reactant.to_rarray(SpeedyWeather.Clock(), track_numbers = true)
```

Key there is the `track_numbers`: Without it, we'd get wrong results in the Reactant runs for the time. But doing it like this automatically converts all numbers to tracked numbers (including the date times) 

So there it tries to copy the clock from the Reactant to non-Reactant simulation in the correctness test script, and it can't do that anymore because there's no such conversion defined for traced Int to normal Int. I defined those type of conversions for the date time types, but maybe those are a bit against Reactant's spirit, because they basically erase information, so it might be the safe thing to not define those. 

For your change here, we might just have to adjust the `copy!` maybe, or be a bit more targeted with the construction of the `Clock` with a `ReactantDevice`

milankl:
Sorry I haven't understood how introducing `orbit_time` and `rotation_time` is different from the `time` that already exists and is traced. The other fields added are `::NF` and do not seem to cause the problem but one of the integer fields, that I haven't touched here.

maximilian-gelbrecht:
Ah sorry, the error doesn't come from the `copy!` but the stack trace points to `dilate` and this also makes sense because `dilate` tries to round a `TracedNumber{Int}` to a `Int` on Reactant now. 

To resolve this you could try to just define in the ReactantExtension 

```
SpeedyWeather.dilate(Δt::ReactantDatesExt.ReactantMillisecond{I}, dilation) = 
= ReactantDatesExt.ReactantMillisecond(round(I, ReactantDatesExt.ReactantMillisecond(Δt).value * dilation))
```

Replace the first `ReactantDatesExt.ReactantMillisecond` in the function signature with the type the actual input has (maybe it's also milliseconds, I don't know.

maximilian-gelbrecht:
@milankl I committed this already and it should be working now.

milankl:
AMDGPU tests are hanging/waiting for a few hours now, given that CUDA is passing I'll merge as I don't see anything here that should make a difference between them

milankl:
Weird, reactant correctness here passed but on main it failed after merge

https://github.com/SpeedyWeather/SpeedyWeather.jl/actions/runs/32158232042/job/95780523695#step:6:23

maximilian-gelbrecht:
I think that might be unrelated

maximilian-gelbrecht:
Already noted here github.com/EnzymeAD/Reactant.jl/issues/3196

PR #986, opened 2026-02-12
Fix ozone absorption: deposit energy into stratospheric layers, not TOA

nviebig:
# Bug Fix: Ozone SW Absorption Deposited into Stratospheric Layers

## Summary

This PR fixes a shortwave (SW) radiation bug where ozone absorption was removed from the downward beam before entering the layer loop, meaning no stratospheric layer ever received the absorbed energy. The energy was lost from the system.

The fix moves ozone absorption inside the per-layer downward sweep and expresses it as a fraction of D_toa, restoring shortwave energy conservation in the stratosphere.

---

## The Bug

Previously, ozone absorption was subtracted from the downward SW beam before entering the layer loop, using parameters in W/m²:

```julia
# OLD — energy silently removed from system, no layer heated
D = max(0, D_toa - ozone_absorp_upper - ozone_absorp_lower)
```

Layers 1 and 2 (upper and lower stratosphere) received no SW heating from ozone.

---

## The Fix

Ozone absorption is now computed as a fraction of D_toa and applied **inside** the downward sweep at layers 1 and 2, so the absorbed flux is captured by the layer tendency:

```julia
# NEW — ozone_absorption_upper/lower are fractions of D_toa (0..1)
for k in 1:nlayers
    if k == 1
        ozone_absorption = ozone_absorption_upper * D_toa
    elseif k == 2
        ozone_absorption = ozone_absorption_lower * D_toa
    else
        ozone_absorption = zero(D)
    end

    D_out = (D - ozone_absorption) * t[ij, k]
    # D - D_out now includes both ozone and transmissivity absorption → heats layer k
    dTdt[ij, k] += flux_to_tendency((D - D_out) / cₚ, pₛ, k, model)
    D = D_out
end
```

Default parameter values: `ozone_absorption_upper = 0.03`, `ozone_absorption_lower = 0.024` (3% and 2.4% of incoming solar).

---

## Physics Validation

A 3-year `PrimitiveWetModel` run (T31, 8 layers) was used to validate.

```julia
# ── Run configuration — edit these lines to change the run ─────────────────
N_YEARS           =  3   # total simulation length [years]
SNAP_EVERY_MONTHS = 12   # spatial snapshot interval [months]  (e.g. 6 = twice/year, 12 = yearly)
TIME_STEP         = 40   # output_dt [minutes]

# ── Ozone configuration (fractions of D_toa) ────────────────────────────────
[...]

PR #987, opened 2026-02-13
SpeedLimitDrag to counteract winds >100m/s

hannahw0od:
I tried to add a drag term to the top atmospheric layer but it is not working right now (sad face)

Milan: For winds >100m/s or so CFL violation instabilities occur in the jetstreams (fastest wind speeds) often looking like this. Sometimes they go away like here but they can cause the model to crash too. Not good.

https://github.com/user-attachments/assets/b9f7e6e5-57b3-4f47-ad4e-92dcce681c0c

This is related to #986  because radiation affects the tropopause/stratospheric temperatures and jet streams are essentially thermal winds proportional to temperature gradients. So lowering the tropopause is essential to slow these winds down but we also implement a SpeedLimitDrag that's a quadratic drag that only kicks in for higher wind speeds. Proposed formulation is to add a drag term $D$ of the form

```math
D = - c\max(0, \|\mathbf{u}\| - U)^2 * \text{sign}~(\mathbf{u})
```

so not like a classic quadratic drag $D = - c \|\mathbf{u}\| \mathbf{u}$ scaling quadratically with the speed but in the opposite direction of the flow. We include a speed limit $U$ here such that with the max-formulation the drag only applies to wind speeds above $U$ where they scale quadratically to kick in smoothly but clearly prevent atmospheric velocities from going much higher than $U$. Depending on $c$ the speed limit really has to be thought of like in the real world, in many situations would try to go higher but the penalty only really kicks after a tolerance zone above $U$. In first experiments setting the speed limit to 80m/s would still allow maximum jet stream speeds above 100m/s but not like 120m/s.

There is a discussion around resolution, with

```math
CFL = U \frac{\Delta t}{\Delta x}
```

a higher resolution (lower $\Delta x$) will be chosen with a shorter time step $\Delta t$ (done automatically). So this characteristic velocity scale $U$ the numerics can tolerate actually remains constant across resolution. With speeds reaching 100m/s that's 360km/h we could simulate a CAT5 hurricane so that's fine.

### DIFF
--- a/SpeedyWeather/src/dynamics/drag.jl
+++ b/SpeedyWeather/src/dynamics/drag.jl
@@ -175,7 +175,7 @@ function drag!(
 end
 
 @kernel inbounds = true function speed_limit_drag_kernel!(
-        Fu, Fv, u, v, @Const(c), @Const(speed_limit)
+        Fu, Fv, u, v, c, speed_limit
     )
     ij, k = @index(Global, NTuple)
 


SpeedyWeather/src/dynamics/atmosphere.jl
```diff
@@ -71,6 +71,9 @@ $(TYPEDFIELDS)"""
 
     "[OPTION] Reference dry-adiabatic temperature lapse rate [K/m]"
     @param dry_lapse_rate::NF = 9.8 / 1000
+
+    "[OPTION] Layer thickness for the shallow water model [m]"
+    @param layer_thickness::NF = 8500 (bounds = Positive,)
```
milankl: Let's take this out here because we just need some strength parameter in `StratosphericDrag` but this does not need to depend on the layer thickness. That is really only done in the ShallowWaterModel

SpeedyWeather/src/dynamics/drag.jl
```diff
@@ -128,6 +128,68 @@ end
     Fv[ij] -= c * speed * v[ij]
 end
 
+## Tropospheric drag
+export TroposphericDrag
+@parameterized @kwdef mutable struct TroposphericDrag{NF} <: AbstractDrag
+    "[OPTION] drag coefficient [1]"
+    @param c_D::NF = 1.0e-8 (bounds = Nonnegative,)    # TODO is this a good default?
+    @param U::NF = 80 (bounds = Nonnegative,)
```
milankl: Adding a `""" description """` above a field will pull this description directly into the help+documentation like `?StratosphericDrag`

SpeedyWeather/src/dynamics/drag.jl
```diff
@@ -128,6 +128,68 @@ end
     Fv[ij] -= c * speed * v[ij]
 end
 
+## Tropospheric drag
+export TroposphericDrag
+@parameterized @kwdef mutable struct TroposphericDrag{NF} <: AbstractDrag
```
milankl: Rename to `StratosphericDrag` because that's where it's applied?

SpeedyWeather/src/dynamics/drag.jl
```diff
@@ -128,6 +128,68 @@ end
...
+$(TYPEDSIGNATURES)
+Tropospheric drag for the momentum equations.
+
+    Fu = -c_D/H*(|(u, v)|- U)^2*u
+
+with `c_D` the non-dimensional drag coefficient as defined in `drag::TroposphericDrag`.
+`c_D` and layer thickness `H` are precomputed in `initialize!(::TroposphericDrag, ::AbstractModel)`
+and scaled by the radius as are the momentum equations."""
+
+function drag!(
```
milankl, line 153: Don't leave a space here because then this docstring isn't associated with the `drag!` method defined below

SpeedyWeather/src/dynamics/drag.jl
```diff
@@ -128,6 +128,68 @@ end
...
+        diagn::DiagnosticVariables,
+        progn::PrognosticVariables,
+        drag::TroposphericDrag,
+        lf::Integer,
+        model::AbstractModel,
+    )
+    k = 0   # only apply to top layer
```
milankl: Julia uses 1-based indexing 😉

SpeedyWeather/src/dynamics/drag.jl
```diff
@@ -128,6 +128,68 @@ end
...
+    u = field_view(diagn.grid.u_grid, :, k)
+    v = field_view(diagn.grid.v_grid, :, k)
+
+    Fu = field_view(diagn.tendencies.u_tend_grid, :, k)
+    Fv = field_view(diagn.tendencies.v_tend_grid, :, k)
+
+    # total drag coefficient with radius scaling
+    c = drag.c_D / model.atmosphere.layer_thickness
```
milankl: Just leave out the / layer thickness here we can just choose the drag coefficient directly

SpeedyWeather/src/dynamics/drag.jl
```diff
@@ -128,6 +128,68 @@ end
...
+function drag!(
+        diagn::DiagnosticVariables,
+        progn::PrognosticVariables,
+        drag::TroposphericDrag,
+        lf::Integer,
+        model::AbstractModel,
+    )
+    k = 0   # only apply to top layer
+    u = field_view(diagn.grid.u_grid, :, k)
+    v = field_view(diagn.grid.v_grid, :, k)
```
milankl: At the moment we are viewing only one layer. I'd suggest to pass on the whole array but don't do anything for k > 2

SpeedyWeather/src/dynamics/drag.jl
```diff
@@ -128,6 +128,68 @@ end
...
+    return launch!(
+        architecture(Fu), LinearWorkOrder, size(Fu), tropospheric_drag_kernel!,
+        Fu, Fv, u, v, c, U
+    )
+end
+
+@kernel inbounds = true function tropospheric_drag_kernel!(
+        Fu, Fv, u, v, @Const(c), @Const(U)
+    )
+    ij = @index(Global, Linear)
```
milankl: If we pass on the whole array then `ij` like this would just index all elements (regardless of k) from beginning to the end, I'll change this to a cartesian index so you can have directly access to `k` as well and make choices depending on that.

SpeedyWeather/src/dynamics/drag.jl
```diff
@@ -128,6 +128,68 @@ end
...
+
+    # Calculate speed at layer k
+    speed = sqrt(u[ij]^2 + v[ij]^2)
+
+    # Apply tropospheric drag, -= as the tendencies already contain forcing
+    delta = speed - U
+    if delta > 0.0
```
milankl: This is an absolutely reasonable approach on the CPU to skip computations where a criterion isn't met to save time. On the GPU that doesn't work quite as well because in some sense you always have to work for the slowest thread anyway. So we generally try to write code that every thread does the same operation but based on different data and avoid `if` whereever possible

hannahw0od:
was wondering why you chose 4e-7 for the drag term? @milankl

milankl:
> was wondering why you chose 4e-7 for the drag term? @milankl

I started several orders of magnitude larger and the model blew up because the moment a velocity hits >80m/s it would see a massive tendency that caused it to crash on its own. Then I chose something orders of magnitude smaller and I would still these instability wiggles from the video above ☝🏼 So I chose the smallest value that would remove them. 1e-7 still had them, 4e-7 seems to allow velocities to touch 100m/s, sometimes a little above but otherwise good!

hannahw0od:
I added the speed limit drag term to the dynamics sections of Julia ```primitive_wet.jl``` and Julia ```primitive_dry.jl``` models so the default is no longer Julia ```drag::DR = nothing``` it is Julia ```drag::DR = SpeedLimitDrag(spectral_grid)```.

milankl:
Great!!! Looks good to me, if all checks pass feel free to merge 🎉

hannahw0od:
@milankl I think something with the GPU failed? Im not so sure what this test does

SpeedyWeather/src/dynamics/drag.jl
```diff
@@ -128,6 +129,65 @@ end
...
+
+    launch!(
+        architecture(Fu), RingGridWorkOrder, size(Fu), speed_limit_drag_kernel!,
+        Fu, Fv, u, v, c, speed_limit
+    )
+    return nothing
+end
+
+@kernel inbounds = true function speed_limit_drag_kernel!(
+        Fu, Fv, u, v, @Const(c), @Const(speed_limit)
```
maximilian-gelbrecht: I know we have this in a few other kernels as well but `@Const(::Integer)` does nothing, this `@Const` notation is for arrays. But there are also some problems with it for arrays at the moment. Best just remove it. Eventually we'll remove it in other kernels as well.

hannahw0od: > I know we have this in a few other kernels as well but `@Const(::Integer)` does nothing, this `@Const` notation is for arrays. But there are also some problems with it for arrays at the moment. Best just remove it. Eventually we'll remove it in other kernels as well.

what would I write instead?

maximilian-gelbrecht: Just `Fu, Fv, u, v, c, speed_limit` without the `@Const`

hannahw0od: > Just `Fu, Fv, u, v, c, speed_limit` without the `@Const`

ok, thanks!

milankl:
For some reason the AMDGPU tests are hanging (haven't even started properly), let's ignore this and merge

PR #989, opened 2026-02-15
Progress bar with additional information

milankl:
At the moment the progress bar just shows 

image

this PR changes this to also visualize current time, performance, maximum absolute zonal velocity and temperature range

image

Temperature range isn't printed for the 2D models.

I wish ProgressMeter.jl was a bit more customisable so we're committing a bunch of type-piracy in order to hook into it's printing. Any request for additional information that should be printed?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Progress bar with additional information [#989](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/989)
+- Bug fix: Ozone absorption now deposited into stratospheric layers (1 and 2) rather than removed from the beam at TOA. Previously ozone energy was silently lost from the system without contributing a temperature tendency. Fix also adds symmetric ozone attenuation on the upward SW beam for energy conservation. [#986](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/986)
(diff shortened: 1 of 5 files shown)


maximilian-gelbrecht (commented):
It's really cool to have some extra info that also allow to quickly see if something's going really wrong while the model still integrates. 

I am just not really that much a fan of hacky solutions like that. Would there be an easy way to propose changes to ProgressMeter.jl that makes it provide a clean interface for this?

milankl:
> I am just not really that much a fan of hacky solutions like that. Would there be an easy way to propose changes to ProgressMeter.jl that makes it provide a clean interface for this?

Neither am I. I will open an issue there to request modularity of the progress information. The type piracy we commit isn't thaaaat bad: We're extending `ProgressMeter.showspeed` and hack in to show a bunch of other stuff. We can't change the function signature without changes to ProgressMeter.jl so I'm using RefValue constants that are read in from SpeedyWeather's scope. So using ProgressMeter when also SpeedyWeather is loaded will only yield weird results if the `showspeed=true` kwarg is used (default `false`).

milankl:
This has been raised here: https://github.com/timholy/ProgressMeter.jl/issues/357. As we are not committing more type piracy than we already do here anyway, I don't think the "type priacy / let's fix it in ProgressMeter jl"-issue should prevent this PR to be merged

maximilian-gelbrecht (approved):
Yeah, fine with me. Will take a while until it would be part of ProgressMeter.jl anyway and then it's easy to adjust in a follow up.

PR #990, opened 2026-02-16
Fix terminology from colatitudes to latitudes

aasdelat:
Latitudes run from -pi/2 (south pole) to pi/2, and colatitudes run from 0 (north pole) to pi (south pole). https://en.wikipedia.org/wiki/Colatitude

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fix docs colatitude -> latitude terminology [#990](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/990)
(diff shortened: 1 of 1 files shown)


milankl:
Good catch. Thanks. I guess normally the spherical harmonics are defined over colatitudes then using cos in its formulation. Here we write sin, changed to latitudes but forgot to remove the `co`!

milankl:
Indeed

image


from [here](https://github.com/SpeedyWeather/SpeedyWeather.jl/commit/dd56e4342495243aafc6c3763469966a96222567#diff-34aa3c4094a73f938d1e35c156e6a34f5bf8eb571dd375bbec6543422cdff8c1)

PR #991, opened 2026-02-16
Bump to v0.18.1

milankl:
@maximilian-gelbrecht @nviebig @gregrmunday @hannahw0od and others any objections?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,11 @@
 
 ## Unreleased
 
+## v0.18.1
+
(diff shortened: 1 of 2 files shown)


nviebig:
I cant think of any! :)

hannahw0od:
No objections :)

gregrmunday:
None from me! The new input system could fit into this as well but if you want a quick release then all good :)

milankl:
> The new input system could fit into this as well but if you want a quick release then all good :)

I want to give this more time to be experimented with before releasing it

PR #993, opened 2026-02-27
Implement CO2 and greenhouse gases

milankl:
Greenhouse gas interface would be

```julia
# a constant
julia> co2 = CO2(280)
CO2{Int64} <: Main.SpeedyWeather.AbstractGreenhouseGas
└ concentration::Int64 = 280

# a function
julia> co2 = CO2(time -> 280 + year(time) - 1850)
CO2{var"#29#30"} <: Main.SpeedyWeather.AbstractGreenhouseGas
└ concentration::var"#29#30" = #29

# a predefined exponential fit to the Keeling curve
julia> co2 = ExponentialCO2(spectral_grid)
ExponentialCO2{Float32} <: Main.SpeedyWeather.AbstractGreenhouseGas
├ base_concentration::Float32 = 280.0
├ start::DateTime = 1850-01-01T00:00:00
├ a::Float32 = 3.85
└ b::Float32 = 0.020833334
```

these can then be wrapped into a named tuple and passed onto the `model.greenhouse_gases` component

```julia
# pass on as named tuple so (; co2, ch4) should work too
julia> model = PrimitiveWetModel(spectral_grid, greenhouse_gases=(; co2))
julia> model.greenhouse_gases.co2
CO2{Int64} <: Main.SpeedyWeather.AbstractGreenhouseGas
└ concentration::Int64 = 280
```

with #969 merged these are now allocated (as refvalues) and time stepped too. The variable name is chosen from the named tuple so `(; co2 = CO2(280), co2_2 = CO2(500))` work too and the variable name can be flexibly chosen

```julia
julia> model = PrimitiveWetModel(spectral_grid, greenhouse_gases=(co2=CO2(280f0), co2_2=ExponentialCO2(spectral_grid)))
PrimitiveWetModel{...} <: PrimitiveWet (1.05 MB)
...

julia> simulation.variables
├ prognostic (421.82 KB)
│ ├ clock: Clock{DateTime, Second, Int64, Millisecond}
│ ├ vor: 560×8×2 LowerTriangularArray{ComplexF32, 3, Array{...}}
│ ├ div: 560×8×2 LowerTriangularArray{ComplexF32, 3, Array{...}}
│ ├ temp: 560×8×2 LowerTriangularArray{ComplexF32, 3, Array{...}}
│ ├ pres: 560×2 LowerTriangularArray{ComplexF32, 2, Array{...}}
│ ├ humid: 560×8×2 LowerTriangularArray{ComplexF32, 3, Array{...}}
│ ├ greenhouse_gases
│ │ ├ co2: Base.RefValue{Float32}
│ │ └ co2_2: Base.RefValue{Float32}
```

In this example initial conditions are

```julia
julia> simulation.variables.prognostic.greenhouse_gases
(co2 = Base.RefValue{Float32}(280.0f0), co2_2 = Base.RefValue{Float32}(367.61813f0))
```

but after 20 days of simulation we have

```julia
julia> simulation.variables.prognostic.greenhouse_gases
(co2 = Base.RefValue{Float32}(280.0f0), co2_2 = Base.RefValue{Float32}(367.718f0))
```

so while one of them stays constant (as defined) the other one increases exponentially.


milankl:
Quick and dirty fit to the Keeling curve via

```math
c = c_0 + a \exp(\Delta t / b)
```

with $c$ being CO2 in ppm, $c_0 = 280$ ppm is the baseline concentration, $a = 3.85$ ppm a scaling factor and $b$ the e-folding time scale of 48 years and $\Delta t$ the time since 1850 in (fractional) years.

image

SpeedyWeather/src/parameterizations/radiation/greenhouse_gases.jl
```diff
@@ -0,0 +1,81 @@
...
+variables(::AbstractCO2, name::Symbol, args...) = (
+    PrognosticVariable(name, ScalarDim(), units="ppm", desc="Carbon dioxide", namespace=:greenhouse_gases),
+)
+
+"""$(TYPEDSIGNATURES)
+Time step greenhouse gas concentration as function of time only."""
+function greenhouse_gases_timestep!(vars::Variables, model::PrimitiveEquation)
+    isnothing(model.greenhouse_gases) && return nothing     # if greenhouse_gases is nothing, then skip this step
+    t = vars.prognostic.clock.time
+    for gas in keys(model.greenhouse_gases)
```
milankl, line 18: @maximilian-gelbrecht I forgot how to loop over NamedTuples in a Reactant-compatible manner, feel free to rewrite this. Is it more about `for i in named_tuple` because that way `i` is possibly a different type every iteration? But `for key in keys(nt)` is fine because then `key::Symbol` always?

SpeedyWeather/src/models/abstract_models.jl
```diff
@@ -18,6 +18,15 @@ finalize!(::Nothing, ::AbstractModel) = nothing
 initialize!(::AbstractModelComponent, ::AbstractModel) = nothing
 finalize!(::AbstractModelComponent, ::AbstractModel) = nothing
 
+# model components that are named tuples of other components,
+# e.g. `greenhouse_gases`, can just call `initialize!` on the elements of the named tuple
+function initialize!(nt::NamedTuple, model::AbstractModel)
+    for value in values(nt)
+        initialize!(value, model)
+    end
```
milankl: So I guess Reactant doesn't like this?

SpeedyWeather/src/parameterizations/radiation/greenhouse_gases.jl
maximilian-gelbrecht: This was actually about Reactant and Enzyme before, both. 
I do think this should be fine as in this case all keys and elements are of the same type, right? 
I can have a bit deeper look at it tomorrow.

maximilian-gelbrecht: Looking again at this a bit closer. This occurred first at `reset_tendencies!` where we loop over a heterogenous tuple that either has LTAs or another NamedTuple in it. Afaik it is enough to define clearly typed inner functions for these loops that can be called. If that still causes problems one can do it either with `first` and `tail` and recursive or with `@generated` functions.

SpeedyWeather/src/dynamics/initial_conditions.jl
```diff
@@ -44,6 +44,9 @@ function initialize!(vars::Variables, model::AbstractModel)
         initialize!(vars.prognostic.particles, vars, model)
     end
 
+    # green house gases are function of time only, just do one "timestep"
+    greenhouse_gases_timestep!(vars, model)
```
maximilian-gelbrecht: Why is the first time step here actually necessary?

SpeedyWeather/src/parameterizations/radiation/greenhouse_gases.jl
```diff
@@ -0,0 +1,81 @@
...
+export ExponentialCO2
+
+"""$(TYPEDSIGNATURES) Exponential CO2 concentration with defaults fitted
+to the Keeling curve, ignoring seasonal variation."""
+@kwdef struct ExponentialCO2{NF} <: AbstractCO2
+    "[OPTION] preindustrial CO2 concentration [ppm]"
+    base_concentration::NF = 280
+
+    "[OPTION] time of preindustrial CO2"
+    start::DateTime = DateTime(1850)
```
maximilian-gelbrecht: Can we start making those things also parametric (for Reactant) when we implement something new. So here just `DT` (or something like this)

maximilian-gelbrecht (approved):

docs/src/radiation.md
```diff
@@ -345,6 +345,40 @@ nothing # hide
...
+```@example radiation
+co2 = ExponentialCO2(spectral_grid)
+```
+
+Step-change scenarios are also available: `TwoTimesCO2` and `FourTimesCO2` double or
+quadruple the preindustrial CO2 at a given date (default: year 2000). Any callable
+`f(t::DateTime) -> concentration_in_ppm` can be wrapped in `CO2(f)` for a fully
+custom trajectory.
+
+!!! warning "Radiation schemes not yet CO2-aware"
```
maximilian-gelbrecht, line 375: Might directly revise this to mention `AnalyticBandRadiation`

maximilian-gelbrecht:
What still left to do here @milankl? I'd love to hook this up with `AnalyticBandRadiation` next.

SpeedyWeather/src/dynamics/initial_conditions.jl
milankl: to set the initial conditions, where else would you set them?

milankl: We currently define greehousegases as function of time (and not itself) only, so that's equivalent

milankl:
I don't think anything is missing, I'll have a look now while I'm on a train and then hopefully good to go!

milankl:
Seems like all is working!! Feel free to merge when passes. I only changed the `timestep` to `time_step` in anticipation of #1035

SpeedyWeather/src/parameterizations/radiation/greenhouse_gases.jl
milankl: Did this!

SpeedyWeather/src/models/abstract_models.jl
milankl: I changed these to loop over the keys btw

SpeedyWeather/src/parameterizations/radiation/greenhouse_gases.jl
milankl: I've avoided looping over the values and instead loop over the keys maybe that already solves the Reactant issues, feel free to change those otherwise

pierluigividale:
This is cool.

maximilian-gelbrecht:
@pierluigividale Glad you like it! If you want to use this to run some forcing experiments, have a look at #1089 . Greg and I also might add some further schemes to that package that support SpeedyWeather with greenhouse gases in the upcoming weeks.

pierluigividale:
Thanks, I shall!

PR #999, opened 2026-03-09
Add CLAUDE.md with codebase summary for AI-assisted development

milankl-claude:
This PR adds a `CLAUDE.md` file at the repository root summarising the package structure, key types, numerical methods, time-step information flow, extensibility patterns, and testing conventions.

The file is intended to give AI coding assistants (e.g. Claude Code) a fast, accurate overview of the codebase so they can assist with development more effectively.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -151,9 +151,9 @@ CPUStatic()  # wraps KernelAbstractions.CPU(; static = true)
 GPU()        # wraps backend-specific GPU (CUDA default; Metal/AMDGPU via extensions)
 ```
 
-Backend-specific array types (`CuArray`, `ROCArray`, `MtlArray`) and
-`convert_to_device` are registered in the extension modules under
-`SpeedyWeatherInternals/ext/`.
+Backend-specific array types (`CuArray`, `ROCArray`, `MtlArray`) are registered
+in the extension modules under `SpeedyWeatherInternals/ext/`.
+Use `on_architecture(device, data)` to transfer data between devices.
 
 ### Work order types
 


milankl:
At the moment I've set up @milankl-claude without any rights as an external collaborator to gather experience first. But I think in the future we can make them a team member but without merging rights?

CLAUDE.md
```diff
@@ -0,0 +1,229 @@
...
+```
+
+All model components (orography, ocean, convection, etc.) inherit from
+`AbstractModelComponent` and implement `initialize!`, optionally `timestep!`
+and `finalize!`.
+
+## Typical Usage
+
+```julia
+spectral_grid = SpectralGrid(trunc=31, nlayers=8)
```
maximilian-gelbrecht: should probably 

```Julia 
arch = SpeedyWeather.CPU() 
spectral_grid = SpectralGrid(trunc=31, nlayers=8, architecture=arch)
```

I've seen Claude get this wrong serval times before, when asking about using other architectures

CLAUDE.md
```diff
@@ -0,0 +1,229 @@
...
+
+```julia
+loop! = kernel!(device(arch), workgroup, worksize)
+loop!(args...)
+```
+
+### Defining a kernel
+
+```julia
+@kernel inbounds = true function my_kernel!(A, B, @Const(c))
```
maximilian-gelbrecht: just remove the `@Const` we don't really want to use it anymore.

CLAUDE.md
milankl-claude: Good catch, fixed! Added `arch = SpeedyWeather.CPU()` and passed it as `architecture=arch` to `SpectralGrid` so the default architecture is explicit. This should prevent confusion when other architectures are used.

CLAUDE.md
milankl-claude: Done, removed `@Const` from the example kernel and also removed the bullet point mentioning it.

milankl-claude:
Yes! Claude Code automatically loads `CLAUDE.md` from the working directory (and parent directories) into the system prompt at the start of every session. This is Claude Code's built-in feature — no configuration needed. So as long as you open Claude Code from within the repo, it will always have this context available.

CLAUDE.md
```diff
@@ -0,0 +1,229 @@
...
+### Architecture abstraction
+
+```julia
+CPU()        # wraps KernelAbstractions.CPU()
+CPUStatic()  # wraps KernelAbstractions.CPU(; static = true)
+GPU()        # wraps backend-specific GPU (CUDA default; Metal/AMDGPU via extensions)
+```
+
+Backend-specific array types (`CuArray`, `ROCArray`, `MtlArray`) and
+`convert_to_device` are registered in the extension modules under
```
maximilian-gelbrecht: For transferring data between devices you should use `on_architecture(device, data)`

milankl-claude: Thanks! Replaced `convert_to_device` with `on_architecture(device, data)` and rewrote the sentence to make it the main way to transfer data between devices.

maximilian-gelbrecht (approved):
Fine with me to merge it. We can still quickly add and edit things as we go

Btw: this is #999, we are very close nr. 1000 ;)

maximilian-gelbrecht:
Actually, one thing more to add that Claude also often gets wrong is how to initialise both zero and random arrays with RingGrids and LowerTriangularArrays for all kinds of tests.

milankl:
> Actually, one thing more to add that Claude also often gets wrong is how to initialise both zero and random arrays with RingGrids and LowerTriangularArrays for all kinds of tests.

I believe there'll be several things it gets often wrong and so I suggest in every PR where we realise this to take the freedom and let Claude add additional points or update the CLAUDE.md file.

PR #1002, opened 2026-03-10
Move get_asset to RingGrids with NCDatasets extension

### LINKED ISSUE #996: Moving input data handling to RingGrids (and generalise it)
I know, I am a bit late after #983 is merged now, but would you mind if we slightly generalise the input data handling and put it in RingGrids? In case we absolutely don't want RingGrids to depend on `NCDatasets` we could also do it in an extension for RingGrids.

With that change, we could just reuse it in Terrarium as well, at least as long as we don't have specialised input data there. With reuse I mean that Terrarium can access SpeedyWeather's input data with only depending on `RingGrids`, not that it set ups another asset repo with the same code. 

cc @bgroenks96 @gregrmunday

milankl-claude:
Closes #996

Moves the asset-downloading code from `SpeedyWeather` into `RingGrids`, addressing all points raised in #996:

### Changes

**`RingGrids/src/get_asset.jl`** (new, no NCDatasets dependency)
- `get_asset` download and artifact-caching logic
- `load_shape_preserving` helpers
- `_get_asset` method for `AbstractFullField` (wraps loaded array into a grid field)
- `const ASSETS_URL` and `const DEFAULT_ASSETS_VERSION` (both now exported from RingGrids)

**`RingGrids/ext/RingGridsNCDatasetsExt.jl`** (new extension)
- NCDatasets is now a **weakdep** of RingGrids — loaded automatically when NCDatasets.jl is also loaded
- `get_nc_variable_name` lives here
- `_get_asset` methods for `NCDataset` and `Array` file formats live here

**`RingGrids/Project.toml`**
- `Artifacts` and `Pkg` added as direct deps
- `NCDatasets` added as a weakdep with extension `RingGridsNCDatasetsExt`

**`SpeedyWeather/src/SpeedyWeather.jl`**
- `include("input/get_asset.jl")` removed; `get_asset` is now available via `using RingGrids`
- `import Artifacts` and `import Pkg` removed

**`SpeedyWeather/Project.toml`**
- `Artifacts` and `Pkg` removed from `[deps]` and `[compat]`

### Points from #996 addressed

| Point | How addressed |
|---|---|
| Move to RingGrids as NCDatasets extension | `RingGridsNCDatasetsExt` — NCDatasets is a weakdep, loaded only when needed |
| Most of `get_asset` is NCDatasets-independent | Core logic in `RingGrids/src/get_asset.jl`; only file-reading in extension |
| Make `ASSETS_URL` a kwarg | Added `assets_url::String = ASSETS_URL` kwarg to `get_asset` |
| Make `DEFAULT_ASSETS_VERSION` configurable | Already a kwarg (`version`); constant now exported from RingGrids |
| Simplify `AVAILABLE_ASSETS_VERSIONS` | Removed entirely — unknown versions now simply fail at download time with a clear error message |

### Notes
- SpeedyWeather callers are unchanged: they still pass `FileFormat = NCDataset` and since SpeedyWeather has NCDatasets as a direct dep, `RingGridsNCDatasetsExt` is loaded automatically
- Artifact storage moves from `SpeedyWeather/Artifacts.toml` to `RingGrids/Artifacts.toml`
- Other projects (e.g. Terrarium) can load only RingGrids + NCDatasets and call `get_asset(..., assets_url="https://...")` pointing to their own asset repo

### DIFF
--- a/RingGrids/ext/RingGridsNCDatasetsExt.jl
+++ b/RingGrids/ext/RingGridsNCDatasetsExt.jl
@@ -29,7 +29,7 @@ function get_nc_variable_name(ncfile::NCDataset, name::String)
 end
 
 # lazy load from NCDataset into netCDF variable (actually CommonDataModel.CFVariable)
-function RingGrids._get_asset(path::String, name::String, ArrayType::Type{<:NCDataset}, FileFormat::Type{<:NCDataset}, fill_value)
+function RingGrids.load_asset(path::String, name::String, ArrayType::Type{<:NCDataset}, FileFormat::Type{<:NCDataset}, fill_value)
     ds = NCDataset(path)
     # TODO also read lat, lon from file and flip array in case it's not as expected
     target_name = get_nc_variable_name(ds, name)
@@ -37,8 +37,8 @@ function RingGrids._get_asset(path::String, name::String, ArrayType::Type{<:NCDa
 end
 
 # load from NetCDF into Array
-function RingGrids._get_asset(path::String, name::String, ArrayType::Type{<:Array}, FileFormat::Type{<:NCDataset}, fill_value)
-    v, ds = RingGrids._get_asset(path, name, NCDataset, FileFormat, fill_value)
+function RingGrids.load_asset(path::String, name::String, ArrayType::Type{<:Array}, FileFormat::Type{<:NCDataset}, fill_value)
+    v, ds = RingGrids.load_asset(path, name, NCDataset, FileFormat, fill_value)
     data = RingGrids.load_shape_preserving(v, Val(ndims(v)))
     if eltype(data) <: AbstractFloat    # exclude case of loading integer data, e.g. land-sea mask
         nc_fill = get(v.attrib, "_FillValue", fill_value)
--- a/RingGrids/src/RingGrids.jl
+++ b/RingGrids/src/RingGrids.jl
@@ -167,10 +167,10 @@ include("vertices.jl")
 include("statistics.jl")
 
 # ASSET DOWNLOADING
-export get_asset, ASSETS_URL, DEFAULT_ASSETS_VERSION
+export get_asset, load_asset, ASSETS_URL, DEFAULT_ASSETS_VERSION
 
-# stub extended by RingGridsNCDatasetsExt
-function _get_asset end
+# load_asset: extension dispatch point implemented by RingGridsNCDatasetsExt (and future format extensions)
+function load_asset end
 
 include("get_asset.jl")
 
--- a/RingGrids/src/get_asset.jl
+++ b/RingGrids/src/get_asset.jl
@@ -40,7 +40,7 @@ function get_asset(
     if !from_assets     # try to load locally
         if isfile(path) # check if path is local (custom input)
             try
-                return _get_asset(path, name, ArrayType, FileFormat, fill_value)
+                return load_asset(path, name, ArrayType, FileFormat, fill_value)
             catch e
                 throw("Local asset loading failed with: $e")
             end
@@ -83,12 +83,12 @@ function get_asset(
 
     asset_path = joinpath(Artifacts.artifact_path(hash), filename)
 
-    return _get_asset(asset_path, name, ArrayType, FileFormat, fill_value)
+    return load_asset(asset_path, name, ArrayType, FileFormat, fill_value)
 end
 
 # load from array into a RingGrid Field
-function _get_asset(path::String, name::String, ArrayType::Type{<:AbstractFullField}, FileFormat, fill_value)
-    data = _get_asset(path, name, Array, FileFormat, fill_value)        # first load as Array
+function load_asset(path::String, name::String, ArrayType::Type{<:AbstractFullField}, FileFormat, fill_value)
+    data = load_asset(path, name, Array, FileFormat, fill_value)        # first load as Array
     return ArrayType(data, input_as = Matrix)
 end
 


milankl-claude:
@milankl @maximilian-gelbrecht @gregrmunday — would appreciate your review on this, given you all contributed to the asset handling and related infrastructure. Happy to adjust based on your feedback!

milankl:
I've just told Claude to listen to you guys and act on it, so don't abuse it 😉 milankl-claude if you read this just give a thumbs up!

RingGrids/src/RingGrids.jl
```diff
@@ -162,4 +166,13 @@ include("interpolation.jl")
 include("vertices.jl")
 include("statistics.jl")
 
+# ASSET DOWNLOADING
+export get_asset, ASSETS_URL, DEFAULT_ASSETS_VERSION
+
+# stubs extended by RingGridsNCDatasetsExt
+function _get_asset end
```
bgroenks96: Methods prefixed with `_` should not be interface methods that are extended by an extension module. The interface should be defined as a dispatch of the `get_asset` method.

RingGrids/src/RingGrids.jl
```diff
@@ -162,4 +166,13 @@ include("interpolation.jl")
...
+function get_nc_variable_name end
```
bgroenks96: This is a leaky abstraction. File format specific methods should not be defined at the top level. Alternatively, if extending this is absolutely necessary, it could be changed to `get_output_variable_name` or something like that.

RingGrids/ext/RingGridsNCDatasetsExt.jl
```diff
@@ -0,0 +1,52 @@
...
+    target_name = RingGrids.get_nc_variable_name(ds, name)
+    return ds[target_name], ds
+end
+
+# load from NetCDF into Array
+function RingGrids._get_asset(path::String, name::String, ArrayType::Type{<:Array}, FileFormat::Type{<:NCDataset})
+    v, ds = RingGrids._get_asset(path, name, NCDataset, FileFormat)
+    data = RingGrids.load_shape_preserving(v, Val(ndims(v)))
+    if eltype(data) <: AbstractFloat    # exclude case of loading integer data, e.g. land-sea mask
+        fill_value = get(v.attrib, "_FillValue", NaN)
```
bgroenks96: The default fill value should be able to be specified by the user.

RingGrids/src/RingGrids.jl
maximilian-gelbrecht: I don't mind this to be honest. It's also not really meant to be a fully fledged out interface for the foreseeable future.

RingGrids/ext/RingGridsNCDatasetsExt.jl
maximilian-gelbrecht: @milankl-claude Let the user specify the fill value in `get_asset` and `_get_asset`, default to `NaN`.

RingGrids/src/RingGrids.jl
milankl-claude: (as requested by @milankl — giving a thumbs up! 👍)

Since `_get_asset` is not exported and Max agrees it is fine as an internal, non-public extension point, I will keep the underscore prefix for now. If the interface is ever formalised we can revisit this.

RingGrids/src/RingGrids.jl
milankl-claude: Good catch. `get_nc_variable_name` is purely an NCDatasets implementation detail and shouldn't exist in the RingGrids namespace at all. Fixed in the latest commit: the function is now module-local to `RingGridsNCDatasetsExt` (no longer a stub in `RingGrids.jl`, no longer prefixed with `RingGrids.`).

RingGrids/ext/RingGridsNCDatasetsExt.jl
milankl-claude: Done — added `fill_value = NaN` as a keyword argument to `get_asset`, threaded through to `_get_asset`. In the extension, the NetCDF file's `_FillValue` attribute is still used to identify missing data in the file, and the user-supplied `fill_value` controls what those entries are replaced with (defaulting to `NaN`).

milankl-claude: Done, see above reply to Brian's comment.

RingGrids/src/RingGrids.jl
bgroenks96: Extension modules are essentially lightweight downstream packages, so there really should not be any difference in how an extension module interacts with the top-level package module (here `RingGrids`) vs. how another Julia library or user package would do so.

I would therefore strongly recommend against violating standard Julia conventions like this, and make the interface, however minimal, well defined, even if it doesn't seem that important right now.

maximilian-gelbrecht: If it gets a different name without a leading underscore that's also fine with me, but I really wouldn't bother to make this any more complicated. What we have already have works, and even with this, it would be possible to define another extension for a new file format and dispatch on the function as it is, the file format is already in the signature of the function.

milankl-claude: Brian, you make a fair point about Julia conventions — a leading underscore does signal "private, do not touch" and using it as an extension dispatch target is indeed a mixed message.

Max is fine with dropping the underscore. Given that, I will rename `_get_asset` → `load_asset` (or a similarly unambiguous name) in the stub definition in `RingGrids.jl` and update the extension accordingly. That way the interface is explicit and public without any ambiguity about its intended role.

I will push the rename in a follow-up commit shortly.

milankl-claude: Done — renamed `_get_asset` → `load_asset` throughout (`RingGrids.jl` stub, `get_asset.jl`, and the NCDatasets extension). The function is now exported as a proper public extension point.

Ref: https://github.com/milankl-claude/SpeedyWeather.jl/commit/8f522dd1

maximilian-gelbrecht:
@bgroenks96 Are you fine with the changes? Then we could merge this.

bgroenks96:
Fine with me, maybe wait for @gregrmunday

RingGrids/Project.toml
```diff
@@ -18,14 +20,17 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 GeoMakie = "db073c08-6b98-4ee5-b6a4-5efafb3259c6"
 Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
+NCDatasets = "85f8d34a-cbdd-5861-8df4-14fed0d494ab"
```
maximilian-gelbrecht, line 23: @milankl-claude move the `NCDatasets` dependency to the weakdeps, it's in an extension.

milankl-claude:
@gregrmunday — @bgroenks96 has approved the changes. Could you take a look when you have a moment? Happy to answer any questions about the implementation.

RingGrids/Project.toml
maximilian-gelbrecht: reminder for @milankl-claude to address the comment above

milankl: It is already! github collapsing the preview just looks like it's under `[deps]` but it's under `[weakdeps]`

gregrmunday:
> @gregrmunday — @bgroenks96 has approved the changes. Could you take a look when you have a moment? Happy to answer any questions about the implementation.

I've taken a look - looks good to me! Although @maximilian-gelbrecht has an unresolved requested change above^

milankl-claude:
Thanks for looking @gregrmunday\! The unresolved thread has been sorted:

- The `maximilian-gelbrecht` comment about `NCDatasets` in `weakdeps` was just resolved by @milankl — it's already under `[weakdeps]` in `RingGrids/Project.toml`, GitHub's collapsed diff view was misleading.
- `bgroenks96`'s original `CHANGES_REQUESTED` (about the `_get_asset` naming) was addressed in a prior commit (renamed to `load_asset`), and Brian subsequently approved.

All three reviewers have now approved and all raised concerns are addressed. The PR is ready to merge whenever @milankl or @maximilian-gelbrecht wants to pull the trigger.

PR #1004, opened 2026-03-13
`get_asset` interpolate option and unit tests

maximilian-gelbrecht:
Adds an option to the `get_asset` to directly interpolate to a specified output grid and also adds units tests


milankl-claude:
Nice addition\! The `interpolate\!` on-the-fly path (which internally creates an `AnvilInterpolator`) handles the Field-to-Field case cleanly.

One thing worth flagging: `interpolate\!(out_field, data)` requires both arguments to be `Field` subtypes. If a user calls `get_asset(...; output_grid=g)` with the default `ArrayType` (plain `Array`), they will get a confusing method-not-found error rather than a helpful message.

Two small suggestions:
1. Add a guard before the interpolation block, e.g. `data isa AbstractField || throw(ArgumentError("output_grid requires a Field ArrayType, e.g. ArrayType=FullGaussianField"))`
2. Mention in the docstring that `output_grid` only works when `ArrayType` is a `Field` subtype.

Otherwise the implementation is clean, the tests cover the important paths (download, local load, version kwarg, interpolation, error on bad variable name), and the CHANGELOG entry is in order. Happy to approve once those two points are addressed.

maximilian-gelbrecht:
I'll merge this relatively quickly, because I actually directly want to use it over at Terrarium ;)

milankl:
Fine with me. Only the submodule can get released I believe in this case with a method that was moved given get asset isn't part yet of a release?

maximilian-gelbrecht:
Ah right I forgot about that part for a second, but yes, this should be fine to just release RingGrids.

PR #1008, opened 2026-03-13
Fix Vitepress docs

milankl:
Thanks to @lazarusA: There was one line missing because we deploy in a separate repo.

### DIFF
--- a/docs/src/installation.md
+++ b/docs/src/installation.md
@@ -63,17 +63,19 @@ section in the Pkg documentation to learn more about the differences between `ad
 
 ## Compatibility with Julia versions
 
-SpeedyWeather.jl requires Julia v1.10 or later. The package is tested on Julia 1.10, and 1.11.
+SpeedyWeather.jl requires Julia v1.10 or later.
+The package is tested on Julia 1.10, 1.11 and 1.12.
 
 ## Extensions
 
-SpeedyWeather.jl has a weak dependency on
+SpeedyWeather.jl and its submodules have several weak dependencies for visualisation (Makie, GeoMakie),
+differentiability (Enzyme, FiniteDifferences) or GPU computing (CUDA, AMDGPU, Metal).
+These are all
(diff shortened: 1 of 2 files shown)


milankl:
TODO: Collapse top

image

milankl:
Preview is here https://speedyweather.github.io/SpeedyWeatherDocumentation/previews/PR1008/

the `documenter/deploy` workflow unfortunately doesn't link correctly anymore.

image

@giordano you fixed this problem before do you know why this link is broken again with DocumterVitepress?

giordano:
I [fixed it in Documenter](https://github.com/JuliaDocs/Documenter.jl/pull/2816), I believe DocumterVitepress does the deployement in a different way.

giordano:
Ok, it uses `Documenter.post_status` https://github.com/LuxDL/DocumenterVitepress.jl/blob/05209bf8d0a7c3a7c67690e3d562b6477a5030ad/src/DocumenterVitepress.jl#L184, but I don't know how `deploy_config = Documenter.auto_detect_deploy_system()` is constructed

giordano:
I'm pretty sure the problem is that `repo` in https://github.com/LuxDL/DocumenterVitepress.jl/blob/05209bf8d0a7c3a7c67690e3d562b6477a5030ad/src/DocumenterVitepress.jl#L184 should be `deploy_repo` (which is probably somewhere hidden in `kwargs`): https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/ffecb5fc05e55290862a63dc9a05ac72918a3706/docs/make.jl#L94

giordano:
But the `repo` argument to `post_status` was removed in https://github.com/JuliaDocs/Documenter.jl/pull/2755, so good luck when a new version of Documenter is out.

lazarusA:
https://github.com/LuxDL/DocumenterVitepress.jl/issues/335

PR #1010, opened 2026-03-19
Rename `variable_reference` to `reference_variable`

navidcy:
I propose renaming these variable to read more like English prose.

x-ref: https://github.com/NumericalEarth/NumericalEarth.jl/pull/133#discussion_r2948302507

cc @maximilian-gelbrecht

### DIFF
--- a/SpeedyWeather/Project.toml
+++ b/SpeedyWeather/Project.toml
@@ -1,6 +1,6 @@
 name = "SpeedyWeather"
 uuid = "9e226e20-d153-4fed-8a5b-493def4f21a9"
-version = "0.18.2"
+version = "0.19.0"
 authors = ["SpeedyWeather contributors"]
 
 [deps]


maximilian-gelbrecht:
A new release would need a new release of the submodules (at the very least of `RingGrids`) as well, because we moved some code from `SpeedyWeather` into `RingGrids` since the last release. 

I wouldn't mind though. 

@milankl any objections to a new release?

milankl:
Can we merge this into #969 instead which renames a bunch of variables anyway? Also this should be considered a breaking change?

I'm very much agree on the renaming though. I do want to name variables as close as possible to their typical brief but unapproximated way of writing or speaking. The prognostic variables remain an exception and that might be good. We currently use vor div temp humid pres u v. Any inputs on how to name them appreciated!

navidcy:
I changed the base to `mk/newvars` and a bunch of conflicts emerged...

@milankl, perhaps it'll be easier if commit directly to `mk/newvars`?

milankl:
There will be conflicts but check whether merge gives you fewer. I can resolve them. Thanks Navid!!!

navidcy:
If you can resolve them, then please do!

navidcy:
I think I resolved them, but a second look wouldn't hurt @milankl.

navidcy:
> Also this should be considered a breaking change?

Yes. I think so. I pushed v0.19 version in Project.toml/CHANGELOG.md

milankl:
Thanks Navid, this is great! However, if I change the base branch there's just too many conflicts now so I manually moved your changes via https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1025/commits/8f37c620f3c3f2ea99f1051add6f87bb81ed5330 into #1025

navidcy:
Great! Thanks!

PR #1013, opened 2026-03-23
Learned + dynamic albedo scheme 

gregrmunday:
Depends on #959.

A more realistic and dynamic albedo scheme. 

**Ocean**:  Following Jin et al. (2011)
**Sea-ice**: tbc
**Land**: Learned albedo representation depending on the local land surface conditions, based on MODIS albedo parameters and ERA5 reanalysis.


gregrmunday:
Ocean albedo seems to be running stable from an initial run, see the wind dependence over the ocean. 

Screenshot 2026-03-23 at 12 34 49

milankl:
An albedo of 1 at night though???

gregrmunday:
> An albedo of 1 at night though???


I think this actually makes sense? If the sun is at zenith angle 90 then you should get 100% reflectance, which then next day will decrease as the sun comes up

milankl:
@bwtimmermans Greg is cooking up a learned albedo scheme here. Feel free to chat about how this is being implemented. At the moment the PR also contains commits from #959 yet that aren't merged yet ...

gregrmunday:
Both the land (NN) and ocean schemes implemented now. Land NN is still a WIP, and I need to make it a fair bit lighter, as it's currently ~300K params. Albedo over the land + ocean is 1 at night, really this is for numerical stability so the fraction of direct/diffuse radiation at night doesn't blow up if you divide by 0, but this could be any value at night as albedo doesn't have an impact. 

Wiggly lines across the Sahara are impacts of the new LAI input maps, which sit inside a single `vegetation.nc` - I'll make a PR in SpeedyWeatherAssets to replace the current file with this one. 

I've also taken the `glacier_mask` variable from ERA5-Land to initialise areas of 'permanent' ice/snow with 10m of snow depth. This includes Antarctica, Greenland and some glaciers - I'll make another PR in SpeedyWeatherAssets if we're happy to do it this way!

Screenshot 2026-03-24 at 10 57 21

Screenshot 2026-03-24 at 11 04 55


I also think we're seeing generally lower albedo than we would normally because I think we're too cloudy/too much convection, and this is the blue-sky albedo we're looking at here.

gregrmunday:
The new albedo scheme, plus the new surface roughness scheme seem quite stable together, though I need to test for a few years. We do have a blowup though, but originating from the inclusion of the new `glacier_mask.nc` initial condition.

I think I've isolated the problem and it seems like because the snow runoff rate is dependent on the snow depth, with new perennial snow packs of depth 10m, the top layer soil temperature just starts to get ridiculously cold. After about a month at either high or low resolution the top layer soil temp in Greenland and the Arctic circle-ish goes to absolute zero:

Screenshot 2026-03-24 at 16 16 50


I'm not exactly sure why this happens only in the northern hemisphere and not Antarctica. Maybe because we initialise in the winter, the NH doesn't get any SW down and so cools a lot quicker? 

Either way, when I set the snow `runoff_time_scale` to `9999999` things are stable again.

Also, why is the soil temperature so hot in the tropics??

Same run on a comparable colour scale (high is actually 70degC)

Screenshot 2026-03-24 at 16 37 24

Compared to the vanilla run:

Screenshot 2026-03-24 at 16 34 00

SpeedyWeather/src/physics/albedo.jl
```diff
@@ -272,3 +272,174 @@ initialize!(albedo::LandSnowAlbedo, model::PrimitiveEquation) = nothing
...
+## OceanAlbedo (Jin et al., 2011)
+export OceanAlbedo
+
+"""
+Ocean surface albedo parameterization based on Jin et al. (2011).
+Calculates broadband albedo as a function of wind speed, surface roughness, 
+and solar zenith angle, partitioned into direct and diffuse components.
+Fields are $(TYPEDFIELDS)
+"""
+@parameterized @kwdef struct OceanAlbedo{NF} <: AbstractAlbedo
```
milankl: Give it a more telling name given it's complexity? What about `JinOceanAlbedo` if it strictly follows that authors' work or `OceanWaveAlbedo` to denote that it's albedo that attemps to quantify the effect of waves, or `OceanWaveAngleAlbedo` to also highlight that it depends on the angle of the incoming light?

SpeedyWeather/src/physics/albedo.jl
```diff
@@ -272,3 +272,174 @@ initialize!(albedo::LandSnowAlbedo, model::PrimitiveEquation) = nothing
...
+Adapt.@adapt_structure OceanAlbedo
+OceanAlbedo(SG::SpectralGrid; kwargs...) = OceanAlbedo{SG.NF}(; kwargs...)
+
+initialize!(::OceanAlbedo, ::PrimitiveEquation) = nothing
+
+"""$(TYPEDSIGNATURES) Compute unpolarized Fresnel reflectance."""
+@inline function fresnel_reflectance(n::NF, mu::NF) where {NF}
+    # Mu is bounded to prevent domain errors
+    mu = clamp(mu, NF(1.0e-5), one(NF))
+    theta_i = acos(mu)
```
milankl: use greek unicode characters here for clarity?

SpeedyWeather/src/physics/albedo.jl
```diff
@@ -272,3 +272,174 @@ initialize!(albedo::LandSnowAlbedo, model::PrimitiveEquation) = nothing
...
+
+    sin_theta_t = sin(theta_i) / n
+    theta_t = asin(sin_theta_t)
+
+    # Perpendicular and parallel polarizations
+    r_s = ((cos(theta_i) - n * cos(theta_t)) / (cos(theta_i) + n * cos(theta_t)))^2
+    r_p = ((n * cos(theta_i) - cos(theta_t)) / (n * cos(theta_i) + cos(theta_t)))^2
```
milankl: is this where division by zero is problematic? Then I would deal with this here instead of setting albedo to 1? E.g. you could do something like `min(1/0, 1)` which simply says you may divide by zero here if the denominator is zero but in that case the definition changes and so you cap it. I prefer this over having a confusing albedo being 1

SpeedyWeather/src/physics/albedo.jl
```diff
@@ -272,3 +272,174 @@ initialize!(albedo::LandSnowAlbedo, model::PrimitiveEquation) = nothing
...
+    # 4. Foam Adjustment (Eqs 16 & 17)
+    f_wc = NF(2.95e-6) * (wind_speed^NF(3.52))
+    f_wc = clamp(f_wc, zero(NF), one(NF))
+
+    alpha_final = f_wc * foam_reflectance + (one(NF) - f_wc) * alpha_b
+
+    # Also account for sea ice if sea ice concentration is available
+    if haskey(progn.ocean, :sea_ice_concentration)
+        sea_ice_conc = progn.ocean.sea_ice_concentration[ij]
+        alpha_ice = NF(0.6)  # Typical sea ice albedo
```
milankl: move into parameter of `OceanAlbedo`?

SpeedyWeather/src/parameterizations/albedo.jl
```diff
@@ -272,3 +272,174 @@ initialize!(albedo::LandSnowAlbedo, model::PrimitiveEquation) = nothing
...
+    alpha_b = f_dir * alpha_dir_s + f_dif * alpha_dif_s + volume_scattering
+
+    # 4. Foam Adjustment (Eqs 16 & 17)
+    f_wc = NF(2.95e-6) * (wind_speed^NF(3.52))
+    f_wc = clamp(f_wc, zero(NF), one(NF))
+
+    alpha_final = f_wc * foam_reflectance + (one(NF) - f_wc) * alpha_b
+
+    # Also account for sea ice if sea ice concentration is available
+    if haskey(progn.ocean, :sea_ice_concentration)
```
milankl: I would try to use `haskey` solely to set a value of a scalar, e.g.
```julia
a = haskey(...) ? some_struct.a : zero(NF)
```

that way you avoid duplicating code below

SpeedyWeather/src/physics/albedo.jl
```diff
@@ -272,3 +272,174 @@ initialize!(albedo::LandSnowAlbedo, model::PrimitiveEquation) = nothing
...
+        diagn.physics.ocean.albedo[ij] = clamp((one(NF) - sea_ice_conc) * alpha_final + sea_ice_conc * alpha_ice, zero(NF), one(NF))
+    else
+        diagn.physics.ocean.albedo[ij] = clamp(alpha_final, zero(NF), one(NF))
+    end
+
+    return
+end
+
+export LearnedBRDF
```
milankl: Give it a more telling name, what's BRDF?

SpeedyWeather/src/physics/albedo.jl
```diff
@@ -272,3 +272,174 @@ initialize!(albedo::LandSnowAlbedo, model::PrimitiveEquation) = nothing
...
+        0.19499032, 0.05033005, 0.01945734,
+        0.36265752, 0.12071748, 0.04039559,
+        0.2922392, 0.08076486, 0.02871396,
+    ]
+    output_stds::Vector{NF} = Float32[
+        0.21011186, 0.02822702, 0.01345009,
+        0.13475, 0.05032686, 0.02287926,
+        0.16457707, 0.03171282, 0.01661654,
+    ]
+
```
milankl: You think the NN normalization should be part of the weights or come with the asset somehow? They are intrinsically linked to what the NN assumes, so "hardcoding" (it's not but it's a long vector so who changes it) the normalization here seems inconsistent. I think many torch models would have weights and normalization stored together or in separate files typically placed in the same folder

SpeedyWeather/ext/SpeedyWeatherLuxExt.jl
```diff
@@ -0,0 +1,403 @@
...
+    NF = typeof(f_iso)
+    wsa_vol_const = NF(0.189184)
+    wsa_geo_const = NF(-1.377622)
+    return f_iso + (wsa_vol_const * f_vol) + (wsa_geo_const * f_geo)
+end
+
+"""
+Calculate SAL using the Lucht et al. (2000) polynomial.
+"""
+Base.@propagate_inbounds function calculate_black_sky_albedo(f_iso, f_vol, f_geo, mu0)
```
milankl: these look like a good opportunity of a functor to me that way you can have a function as part of your LearnedBRDF and can still dispatch over it and easily swap it out for another way of calculating the black sky albedo?

milankl (commented):
Just some WIP comments, maybe they are helpful!

SpeedyWeather/src/physics/albedo.jl
gregrmunday: I like `JinOceanAlbedo` as it does follow this work exactly

gregrmunday:
At the moment this runs for ~1yr at `trunc=127` resolution until it blows up. Looks like there are a few interesting things happening, but seems like the blowup happens with an instability in advection?

Both learned surface roughness and albedo running here.

Plots below are in the timestep before blowup

U-wind here, top layer
Screenshot 2026-04-02 at 18 36 26

Instability seen more clearly in v-wind, top layer
Screenshot 2026-04-02 at 18 36 18

Soil moisture is saturated
Screenshot 2026-04-02 at 18 36 35

I think, because of soil moisture saturation, albedo has decreased globally from this:
Screenshot 2026-04-02 at 18 37 45

to this:
Screenshot 2026-04-02 at 18 36 45


This run is without the 10m snow cap (just set runoff timescale to inf) so about a metre of snow has accumulated on top of Antarctica and Greenland. Not sure what impacts that's having.

milankl:
Indeed I think you might be violating CFL here, this looks like the speedlimit drag isn't kicking in, what is `model.drag` ?

SpeedyWeather/src/parameterizations/land/snow.jl
```diff
@@ -79,38 +84,45 @@ function initialize!(
     interpolator = RingGrids.interpolator(snow_depth, glm, NF = Float32)
     interpolate!(snow_depth, glm, interpolator)
 
-    return set!(progn, model.geometry, snow_depth = glm)
+    return set!(vars, model.geometry, snow_depth = glm)
```
milankl: this shouldn't work because `snow_depth` should be in the `land` namespace? I reckon you have to add `namespace = :land` as kwarg here

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
```diff
@@ -87,24 +87,24 @@ Adapt.@adapt_structure ConstantSurfaceRoughness
 ConstantSurfaceRoughness(SG::SpectralGrid, kwargs...) = ConstantSurfaceRoughness{SG.NF}(; kwargs...)
 initialize!(::ConstantSurfaceRoughness, ::PrimitiveEquation) = nothing
 
-@propagate_inbounds function parameterization!(ij, diagn, progn, scheme::AbstractSurfaceRoughness, model)
-    surface_roughness!(ij, diagn, progn, scheme, model.land_sea_mask)
-    return surface_roughness!(ij, diagn, progn, scheme, model.land_sea_mask)
+@propagate_inbounds function parameterization!(ij, vars, scheme::AbstractSurfaceRoughness, model)
+    surface_roughness!(ij, vars, scheme, model.land_sea_mask)
+    return surface_roughness!(ij, vars, scheme, model.land_sea_mask)
```
milankl: why is this called twice?

milankl (commented):
Just some comments!

gregrmunday:
I've made a slight change to the `albedo!` call where instead of passing in the albedo as an argument to `albedo!`, it leaves it to the method to decide to use it, which it can through `vars`. Feels slightly cleaner this way to me

SpeedyWeather/src/parameterizations/albedo.jl
```diff
@@ -40,6 +40,25 @@ function OceanLandAlbedo(
...
+    properties = propertynames(A)
+    n = length(properties)
+    for (i, key) in enumerate(properties)
+        val = getfield(A, key)
+        s = i == n ? "└" : "├"  # choose ending └ for last property
+        p = i == n ? print : println
+        p(io, "$s $key: $(typeof(val))")
+    end
+    return
+end
```
milankl, line 54: This is already printed for me as

```julia
julia> OceanLandAlbedo(spectral_grid)
OceanLandAlbedo{OceanSeaIceAlbedo{Float32}, L...} <: Main.SpeedyWeather.AbstractAlbedo
├ ocean::OceanSeaIceAlbedo{Float32}
└ land::LandSnowAlbedo{Float32, SaturatingSnowCover}
```

as #969 minimized the amount of custom `Base.show` methods and implements a generic one that hopefully works for most model components. So I reckon you can remove this here

SpeedyWeather/src/parameterizations/albedo.jl
```diff
@@ -52,25 +71,22 @@ function variables(albedo::OceanLandAlbedo)
...
+    return albedo!(ij, vars, albedo.land, model)
 end
 
 """$(TYPEDSIGNATURES) Single Albedo: Call same albedo over ocean and land."""
 @propagate_inbounds function parameterization!(ij, vars::Variables, albedo::AbstractAlbedo, model)
-    albedo!(ij, vars.parameterizations.ocean.albedo, vars, albedo, model)
-    return albedo!(ij, vars.parameterizations.land.albedo, vars, albedo, model)
+    albedo!(ij, vars, albedo.ocean, model)
+    return albedo!(ij, vars, albedo.land, model)
 end
```
milankl, line 90: As just discussed, I'd be happy for you to remove the co-existing single+custom albedos we currently have whereby e.g. `AlbedoClimatology` can be used as is but also inside the `OceanLandAlbedo`. I think it's just important to balance

- consistency, at the moment the albedo is actually an exception as they're supposed to extend `albedo!` not `parameterization!`  which isn't nice
- user interface: something as simple as `PrimitiveWetModel(..., albedo = GlobalConstantAlbedo(0.3))` should just work
- flexibility, so you can plug in your albedo as you like

SpeedyWeather/src/parameterizations/albedo.jl
```diff
@@ -264,3 +281,161 @@ initialize!(albedo::LandSnowAlbedo, model::PrimitiveEquation) = nothing
...
+    norm_stds::Vector{Float32} = zeros(Float32, 10)
+    unnorm_means::Vector{Float32} = zeros(Float32, 6)
+    unnorm_stds::Vector{Float32} = zeros(Float32, 6)
+
+    brdf_nn::LNN
+    brdf_params::LP
+    brdf_states::LS
+end
+
+function Base.show(io::IO, scheme::LearnedLandAlbedo)
```
milankl, line 436: remove? see above

SpeedyWeather/src/parameterizations/radiation/shortwave_radiation.jl
```diff
@@ -211,6 +217,19 @@ One-band shortwave radiative transfer with cloud reflection and ozone absorption
...
 
+    # Calculate direct/diffuse radiation fractions at the surface
+    mu = max(cos_zenith, 0.05)
+    
+    # Baseline Rayleigh scattering increases at lower solar angles (longer path)
+    clear_sky_diffuse = clamp(0.15 + 0.1 / mu, 0.0, 1.0)
+    
+    # Linearly scale between clear sky diffuse and 100% diffuse based on cloud cover
+    fraction_diffuse = clear_sky_diffuse * (1 - cloud_cover) + 1.0 * cloud_cover
+    fraction_direct = 1.0 - fraction_diffuse
```
milankl, line 228: this isn't type stable code. Create `model.shortwave_radiation.scattering` as a component that defines this calculation, e.g. as a functor?

Otherwise how to write this type stable

```julia
max(a, convert(typeof(a), 0.05))   # if type of a is known at compile time the conversion is compiled away
clamp(convert(typeof(a), 0.15) + 1 / 10mu, 0, 1) # clamp automatically does type promotion
1 * a   # for integer 1 this will be compiled away, for 1.0 * a this triggers conversion to double ... bad!
1 - a   # this will convert 1 to typeof a at compile time
```

some fractions like 0.15 can also be neatly written as rationals, e.g. `3 // 20` if they're compiled away that's great, if they're not that's bad, so better check, e.g.

```julia
julia> f(x) = 3 // 20 + 1 / 10x

julia> @code_llvm f(2f0)
; Function Signature: f(Float32)
;  @ REPL[1]:1 within `f`
define float @julia_f_1220(float %"x::Float32") #0 {
top:
; ┌ @ promotion.jl:434 within `*` @ float.jl:497
   %0 = fmul float %"x::Float32", 1.000000e+01
; └
; ┌ @ promotion.jl:436 within `/` @ float.jl:498
   %1 = fdiv float 1.000000e+00, %0
; └
; ┌ @ promotion.jl:433 within `+` @ float.jl:495
   %2 = fadd float %1, 0x3FC3333340000000
   ret float %2
; └
}
```

So the rational is compiled to `0x3F...` and only multiplication and division happen in the desired format, no conversion!

milankl (commented):
Maybe as an idea, not sure it would be convenient, what about defining all albedos as 

```julia
@kwdef struct Albedo{O, L, E} <: AbstractAlbedo
    ocean::O
    land::L
    global::G = nothing
end

function parameterization!(ij, vars, albedo::Albedo, model)
    parameterization!(ij, vars, albedo.ocean, model)
    parameterization!(ij, vars, albedo.land, model)
    parameterization!(ij, vars, albedo.global, model)
    return nothing
end

# define a constructor that returns Albedo though with ocean, land set to nothing
ManualAlbedo(...) = Albedo(ocean=nothing, land=land, global=ManualAlbedo(...))
```

There's still some ambiguity if `ManualAlbedo` doesn't know where it's supposed to be applied then it doesn't know which array to write into. But you could also distinguish that by type

```julia
abstract type AbstractOceanAlbedo <: AbstractAlbedo end
abstract type AbstractLandAlbedo <: AbstractAlbedo end
abstract type AbstractGlobalAlbedo <: AbstractAlbedo end

struct ConstantGlobalAlbedo{NF} <: AbstractGlobalAlbedo albedo::NF end
struct ConstantOceanAlbedo{NF} <: AbstractOceanAlbedo albedo::NF end
struct ConstantLandAlbedo{NF} <: AbstractLandAlbedo albedo::NF end

which_albedo(vars::Variables, ::AbtractOceanAlbedo) = vars.parameterization.ocean.albedo
which_albedo(vars::Variables, ::AbtractLandAlbedo) = vars.parameterization.land.albedo

function parameterization!(ij, vars, albedo::AbstractAlbedo, ...)
    albedo_field = which_albedo(vars, albedo)
    ...
```

And then you could build some function as outlined here to minimize code redundancy between them.

SpeedyWeather/ext/SpeedyWeatherLuxExt.jl
```diff
@@ -0,0 +1,541 @@
...
+    for ij in eachindex(model.land_sea_mask.mask) 
+        SpeedyWeather.albedo!(ij, vars, albedo.ocean, model)
+    end
+
+    SpeedyWeather.albedo!(vars, albedo.land, model)
+
+    # Blend ocean and land albedo
+    mask = model.land_sea_mask.mask
+    @. vars.parameterizations.albedo = mask * vars.parameterizations.land.albedo + 
+                                       (1 - mask) * vars.parameterizations.ocean.albedo
```
milankl, line 527: at the moment the shortwave radiation scheme does that but maybe it shouldn't!

SpeedyWeather/ext/SpeedyWeatherLuxExt.jl
```diff
@@ -0,0 +1,541 @@
...
+    @views @. X[1, :] = normalise(vegh, scheme.norm_means[1], scheme.norm_stds[1])
+    @views @. X[2, :] = normalise(vegl, scheme.norm_means[2], scheme.norm_stds[2])
+    @views @. X[3, :] = normalise(sm1, scheme.norm_means[3], scheme.norm_stds[3])
+    @views @. X[4, :] = normalise(st1, scheme.norm_means[4], scheme.norm_stds[4])
+    @views @. X[5, :] = normalise(sm2, scheme.norm_means[5], scheme.norm_stds[5])
+    @views @. X[6, :] = normalise(st2, scheme.norm_means[6], scheme.norm_stds[6])
+    @views @. X[7, :] = normalise(geopotential, scheme.norm_means[7], scheme.norm_stds[7])
+    @views @. X[8, :] = normalise(lai_hv, scheme.norm_means[8], scheme.norm_stds[8])
+    @views @. X[9, :] = normalise(lai_lv, scheme.norm_means[9], scheme.norm_stds[9])
+    @views @. X[10, :] = normalise(snow_cover, scheme.norm_means[10], scheme.norm_stds[10])
```
milankl, line 472: not sure how to do this better but this is non-contiguous memory access, in essence you need to read and write the array X 10x because `[i, :]` indexes only ever 10th element. Wondering whether we can pull this together into one kernel ....

milankl (commented):
More comments!

PR #1017, opened 2026-03-26
Add snow depth cap

gregrmunday:
It was found that the runoff melt rate implemented in the snow model in #833 caused issues when the PrimitiveWetModel was initialised with permanent ice caps and glaciers present. Large snow depth over perm snow --> large runoff rate --> huge latent heat sink for land temperature budget, so the soil went to absolute zero! As documented in https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1013#issuecomment-4119748116

A fix for this for the moment is to follow how IFS implement their permanent snow area and cap the equivalent liquid water depth at 10m - that's what we do here.

- [x] Update code
- [x] Update docs
- [x] Update changelog.md


gregrmunday:
I've added the snow depth cap as a parameter to the `struct` but not to the parameter scheme as I don't think it should be included there?

docs/src/land.md
```diff
@@ -282,24 +281,19 @@ this formulation allows to melt more snow than there actually is, so we need
...
-```
-
-in ``kg/m²/s``, where the `excess` term (negative for melting/runoff trying to remove more snow than there is)
+in ``kg/m²/s``, where the `excess` term (negative for melting trying to remove more snow than there is)
 only appears when the naive tendency would overdraw the bucket.
-`snow_melt_rate` therefore combines true melt with the runoff leak and is zero over ocean points.
-Snow depth is clipped to zero and stored as equivalent liquid water height, not physical snow thickness.
+`snow_melt_rate` is zero over ocean points. Snow depth is clipped to zero and stored 
+as equivalent liquid water height, not physical snow thickness. The accumulation is capped at 10m equivalent liquid water height, following how permanent snow area is treated 
+in IFS Cycle 49r1.
```
milankl: provide a reference here if you already cite other models?

milankl: (also try to stick to a line length of not more than 120 characters or so, this is not a requirement by runic but would be nice)

SpeedyWeather/src/physics/land/snow.jl
```diff
@@ -3,20 +3,19 @@ abstract type AbstractSnow <: AbstractLandComponent end
 export SnowModel    # maybe change for a more concise name later
 
 """
-    SnowModel(; melting_threshold=275, runoff_time_scale=Year(1))
+    SnowModel(; melting_threshold=275, snow_depth_cap=10)
```
milankl: use `($TYPEDSIGNATURES)` to have this filled in automatically

milankl (commented):
This looks great. One point is conservation of water / energy: The cap should violate that and that's fine but I want to avoid a situation where any additional meter of snow above 10m is forced to melt (cause melt water where it shouldn't melt). I believe the way you introduced this it's correct but could you double check with this in mind and add a comment or two in the code? 

For example we introduced `dsnow` to not melt more snow then there is (even though temperature would allow to melt more).

milankl:
I've increased transform test tolerances slightly as they kept failing (1 out of many)
```julia
  Expression: ≈(oro_spec1[lm], oro_spec2[lm], atol = tol, rtol = tol)
   Evaluated: 0.9082879f0 + 0.0f0im ≈ 0.90865266f0 + 0.0f0im (atol=0.00034526698, rtol=0.00034526698)

Stacktrace:
...
Test Summary:                       |   Pass  Fail   Total  Time
Transform: Orography (exact grids)  | 138511     1  138512  3.4s
  trunc = 31                        |  46112         46112  1.6s
  trunc = 42                        |  92399     1   92400  1.8s
    NF = Float64                    |  46200         46200  0.4s
    NF = Float32                    |  46199     1   46200  1.4s
      Grid = FullGaussianGrid       |   9138          9138  0.2s
      Grid = FullClenshawGrid       |  19185     1   19186  1.2s
      Grid = OctahedralGaussianGrid |   6194          6194  0.0s
      Grid = OctahedralClenshawGrid |  11682         11682  0.1s
ERROR: LoadError: Some tests did not pass: 138511 passed, 1 failed, 0 errored, 0 broken.
```

milankl:
~~Also can you throw in the figures that made you conclude that this is important to change? Greenland cooling down below absolute zero~~ updated your first comment above ☝🏼

docs/src/land.md
gregrmunday: Thanks Milan, both of these done in the latest commit!

gregrmunday: I've added a couple of explanatory comments in the latest, but otherwise I think it looks good re: not adding extra melt water which could lead to a blowup.

milankl:
I think the primitive dry model is still unstable when run for longer but that's not the fault of this PR

```julia
┌ Warning: NaN or Inf detected at time step 3605 (2000-04-10T04:00:00)
└ @ SpeedyWeather ~/work/SpeedyWeather.jl/SpeedyWeather.jl/SpeedyWeather/src/output/feedback.jl:123
run defaults for a year, no blow up: Test Failed at /home/runner/work/SpeedyWeather.jl/SpeedyWeather.jl/SpeedyWeather/test/dynamics/run_defaults.jl:7
  Expression: simulation.model.feedback.nans_detected == false
   Evaluated: true == false
```

milankl:
Skip submodules CI again. While there's a change to SpeedyTransforms this has been tested and passes (increase test tolerance).

milankl (approved):
Almost all checks are passing now those that don't has nothing to do with this PR. So I suggested to bypass and merge

PR #1019, opened 2026-03-27
Minor fixes for Metal

maximilian-gelbrecht:
Metal overall is still nonworking, but this makes e.g. the MatrixSpectralTransform work.

### DIFF
--- a/SpeedyTransforms/src/matrix_transform.jl
+++ b/SpeedyTransforms/src/matrix_transform.jl
@@ -62,9 +62,9 @@ function MatrixSpectralTransform(
     ArrayType_ = nonparametric_type(ArrayType)      # drop parameters of ArrayType
 
     # LATITUDE VECTORS (based on Gaussian, equi-angle or HEALPix latitudes)
-    latd = RingGrids.get_latd(grid)                         # latitude in degrees (90˚Nto -90˚N)
-    coslat = on_architecture(architecture, cosd.(latd))     # cos(lat)
-    coslat⁻¹ = on_architecture(architecture, inv.(coslat))  # 1/cos(lat)
+    latd = RingGrids.get_latd(grid)                                     # latitude in degrees (90˚Nto -90˚N)
+    coslat = on_architecture(architecture, NF.(cosd.(latd)))            # cos(lat)
+    coslat⁻¹ = on_architecture(architecture, NF.(inv.(cosd.(latd))))   # 1/cos(lat)
 
     # Create another SpectralTransform to calculate the transform matrices from (do this on the CPU)
     spectrum_cpu = on_architecture(CPU(), spectrum)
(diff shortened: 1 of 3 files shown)


SpeedyWeatherInternals/ext/SpeedyWeatherInternalsMetalExt.jl
```diff
@@ -36,6 +36,9 @@ Architectures.on_architecture(::MtlGPU, a::SubArray{<:Any, <:Any, <:MtlArray}) =
 Architectures.on_architecture(::MtlGPU, a::SubArray{<:Any, <:Any, <:Array}) = MtlArray(a)
 Architectures.on_architecture(::MtlGPU, a::StepRangeLen) = a
 
+# convert Float64 arrays to Float32 first for Metal as F64 isn't supported
+Architectures.on_architecture(::MtlGPU, a::Array{Float64}) = MtlArray(Float32.(a))
```
milankl, line 40: That's a silent conversion, I'd be happy with that, just maybe good to know or add that information at some point to the docs?

SpeedyTransforms/src/matrix_transform.jl
```diff
@@ -62,9 +62,9 @@ function MatrixSpectralTransform(
     ArrayType_ = nonparametric_type(ArrayType)      # drop parameters of ArrayType
 
     # LATITUDE VECTORS (based on Gaussian, equi-angle or HEALPix latitudes)
-    latd = RingGrids.get_latd(grid)                         # latitude in degrees (90˚Nto -90˚N)
-    coslat = on_architecture(architecture, cosd.(latd))     # cos(lat)
-    coslat⁻¹ = on_architecture(architecture, inv.(coslat))  # 1/cos(lat)
+    latd = RingGrids.get_latd(grid)                                     # latitude in degrees (90˚Nto -90˚N)
+    coslat = on_architecture(architecture, NF.(cosd.(latd)))            # cos(lat)
+    coslat⁻¹ = on_architecture(architecture, NF.(inv.(cosd.(latd))))   # 1/cos(lat)
```
milankl, line 67: Isn't that redundant with the F64->F32 conversion you added below?

maximilian-gelbrecht: I figured that's actually good to have across all architectures to be sure they are the same number type, as this can cause problems

SpeedyWeatherInternals/ext/SpeedyWeatherInternalsMetalExt.jl
maximilian-gelbrecht: Yeah, true could add an `@info`, but also note that before you just got a straight error here. `MtlArray{Float64}` don't exist.

milankl: Happy with this, just good to know

PR #1025, opened 2026-04-02
Rename prognostic variables

milankl:
Renames the prognostic variables to

```julia
julia> simulation.variables
├ prognostic (421.80 KB)
│ ├ clock: Clock{DateTime, Second, Int64, Millisecond}
│ ├ vorticity: 560×8×2 LowerTriangularArray{ComplexF32, 3, Array{...}}
│ ├ divergence: 560×8×2 LowerTriangularArray{ComplexF32, 3, Array{...}}
│ ├ temperature: 560×8×2 LowerTriangularArray{ComplexF32, 3, Array{...}}
│ ├ pressure: 560×2 LowerTriangularArray{ComplexF32, 2, Array{...}}
│ ├ humidity: 560×8×2 LowerTriangularArray{ComplexF32, 3, Array{...}}
```

Following the logic to name this as you speak them. While relative vorticity is _relative_ vorticity, we say "vorticity" as is in most cases unless we need to distinguish it from planetary, potential etc. Pressure is actually the logarithm of surface pressure, so I was debating `log_pressure` here, I'd be happy with that too. Specific humidity drops "specific". Open for any other suggestions

Also includes the changes Navid suggested in #1010 manually added via https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1025/commits/8f37c620f3c3f2ea99f1051add6f87bb81ed5330:
- renamed `pressure_reference` to `reference_pressure
- renamed `temperature_reference` to `temperature_reference`

Also

- SpeedyWeather with 1.12 instead of 1.11, 1.10 LTS remains tested
- Documentation is build with 1.12 instead of 1.11


milankl:
So I've decided against `log_pressure` and prefer simply `pressure` instead. The reason is that a) we also don't specify `relative_vorticity` for example and more importantly b) whether it's the logarithm or not feels like a question of units. We also don't say `temperature_in_kelvin` for example. So I'd drop it but provide comment to denote that this is actually the logarithm of surface pressure. Feel free to disagree

Issue #988, opened 2026-02-13
ML-compatible output / array output


maximilian-gelbrecht:
Progressing with Reactant, I am thinking a bit about ML tasks. 

How do we handle model output best for ML tasks? We want more than the final state when training or calibrating a model. 

The `NetCDFOutput` doesn't work with Reactant (yet), and might also just not what we want in these situations. We don't want to archive the results, we want to use them for the training / calibration itself, and then we most likely don't need them anymore. 

The `JLDOutput` already exists and could definitely be useful, but does this even work for an ML task? Saving to disk and then loading again? Intuitively I would say no. 

So I think we need and want an additional `ArrayOutput` that keeps the output in RAM. This could be quite useful and efficient for small problems (low resolutions and/or short time series), if we take a little care it will also work with AD as well. 

I do wonder though how we handle this in the best way when actually calibrating/training very memory intensive problems. Would it work to compute the accumulate the objective function on the fly (e.g. as a callback) and differentiate it? We should test this. Is the current callback system already enough for this? Do we need something else? 

What are your thoughts?

gregrmunday:
Hi Max, I would be pretty keen to work with you on this, as it directly relates to my upcoming research. I'd be learning along the way as I haven't used Enzyme or done much differentiable programming before, but am up for the challenge! It would probably be good to have an intro to how we've integrated Enzyme into SpeedyWeather, if either you or @milankl  have some spare time to help me get up to speed a bit...

maximilian-gelbrecht:
Great! 

I think they are two separate things we could implement / look into directly 

1. Implementing a `ArrayOutput` that just outputs into an array that's hold in RAM. I think that's really easy to do, as it it can just reuse some of the logic from the existing `JLD2Output` 
2. Testing if output callbacks work with Enzyme. There are the existing Enzyme script where we investigate and test differentiability. We could just modify them to test if you can (hopefully) also differentiate through the callback logic. If not, we have to adjust it so that it works 

I'll also write you on Slack, happy to set up a meeting about what we did so far with Enzyme (and Reactant).

milankl:
This does sound good to me. I guess most of the output infrastructure can be reused, like all the variables. Let me know if you need any input from me here

maximilian-gelbrecht:
I had a brief thought about this again. Now I would actually just wait for #969, then we revise the current `JLD2Output` together with a new `ArrayOutput`. With #969 it will be much easier to actually select which variables to save.

PR #1026, opened 2026-04-03
`JLD2Output` revised, `ArrayOutput` introduced

maximilian-gelbrecht:
This partially addresses #988 (cc @gregrmunday) 

Based on the new `Variables` it extends the current `JLD2Output` so that we can select which groups (prognostic, grid, tendencies, ...) to output. It also add a new `ArrayOutput` that does basically the same as `JLD2Output` just that it saves everythign to RAM. Both share a part of the implementation through a new `AbstractVariablesOutput <: AbstractOutput` type.

I did this very quickly with Claude, so this was not much effort, but also just a draft. 

Now added an additional thing, so that `JLD2Output` is actually GPU compatible (it wasn't before). For that, we can now also use `on_architecture(device, ::Variables)`. To quickly transfer all variables between devices. 

Documentation is still missing, will do that once everything else is clear.


SpeedyWeather/test/variables.jl
```diff
@@ -1,3 +1,24 @@
...
+    @test parent(vars_gpu.prognostic.vor) isa JLArray
+
+    # round-trip back to CPU should reproduce original values
+    vars_back = on_architecture(SpeedyWeather.CPU(), vars_gpu)
+    @test Array(parent(vars_back.prognostic.vor)) == Array(parent(vars_cpu.prognostic.vor))
+    @test Array(parent(vars_back.prognostic.temp)) == Array(parent(vars_cpu.prognostic.temp))
+end
+
+#=
 @testset "Variables initialize, zero, fill!, one" begin
```
maximilian-gelbrecht: @milankl This file was already here from #969, but probably by mistake, right? Because we didn't end up implementing those routines.

SpeedyWeather/src/output/variables_output.jl
maximilian-gelbrecht, line 1: Most parts of this are just the old JLD2Output just moved to a new file.

SpeedyWeather/src/output/variables_output.jl
```diff
@@ -0,0 +1,311 @@
+export JLD2Output, ArrayOutput
+
+"""Abstract supertype for output writers that save `Variables` (or subsets thereof)
+at regular intervals. Subtypes must have fields: `active`, `path`, `run_prefix`, `id`,
+`run_number`, `run_digits`, `run_folder`, `run_path`, `overwrite`, `write_restart`,
+`write_parameters_txt`, `write_progress_txt`, `output_dt`, `output_every_n_steps`,
+`timestep_counter`, `output_counter`, `groups`."""
```
milankl: I've been wondering whether we should just introduce a `OutputWriterCore` type that holds and determines this information that's independent of the output format or variables

SpeedyWeather/src/output/variables_output.jl
```diff
@@ -0,0 +1,311 @@
...
+abstract type AbstractVariablesOutput <: AbstractOutput end
+
+# GROUPS FILTERING
+
+const ALL_VARIABLE_GROUPS = (:prognostic, :grid, :tendencies, :dynamics, :parameterizations, :particles, :scratch)
```
milankl: Move to `src/variables/variables.jl` ? Seems to be odd to define this here whereas it's actually already hardcoded in that file

SpeedyWeather/src/output/variables_output.jl
```diff
@@ -0,0 +1,311 @@
...
+    selected = Tuple(g for g in ALL_VARIABLE_GROUPS if g in groups)
+    return NamedTuple{selected}(Tuple(getfield(vars, g) for g in selected))
+end
+
+# SHARED INITIALIZE
+
+"""$(TYPEDSIGNATURES)
+Shared initialization for `AbstractVariablesOutput` subtypes: determine run folder,
+compute output frequency, reset counters, add standard callbacks."""
+function initialize_variables_output!(
```
milankl: I think all of this should be in `initialize!(output::OutputWriterCore, model)` ? Otherwise it's a duplication of the essentially same code for NetCDFOutput no?

SpeedyWeather/src/output/variables_output.jl
```diff
@@ -0,0 +1,311 @@
...
+
+    return true
+end
+
+# SHARED OUTPUT STEPPING
+
+"""$(TYPEDSIGNATURES)
+Shared output stepping logic: increment timestep counter and check whether output
+should be written on this step. Returns `true` if output should proceed."""
+function should_output!(output::AbstractVariablesOutput)
```
milankl: Same

SpeedyWeather/test/variables.jl
milankl: yeah feel free to delete if not needed, I may have commented them out and then forgot about it

milankl (commented):
Looks good to me! For any future output writer e.g. `ZarrOutputWriter` I propose to introduce an `OutputWriterCore` see below that holds the information about output frequency and paths. That way this can all be shared between all output writers?

maximilian-gelbrecht:
> Looks good to me! For any future output writer e.g. `ZarrOutputWriter` I propose to introduce an `OutputWriterCore` see below that holds the information about output frequency and paths. That way this can all be shared between all output writers?

Yeah, that's a good idea, I like that. I'll try to implement that already here.

maximilian-gelbrecht:
Ok, I introduced the `OutputWriterCore` now.

SpeedyWeather/src/output/variables_output.jl
maximilian-gelbrecht: I've done that now, but in order to be easily configurable for the user we still need all `[OPTION]` fields still in the main structs (and those are repeated). The `OutputWriterCore` basically only holds all derived variables that are needed for the output logic. At least that's the only option I saw so far.

SpeedyWeather/src/output/variables_output.jl
maximilian-gelbrecht: Yep done that now.

SpeedyWeather/src/output/variables_output.jl
maximilian-gelbrecht: Ok moved it there. btw: `fieldnames(Variables)` isn't type stable according to `code_warntype` so that's why that was there.

SpeedyWeather/test/variables.jl
maximilian-gelbrecht: ok deleted now

SpeedyWeather/src/output/variables_output.jl
milankl: Yeah, that's good! I guess we could also encourage a `set!(model.output, frequency = Day(1))` usage and break with the `model.output.output_dt = Day(1)` syntax. It's breaking I agree but if we generally move to a user interface via the `set!` (or some other method) then we can change the paths within it without making it a breaking change to the user?

maximilian-gelbrecht: I don't like it. It would make things more complicated during model setup. 

```Julia 
spectral_grid = SpectralGrid(nlayers = 1)
output = NetCDFOutput(spectral_grid, output_dt = Hour(1)) 
model = PrimitiveWetModel(output) 
``` 
just is much more intuitive compared to 

```Julia 
spectral_grid = SpectralGrid(nlayers = 1)
output = NetCDFOutput(spectral_grid) 
set!(output, output_dt = Hour(1))
model = PrimitiveWetModel(output) 
```

SpeedyWeather/src/output/variables_output.jl
milankl: Isn't that because `Variables` itself is type unstable in a sense that it's an abstract super type for all `Variables{A, B, C ...}` ?

Define a parametric type
```julia
julia> struct S{A}
           a::A
       end
```

Now `fieldnames(S)` returns red `::Any` 

```julia
julia> @code_warntype(fieldnames(S))
MethodInstance for fieldnames(::Type{S})
  from fieldnames(t::UnionAll) @ Base runtime_internals.jl:338
Arguments
  #self#::Core.Const(fieldnames)
  t::Type{S}
Body::Any
1 ─ %1 = Base.fieldnames::Core.Const(fieldnames)
│   %2 = Base.unwrap_unionall::Core.Const(Base.unwrap_unionall)
│   %3 = (%2)(t)::Any
│   %4 = (%1)(%3)::Any
└──      return %4
```

But `S{Int}` doesn't

```julia
julia> @code_warntype(fieldnames(S{Int}))
MethodInstance for fieldnames(::Type{S{Int64}})
  from fieldnames(t::DataType) @ Base runtime_internals.jl:336
Arguments
  #self#::Core.Const(fieldnames)
  t::Core.Const(S{Int64})
Body::Tuple{Symbol}
1 ─ %1  = Base.fieldcount::Core.Const(fieldcount)
│         (%1)(t)
│   %3  = Base._fieldnames::Core.Const(Base._fieldnames)
│   %4  = (%3)(t)::Core.Const(svec(:a))
│   %5  = Core._apply_iterate(Base.iterate, Core.tuple, %4)::Core.Const((:a,))
│   %6  = Base.Tuple::Core.Const(Tuple)
│   %7  = Base.Vararg::Core.Const(Vararg)
│   %8  = Base.Symbol::Core.Const(Symbol)
│   %9  = Core.apply_type(%7, %8)::Core.Const(Vararg{Symbol})
│   %10 = Core.apply_type(%6, %9)::Core.Const(Tuple{Vararg{Symbol}})
│   %11 = Core.typeassert(%5, %10)::Core.Const((:a,))
└──       return %11
```

but maybe I'm misunderstanding this?

maximilian-gelbrecht: Ah yeah that makes sense, but it's also results in the odd thing that here we could replace this with just dummy types, so doing `fieldnames(Variables{Int32, Int32, Int32, Int32, Int32, Int32, Int32})` would actually work even though it does look like nonsense. I am not even sure I'd want that abomination in our code ;)

SpeedyWeather/src/output/variables_output.jl
milankl: But this we can still do by declaring some keyword arguments that are passed from the generator function to the actual construction like

```julia
function NetCDFOutput(::SpectralGrid;
    # core key word arguments
    filename = "output.nc",
    frequency = Day(1),
    # other kwargs
    kwargs...)

    core = OutputWriterCore(filename, frequency)
    return NetCDFOutput(core, kwargs...)
end
```

milankl: But what it would break is `model.output.output_dt = Day(1)` as it turns that into `model.output.core.output_dt = Day(1)` ... for which one then would need a "core"-independent interface, e.g. the `set!` function no?

maximilian-gelbrecht: For that we could just define a `Base.setproperty!(output::AbstractOutput, s::Symbol, v)`. I already do that anyway here for `ArrayOutput`

maximilian-gelbrecht:
Okay merged main in here with the change from #1025. Locally tests pass. From my side this could be merged.

PR #1027, opened 2026-04-03
RossbyHaurwitz wave example without forcing or drag

milankl:
The Rossby Haurwitz wave example initial conditions in the docs

image

don't actually maintain shape while propagating longitudinally at the moment:

image

Likely because we haven't actually switched off forcing and drag in this example

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


milankl:
Okay works like a charm again

image

Will merge as it's essentially just a followup from #969

Issue #919, opened 2025-11-18
SpeedyWeather becomes not-so-speedy when FFTW multithreading is enabled


bgroenks96:
I recently discovered that SpeedyWeather becomes slow when Oceananigans is loaded in a Julia session with multiple threads. This is because Oceananigans [sets the number of FFTW threads](https://github.com/CliMA/Oceananigans.jl/issues/1113) in multi-threaded mode. It's not clear yet why exactly this is happening, but probably there is some bad interaction between Speedy's spectral transforms and the internal FFTW multi-threading.

Steps to reproduce:

1. Start Julia with `--threads=auto` (assuming you aren't running Speedy on a Pentium II single-core processor or something...)
2. Run the following example

```julia
using SpeedyWeather

spectral_grid = SpectralGrid(trunc=31)
primitive_dry = PrimitiveDryModel(spectral_grid)
sim = initialize!(primitive_dry)
run!(sim) # everything is speedy
```
Everything should run normally since FFTW does not multithread by default:
```
Weather is speedy: 100%|████████████████████████████████████████████████████████| Time: 0:00:01 (1479.70 years/day)
```
3. Call `FFTW.set_num_threads(Base.Threads.nthreads())`
4. Re-run the above code and watch Speedy impersonate a turtle

```
Weather is (not so) speedy: 100%|████████████████████████████████████████████████████████| Time: 0:00:14 (166.00 years/day)
```

This is verified on both  Julia v1.10 and v1.11.

```
Julia Version 1.11.7
Commit f2b3dbda30a (2025-09-08 12:10 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 12 × 13th Gen Intel(R) Core(TM) i7-1365U
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, goldmont)
Threads: 12 default, 0 interactive, 6 GC (on 12 virtual cores)
```

bgroenks96:
I am marking this as a bug for now since there is no obvious reason why Speedy should become so much slower when FFTW uses multi-threading.

maximilian-gelbrecht:
Thanks! We have to break this down a bit further. This is weird and counterintuitive. 

I lack the time right now, but I would start by really testing and benchmarking this on the transforms and `_fourier!` directly.

maximilian-gelbrecht:
[...]

PR #1032, opened 2026-04-08
Multithread CPU parametrizations

maximilian-gelbrecht:
Some quick benchmark just for the `parametrization_tendencies!`:

```
threads=1  time=9.076 ms  allocs=91  memory=65.28 KiB
threads=2  time=4.963 ms  allocs=156  memory=129.95 KiB
threads=3  time=3.480 ms  allocs=221  memory=194.42 KiB
threads=4  time=2.896 ms  allocs=286  memory=259.09 KiB
```


maximilian-gelbrecht:
For the full model we would also need to multithread the transform, so `legendre!`. Which should also be fairly simple.

maximilian-gelbrecht:
@milankl Did we actually had a proper solution in the end for this multithreading problem when coupling to Oceananigans? 

The only thing I somewhat worry about with multithreading loops is if there's some interference when using other multithreaded code at the same time.

milankl:
I can't find the issue about this anymore but @bgroenks96 discovered this. It was something about FFTW though not sure this still applies ...

bgroenks96:
It was #919 and the issue was due to Oceananigans and Speedy both relying on the global FFTW `num_threads`.

maximilian-gelbrecht:
The Enzyme CI keeps failing here. I don't see how this is connected to the multithreading, but this is the only thing that changed here really 🤷

maximilian-gelbrecht:
Tried it again, multithreading still seems to cause an error with Enzyme here. I won't invest any time into this in the near future, just occasionally checking if this gets fixed upstream by any chance. A very naive MWE with threads did work, so this isn't something super obvious.

PR #1035, opened 2026-04-08
NCycleLorenz time stepping part 2

### LINKED ISSUE #1069: Unalias copy missing for Field
This works for LowerTriangularArrays: Create a 2-step 2D variable, say old = first step, new = 2nd step then copy from one to another

```julia
julia> coeff = rand(LowerTriangularArray, 3, 3, 2)
6×2 LowerTriangularArray{Float64, 2, Array{...}}
 0.6016016468923711    0.8502382968032102
 0.6534794370647571    0.9343438183931021
 0.047221672535886716  0.7006885991148764
 0.5748757134660201    0.44718501418541856
 0.6079247435685476    0.27108254478408567
 0.430220145623176     0.8609971000540944

julia> old, new = SpeedyWeather.get_steps(coeff);

julia> old .= new
6-element, 3x3 LowerTriangularMatrix{Float64, Array{...}}
 0.850238  0.0       0.0
 0.934344  0.447185  0.0
 0.700689  0.271083  0.860997

julia> coeff
6×2 LowerTriangularArray{Float64, 2, Array{...}}
 0.8502382968032102   0.8502382968032102
 0.9343438183931021   0.9343438183931021
 0.7006885991148764   0.7006885991148764
 0.44718501418541856  0.44718501418541856
 0.27108254478408567  0.27108254478408567
 0.8609971000540944   0.8609971000540944
```

But it doesn't work with `Field`

```julia
julia> field = rand(HEALPixField, 2, 2);

julia> old, new = SpeedyWeather.get_steps(field);

julia> old .= new
ERROR: ArgumentError: an array of type `Field` shares memory with another argument and must make a preventative copy of itself in order to maintain consistent semantics, but `copy(::Field{Float64, 1, SubArray{Float64, 1, Matrix{Float64}, Tuple{Base.Slice{Base.OneTo{Int64}}, Int64}, true}, HEALPixGrid{CPU{KernelAbstractions.CPU}, Vector{UnitRange{Int64}}, Vector{Int64}, Int64}})` returns a new array of type `Field{Float64, 1, Vector{Float64}, HEALPixGrid{CPU{KernelAbstractions.CPU}, Vector{UnitRange{Int64}}, Vector{Int64}, Int64}}`.
To fix, implement:
`Base.unaliascopy(A::Field)::typeof(A)`
Stacktrace:
 [1] _unaliascopy(A::Field{Float64, 1, SubArray{…}, HEALPixGrid{…}}, C::Field{Float64, 1, Vector{…}, HEALPixGrid{…}})
   @ Base ./abstractarray.jl:1551
 [2] unaliascopy(A::Field{Float64, 1, SubArray{…}, HEALPixGrid{…}})
   @ Base ./abstractarray.jl:1547
 [3] unalias
   @ ./abstractarray.jl:1530 [inlined]
 [4] copyto!
   @ ./abstractarray.jl:1069 [inlined]
 [5] copyto!
   @ ./broadcast.jl:988 [inlined]
 [6] copyto!
   @ ./broadcast.jl:947 [inlined]
 [7] materialize!
   @ ./broadcast.jl:905 [inlined]
 [8] materialize!(dest::Field{…}, bc::Base.Broadcast.Broadcasted{…})
   @ Base.Broadcast ./broadcast.jl:902
 [9] top-level scope
   @ REPL[40]:1
```

This is needed for #1035

milankl:
Follow-up from #937. Changes

1. Time steppers decide the number of prognostic/tendency steps

E.g. with `Leapfrog` we would have 2 prognostic steps in spectral but else 1

```julia
julia> simulation.variables
Variables{@NamedTuple{...}, ...} (158.00 KB)
...
│ └ vorticity: 560×1×2 LowerTriangularArray{ComplexF32, 3, Array{...}}
├ grid (64.50 KB)
│ ├ vorticity: 3168×1×1, 48-ring OctahedralGaussianField{Float32, 3} as Arr...
...
├ tendencies (79.24 KB)
│ ├ vorticity: 560×1×1 LowerTriangularArray{ComplexF32, 3, Array{...}}
│ └ grid
│   ├ vorticity: 3168×1×1, 48-ring OctahedralGaussianField{Float32, 3} as A...
...
```

but with `NCycleLorenz` we now have 2 spectral tendency steps else 1

```julia
julia> simulation.variables
Variables{@NamedTuple{...}, ...} (158.00 KB)
...
│ └ vorticity: 560×1×1 LowerTriangularArray{ComplexF32, 3, Array{...}}
├ grid (64.50 KB)
│ ├ vorticity: 3168×1×1, 48-ring OctahedralGaussianField{Float32, 3} as Arr...
...
├ tendencies (83.72 KB)
│ ├ vorticity: 560×1×2 LowerTriangularArray{ComplexF32, 3, Array{...}}
│ └ grid
│   ├ vorticity: 3168×1×1, 48-ring OctahedralGaussianField{Float32, 3} as A...
```

These are determined by extending

```julia
prognostic_steps(TS::AbstractTimeStepper, ::AbstractModel)
prognostic_grid_steps(TS::AbstractTimeStepper, ::AbstractModel)
prognostic_spectral_steps(TS::AbstractTimeStepper, ::AbstractModel)

tendency_steps(TS::AbstractTimeStepper, ::AbstractModel)
tendency_grid_steps(TS::AbstractTimeStepper, ::AbstractModel)
tendency_spectral_steps(TS::AbstractTimeStepper, ::AbstractModel)
```

such that both time stepper and model can decide what's needed. Consequence: All prognostic+grid+tendencies have now a mandatory "step" dimension that in many cases is just a singleton dimension.

2. Time steppers decide which step to use per variable and component

Throughout the code we now have lines like

```julia
    humid = get_prognostic_step(vars.grid.humidity, time_stepping, condensation)                    # specific humidity [kg/kg]
    temp_tend = get_tendency_step(vars.tendencies.grid.temperature, time_stepping, condensation)    # temperature tendency [K/s]
    tracer_grid = get_prognostic_step(vars.grid.tracers[name], time_stepping, S, model)
```

such that `get_prognostic_step` and `get_tendency_step` create a view selecting the desired step index depending on the `time_stepping`, the model component `condensation` and, optionally, the `model` itself.

For the time steppers this is implemented via dispatch over `which_prognostic_step` and `which_tendency_step`, e.g.

```julia
@inline which_prognostic_step(var, ::AbstractLeapfrog, ::AbstractDynamicalCoreComponent) = 2
@inline which_prognostic_step(var::AbstractField, ::AbstractLeapfrog, ::AbstractDynamicalCoreComponent, ::TwoDModels) = 1
```

neat: these are compile time decisions so the actual code will have a compiled `view(var, :, 1)` or `view(var, :, 2)` in it.

3. Optional reinitialize model components

Models can now decide whether some of its components should be reinitialized, e.g. the primitive models do

```julia
function reinitialize!(model::PrimitiveDryModel, vars::AbstractVariables)
    reinitialize!(model.implicit, model, vars)
    return nothing
end
```

to be able to update the time step for the implicit solver whenever that changes. Fallback for all model components is to do nothing but e.g. the implicit solver implements

```julia
function reinitialize!(
        implicit::ImplicitPrimitiveEquation,
        model::PrimitiveEquation,
        vars::Variables,
    )
    (; time_stepping, geometry, geopotential, atmosphere, adiabatic_conversion) = model
    Δt = time_step(time_stepping, vars.prognostic.clock) 
    implicit.Δt[] == Δt && return nothing                   # if time step has not changed no need to reinitialize
    ...
end
```

to check whether reinitialization is necessary (and do it if that's the case)

4. The Barotropic model uses NCycleLorenz now by default

Others still use Leapfrog for stability reasons, it's not clear to me yet exactly how much shorter the NCycleLorenz time step should be depending on model and complexity (shallow water, dycore, dry core + parameterizations, wet). Will be added in follow-up PR.


RingGrids/Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "RingGrids"
 uuid = "d1845624-ad4f-453b-8ff4-a8db365bf3a7"
-version = "0.1.5"
+version = "0.1.5+DEV"
```
maximilian-gelbrecht: I am still a bit confused so how it this work now with the `+DEV`? It's just basically a marker for us for the current `main` version, and then we change it to e.g. `0.1.6` here when releasing the next version? So also basically every time we update a submodule after a release we add a `+DEV` now?

SpeedyWeather/src/dynamics/drag.jl
```diff
@@ -106,23 +103,23 @@ function drag!(
...
-        Fu, Fv, u, v, c
+        architecture(Fu), LinearWorkOrder, (size(Fu, 1),), quadratic_drag_kernel!,
+        Fu, Fv, u, v, c, k
     )
     return nothing
 end
 
 @kernel inbounds = true function quadratic_drag_kernel!(
-        Fu, Fv, u, v, c
+        Fu, Fv, u, v, c, k
```
maximilian-gelbrecht, line 114: what's actually the reason for this change to hand over the index instead of the view of the field? Is this universal for all kernels with the new `get_tendency_step` and `get_prognostic_step`

maximilian-gelbrecht (commented):
Just some initial comments, I will continue later

SpeedyWeather/src/dynamics/horizontal_diffusion.jl
```diff
@@ -160,9 +156,9 @@ end
         ∇²ⁿ[l_plus_1, k] = -eigenvalue_norm^power / time_scale
         ∇²ⁿ_div[l_plus_1, k] = -eigenvalue_norm^p / time_scale_div
 
-        # and implicit part of the diffusion (= 1/(1-2Δtν∇²ⁿ))
-        ∇²ⁿ_implicit[l_plus_1, k] = 1 / (1 - 2Δt * ∇²ⁿ[l_plus_1, k])
-        ∇²ⁿ_div_implicit[l_plus_1, k] = 1 / (1 - 2Δt * ∇²ⁿ_div[l_plus_1, k])
+        # and implicit part of the diffusion (= 1/(1-Δtν∇²ⁿ))
+        ∇²ⁿ_implicit[l_plus_1, k] = 1 / (1 - Δt * ∇²ⁿ[l_plus_1, k])
```
maximilian-gelbrecht, line 160: Is this connected to the typical halved timestep of the timestepper as a whole or independent?

SpeedyWeather/src/dynamics/tendencies.jl
```diff
@@ -47,43 +44,38 @@ end
 Calculate all tendencies for the PrimitiveEquation model (wet or dry)."""
 function dynamics_tendencies!(
         vars::Variables,
-        lf::Integer,                # leapfrog index for tendencies
```
maximilian-gelbrecht, line 50: Nice co-benefit to take the `lf` out of the function arguments. Makes it cleaner and easier I think, even if the time stepper is now an argument as well!

SpeedyWeather/src/dynamics/tendencies.jl
```diff
@@ -443,19 +440,21 @@ function vordiv_tendencies!(
...
-    (; u, v) = vars.grid                                    # velocities, vorticity, temperature
-    vor = vars.grid.vorticity
-    temp = vars.grid.temperature
+    u_tend_grid = get_tendency_step(vars.tendencies.grid.u, time_stepping, DynamicalCore())
+    v_tend_grid = get_tendency_step(vars.tendencies.grid.v, time_stepping, DynamicalCore())
+    u = get_prognostic_step(vars.grid.u, time_stepping, DynamicalCore())
+    v = get_prognostic_step(vars.grid.v, time_stepping, DynamicalCore())
+    vor = get_prognostic_step(vars.grid.vorticity, time_stepping, DynamicalCore())
+    temp = get_prognostic_step(vars.grid.temperature, time_stepping, DynamicalCore())
     vars.scratch.grid.a .= 0
```
maximilian-gelbrecht: Technically, this only needs to be restated for the dry model, right?

SpeedyWeather/src/models/simulation.jl
```diff
@@ -71,31 +71,21 @@ function initialize!(
...
-
-    # transform variables from spectral to grid (= set the diagnostic variables in the correct initial state)
-    transform!(variables, lf, model, initialize = true)
+    
+    # TRANSFORM variables from spectral to grid (= set the diagnostic variables in the correct initial state)
+    transform!(variables, model, initialize = true)
     haskey(progn, :particles) && initialize!(variables, progn.particles, model)     # initialize particle work arrays
+    
+    # IMPLICIT initialization now in case that depends on the variables (e.g. a background temperature profile)
+    initialize!(model.implicit, variables, model)
```
maximilian-gelbrecht: before this was part of `first_timesteps!` right? This is not there anymore but now only here? Makes sense to me

SpeedyWeather/src/output/feedback.jl
```diff
@@ -83,24 +85,21 @@ function initialize!(feedback::Feedback, clock::Clock, model::AbstractModel)
     return nothing
 end
 
-# fallback if feedback is set to nothing
-initialize!(::Nothing, clock::Clock, model::AbstractModel) = nothing
```
maximilian-gelbrecht, line 87: We needed this for Reactant compatibility, or is this replaced somewhere else?

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
```diff
@@ -97,14 +98,15 @@ end
...
-    surface = size(vars.grid.temperature_prev, 2)     # surface index = nlayers
+    temp = get_prognostic_step(vars.grid.temperature, time_stepping, drag)
+    NF = eltype(temp)
+    surface = size(temp, 2)     # surface index = nlayers
 
     Vₛ = vars.parameterizations.surface_wind_speed[ij]
-    T = vars.grid.temperature_prev[ij, surface]
-    q = haskey(vars.grid, :humidity_prev) ? vars.grid.humidity_prev[ij, surface] : zero(NF)
+    T = temp[ij, surface]
+    q = haskey(vars.grid, :humidity) ? get_prognostic_step(vars.grid.humidity, time_stepping, drag)[ij, surface] : zero(NF)
```
maximilian-gelbrecht, line 140: Why is this actually changed from using the previous time step to the current?

SpeedyWeather/src/parameterizations/parameterizations.jl
```diff
@@ -5,6 +5,8 @@ order to use the parameterization in a model, add it to the `parameterizations`
 at definition."""
 abstract type AbstractParameterization <: AbstractModelComponent end
 
+struct DummyParameterization <: AbstractParameterization end
```
maximilian-gelbrecht: That's probably just here from some debugging?

SpeedyWeather/src/parameterizations/vertical_diffusion.jl
```diff
@@ -266,16 +269,18 @@ For vertical stability in the boundary layer."""
...
     Φ = vars.grid.geopotential
-    T = vars.grid.temperature_prev
+    T = get_prognostic_step(vars.grid.temperature, time_stepping, diffusion)
 
     # for dry models, use scratch array to bypass access to non-existing humidity variable
     for k in 1:nlayers
         vars.scratch.grid.b[ij, k] = 0
     end   # reset to zero humidity
-    q = haskey(vars.grid, :humidity_prev) ? vars.grid.humidity_prev : vars.scratch.grid.b
+    q = haskey(vars.grid, :humidity) ?
```
maximilian-gelbrecht, line 281: here as well, why was is previously the previous time step and now not anymore?

SpeedyWeather/src/parameterizations/vertical_diffusion.jl
```diff
@@ -266,16 +269,18 @@ For vertical stability in the boundary layer."""
...
-    v = vars.grid.v_prev
+    u = get_prognostic_step(vars.grid.u, time_stepping, diffusion)
+    v = get_prognostic_step(vars.grid.v, time_stepping, diffusion)
     Φ = vars.grid.geopotential
-    T = vars.grid.temperature_prev
+    T = get_prognostic_step(vars.grid.temperature, time_stepping, diffusion)
 
     # for dry models, use scratch array to bypass access to non-existing humidity variable
     for k in 1:nlayers
         vars.scratch.grid.b[ij, k] = 0
```
maximilian-gelbrecht, line 279: It's not really necessary for the wet model, or is it?

SpeedyWeather/src/variables/dimensions.jl
```diff
@@ -14,24 +14,30 @@ struct Grid2D <: AbstractVariableDim end
 Base.zero(::AbstractVariable{Grid2D}, model::AbstractModel) = zeros(model.spectral_grid.GridVariable2D, model.spectral_grid.grid)
 
 """Dimension for 3D grid variables on the horizontal grid with vertical levels."""
-struct Grid3D <: AbstractVariableDim end
-Base.zero(::AbstractVariable{Grid3D}, model::AbstractModel) = zeros(model.spectral_grid.GridVariable3D, model.spectral_grid.grid, get_nlayers(model))
+@kwdef struct Grid3D <: AbstractVariableDim
+    n::Int = 0                                                  # length of 3rd dimension, use nlayers for 0
```
maximilian-gelbrecht: The comments are the wrong way around "use 0 for nlayers"

SpeedyWeather/src/output/feedback.jl
maximilian-gelbrecht: Ah, this is now done in `abstract_models.jl` by 
```
initialize!(::Nothing, ::AbstractVariables, ::AbstractModel) = nothing
```
correct?

SpeedyWeather/src/time_stepping/implicit/implicit_primitive_equations.jl
```diff
@@ -459,4 +349,4 @@ end
         pres_correction += ξ * W[k] * div_tend[lm, k]
     end
     pres_tend[lm] += pres_correction
-end
+end
```
maximilian-gelbrecht, line 344: For the primitive equation model we would need a NCycle implicit scheme here, but this will be done in a follow up PR, correct?

SpeedyWeather/src/time_stepping/implicit/implicit_shallow_water.jl
```diff
@@ -0,0 +1,160 @@
...
+    # GPU kernel launch
+    arch = architecture(F_div)
+    launch!(
+        arch, SpectralWorkOrder, size(F_div), implicit_ncycle_lorenz_shallow_water_kernel!,
+        F_div, G_div, F_η, G_η, l_indices, H, g, ξ, w
+    )
+    return nothing
+end
+
+@kernel inbounds = true function implicit_ncycle_lorenz_shallow_water_kernel!(
```
maximilian-gelbrecht, line 134: Could you just point to where this is outlined/defined in the paper?

SpeedyWeather/src/time_stepping/steppers/leapfrog.jl
```diff
@@ -0,0 +1,335 @@
...
+    Δt_millisec::MS
+
+    "[DERIVED] Time step Δt [s] at specified resolution"
+    Δt_sec::NF
+
+    "[DERIVED] Time step Δt [s/m] at specified resolution, scaled by 1/radius"
+    Δt::NF
+end
+
+Adapt.adapt_structure(to, L::Leapfrog) = LeapfrogCore(L.Δt_millisec, L.Δt_sec, L.Δt, L.step_counter)
```
maximilian-gelbrecht: Did you already test this on GPU? I think you might need to wrap this into a 
```
adapt_structure(to , LeapfrogCore(L.Δt_millisec, L.Δt_sec, L.Δt, L.step_counter))
```
to work correctly.

SpeedyWeather/src/time_stepping/steppers/ncycle_lorenz.jl
```diff
@@ -0,0 +1,182 @@
...
+    Δt_millisec::MS
+
+    "[DERIVED] Time step Δt [s] at specified resolution"
+    Δt_sec::NF
+
+    "[DERIVED] Time step Δt [s/m] at specified resolution, scaled by 1/radius"
+    Δt::NF
+end
+
+Adapt.adapt_structure(to, L::NCycleLorenz) = NCycleLorenzCore(L.Δt_millisec, L.Δt_sec, L.Δt)
```
maximilian-gelbrecht: Here as well, it might need to be wrapped in `adapt_structure(... )`

SpeedyWeather/src/time_stepping/steppers/leapfrog.jl
```diff
@@ -0,0 +1,335 @@
...
+        #             var_old, var_new = get_steps(getfield(namespace, varname2))
+        #             var_new .= var_old
+        #         end
+        #     end
+        end
+    end
+end
+
+# copy grid prognostics from 2nd to 1st step to retain these fields for the previous time step
+function move_prognostic_grid_variables_back!(
```
maximilian-gelbrecht, line 102: would this just be a no-op for NCycleLorenz with just one prognostic step for primitive models? Would we run the parameterizations and vertical diffusion on the current timestep?

maximilian-gelbrecht, line 102: would this just be a no-op for NCycleLorenz with just one prognostic step for primitive models? Would we run the parameterizations and vertical diffusion on the current timestep?

SpeedyWeather/src/time_stepping/steppers/leapfrog.jl
```diff
@@ -0,0 +1,335 @@
...
+    return default_time_step(L)                 # later steps leapfrog with 2Δt
+end
+
+function prognostic_step(::Leapfrog, clock::Clock)
+    clock.step_counter == 0 && return 1         # first Euler step disable filters
+    clock.step_counter == 1 && return 1         # 2nd step: Leapfrog also disable filters
+    return 2                                    # later steps: with RAW filters
+end
+
+function update_prognostic!(
```
maximilian-gelbrecht, line 219: So `update_prognostic!` is now the function that we dispatch on for the actual timestep given the tendencies. Correct?

SpeedyWeather/src/time_stepping/time_integration.jl
```diff
@@ -1,206 +1,152 @@
...
 """$(TYPEDSIGNATURES)
-Perform one single time step of `simulation` including
-possibly output and callbacks."""
-function timestep!(simulation::AbstractSimulation)
-    (; clock) = simulation.variables.prognostic
-    @trace if clock.timestep_counter == 0
-        first_timesteps!(simulation)
-    else
-        later_timestep!(simulation)
+Leapfrog time stepping for all prognostic variables in `vars` using their tendencies.
```
maximilian-gelbrecht: The method is for `AbstractTimeStepper` but the docstring says Leapfrog

SpeedyWeather/src/time_stepping/time_integration.jl
maximilian-gelbrecht, line 1: So `first_timestep!` doesn't exist anymore, and instead this is all inside the `Leapfrog` logic with the spin up time steps, correct?

SpeedyWeather/src/time_stepping/steps.jl
```diff
@@ -0,0 +1,92 @@
...
+        prognostic_spectral = prognostic_spectral_steps(time_stepping, model),
+        tendency_grid = tendency_grid_steps(time_stepping, model),
+        tendency_spectral = tendency_spectral_steps(time_stepping, model),
+    )
+end
+
+"""$(TYPEDSIGNATURES)
+Get all steps of a variable as a tuple of views (wrapped into the same type as the input variable)
+as defined by `get_step`. "Steps" refer to the last dimension, for prognostic variables
+e.g. used for the leapfrog time step."""
```
maximilian-gelbrecht, line 39: Tuple of views is the best way to do everything non-allocating, right? Or why is it light this?

RingGrids/Project.toml
milankl: yes `0.19+DEV` for commit that don't have breaking changes, `0.20-DEV` for commits that introduce breaking changes before 0.20 is released. To mark that the next version should bump the version according to a breaking change

SpeedyWeather/src/time_stepping/steps.jl
```diff
@@ -0,0 +1,92 @@
...
+Get the i-th step of a 4D field as a view (wrapped into the same type as the input variable).
+"step" refers to the last dimension, for prognostic variables e.g. used for the leapfrog time step.
+This method is for a 3D field (horizontal + vertical) with steps in the 4rd dimension."""
+@inline get_step(var::AbstractField{T, 3}, step::Integer) where {T} = field_view(var, :, :, step)
+
+# anything that can decide which variable step to get
+const STEP_COMPONENT = Union{AbstractModelComponent, SpeedyTransforms.AbstractSpectralTransform}
+
+@inline get_step(var, TS::AbstractTimeStepper, C::STEP_COMPONENT) = get_step(var, which_step(var, TS, C))
+@inline get_prognostic_step(var, TS::AbstractTimeStepper, C::STEP_COMPONENT) = get_step(var, which_prognostic_step(var, TS, C))
```
maximilian-gelbrecht, line 122: I think with #1083 merged and #1099 started, in the end we could probably merge those things together in #1099. There I would probably just declare 

```Julia 
@inline get_prognostic_step(var::Variables, name::Symbol, TS::AbstractTimeStepper, C::STEP_COMPONENT) = wrapped_view(var.fused.prognostic.parent, :, :, var.fused.prognstic.slot_map[name], which_prognostic_step(var, TS, C))
```
and similar for the tendendencies. So it's actually great that we have these functions introduced here. 

This should work to get a contiguous view on GPU and CPU. It's hard coded to be fused. But that's the easiest thing to do, everything else would need some unsafe pointer reinterpretation on GPU that I'd like to avoid.

SpeedyWeather/src/time_stepping/steppers/leapfrog.jl
milankl: done this to be explicit

SpeedyWeather/src/time_stepping/steppers/leapfrog.jl
milankl: Yes, NCycleLorenz is supposed to only use one step for all prognostic variables (grid + spectral). This function should not be called for NCycleLorenz

SpeedyWeather/src/time_stepping/steps.jl
milankl: Yes, this should return tuples of views to be non-allocating, is it not?

SpeedyWeather/src/time_stepping/steps.jl
milankl: Okay so we could have both methods coexisting, `get_prognostic_step(::FieldOrLTA)` as this is currently expected or by providing the full `::Variables` and then also provide `name`, that's sounds good and means we can stick to the current `get_step` logic in most places but where we do the fused transforms we'll do as you suggest

SpeedyWeather/src/time_stepping/steps.jl
maximilian-gelbrecht: Yeah, it is. I am just everything anew confused by what causes allocations and what not

SpeedyWeather/src/time_stepping/steppers/ncycle_lorenz.jl
milankl: done!

SpeedyWeather/src/variables/dimensions.jl
milankl: well caught

SpeedyWeather/src/parameterizations/parameterizations.jl
milankl: yeah we do need those though to be able to dispatch in get_..._step but moved this now and commented correctly

milankl:
With the last commits we now have

- Barotropic + ShallowWater running with Leapfrog or NCycleLorenz (default shorter time step of 30 instead of 40min)
- Primitive models (no dynamics) running also with both steppers (stable at 40min)
- Primitive models with dycore still blows up (WIP)

maximilian-gelbrecht:
> * Primitive models with dycore still blows up (WIP)

For both leapfrog and NCycleLorenz?

maximilian-gelbrecht:
I let Claude iterate over it while I was doing something else and it definitely found some things that are wrong. Not sure this is everything. I didn't try to fix them. 

| Source | Impact | Affects default? | Root cause |
|---|---|---|---|
| **C** Euler step transforms the wrong leapfrog index to the grid | **dominant** | yes | `which_prognostic_step(::AbstractSpectralTransform)=2` always transforms step 2 (new state); `main` transforms step **1** (the IC) after the Euler step. Bit-exact: `main`'s post-Euler grid == branch grid step 1 == `0.000e+00`. Fixing it alone shrinks wind/humidity divergence ~17×. |
| **D+E** surface-pₛ tendency seed → divergence equation | small seed → **large** | yes | branch stores `tendencies.grid.pressure` with `nlayers` (`(3168,8,1)`) vs `main`'s surface-only `(3168,)`, giving a `1.3e-5` lnpₛ error after the Euler step that amplifies via `-∇²(R_dry·Tₖ·lnpₛ)` + implicit into `tendencies.divergence ≈ 2.4e4`. |
| **A** implicit `S⁻¹` not re-initialized per step | minor | yes | built once at `ξ=centering·Δt/2`, never updated, while `implicit_correction!` recomputes `ξ` per step. `main` rebuilds `S⁻¹`+`ξ` together (Δt/2, Δt, 2Δt). |
| **B** linear dycore terms always at step 1 | n/a in default | only `α=0` | branch ignores `main`'s `lf_implicit = α==0 ? lf : 1` (`dynamics/tendencies.jl:61`). |

<details> 
<summary> Detailed report </summary> 


## Source C (DOMINANT, affects the default config) — Euler step transforms the wrong leapfrog index to the grid

`main` (`first_timesteps!`) runs the Euler half-step as
`timestep!(vars, Δt/2, model, lf1=1, lf2=1)`, whose final
`transform!(vars, lf2=1, model)` puts **leapfrog step 1 (the IC)** on the grid
(`main`'s `transform.jl` saves the old grid into `*_prev`, then transforms
spectral step `lf`).

The branch's unified transform always uses the *current* step:
`which_prognostic_step(var, ::AbstractLeapfrog, ::AbstractSpectralTransform) = 2`
(`steppers/leapfrog.jl:53`), so after the Euler step it transforms **step 2 (the
new post-Euler state)** to the grid.

**Bit-exact evidence** (post-Euler, `dynamics_only`, `α=0`):

```
main grid.u  ==  main grid.u_prev  ==  branch grid.u[:,:,1]   (diff = 0.000e+00)
branch grid.u[:,:,2]  (the new state) differs from main by    8.97e+00
grid.divergence:  branch[:,:,1] == main (0.0) ;  branch[:,:,2] vs main = 2.28e+02
```

So `main`'s grid after the Euler step equals the branch's **previous** grid step
exactly; the branch's **current** grid step (which the dynamical core reads for
the next step) holds a different state. The first leapfrog step therefore
evaluates its nonlinear/grid terms at a different state on the two branches —
this is the largest contribution to the divergence and grows every step.

- **Where:** `which_prognostic_step(..., ::AbstractSpectralTransform) = 2` does
  not reproduce `main`'s special Euler-step transform of `lf=1`.

## Source A (real, minor) — semi-implicit operators not re-initialized per step

- **`main`** stores `ξ = α·dt` *inside* the implicit struct (`implicit.ξ[]`) and
  rebuilds **both** `S⁻¹` and `ξ` together, re-initializing the implicit terms
  with `Δt/2`, then `Δt`, then `2Δt` across the first steps (`first_timesteps!`).
  `S⁻¹` and `ξ` are always mutually consistent.
- **Branch** initializes the implicit operators **once**, in
  `initialize!(simulation)` while `clock.step_counter == 0`, i.e. for
  `ξ = centering·Δt/2` (`implicit/implicit_primitive_equations.jl:93`), and
  **never re-initializes** them (the per-step re-init is a `# TODO` in
  `time_integration.jl` and is commented out in `steppers/leapfrog.jl`), while
  `implicit_correction!` recomputes `ξ = centering·time_step(L, clock)` **per
  step** (`implicit_primitive_equations.jl:256`). ⇒ `S⁻¹` (fixed at `Δt/2`)
  becomes inconsistent with `ξ` (`Δt`, then `2Δt`) from the first leapfrog step.

**Evidence:** a test patch re-initializing the implicit each step changed the
results (e.g. `grid.geopotential` diff at `+40 min` dropped `1.78e3 → 1.17e3`)
but did **not** remove the dominant divergence — so A is real but secondary to C.

## Source B (real, but NOT triggered by the default config) — linear dycore step in the explicit case

`main`'s `dynamics_tendencies!` uses
`lf_implicit = implicit.α == 0 ? lf : 1` (`dynamics/tendencies.jl:61`) for
`linear_virtual_temperature!`, `vertical_integration!`,
`linear_pressure_gradient!` — i.e. linear gravity-wave terms at the **previous**
step `1` for the default `α=0.5`, but at the **current** step `lf` when `α=0`.

The branch hardcodes these at step 1 (`LinearDynamicalCore` →
`which_prognostic_step = 1`, `steppers/leapfrog.jl:66`) **regardless of α**.

- Default `α=0.5`: both use step 1 ⇒ **agree** (B does not affect the default).
- `α=0` (explicit): `main` uses the current step, the branch the previous step ⇒
  diverge. This is why the `α=0` isolation experiment showed an *extra*
  divergence that is absent from the default config.

## Source E (real, default config, NOT explained by C/A) — divergence-equation tendency at the first leapfrog step

A combined test patch (Euler step transforms step 1 like `main` **and** implicit
re-initialized per step) was applied to the branch and compared to `main` in the
default config. It **massively reduced** the divergence — confirming C dominant
and A real — e.g. at `+40 min`:

```
                          no fix        C+A fix
tendencies.grid.v        1.27e+05  ->   7.60e+03   (~17x smaller, source C)
grid.humidity            5.07e-04  ->   8.4e-07    (≈ bit-identical now)
prognostic.humidity      2.73e-05  ->   4.1e-08
```

…but a **stable residual remains**, concentrated in the **divergence /
geopotential / wind-tendency** channel and essentially unchanged by the C and A
fixes or by turning parameterizations off:

```
tendencies.divergence @ +40 min:   no-fix 2.515e4 | C+A 2.417e4 | dyn-only+C+A 2.419e4
grid.geopotential      @ +40 min:   ~1.6e3 in all cases
```

So there is a **fifth source** in the dynamical core's **divergence equation**,
specific to the **first leapfrog step** (the Euler-step divergence tendency is
bit-identical with `α=0`, so the raw divergence equation is fine when evaluated
at the IC). It is robust to the C and A fixes and present with
`dynamics_only=true`. Root cause not yet isolated — candidates are a step-reading
or evaluation-order difference in `vordiv_tendencies!` / `bernoulli_potential!` /
`linear_pressure_gradient!` / the implicit divergence correction when the
tendencies are evaluated at the post-Euler (leapfrog step 2) state. Needs
within-step instrumentation to pin down.

## Source D (small, ~1e-5; likely the seed of E) — surface-pressure tendency at the Euler step

Post-Euler, with identical IC grids and `α=0`, all spectral prognostics match
bit-for-bit **except `pressure`** (`prognostic.pressure` diff `1.28e-5`,
`tendencies.pressure` diff `6.8e-2` in scaled units). `surface_pressure_tendency!`
and `pressure_gradient_flux!` are otherwise identical to `main` and their inputs
at the Euler step are identical — the remaining suspect is that the branch stores
**`tendencies.grid.pressure` with shape `(npoints, nlayers, 1)` (8 layers)**,
whereas `main` uses the surface-only `(npoints,)`. The `@. pres_tend_grid +=
u_mean_grid*dpres_dx + v_mean_grid*dpres_dy` accumulation then broadcasts/sums
differently, producing the small seed.

The arithmetic links D → E: scaled `Δt/2 ≈ 1.9e-4`, and `6.8e-2 × 1.9e-4 ≈
1.3e-5` (the observed post-Euler `pressure` diff). That `1.3e-5` log-pₛ seed is
then amplified by the linear pressure-gradient term
`-∇²(R_dry·Tₖ·lnpₛ)` (eigenvalues up to ~`l²`) and the implicit divergence
correction into the large `tendencies.divergence` residual seen as **E**. So
**D and E are very likely the same underlying issue** (grid surface-pressure
tendency handling), surfacing as a tiny pressure error that amplifies into the
divergence equation.

## Note — `spectral_truncation!` of tendencies (latent, no-op here)

`main`'s `leapfrog!` calls `SpeedyTransforms.spectral_truncation!(var_tend)` on
every tendency before the update; the branch's `update_prognostic!` does not.
A test patch adding it changed **nothing** at this config (the tendencies are
already truncated), so it is **not** a divergence source here — but it is a
genuine code-level difference that could matter for other resolutions/setups.

---

## Summary

| Source | Affects default config? | Magnitude | Root cause |
|--------|:----:|-----------|------------|
| **C** Euler-step transforms step 2 (not step 1) to grid | **yes** | **dominant** | `which_prognostic_step(::AbstractSpectralTransform)=2` ignores the Euler step |
| **D+E** surface-pₛ tendency seed (Euler) amplified into the divergence eq (first leapfrog) | **yes** | small seed → **large** in div eq | `tendencies.grid.pressure` stored with `nlayers` (vs surface-only); `1.3e-5` lnpₛ amplified by `-∇²(R_dry·Tₖ·lnpₛ)` + implicit |
| **A** implicit `S⁻¹` not re-initialized per step | **yes** | minor | implicit built once at `ξ=centering·Δt/2`, `ξ` recomputed per step |
| **B** linear dycore terms always at step 1 | no (only `α=0`) | n/a in default | branch ignores `main`'s `lf_implicit = α==0 ? lf : 1` |

</details>

milankl:
Many thanks this is really helpful, comments

### Source A, Implicit not reinitialized

This has been fixed now, introducing a new function `reinitialize!` that's by default called before every time step. Fallback is to do nothing, the primitive models by default call it on `model.implicit` which then can decide what to do. So what to "reinitialize" is up to the model to decide, when to reinitialize has a component to decide. 

I still find the terminology "reinitialize" confusing as it's used for implicit to initialize in the first place and also one may assume that `reinitialize!` a model is the same as `initialize!` a model (it's not, the former only does implicit the latter everything else)

But this made the drycore simulation blow up in 77 instead of 11 steps so clearly a step in the right direction.

### Source B, remove implicit centering = 0

This is correctly flagged as alpha=0 doesn't do what it used to. I decided it's not anything anyone would ever actually want and if they do then just `implicit=nothing` instead of supporting this. The reason is that I don't want a parameter in a model component to decide `which_step` to use. That just gets out of hand. Now we have an `@assert 0.5 <= alpha <= 1` instead.

### Source C, Euler step uses wrong leapfrog index

I believe this isn't true. I didn't want the `which_prognostic_step` to depend on `clock.step_counter` and so thought it would be an easier way if for Leapfrog the prognostic variables `old, new` are initialized as

```julia
new <- old
```

That way both are initially identical but the Euler step is still performed as normal with the difference that we can always `get_prognostic_step` 2 and don't have to change from leapfrog index 1 to 2. But I need to double check that. Maybe Claude is indeed flagging something important here but concluded incorrectly why.

### Source D, pressure tendency on the grid dimensions

This field indeed had a vertical dimension which it shouldn't have. I'm still surprised why this wasn't flagging an error anywhere but this has been fixed now and drycore now blows up after 80 instead of 77 steps.

### Source E, combined error

Not sure I can follow Claude's conclusion here, looking further into it

maximilian-gelbrecht:
Confirmed on the current code — and your pushback sharpened the "why". Here's the precise picture:

## Source C: the flag is right, here's the exact mechanism
Post-Euler state (current code, da4bc32c, dynamics-only):

```
                       main vs branch[step1]   main vs branch[step2=current]
grid.u                    0.000e+00               7.29e+00
grid.v                    0.000e+00               1.13e+01
grid.vorticity            0.000e+00               1.10e+01
grid.divergence           0.000e+00               1.70e+02
grid.temperature          0.000e+00               1.96e+00

prognostic.vorticity   step1: 0.0   step2: 0.0     ← spectral state is bit-identical
prognostic.divergence  step1: 0.0   step2: 2.9e-2  ← (the only spectral diff is residual D)
```

So your `new <- old` init is working — the Euler dynamics are bit-identical (spectral step 1 and step 2 match main exactly). The divergence is purely in the transform after the Euler step:

main transforms leapfrog step 1 (the IC, because its Euler call uses `lf2=1`) → its grid = transform(IC).
branch transforms step 2 (which_prognostic_step(::AbstractSpectralTransform)=2) → its current grid = transform(new state).
Bit-exact: main's grid == branch's grid[step 1] (previous); main's grid ≠ branch's grid[step 2] (current).
Then the first leapfrog reads the current grid (step 2): main reads transform(IC) (t=0), the branch reads transform(new state) (t=Δt/2). The nonlinear/grid terms of the first leapfrog are evaluated at a different time level → divergence from the first leapfrog onward.

Why new <- old doesn't cover it: it equalizes step 1 and step 2 only before the Euler step. After the Euler step, step 2 (new) ≠ step 1 (IC), and since the branch always transforms/reads step 2, its grid carries the new state while main's carries the IC.

### The real question is a design one (your call)
It's genuinely arguable that the branch is more correct here — evaluating the first leapfrog's nonlinear terms at the midpoint t=Δt/2 (step 2) rather than main's t=0 (step 1). But to be bit-identical to main, the post-Euler current grid must be transform(step 1), which is inherently the Euler-step special case — exactly the clock.step_counter-dependence you wanted to avoid.

So: the flag stands (real, from the first leapfrog), but whether to "fix" it depends on whether you want to reproduce main's bootstrap or keep the branch's (arguably better) midpoint evaluation and accept that the first ~2 bootstrap steps can't be bit-identical. That's a call only you can make — it's not a clear-cut bug like A or D.

Want me to look into Source E next (you mentioned you're still digging), or help design a clock-free way to get main's bootstrap (e.g. a one-shot flag the time stepper flips after the first step, rather than reading step_counter in which_prognostic_step)?

milankl:
Okay turned out Claude's "note" was the culprit: Previously we had spectral_truncation! in the time stepping on all tendencies so that the last degree is zeroed. I felt this wasn't responsibility of the time stepping and redundant anyway because the horizontal diffusion is already setting this to zero anyway. Forgot that pressure isn't subject to diffusion so that still had something in the last degree that kept piling up.

Now everything seem stable again. Trying to clean things up now and get all the broken tests to pass again ...

maximilian-gelbrecht:
I just let Claude continue on this as I still had this session open anyway. It claims that leapfrog time steps are now bit-identical but that `Source C` still is a problem and proposed an ad-hoc solution: 

Here's the summary and suggestion: 

## Source C — Euler spin-up step transforms the wrong leapfrog index to the grid

### Symptom
The default `PrimitiveWetModel` (which still uses `Leapfrog`) should bit-reproduce `main`, but it diverges starting from the **first leapfrog step**. After the implicit-reinit, `zero_last_degree!` and grid-pressure-tendency fixes, **this is the dominant remaining difference**: with a one-line diagnostic patch that fixes only this, `tendencies.grid.v` at the first leapfrog drops from `1.1e5` → `3.5` and `tendencies.vorticity` from `7343` → `0.85`.

### Diagnosis
After the Euler half-step the **spectral** state is bit-identical to `main` on both leapfrog steps (`step1 = IC`, `step2 = new`). The divergence is purely in **which step gets transformed to the grid**:

- **`main`** transforms leapfrog **step 1 (the IC)** after the Euler step — its Euler call uses `lf2 = 1`.
- **branch** always transforms **step 2 (the new state)** — `which_prognostic_step(var, ::AbstractLeapfrog, ::AbstractSpectralTransform) = 2`.

Bit-exact evidence (post-Euler, dynamics-only):

                   main vs branch[step 1]   main vs branch[step 2 = current]
```
grid.u                    0.000e+00               7.29e+00
grid.divergence           0.000e+00               1.70e+02
grid.vorticity            0.000e+00               1.10e+01
prognostic.vorticity   step1: 0.0  step2: 0.0   (spectral state matches exactly)
```


i.e. **`main`'s grid == the branch's *previous* grid step (1); `main`'s grid ≠ the branch's *current* grid step (2)** — yet the dynamical core reads the current step. So the first leapfrog evaluates its **nonlinear/grid terms at a different time level** (`main`: t=0 / IC, branch: t=Δt/2 / new state), and the state diverges from the first leapfrog on.

The `new <- old` initialization only equalizes step 1 and step 2 *before* the Euler step; after it, `step2 (new) ≠ step1 (IC)`, and since the branch always transforms/reads step 2, its grid carries the new state while `main`'s carries the IC.

> Note: it's arguable the branch is *more* correct here (midpoint evaluation at step 2). But to be **bit-identical to `main`**, the post-Euler *current* grid must be `transform(step 1)` — which is precisely the per-step special case. Verified: once an identical post-Euler state is injected into both, **every subsequent leapfrog step is bit-identical**, so C is the only thing left (besides a tiny residual at the Euler step itself).

### Proposed fix (clock-free)
Have the grid transform read **leapfrog step 1 for its spectral source only during the Euler spin-up step**, then step 2 from the first leapfrog on — driven by a transient flag on the stepper (not by `clock.step_counter` inside `which_prognostic_step`):

```julia
# Leapfrog (mutable struct): add a transient flag
"[DERIVED] transform leapfrog step 1 (IC) to the grid during the Euler spin-up step"
euler_grid_step::Bool

@inline transform_source_step(L::AbstractLeapfrog) = L.euler_grid_step ? 1 : 2

# in transform!(vars, ::AbstractLeapfrog, model::PrimitiveEquation; ...)
# only the SPECTRAL SOURCE step changes; the grid *targets* (u_grid, …) stay at step 2 (current slot)
lf   = transform_source_step(time_stepping)
vor  = get_step(vars.prognostic.vorticity,   lf)
div  = get_step(vars.prognostic.divergence,  lf)
pres = get_step(vars.prognostic.pressure,    lf)
temp = get_step(vars.prognostic.temperature, lf)
# humid likewise for PrimitiveWet

# flip the flag after the Euler step (transform has already run by then)
function time_step!(clock::Clock, L::Leapfrog)
    if clock.step_counter == 0
        time_step!(clock, L.Δt_millisec ÷ 2, increase_counter = false)
        L.euler_grid_step = false          # transform step 2 from the first leapfrog on
    elseif ...
    end
end

# reset at initialize!: L.euler_grid_step = L.start_with_euler   (false on restart → no Euler step)
```
Notes:

Only the spectral source step is special-cased; the grid target arrays keep which_prognostic_step = 2 (we still write into the current grid slot, just from transform(step 1)).
ShallowWater and Barotropic transforms need the same one-line treatment.
The flag is set/reset by the stepping logic, so which_prognostic_step never reads the clock.
Alternative
If the branch's midpoint evaluation is preferred over reproducing main's bootstrap, then leave it as-is and document that the first ~1 leapfrog step of the spin-up is intentionally not bit-identical to main — every later leapfrog step already is.

SpeedyWeather/src/dynamics/geopotential.jl
```diff
@@ -60,16 +60,19 @@ function geopotential!(
...
     )
-    T = vars.grid.temperature
-    Φ = vars.grid.geopotential
+    TS = model.time_stepping
+    G = model.geopotential
+    T = get_prognostic_step(vars.grid.temperature, TS, G)
+    Φ = vars.dynamics.geopotential
 
     # use zero scratch for humidity in dry models to not distinguish in kernels below
     vars.scratch.grid.a .= 0
```
maximilian-gelbrecht: could include the zero-ing in the statement below

SpeedyWeather/src/dynamics/geopotential.jl
```diff
@@ -60,16 +60,19 @@ function geopotential!(
...
-    q = haskey(vars.grid, :humidity) ? vars.grid.humidity : vars.scratch.grid.a
+    q = haskey(vars.grid, :humidity) ?  get_prognostic_step(vars.grid.humidity, TS, G) : vars.scratch.grid.a
```
maximilian-gelbrecht:
```suggestion
    q = haskey(vars.grid, :humidity) ?  get_prognostic_step(vars.grid.humidity, TS, G) : (vars.scratch.grid.a; fill!(vars.scratch.grid.a, 0))
```

SpeedyWeather/src/dynamics/horizontal_diffusion.jl
```diff
@@ -262,34 +304,43 @@ function horizontal_diffusion!(
...
     horizontal_diffusion!(vor_tend, vor, expl, impl)
     horizontal_diffusion!(div_tend, div, expl_div, impl_div)
     horizontal_diffusion!(temp_tend, temp, expl, impl)
 
+    # horizontal diffusion is also responsible to set a tendency's last degree to zero
+    # all variables carry this degree but it's not used for the scalar quantities that
+    # all our prognostic variables are, so set to zero now for those variables that
+    # aren't subject to diffusion
+    pres_tend = get_tendency_step(vars.tendencies.pressure, model.time_stepping, diffusion)
+    zero_last_degree!(pres_tend)
```
maximilian-gelbrecht, line 331: here it's just pressure that has its last degree zeroed. What about the other prognostic variables?

SpeedyWeather/src/dynamics/tendencies.jl
```diff
@@ -443,19 +443,21 @@ function vordiv_tendencies!(
...
-    humid = haskey(vars.grid, :humidity) ? vars.grid.humidity : vars.scratch.grid.a
+    humid = haskey(vars.grid, :humidity) ? get_prognostic_step(vars.grid.humidity, time_stepping, DynamicalCore()) : vars.scratch.grid.a
```
maximilian-gelbrecht:
```suggestion
    humid = haskey(vars.grid, :humidity) ? get_prognostic_step(vars.grid.humidity, time_stepping, DynamicalCore()) : (vars.scratch.grid.a; fill!(vars.scratch.grid.a, 0))
```

SpeedyWeather/src/dynamics/tendencies.jl
```diff
@@ -443,19 +443,21 @@ function vordiv_tendencies!(
...
-    (; u, v) = vars.grid                                    # velocities, vorticity, temperature
-    vor = vars.grid.vorticity
-    temp = vars.grid.temperature
+    u_tend_grid = get_tendency_step(vars.tendencies.grid.u, time_stepping, DynamicalCore())
+    v_tend_grid = get_tendency_step(vars.tendencies.grid.v, time_stepping, DynamicalCore())
+    u = get_prognostic_step(vars.grid.u, time_stepping, DynamicalCore())
+    v = get_prognostic_step(vars.grid.v, time_stepping, DynamicalCore())
+    vor = get_prognostic_step(vars.grid.vorticity, time_stepping, DynamicalCore())
+    temp = get_prognostic_step(vars.grid.temperature, time_stepping, DynamicalCore())
     vars.scratch.grid.a .= 0
```
maximilian-gelbrecht:
```suggestion
```

SpeedyWeather/src/dynamics/tendencies.jl
```diff
@@ -576,20 +567,19 @@ Compute the temperature tendency.
...
+
+    temp_tend = get_tendency_step(vars.tendencies.temperature, time_stepping, DynamicalCore())
+    temp_tend_grid = get_tendency_step(vars.tendencies.grid.temperature, time_stepping, DynamicalCore())
+    div_grid = get_prognostic_step(vars.grid.divergence, time_stepping, DynamicalCore())
+    temp = get_prognostic_step(vars.grid.temperature, time_stepping, DynamicalCore())
 
     # use scratch array with zeros in case humidity doesn't exist
     vars.scratch.grid.a .= 0
-    humid = haskey(vars.grid, :humidity) ? vars.grid.humidity : vars.scratch.grid.a
+    humid = haskey(vars.grid, :humidity) ? get_prognostic_step(vars.grid.humidity, time_stepping, DynamicalCore()) : vars.scratch.grid.a
```
maximilian-gelbrecht:
```suggestion
    humid = haskey(vars.grid, :humidity) ? get_prognostic_step(vars.grid.humidity, time_stepping, DynamicalCore()) : (vars.scratch.grid.a; fill!(vars.scratch.grid.a, 0))
```

SpeedyWeather/src/dynamics/tendencies.jl
```diff
@@ -576,20 +567,19 @@ Compute the temperature tendency.
...
+
+    (; adiabatic_conversion, atmosphere, implicit, spectral_transform, time_stepping) = model
+
+    temp_tend = get_tendency_step(vars.tendencies.temperature, time_stepping, DynamicalCore())
+    temp_tend_grid = get_tendency_step(vars.tendencies.grid.temperature, time_stepping, DynamicalCore())
+    div_grid = get_prognostic_step(vars.grid.divergence, time_stepping, DynamicalCore())
+    temp = get_prognostic_step(vars.grid.temperature, time_stepping, DynamicalCore())
 
     # use scratch array with zeros in case humidity doesn't exist
     vars.scratch.grid.a .= 0
```
maximilian-gelbrecht:
```suggestion
```

SpeedyWeather/src/models/abstract_models.jl
```diff
@@ -27,6 +33,12 @@ function initialize!(nt::NamedTuple, model::AbstractModel)
     return nothing
 end
 
+# some components may need to be reinitialized, e.g. implicit with changing time step
+# fall back to nothing, models can implement their own reinitialize! method if needed
+reinitialize!(model::AbstractModel, vars::AbstractVariables) = nothing
```
maximilian-gelbrecht: maybe `@inline` the fallbacks just to be safe

SpeedyWeather/src/output/variables/dynamics.jl
```diff
@@ -116,10 +116,11 @@ Fields are: $(TYPEDFIELDS)"""
     compression_level::Int = 3
     shuffle::Bool = true
     keepbits::Int = 12
-    transform::F = (x) -> exp(x) / 100     # log(Pa) to hPa
+    transform::F = (x) -> x / 100     # Pa to hPa
```
maximilian-gelbrecht, line 119: I remember we talked about this. So now, we have both steps in the regular pressure and just take the logarithm if needed, or was it the other way around?

SpeedyWeather/src/time_stepping/steppers/leapfrog.jl
```diff
@@ -0,0 +1,267 @@
...
+    if L.start_with_euler
+        L.first_step_euler = true
+    end
+    return nothing
+end
+
+"""$(TYPEDSIGNATURES) Leapfrog is spun up with 1 Euler forward step that doesn't count for clock + output"""
+spin_up_steps(::Leapfrog) = 1
+
+function time_step!(clock::Clock, time_stepping::Leapfrog)
```
maximilian-gelbrecht, line 187: this here is the clock side to replace the `first_timestep!` / `later_timestep!` logic, right? 

Did I overlook where the initial Euler step logic is?

SpeedyWeather/src/dynamics/geopotential.jl
milankl: Actually `fill!(vars.scratch.grid.a, 0)` this already returns the array, so that's enough I think!

SpeedyWeather/src/dynamics/horizontal_diffusion.jl
milankl: The horizontal diffusion operators have a guaranteed zero for the last degree, so that happens automatically!

CHANGELOG.md
```diff
@@ -35,6 +35,8 @@
 - Documentation hero page and docs rendering [#1071](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1071) [#1072](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1072)
 - Bump julia-actions/cache and setup-julia v2 to v3 [#1073](ttps://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1073)
```
maximilian-gelbrecht, line 59:
```suggestion
```

SpeedyWeather/src/dynamics/horizontal_diffusion.jl
milankl: It's just that pressure is the only variable that isn't diffused (because not advected) so hence it had something non-zero in the last degree!

SpeedyWeather/src/dynamics/geopotential.jl
maximilian-gelbrecht: ah true, yes.

SpeedyWeather/src/output/variables/dynamics.jl
milankl: `simulation.variables.grid.pressure` is like `simulation.variables.prognostic.pressure` the logarithm of surface pressure but we now have `simulation.variables.parameterizations.surface_pressure` that's in Pa and just computed once in `transform!`

maximilian-gelbrecht: okay understood

maximilian-gelbrecht:
I tried to give everything a look and I am fine with it. 

I am still a little bit confused about the different handling of the first time steps and the whole thing Claude wrote about that also adds to my further confusion. If we could still address that in some way or be sure we do exactly what we want, that would be great. 

Otherwise I have not much to add, this look good. I hope we can also adjust all the tests and then merge this quickly :)

milankl:
The 2nd last commit introduces one breaking change, `set!` had `lf` as keyword argument now renamed to `step` this in favour of the general step dimension indexed with `step`. And the initial conditions are always written to the first step which is the default in `set!`.

milankl:
The last commit now developed a clearer definition of what a prognostic variable is vs what a time stepping variable is

- `prognostic` is any variable that retains memory from one time step to the next that influences other variables at some later step
- in most cases this means we solve a differential equation with d/dt for these variables (e.g. clock is time += dt*1) but a random process can be formulated as AR1 process too and is therefore a discrete map and not continuous -- still a prognostic variable 
- exceptions are multi-step methods where also the tendencies contain memory

In contrast we define a time stepping variable, i.e. one that's stepped forward using `model.time_stepping` as one that is present in

- in `simulation.variables.tendencies`
- or `simulation.variables.tendencies.land`
- or `simulation.variables.tendencies.ocean`

which deliberately excludes `tendencies.grid` as these are thought of to be copies (potentially of only some terms) in grid space and therefore not used by the time stepper directly. This also means that land/ocean variables can decide to make their variables subject to time stepping or implement their own. These variables are now available via

- `tendency_names`
- `land_tendency_names`
- `ocean_tendency_names`

e.g. 

```julia
    for varname in tendency_names(vars)
        var = getfield(prognostic, varname)
        tendency = getfield(tendencies, varname)
        update_prognostic!(var, tendency, vars, time_stepping, model.implicit, model)
    end
```

is now used to retrieve all atmospheric variables subject to time stepping as defined by their existence in the tendencies

SpeedyWeather/src/time_stepping/steppers/leapfrog.jl
milankl: Yes, the clock has it's own `time_step!` which generally is

```julia
time_step!(clock::Clock, time_stepping::AbstractTimeStepper) =
    time_step!(clock, time_stepping.Δt_millisec)

function time_step!(clock::Clock, Δt; increase_counter::Bool = true)
    clock.time += Δt
    clock.step_counter += 1                     # always increased, counts time stepper steps
    clock.time_step_counter += increase_counter # spin up steps may not count for clock
    return nothing
end
```

see `time_stepping/clock.jl` (moved from `dynamics/clock.jl`) but here Leapfrog implements it's own version of it by extending the dispatch over `::Leapfrog`

SpeedyWeather/src/models/abstract_models.jl
milankl: did that

SpeedyWeather/src/dynamics/geopotential.jl
milankl: applied this to all places

milankl:
The CI always contained a test of solving the oscillation equation (once around the complex unit circle in 192 time steps), now also with NCycleLorenz which looks really good I think

(step), time stepper, error, magnitude (in Float32/64)

```julia
[ Info: (Leapfrog, 0.0011637651f0, 0.9999415f0)
[ Info: (Leapfrog, 0.0011636298171382963, 0.9999414811407209)
[ Info: (3, NCycleLorenzA, 0.0002481685f0, 0.9997526f0)
[ Info: (4, NCycleLorenzA, 1.5246626f-5, 0.99999785f0)
[ Info: (3, NCycleLorenzB, 0.00024817637f0, 0.9997526f0)
[ Info: (4, NCycleLorenzB, 1.5526313f-5, 0.99999845f0)
[ Info: (3, NCycleLorenzAB, 0.00024812508f0, 0.99975264f0)
[ Info: (4, NCycleLorenzAB, 1.5117451f-5, 0.9999984f0)
[ Info: (3, NCycleLorenzABBA, 0.0002481813f0, 0.9997526f0)
[ Info: (4, NCycleLorenzABBA, 1.5183154f-5, 0.99999833f0)
[ Info: (3, NCycleLorenzA, 0.00024766219524000786, 0.9997531012816103)
[ Info: (4, NCycleLorenzA, 1.5370180525335732e-5, 0.9999983266800245)
[ Info: (3, NCycleLorenzB, 0.0002476621952400182, 0.9997531012816103)
[ Info: (4, NCycleLorenzB, 1.5370180524692475e-5, 0.9999983266800249)
[ Info: (3, NCycleLorenzAB, 0.00024766219523981206, 0.9997531012816105)
[ Info: (4, NCycleLorenzAB, 1.537018052482872e-5, 0.9999983266800248)
[ Info: (3, NCycleLorenzABBA, 0.0002476621952402254, 0.9997531012816101)
[ Info: (4, NCycleLorenzABBA, 1.537018052538743e-5, 0.9999983266800248)
```

Interesting that there's no obvious improvement over using the combined variants AB, ABBA.
Errors with 4-step (=4th order) drops much more quickly with dt than the 3-step variants. Check.
And everything is always stable (magnitude <= 1). Check.

For error and magnitude it's important to complete a cycle. Hence using 192 time steps as divisible by 3, 4, 12, 16.

milankl:
I had some issues with `heatmap` extensions (Makie/UnicodePlots) they now by default drop any additional dimensions and pick index 1 for those, raising a warning if that occurs

maximilian-gelbrecht (approved):

PR #1036, opened 2026-04-09
Fix link for examples in README

gregordecristoforo:
Clicking on the link in the Readme currently doesn't work. This should fix it. This is pedantic, I know.


gregordecristoforo:
Actually, it seems like multiple links in the Readme are dead. Let me commit the fixes to this PR.

milankl:
No this is great! Can you name branches starting with your initials so `gd/some_branch` that makes them easier to distinguish (for the future). Also if you only change the README then you can switch off a bunch of CI tests, they are all labels that start with `skip-...`

PR #1042, opened 2026-04-13
Held-Suarez forcing missed scaling and pressure normalisation

milankl:
Somehow the Held-Suarez kernel had a `pres` as argument but that must have fallen overboard. We therefore have a Held-Suarez forcing towards too low temperatures. Also a radius scaling was missing from the forcing because the temperature equation are scaled by the radius! This got probably missing when moving HeldSuarez from parameterization to forcing.

Now surface temperatures after 1 year look good

image


maximilian-gelbrecht (approved):

SpeedyWeather/src/dynamics/forcing.jl
```diff
@@ -359,41 +362,43 @@ function forcing!(
...
+        log_σ,
         whichring,
     )
     ij, k = @index(Global, NTuple)
     j = whichring[ij]                   # latitude ring index
-    kₜ = temp_relax_freq[k, j]           # (inverse) relaxation time scale
+    kₜ = temp_relax_freq[k, j]          # (inverse) relaxation time scale
+
+    log_p = log_pₛ[ij] + log_σ[k]        # p/pₛ = σ but in log space
+    p = exp(log_p[ij])                  # pressure [Pa]
```
milankl: @maximilian-gelbrecht I just realised that this indexes a scalar. Why does this not throw an error inside a kernel?

maximilian-gelbrecht: Huh, good question, this does def seem like an error. Initially I was thinking because of the `inbounds=true` but this is deactivated in the tests.

milankl: I've corrected this bug and so I'm happy to merge it but it might be nice to figure out how to make sure that such a bug actually throws an error

maximilian-gelbrecht: Could you ask Claude for an MWE of that?

maximilian-gelbrecht (approved):

PR #1048, opened 2026-04-14
Documentation sidebar configuration

maximilian-gelbrecht:
This tries to configure the sidebar in the documentation in the same style as Lux and Reactant have it, so that it only shows the pages of the current section, not all. 

Quickly done with Claude, I want to see the preview build here to confirm its working


maximilian-gelbrecht:
@milankl Do you have an idea why the preview doesn't work here?

milankl:
The preview does work https://speedyweather.github.io/SpeedyWeatherDocumentation/previews/PR1048/ but the link in the `documenter/deploy` workflow is wrong. Documenter.jl places the correct link, DocumenterVitePress.jl doesn't adapt the link to out-of-repo deployment ... 

Raised an issue
https://github.com/LuxDL/DocumenterVitepress.jl/issues/346

maximilian-gelbrecht:
Ah great! 

So also the PR works in what I intended. What do you think? I would prefer it like this.

milankl:
I like it too! Happy for this to be merged. While we're at it, I realised that models.md was linked from other pages but not indexed, just changed that in the last commit

maximilian-gelbrecht:
@bgroenks96 this is also what I mentioned on Zulip, maybe you like this style a bit more

bgroenks96:
Yes much better. I think personally I still prefer having the various sections on the left sidebar rather than in dropdown menus from the top because it makes them more visible and requires less hunting. But that's a matter of taste I guess.

PR #1049, opened 2026-04-14
Architectures: Add tests and fix subarray bug

maximilian-gelbrecht:
Milan: Motivation, this

```julia
    try
        v = path(variable, simulation)
        ts = simulation.model.time_stepping
        has_step = (is3D(variable) && ndims(v) == 3) ||
                    (is2D(variable) && ndims(v) == 2)
        v = has_step ? get_prognostic_step(v, ts, output) : v

        # TODO somehow this throws an error with a view
        raw = on_architecture(CPU(), v)
        RingGrids.interpolate!(var, raw, output.interpolator)
    catch FieldError
        var .= variable.missing_value
    end
```

threw an error in #1035 because there we need to view an array before passing to `on_architecture`. This was actually tricky to find out because of the `try` block no error was thrown.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fixed a bug for `SubArrays` not working with `Architectures.ismatching` [#1049
+](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1049)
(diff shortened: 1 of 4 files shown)


milankl (approved):
So quick! 🚀

PR #1056, opened 2026-04-15
Architectures.ismatching for `Type{<:SubArray}`

maximilian-gelbrecht:


milankl:
MWE was

```julia
julia> field = rand(OctahedralGaussianField, 24);

julia> on_architecture(CPU(), RingGrids.field_view(field, :));
ERROR: MethodError: no method matching architecture(::Type{SubArray{Float64, 1, Vector{Float64}, Tuple{Base.Slice{Base.OneTo{Int64}}}, true}})
```

PR #1057, opened 2026-04-16
Williams' Simple Spectral Model for Longwave Radiation

maximilian-gelbrecht:
This is an implementation of Andrew Williams' Simple Spectral Model https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2025MS005405

So far I've just done this as a little experiment how far I can take this just with Claude. 

I gave Claude quite detailed instructions on how it should implement it, and our documentation on parametrization, the pdf of the paper and the Fortran code. 

The unit tests pass, but I also tested a one year rollout of the `PrimitiveWetModel` with it and it seems stable. As expected with CO2 turned off over the course of the integration the atmospheres slowly cools down, at the end after one year I reached $$T_{max}$$ of just 12 degrees C. With CO2 turned on and 280ppm the temperature remains stable.

I've just given the code a quick check but no really detailed review yet.


milankl:
@AndrewILWilliams it's coming!!

maximilian-gelbrecht:
I added documentation and looked over this and I am quite happy with it, but I also lack some radiation knowledge to really judge it well. But I think in principle, this is ready for a review. 

Here are the absorption spectra as implemented (code in the docs): 
kappa_spectrum

I am a bit confused why in the paper (dashed line in Fig 3) the lines look so smooth when it's just exponential functions in a log plot. But given the equations in the paper, this plot looks right.

maximilian-gelbrecht:
With #993, we would just change it so that `co2_ppmv` becomes a variable, but this seems like a super easy change. And then we can run some idealised forcing scenarios :tada:

glwagner:
Hi @maximilian-gelbrecht and @milankl --- after a conversation with @milankl I vibe coded this functionality into a new package here: https://github.com/NumericalEarth/AnalyticBandRadiation.jl

The idea is that if we put this stuff in an independent package, then its features can be drawn into either Breeze or SpeedyWeather, which might increase how much its used and tested.

Curious what you all think! It was just a few prompts, so not too much work.

maximilian-gelbrecht:
Cool! I actually already wanted to talk to you about sharing some parameterizations with Breeze and vice versa. So this is great. I am on vacation next week, and then there's EGU, so some more detailed comments will have to wait a bit.

AndrewILWilliams:
Hi! Sorry, I missed this notification while I was traveling last week. This is very exciting though! @maximilian-gelbrecht the reason for the smooth dashed line in Fig 3 of the paper is that what I plotted there is the sum of H2O line + continuum absorption.

Let me know if there's anything I can help with! If I don't respond quickly then feel free to nudge me via email

maximilian-gelbrecht:
This PR is superseded by our work on the stand alone package https://github.com/NumericalEarth/AnalyticBandRadiation.jl 

and PR #1089 that documents it here in our repo.

PR #1063, opened 2026-04-21
fix potential energy analysis calculation and factor in code

### LINKED ISSUE #1055: Documentation inconsistency in the energy derivation in the shallow water model
Hi,

I would like to discuss the derivation of the energy calculation of the shallow water model in the Analysis section: [Link to section](https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/analysis#Energy)

I think there are some issues/mixups:

First, as stated before equations (1) and (2), the potential energy density is $gz$ (ignoring density here). To obtain the potential energy, one must integrate over the volume, i.e. over $z$ from the bottom $H_B$ to the fluid height $H_b + h = H + \eta$.

Therefore, equation (1) should read instead: 

$$E = \iint \left[ \int_{H_b}^{H_b + h} \frac{1}{2} (u^2 + v^2 + 2gz) \ dz \right]  dA$$

This then leads to 

$$E = \iint \frac{1}{2} (u^2 + v^2 + gh + 2H) h \ dA$$

opposed to equation (2). 

However, since we are only interested in changes of total energy over time (in this section), the constant term $Hh$, which is proportional to the total mass (shown constant in the previous section), can be canceled. This finally (after a few intermediate steps) becomes equivalent to equation (2) in the documentation.

Also, in the code snippet following the derivation:

`E = @. h/2*(u^2 + v^2) + g*h^2  # vertically-integrated mechanical energy`

a factor of `1/2` seems to be missing in front of `g*h^2`. But this somehow almost doesn't matter, although `1/2` leads to slightly better energy conservation over time (in my own testing).

SieglStefan:
Fixes the potential energy expression in the analysis by changing integral boundaries and adding a constant energy shift and short description and fixes missing (1/2) factor in the code section below according to [Issue #1055](https://github.com/SpeedyWeather/SpeedyWeather.jl/issues/1055).

### DIFF
--- a/docs/src/analysis.md
+++ b/docs/src/analysis.md
@@ -125,11 +125,12 @@ total volume (times ``h=\eta+H-H_b`` for the vertical then integrated over the s
 
 ```math
 \begin{align}
-E & = \iint \left[ \int_{H_b}^{\eta}\frac{1}{2}\left(u^2 + v^2 + gz\right)dz \right]dA \\
-  & = \iint \frac{1}{2}\left(u^2 + v^2 + gh\right)h dA
+E & = \iint \left[ \int_{H_b}^{H_b+h}\frac{1}{2}\left(u^2 + v^2 + 2gz\right)dz \right]dA \\
+  & = E_0 + \iint \frac{1}{2}\left(u^2 + v^2 + gh\right)h dA
 \end{align}
 ```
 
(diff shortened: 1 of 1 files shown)


miniufo (approved):

maximilian-gelbrecht:
Thanks for your first PR Stefan! 

As @miniufo approved, I'll merge this once the CI is passing.

PR #1067, opened 2026-04-23
Version bumps 0.19.0 (+ submodules) 

maximilian-gelbrecht:


lazarusA:
Just before tagging a new release, symlinks need to be removed from the gh-pages documentation branch. For more info see:
https://luxdl.github.io/DocumenterVitepress.jl/stable/manual/get_started#5.-Update-Your-Deployment

delete_s

maximilian-gelbrecht:
Thanks Lazaro, will do that! 

@milankl any other PR you want to merge first?

milankl (approved):
I see you sorted the `[BREAKING]` logs to the top, I think I like!

milankl:
I'm generally wondering how to best keep track of whether a PR is breaking for any of the submodules. We renamed things in SpeedyWeatherInternals so I had already bumped that to 0.2 but I think what people generally do is

- `0.1+DEV` i.e. contains non-breaking commits since 0.1 
- `0.2-DEV` i.e. contains breaking commits since 0.1 such that the next release would need to be 0.2

So maybe we can adapt that too? Just checking in the Julia slack that this doesn't cause problems in Pkg.jl ...

milankl:
Otherwise we'd need to write [BREAKING SW], [BREAKING ST], [BREAKING RG] etc to say for which submodule it's actually breaking. I'd be happy for now to mean with BREAKING that it's breaking for SpeedyWeather or any of its submodules?

maximilian-gelbrecht:
What's even exactly breaking ;) 

Some of these changes are breaking for some internal function but it will usually not affect the typical user. 

Are you okay with the version names for this PR or shall I wait?

SpeedyWeatherInternals/Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "SpeedyWeatherInternals"
 uuid = "34489162-d270-4603-8b96-37b04f830d73"
-version = "0.2"
+version = "0.2.1"
```
milankl: Ah no, this was already bumped in #969! So should be just `0.2.0` as we haven't released since!

maximilian-gelbrecht: Ah sorry. Okay

milankl:
Happy for this to be merged, then we do release marathon bottom up, right? So SpeedyWeatherInternals first, then RingGrids+LowerTriangularArrays, then SpeedyTransforms, then SpeedyWeather?

maximilian-gelbrecht:
> Happy for this to be merged, then we do release marathon bottom up, right? So SpeedyWeatherInternals first, then RingGrids+LowerTriangularArrays, then SpeedyTransforms, then SpeedyWeather?

Yes exactly.

PR #1072, opened 2026-04-26
Fix some docs rendering

navidcy:
This PR fixes some math rendering in the docs.

The only non-trivial change is that in https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1072/changes#diff-711744623cb8d3002db85edfec455266b612c1179402a67eaa18630b7764b3cf where a `+ -` is replaced with `-`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
-## Unreleased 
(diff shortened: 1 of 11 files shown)


milankl (approved):
Awesome, thanks Navid!!

PR #1075, opened 2026-05-03
`set!(output, model; interval)` introduced to change output frequency after model init 

maximilian-gelbrecht:
As discussed over in https://github.com/NumericalEarth/NumericalEarth.jl/issues/185


maximilian-gelbrecht:
@milankl to fix the issue in NumericalEarth, I d also directly release a 0.19.1 so that this function can be used.

PR #1076, opened 2026-05-05
ZarrOutput 

maximilian-gelbrecht:
This implements `ZarrOutput` in an extension. It follows the same output logic as `NetCDFOutput`.


SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
```diff
@@ -0,0 +1,366 @@
...
+        output::ZarrOutput,
+        variable::AbstractOutputVariable,
+        simulation::AbstractSimulation,
+    )
+    ~hastime(variable) && output.output_counter > 1 && return nothing
+
+    # interpolate 2D/3D variables
+    var = is3D(variable) ? (is_land(variable) ? output.field3Dland : output.field3D) : output.field2D
+
+    try
```
maximilian-gelbrecht: @milankl Why do we actually have that `try`/`catch` statement in the `NetCDFOutput`? I've also seen that you removed that in the `mk/lorenz2` branch.

maximilian-gelbrecht:
Obviously mostly vibe coded, but I reviewed it and I am reasonably happy with it. I also checked xarray compatibility.

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
milankl: Yes I would also remove this here. I can't remember why it was originally written but it easily masks errors in the `interpolate!` function. E.g. in `mk/lorenz2` we're passing on views, which threw an error and therefore just overwrote the entire array with missing before output. So while the simulation was running fine I was only able to see NaNs in the output

milankl (commented):
Looks great! Questions

- is `append` really the recommended approach to write into an infinite time dimension?
- can we define more functions in `netcdf.jl` for `::AbstractOutput` to avoid duplication here?

maximilian-gelbrecht:
Yeah you are completely right!

> * is `append` really the recommended approach to write into an infinite time dimension?

I changed it now to pre-allocate everything. That's better. It doesn't really seem necessary for us to continue output in the same file when hitting multiple `run!` anyway at the moment. In this cases users can still merge the files/folder afterwards anyway. 

> * can we define more functions in `netcdf.jl` for `::AbstractOutput` to avoid duplication here?

Yes! Absolutely, I removed some duplication here now.

maximilian-gelbrecht:
With T31L8, integrated for 10 days and hourly output, we are 18% faster with `ZarrOutput` compared to `NetCDFOutput`. 

@milankl When you can open it in Browzarr, I'd be happy to merge this

Format	Min	Median	Max	On-disk
NetCDF	3.455 s	3.590 s	3.810 s	28.9 MiB
Zarr	2.721 s	2.954 s	3.890 s	36.1 MiB

Script: 

```Julia 
# Quick benchmark: NetCDFOutput vs ZarrOutput total `run!` wall time
# over period = Day(3) with hourly output. Default model + spectral grid.
# Run from the SpeedyWeather.jl root with:
#   julia --project=SpeedyWeather/test zarr_vs_netcdf_bench.jl

using SpeedyWeather, Zarr, Printf, Statistics

const PERIOD = Day(10)
const INTERVAL = Hour(1)
const N_REPEATS = 3

tmp = mktempdir(@__DIR__, prefix = "bench_")
println("Tempdir: ", tmp)

function build_simulation(output)
    spectral_grid = SpectralGrid()                    # defaults
    return initialize!(model)
end

function on_disk_size(path::String)
    isdir(path) || isfile(path) || return 0
    total = 0
    if isfile(path)
        total += filesize(path)
    else
        for (root, _, files) in walkdir(path)
            for f in files
                total += filesize(joinpath(root, f))
            end
        end
    end
    return total
end

human(b) = b < 1024 ? "$(b) B" :
    b < 1024^2 ? @sprintf("%.1f KiB", b / 1024) :
    b < 1024^3 ? @sprintf("%.1f MiB", b / 1024^2) :
    @sprintf("%.2f GiB", b / 1024^3)

function bench(label::String, make_output)
    times = Float64[]
    last_path = ""

    # warmup run, not timed
    output = make_output("warmup")
    sim = build_simulation(output)
    run!(sim, period = Hour(1), output = true)

    for i in 1:N_REPEATS
        output = make_output("$(label)-$i")
        sim = build_simulation(output)
        t = @elapsed run!(sim, period = PERIOD, output = true)
        push!(times, t)
        last_path = joinpath(output.run_path, output.filename)
        @printf("  %s run %d: %.3f s\n", label, i, t)
    end

    println("  └ store: $(human(on_disk_size(last_path))) at $(last_path)")
    return times
end

println("\nWarmup + benchmarks (period = $(PERIOD), interval = $(INTERVAL))\n")

println("NetCDFOutput")
nc_times = bench("nc", id -> NetCDFOutput(
    SpectralGrid(); interval = INTERVAL,
    path = tmp, id = id, write_restart = false,
    write_parameters_txt = false, write_progress_txt = false,
))

println("\nZarrOutput")
zarr_times = bench("zarr", id -> ZarrOutput(
    SpectralGrid(); interval = INTERVAL,
    path = tmp, id = id, write_restart = false,
    write_parameters_txt = false, write_progress_txt = false,
))

println("\n--- Summary ---")
@printf("NetCDF: min %.3f s, median %.3f s, max %.3f s\n",
    minimum(nc_times), median(nc_times), maximum(nc_times))
@printf("Zarr:   min %.3f s, median %.3f s, max %.3f s\n",
    minimum(zarr_times), median(zarr_times), maximum(zarr_times))

ratio = median(zarr_times) / median(nc_times)
if ratio < 1
    @printf("Zarr is %.1f%% faster than NetCDF (median).\n", (1 - ratio) * 100)
else
    @printf("NetCDF is %.1f%% faster than Zarr (median).\n", (1 - 1 / ratio) * 100)
end
```

milankl:
Also works in Browzarr, @lazarusA 

image

milankl:
Hold on, I just found an error

```julia
julia> add!(model.output, SpeedyWeather.SurfaceTemperatureOutput())
ZarrOutput{Field{Float32, 1, Vector{Floa...}
├ status: active
├ write restart file = true (if active)
├ interpolator::AnvilInterpolator{Float32, RingGrids.GridGeometry{OctaHEALPixGrid{CPU{...}
├ path = /Users/milan/run_0002/output.zarr (overwrite=false)
├ frequency = 21600 seconds
├ time chunk = 1
└ variables
  ├ tsurf: Surface air temperature [degC]
  ├ v: meridional wind [m/s]
  ├ u: zonal wind [m/s]
  └ vor: relative vorticity [s^-1]

julia> run!(simulation, period=Day(1), output=true)
ERROR: DimensionMismatch: 32512-element, 127-ring FullOctaHEALPixField{Float32, 1} as Array on CPU does not match 16384×8, 127-ring OctaHEALPixField{Float32, 2} as Array on CPU
Stacktrace:
  [1] #eachlayer#3
    @ ~/git/SpeedyWeather/RingGrids/src/field.jl:153 [inlined]
  [2] eachlayer
    @ ~/git/SpeedyWeather/RingGrids/src/field.jl:152 [inlined]
  [3] interpolate!(Aout::Field{…}, A::Field{…}, locator::RingGrids.AnvilLocator{…}, geometry::RingGrids.GridGeometry{…})
```

maximilian-gelbrecht:
Yeah, I see it now as well. There are still a few too many output function that are specific to `NetCDFOutput` that take care of this. I am looking into it.

maximilian-gelbrecht:
I think I fixed it. Let's see if it works.

milankl:
I just sorted the code into `output/writers/general.jl` for all methods that weren't specific to netcdf hope this didn't interfere with your commit

lazarusA:
A full chunk in time makes sense when you want to analyse time series, for maps the chunking goes in space. I didn't analyse the full code, just wondering if there is an option for that. To pass your own chunking preferences.

maximilian-gelbrecht:
> A full chunk in time makes sense when you want to analyse time series, for maps the chunking goes in space. I didn't analyse the full code, just wondering if there is an option for that. To pass your own chunking preferences.

Good point. thanks. So far we just had chunking in space, now I made it an option to freely configure this.

maximilian-gelbrecht:
> A full chunk in time makes sense when you want to analyse time series, for maps the chunking goes in space. I didn't analyse the full code, just wondering if there is an option for that. To pass your own chunking preferences.

What would you recommend as a default chunking for ESM outputs?

lazarusA:
Different use cases:

- Optimise for IO (we let DiskArrayEngine.jl do that for us). Note, that is beneficial for high-resolution, <1km. Not familiar enough with SpeedyWhether to judge if this case can actually run.
- Manual chunking
  -  set big chunks in space (for maps, spatial analysis). I like this one for Browzarr, downloads speeds are fast in this case.
  - or set chunks in time for time series.

but, since you are using directly zarr, maybe going with the same heuristics as the netcdf case is the best course of action. And having input arguments to change chunk sizes is good for users to tweak for their own use cases, hence my previous comment.

maximilian-gelbrecht:
Ok, thanks!

milankl:
Lazaro, when you have a time chunk of length N and compression on, when will the compression be done? Ideally after N steps, or would Zarr.jl decompress at every write step to modify the data according to the new step and then compress again? I remember netCDF easily doing the latter if you're not careful.

lazarusA:
Don't know at the moment. Will need to look into this.

SpeedyWeather/Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "SpeedyWeather"
 uuid = "9e226e20-d153-4fed-8a5b-493def4f21a9"
-version = "0.20.0"
 authors = ["SpeedyWeather contributors"]
+version = "0.20.0+DEV"
```
lazarusA, line 4: why this `+DEV`. I remember seeing this in packages before Julia 1.0. Is this on purpose? just curious.

milankl: Yeah when you have several (in our case 5) registered packages within one repo you easily lose track of whether the next release should be breaking or not. So when you add a commit that is non breaking you do +DEV, when you add a breaking commit you bump and -DEV. 

We had the problem of having introduced a new feature as a patch which caused problems as this wasn't present in a compatible pre-patch release

milankl: Minus is semver standard for prereleaae, plus is for built. So Pkg understands that.

lazarusA: I see.

milankl:
Okay Browzarr works with our `output.zarr` directories! Exploring the video export functions in the mean time

https://github.com/user-attachments/assets/f3dbb1ce-5336-47f2-9e42-55bbdf1ba8e5

@pierluigividale and @nviebig I think you'd like that!

milankl (approved):

pierluigividale:
> With T31L8, integrated for 10 days and hourly output, we are 18% faster with `ZarrOutput` compared to `NetCDFOutput`.
> ...

Has anyone tried this at higher resolution? I am happy to try, as I am very interested in the future of zarr vs netcdf for km-scale. But I need to know what code base I need for that.

Also, is HEALPix output on the development horizon?

maximilian-gelbrecht:
Hi @pierluigividale, 

We just merged this, so this is available on `main` and not yet in a released version. Output is currently only supported on full grids. But I think we could easily support Zarr output on Healpix as well, right @milankl ?
So far none of us has tried this on very high resolution. For this it might make sense to think about how to chunk the data, see @lazarusA comment here for that. You can find example code in the updated docs https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/other_output#Zarr-Output

maximilian-gelbrecht:
`OctaHEALPix` in particular would be very easy to output. I am just not quite sure about what are actually the standards on this. We should output it in a way that also other libraries like `xarray` can deal with it easily.

maximilian-gelbrecht:
Let's discuss possible HEALPix output in the new issue though #1080

pierluigividale:
Hi. So, we tried lots and lots of things at the km-scale hackathon last year, and there were recommendations on chunking etc. etc. This is the draft paper: 
https://digital-earths-global-hackathon.github.io/gmd-paper/

PR #1077, opened 2026-05-06
Version bump 0.20

maximilian-gelbrecht:


maximilian-gelbrecht:
The Reactant CI error seems to be a Reactant issue and not from us.

maximilian-gelbrecht:
I can't reproduce this pre-compilation error from `ReactantCUDAExt` locally even with the same CUDA and Reactant version. Strange.

maximilian-gelbrecht:
Ok, they are on it to fix that issue over at Reactant.

PR #1078, opened 2026-05-06
Also have non-unicode spectral gradient operators

milankl:
As requested by William Greenall (@wegreenall)

introduces the following aliases

```julia
laplace = ∇²
laplace! = ∇²!
inverse_laplace = ∇⁻²
inverse_laplace! = ∇⁻²!
gradient = ∇
gradient! = ∇!
```

for the (only) unicode-based exported functions we have. I think it's a good philosophy to have user-facing code ascii only. Internally I'm still a big fan of greek symbols etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Add non-unicode function name aliases laplace, inverse_laplace, gradient [#1078](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1078)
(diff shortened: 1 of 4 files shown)


maximilian-gelbrecht (approved):

milankl:
@wegreenall Any further or comments that should go in here? All tests pass so this can be merged otherwise

wegreenall:
I have not had chance to do any checks for myself, but if you are happy
with adding these then of course you should go ahead. I will try next week
when back in the office to implement what we were discussing RE linearised
model/adjoint and I will let you know if I hit any unforeseen problems.

Thanks for this! Greatly appreciated and lovely to meet you at EGU.

PR #1082, opened 2026-05-08
Multithread legendre transform

maximilian-gelbrecht:
Work in progress for our speedy friends from Bern.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Multithread `_legendre!` previously the major bottleneck for multithread CPU performance [#1082](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1082)
 - `ZarrOutput` for writing simulation output to a Zarr store, implemented as a Zarr.jl extension [#1076](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1076)
 - Add non-unicode function name aliases laplace, inverse_laplace, gradient [#1078](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1078)
 
--- a/SpeedyTransforms/src/legendre.jl
+++ b/SpeedyTransforms/src/legendre.jl
@@ -56,43 +56,71 @@ function _legendre!(
     @boundscheck ismatching(S, specs) || throw(DimensionMismatch(S, specs))
     @boundscheck size(g_north) == size(g_south) == (S.nfreq_max, S.nlayers, nlat_half) || throw(DimensionMismatch(S, specs))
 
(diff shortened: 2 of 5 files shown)


maximilian-gelbrecht:
We get some warnings from Enzyme that I'll have to check explicitly. In #1032 we also have some trouble with multi threading and Enzyme

maximilian-gelbrecht:
For low resolution grids like T21 the overhead of multithreading the Legendre transform is quite significant. Low-overhead multithreading via e.g. Polyester.jl or LoopVectorization.jl is fundamentally incompatible with Enzyme as they do some fishy things (https://github.com/EnzymeAD/Enzyme.jl/issues/3093). So that's ruled out as well. 

For mid-high resolutions the multithreading is certainly worth it though, so we might still just merge this. 

Here are some benchmarks

trunc | dir | t=1 | t=2 | t=3 | t=4
-- | -- | -- | -- | -- | --
T21 | inverse | 0.015 (0.90×) | 0.013 (1.06×) | 0.010 (1.37×) | 0.019 (0.75×)
T21 | forward | 0.016 (0.94×) | 0.021 (0.73×) | 0.017 (0.90×) | 0.025 (0.60×)
T31 | inverse | 0.042 (0.95×) | 0.034 (1.16×) | 0.021 (1.96×) | 0.019 (2.10×)
T31 | forward | 0.044 (0.97×) | 0.037 (1.15×) | 0.030 (1.41×) | 0.032 (1.34×)
T63 | inverse | 0.325 (1.00×) | 0.184 (1.77×) | 0.127 (2.57×) | 0.098 (3.32×)
T63 | forward | 0.326 (0.98×) | 0.215 (1.49×) | 0.161 (1.99×) | 0.153 (2.09×)
T127 | inverse | 2.556 (1.00×) | 1.335 (1.91×) | 0.919 (2.79×) | 0.720 (3.56×)
T127 | forward | 2.462 (0.99×) | 1.452 (1.67×) | 1.052 (2.31×) | 0.941 (2.59×)

So T21 the overhead is just too much. T31 is giving a speed up but doesn't scale super well, T63 and above is good with multithreading. 

I just tested the complete model as well and at T31L8 you don't get much of a speed up from multithreading, especially also considering that the FFT still has to be single threaded due to some bug we currently don't understand (https://github.com/SpeedyWeather/SpeedyWeather.jl/issues/919). Higher resolutions give speed ups with multithreading though.

maximilian-gelbrecht:
Hey @christianwirths @ursho2552, so multithreading Speedy in low resolutions is probably not directly worth it for you. However, there probably still things in the set up you can do to speed it up: like for example a different reduced grid, disabling parameterisations you don't need, less horizontal layers etc

PR #1083, opened 2026-05-10
Allow variables to be fused for GPU optimisation 

maximilian-gelbrecht:
This is some preliminary work for the dycore rewrite to make it easier to batch transforms together. 

This allows variables to be fused together into one larger array, with views defined so that the original variables still work as intended (hopefully). It does that by adding a field `fuse` to all variable types. All variables with the same `fuse` are fused together. 

We can still think about where actually we actually want to put those fused variables in our variable hierarchy. Currently they live in `variables.scratch.fused` and the views live in their original spot to ensure that everything else is still working exactly as before.


maximilian-gelbrecht:
Over at #1084, I tested actually using this. 

1) ~~The GPU CI error seems like an unrelated Reactant issue~~ GPU error got fixed externally 
2) On GPU a contiguous view on a `CuArray` results in a `CuArray`, not a `SubArray`. They do share their memory though and don't allocate, so it is a view. However, this does make things a bit intransparent. I confirmed that memory usage on GPU in #1084 is exactly equal to `main` 
3) JLDOutput currently destroys this view - parent relationship. They are now explicitly materialised (otherwise it gives an error and doesn't work at all).

Everything else seems to work and CPU benchmarks look fine as well.

SpeedyWeather/src/variables/dimensions.jl
```diff
@@ -83,3 +83,71 @@ Base.zero(::AbstractVariable{TransformScratchMemory}, model::AbstractModel) = mo
...
+    parent = zeros(model.spectral_grid.SpectralVariable3D, model.spectral_grid.spectrum, length(vars))
+    views = [lta_view(parent, :, k) for k in 1:length(vars)]
+    return parent, views
+end
+
+# fallback: clear error message for unsupported dim types
+allocate_fused(::AbstractVector{<:AbstractVariable{D}}, ::AbstractModel) where {D} =
+    error("Fusion is not implemented for dim type $D. Supported: Grid2D, Grid3D, Spectral2D, Spectral3D.")
+
+function _split_views_grid3d(parent, vars, model)
```
maximilian-gelbrecht: This here is the critical and somewhat fragile point of the code. 
It's not completely clear to me what determines the order of `vars` and this determined than also the split into the views. Currently this works as intended for the example in #1084, but we should maybe make this a bit safer.

maximilian-gelbrecht: Added some additional checks now.

SpeedyWeather/test/variables/fuse.jl
maximilian-gelbrecht, line 1: These are currently just Claude generated. I still have to review them myself properly

maximilian-gelbrecht:
@milankl Do you agree with the strategy of merging this first, and then do the actual batching in follow up PRs? I would like to avoid another mega PR.

milankl:
Yes happy to split it into smaller PRs! Good strategy

SpeedyWeather/src/variables/dimensions.jl
```diff
@@ -83,3 +83,87 @@ allocate(::AbstractVariable{TransformScratchMemory}, model::AbstractModel) = mod
...
+fused_slots(::Grid3D, model::AbstractModel) = get_nlayers(model)
+fused_slots(::Grid2D, ::AbstractModel) = 1
+fused_slots(::Spectral2D, ::AbstractModel) = 1
+fused_slots(d::Spectral3D, model::AbstractModel) = d.n == 0 ? get_nlayers(model) : d.n
+
+fuse_family(::Grid2D) = :grid
+fuse_family(::Grid3D) = :grid
+fuse_family(::Spectral2D) = :spectral
+fuse_family(::Spectral3D) = :spectral
+fuse_family(d::AbstractVariableDim) = error(
```
milankl, line 128: With #1035 several 3D variables become 4D variables, would it be a pain to unravel those dimensions into the "slot" dimension?

SpeedyWeather/src/variables/dimensions.jl
```diff
@@ -83,3 +83,87 @@ allocate(::AbstractVariable{TransformScratchMemory}, model::AbstractModel) = mod
...
+            field_view(parent, :, offset + 1) :    # scalar → 2D Field (no layer dim)
+            field_view(parent, :, slots[i])        # range  → 3D Field (with layer dim)
+        offset += n
+    end
+    @assert offset == size(parent.data, 2) "Fused grid parent has $offset slots assigned but parent has $(size(parent.data, 2)) layers"
+    return views, slots
+end
+
+# Mirror of `_split_grid_views` for the spectral family.
+function _split_spectral_views(parent, vars, model)
```
milankl: These two methods `_split_grid_views` and `_split_spectral_views` could be merged if we actually extend `Base.view` with `field_view` and `lta_view` right? I remember us discussing this but we wanted to be careful. Not sure that's care is needed still

SpeedyWeather/src/variables/variables.jl
```diff
@@ -187,46 +193,209 @@ end
...
-    grid              = allocate(filter_variables(all_vars,             GridVariable), model)
-    tendencies        = allocate(filter_variables(all_vars,         TendencyVariable), model)
-    dynamics          = allocate(filter_variables(all_vars,         DynamicsVariable), model)
-    parameterizations = allocate(filter_variables(all_vars, ParameterizationVariable), model)
-    particles         = allocate(filter_variables(all_vars,         ParticleVariable), model)
-    scratch           = allocate(filter_variables(all_vars,          ScratchVariable), model)
+
+    # First pass (cross-type): allocate one parent per (namespace, fuse_symbol) group, spanning
+    # all variable types. Returns a Dict keyed by (namespace, fuse) with values (parent, view-by-identifier).
+    fuse_parents = build_fuse_parents(all_vars, model)
```
milankl, line 239: Is that costly for compile time? I remember with our dynamic variable system the compile time jumped to 20s or so. Not the worst, but just checking that this doesn't get out of hand as we hardcode less and less and offload more and more to the compiler

maximilian-gelbrecht:
Than I am glad if you give it a quick review and we can go ahead ;) 

Currently this is completely non-instrusive though, so nothing really that can go wrong.

milankl (commented):
I like the idea to fuse variables but also there's some conflicts that will arise in comparison to #1035 and I don't think I've understood yet how we anticipate to solve them. E.g. in Leapfrog we have variables A, B, C but 2 steps of each (call 1, 2)

1. currently in memory we have A1, A2 as one array
2. naively fusing them would yield A12-B12-C12 but transform have to operate on A2-B2-C2 only, so skipping every other 3D array
3. this is a problem for the spectral -> grid transforms at the end of a time step
4. but not for the grid -> spectral transforms at the end of the dynamical core before the prognostic update

In N-Cycle Lorenz we only have prognostic step but two tendency steps so problem 3 goes away but the same problem in the form of 4 appears here. 

Currently I see options

- we could only fuse on where possible given step constraints but it seems we'd likely only make use of half of the potential
- we could split the variables, e.g. A1-B1-C1 as one fused array, A2-B2-C2 as another fused array however that makes the `get_prognostic_step` function unwieldy
- we could reorder the variables, e.g. A1-B1-C1-A2-B2-C2 as one fused array and then A is a view with a stride but that wouldn't allow a mix of 2D and 3D variables as we'd have different stride lengths

SpeedyWeather/src/variables/dimensions.jl
maximilian-gelbrecht: Yes, true. However, I am still somewhat opposed to extending `Base.view` as I would expect that `view` returns a `SubArray` directly. Even if CUDA.jl also ignores it. 

However, we could just rename `field_view` and `lta_view` to `wrapped_view` with dispatch defined. Then it's also just one `_split_views`.

milankl: Yeah I like that! Removes code redundancy. `wrapped_view` would need to be defined in SpeedyWeatherInternals though with both RingGrids and LTAs extending it correct?

milankl: But maybe just easier to define it at the SpeedyWeather level simply doing

```julia
wrapped_view(field::Abstractfield, args...) = field_view(field, args...)
wrapped_view(coeff::LowerTriangularArray, args...) = lta_view(coeff, args...)
```

that way we would consider `wrapped_view` to be private for SpeedyWeather and we could use it internally instead of field/lta_view?

maximilian-gelbrecht: It could also live in `SpeedyTransforms`. There are already one or two utilises for the array types that need both packages imported anyway. 

I'll address this in another commit later, and also think about your main comment a bit more ;)

SpeedyWeather/src/variables/variables.jl
maximilian-gelbrecht: I still have to benchmark it, but the overall test time of the complete unit tests didn't go up with it, so I can't be that bad.

maximilian-gelbrecht:
> I like the idea to fuse variables but also there's some conflicts that will arise in comparison to #1035 and I don't think I've understood yet how we anticipate to solve them. E.g. in Leapfrog we have variables A, B, C but 2 steps of each (call 1, 2)

Oh yes that are some really good points. To be honest, I wasn't considering that NCycle needs two tendency steps. I kind of forgot about that. 

> * we could reorder the variables, e.g. A1-B1-C1-A2-B2-C2 as one fused array and then A is a view with a stride but that wouldn't allow a mix of 2D and 3D variables as we'd have different stride lengths

I am in favour of that option. But I don't think mixing 2D and 3D or 3D and 4D is really a problem, it just doesn't work to mix 2D, 3D AND 4D together, but this also doesn't exist anywhere, right?

So let's make it concrete: 

```Julia 
N_grid = 40
N_levels = 8
N_steps = 2 

A = rand(N_grid, N_levels, N_steps)
B = rand(N_grid, N_levels, N_steps)
C = rand(N_grid, 1, N_steps) 

# Concatenate along columns/layers (dim 2)
# Dimensions: (40, 8 + 8 + 1, 2) -> (40, 17, 2)
final_array = cat(A, B, C; dims=2)

vstep1 = @view final_array[:, :, 1] # that's contiguous in memory 
vstep2 = @view final_array[:, :, 2] # that's contiguous in memory 

vA = @view final_array[:, 1:N_levels, :] # that's strided, but is that bad? 
vB = @view final_array[:, N_levels+1:2*N_levels, :]
vC = @view final_array[:, 2*N_levels+1, :]

vA1 = @view final_array[:, 1:N_levels, 1] # that's also contigious
```

This would give a contagious A1, B1, C1 and A2, B2, C2 and also work out with the views. It works with CUDA as well `vA1` is a `CuArray` again (indicating it's a contiguous view), while `vA` is a `SubArray` (indicating it's a non-contiguous view on a `CuArray`).

So the question is, is there a point in the time stepping where we actually wouldn't want this memory layout? I do think it's fine for the leapfrog, how is it with the NCycle?

SpeedyWeather/src/variables/dimensions.jl
maximilian-gelbrecht: I did that now, but I didn't adjust it across the whole code base. `field_view` and `lta_view` are still used in plenty of places, but I guess that's also fine.

maximilian-gelbrecht:
Ok, the last commit actually added the behaviour described above. So it should work with 4D variables as well now. 

Now we can fuse together 
* 2D (ij/lm) and 3D (ij/lm, n_layers)
* 3D (ij/lm, n_steps) and 4D(ij/lm, n_layers, n_steps)

3D can mean two different things, that's also what makes the code slightly more complex.

I need to give this a proper test that it actually works in a model again when used. One thing in particular I need to check is if the `get_*_step` routines really work as intended and efficiently, also on GPU. Because on GPU strided views and contiguous views are handled slightly differently. In the end every access to a concrete A1, B2, etc is contagious again but CUDA also has to get that.

SpeedyWeather/src/variables/dimensions.jl
maximilian-gelbrecht: See the my other comment, with the code example, I think there the logic is clear.

maximilian-gelbrecht:
As I feared views on views on GPU lose the ability to notice if an array is contiguous or not easily. 

Example:
```Julia 
using CUDA

A = CUDA.rand(10, 5, 2)
B = @view A[:,1:2, 1] # -> CuArray contigous
C = @view A[:,1:2, :] # -> SubArray non-contiguous
B_2 = @view C[:,:, 1] # -> SubArray even though it is contiguous (same as B)
```

We can either do some unsafe reinterpreting of the pointers. That's probably more of a last resort. 

My preferable option is just to _always_ fuse all atmospheric prognostic variables and tendencies I think. Then the `get_prognostic_step` and `get_tendency_step` would be defined only on those fused arrays and the problem is circumvented. This does interfere a bit with #1035, so what are your thoughts @milankl ? Immediately I don't see a direct downside of this, but maybe I am also overlooking something. 

(The failing unit tests for the restart is somewhat related to this and I'll resolve it once we make a decision)

SpeedyWeather/src/variables/variables.jl
milankl: Bit confused how `summarysize` works for views

```julia
julia> A = rand(100, 200);

julia> B = view(A, 1:50, 1:100); # should be 1/4 of the size of A

julia> sizeof(B) / sizeof(A)
0.25

julia> Base.summarysize(B) / Base.summarysize(A)
1.0003498950314906
```

So it seems the summarysize of a view is always the summarysize of the parent array (regardless the size of the view). This probably means that the size summary we give here is correct even though it's not the sum of its parts (which is correct with fused variables) -- but might be worth checking!

docs/src/variable_system.md
```diff
@@ -130,4 +130,8 @@ simulation = initialize!(model)
 simulation.variables.scratch.my_dict
 ```
 
-a dictionary as a variable!
\ No newline at end of file
+a dictionary as a variable!
+
+## Variable fusion 
+
+Variables can be "fused" together to be allocated as one block of contigous memory. This is done primarly with GPU-optimiziation in mind, so that e.g. `transform!` calls can be batched together not just across levels, but also across different variables. All variables that are defined with the same `fuse=:fuse_name` keyword argument are allocated together. At the same time a view is defined that allows for the variable to be used as it would without the variable fusion. As this is primarily a performance optimization, it is not required to run basic custom parameterizations or model components.
```
milankl:
```suggestion
Variables can be "fused" together to be allocated as one block of contiguous memory. This is done primarly with GPU-optimiziation in mind, so that e.g. `transform!` calls can be batched together not just across levels, but also across different variables. All variables that are defined with the same `fuse = :fuse_name` keyword argument are allocated together. At the same time a `view` is defined that allows for the variable to be used as it would without the variable fusion. As this is primarily a performance optimization, it is not required to run basic custom parameterizations or model components. 
```

SpeedyWeather/src/variables/variables.jl
```diff
@@ -98,14 +104,32 @@ end
...
+and direct assignment for `Ref` values.
+
+View leaves (Fields/LTAs whose `.data` is a `SubArray`, or bare `SubArray`s) are
+skipped: their underlying storage lives in a fuse parent that is itself reachable
+through `vars.scratch.fused.*`, and that parent is copied directly. After the
+parent is updated, every view that aliases it sees the new data automatically.
+Copying through both view and parent is unnecessary work, and on cross-architecture
+copies (CPU ↔ Reactant) the broadcast-aliasing check inside `copyto!` fails.
+Skipping view leaves avoids both problems."""
 function Base.copy!(dest::Variables, src::Variables)
```
milankl, line 119: What happens if we copy variables across and while the views exist, the fused parent array exists in only one? Just wondering what happens in restart purposes: If we use JLD2 for restart file output which materialises the views then do we skip writing the parent (which would be a redundant copy) but running an identically initialized simulation would then allocate the fused parent array?

milankl (commented):
Almost there! Thanks for implementing the `fuse` group!

SpeedyWeather/src/variables/variables.jl
maximilian-gelbrecht: Yeah, good point. 

When the fused parent exists in both `src` and `dest` it is no problem. But it is right that you question what happens if it doesn't. 

First of all, we could also just naively copy everything always. It's just that this is a bit wasteful if you copy both with the view and the parent because data gets copied twice. 

I think we could just flip the logic that is implemented there just below. Currently the copy is skipped once the `dest` is a view 
```Julia 
_copy_entry!(dest::AbstractArray, src::AbstractArray) =
    is_view_entry(dest) ? dest : copyto!(dest, src)
```

But what we actually want is that the copy is skipped if the `src` is a view: When the `src` is a view, then a parent has to exist. Can we then assume that the same parent exists in `dest`? In almost all situations yes, and it would be fine. I am just wondering if we want to be extra certain. We could either just add an extra check that the same fused parents exist in both or we just accept that we copy some data twice and don't bother with skipping over views at all.

SpeedyWeather/src/variables/variables.jl
maximilian-gelbrecht: Ok, I didn't consider `summarysize` at all. 

If the `view`s count fully as their parents then we want to exclude them, right? `Base.summarysize` already has a built-in option for excluding certain types, so that easy to do. Well at least on CPU, on GPU it isn't. But memory information from CPU routines for GPU are super unreliable across the board, so maybe we just shouldn't bother with it in this case.

SpeedyWeather/src/variables/variables.jl
milankl: I think it's okay for now to restrict ourselves to three cases

1. Both `src` and `dest` have the same views and parents (either copy views or parents is fine and we can assume parents to match in size)
2. `src` has only materialised arrays, no parents (copy arrays to views)
3. `dest` has only materialised arrays, no parents (copy views to arrays)

If I see this right "skip if the `src` is a view" would go against supporting case 3? And "skip if `dest` is a view" would go against case 2. But looking at JLD2 output arguably case 2 is more important?

SpeedyWeather/src/variables/variables.jl
maximilian-gelbrecht: Ok done in the last commit.

SpeedyWeather/src/variables/variables.jl
maximilian-gelbrecht: So I think the proper logic is. When looking at two arrays from the same entry:  

A) `src` and `dest` are both views -> skip copy (parent is copied)
B) `src` is a view and `dest` isn't   -> copy 
C) `src` isn't a view but `dest` is    -> copy 
D) Neither `src` nor `dest` are views -> copy

That's what I implemented now. 

Situation B and C) are avoiding a double copy because if one doesn't have a fused entry, it also isn't copied.

milankl: That sounds great and should cover my three use cases 1, 2, 3?

maximilian-gelbrecht: Yep.

PR #1087, opened 2026-05-11
Variable system name `allocate` instead of `Base.zero`

maximilian-gelbrecht:
`allocate` is a bit more general than extending `Base.zero`

cc @bgroenks96

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


bgroenks96:
To be clear, it's still fine and helpful to extend `Base.zero`, but then it should actually set everything (I guess also the `clock`) to its natural zero.

milankl (approved):
Yes, sounds great to me! Otherwise we do end up abusing `Base.zero...` I agree with Brian

milankl:
Reminder to bump versions to `+DEV`, this should have been "0.20.1+DEV"

PR #1088, opened 2026-05-11
logo is in public now

lazarusA:
this was not updated when doing the hero section. Now, it should render correctly.


maximilian-gelbrecht:
Thank you!

PR #1089, opened 2026-05-12
Document AnalyticBandRadiation

maximilian-gelbrecht:
This documents the usage of AnalyticBandRadiation.jl https://github.com/NumericalEarth/AnalyticBandRadiation.jl and also provides a CO2 forcing experiment. The script is quite slow though (I need to look into it if there's something wrong, even though it is expected that it is slower than our regular default model of course as the radiation is significantly more expensive)

### DIFF
--- a/docs/src/co2_forcing.md
+++ b/docs/src/co2_forcing.md
@@ -1,0 +1,127 @@
+# CO₂ forcing experiments 
+
+With our [greenhouse gases](radiation.md#greenhouse-gases) and [radiation interface](@ref radiation-ssm), we can also run idealized climate forcing scenarios. We will perform an experiment in which we let the CO₂ concentration rise exponentially starting from 280 ppmv and compare the globally averaged surface temperature to a control run in which we keep the CO₂ concentration constant at 280 ppmv. 
+
+To track the globally averaged surface temperature, we first set up a callback that computes and saves it. Then we set up the two simulations, both using the Simple Spectral Radiation Model of [`AnalyticBandRadiation.jl`](@ref radiation-ssm). 
+
+```julia 
+using Statistics
+using Dates
+using Makie
+
+using SpeedyWeather
+using AnalyticBandRadiation
+const SpeedyAnalyticBand = Base.get_extension(AnalyticBandRadiation, :AnalyticBandRadiationSpeedyWeatherExt)
+
+# Custom callback to compute and save global mean surface temperature
+mutable struct GlobalMeanTemperatureCallback <: SpeedyWeather.AbstractCallback
+    temperatures::Vector{Float32}
+    timestep_counter::Int
+    save_interval::Int  # save every N timesteps
+end
+
+GlobalMeanTemperatureCallback(; save_interval=1) =
+    GlobalMeanTemperatureCallback(Float32[], 0, save_interval)
+
+function SpeedyWeather.initialize!(
+    callback::GlobalMeanTemperatureCallback,
+    ::SpeedyWeather.Variables,
+    ::SpeedyWeather.AbstractModel,
+)
+    # Pre-allocate storage for one sample per day
+    # Assuming ~24 timesteps per day with typical parameters
+    n_days = 365 * 10  # 10 years
+    callback.temperatures = zeros(Float32, n_days)
+    callback.timestep_counter = 0
+    return nothing
+end
+
+function SpeedyWeather.callback!(
+    callback::GlobalMeanTemperatureCallback,
+    vars::SpeedyWeather.Variables,
+    model::SpeedyWeather.AbstractModel,
+)
+    callback.timestep_counter += 1
+
+    # Save every N timesteps
+    if mod(callback.timestep_counter, callback.save_interval) == 0
+        save_index = div(callback.timestep_counter, callback.save_interval)
+
+        # Get surface layer temperature (last layer)
+        nlayers = model.geometry.nlayers
+        temp_grid = vars.grid.temperature[:, nlayers]
+
+        # Compute global mean
+        global_mean = mean(temp_grid)
+
+        # Store result
+        if save_index <= length(callback.temperatures)
+            callback.temperatures[save_index] = global_mean
+        else
+            push!(callback.temperatures, global_mean)
+        end
+    end
+    return nothing
+end
+
+# Setup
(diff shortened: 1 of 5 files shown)


docs/src/co2_forcing.md
```diff
@@ -0,0 +1,127 @@
...
+To track the globally averaged surface temperature, we first set up a callback that computes and saves it. Then we set up the two simulations, both using the Simple Spectral Radiation Model of [`AnalyticBandRadiation.jl`](@ref radiation-ssm). This script takes significant compute, so that it is not built live in the documentation. 
+
+```julia 
+using Statistics
+using Dates
+using Makie
+
+using SpeedyWeather
+using AnalyticBandRadiation
+const SpeedyAnalyticBand = Base.get_extension(AnalyticBandRadiation, :AnalyticBandRadiationSpeedyWeatherExt)
```
maximilian-gelbrecht, line 14: I consider this a temporary interface. In the long run we might either define function stubs in `AnalyticBandRadiation` or make `AnalyticBandRadiation` a dependency of SpeedyWeather and just do that here, so it is easier to import. 

I find it annoying that Julia forbids to define new exported symbols in extensions...

maximilian-gelbrecht: @milankl What do you think about just defining a function stub `Parameterization` in SpeedyWeather. And then all external libraries that define parameterisations can use that to extend it and we avoid this annoying no-exports-from-extensions problem like that? Basically:

```Julia 
external_param = CustomParam(....) # defined in main code of package like AnalyticRadiation without SpeedyWeather dependency 
param = SpeedyWeather.Parameterization(spectral_grid, external_param; kwargs...) # load into SpeedyWeather, extended in extension.

**Review by milankl-claude — COMMENTED** [review]
Thanks for documenting this, it's a nice showcase. A few things need fixing before it can go in.

**Things that break or give wrong results**
1. **The package has been renamed.** AnalyticBandRadiation.jl's `Project.toml` is now `name = "NumericalRadiation"` (same UUID), and the extension is `NumericalRadiationSpeedyWeatherExt`. So `using AnalyticBandRadiation` and `:AnalyticBandRadiationSpeedyWeatherExt` are out of date. That will probably break the `@example radiation` block in `radiation.md` when the docs build.
2. **The CO₂ experiment never raises CO₂.** `model_exp` is set up exactly like the control run (`radiation` is even constructed twice), so the plot compares two control runs. It needs the exponential CO₂ trajectory, e.g. via `CO2(f)` from the greenhouse gas interface.
3. **Callback order and Δt differ between runs.** In the experiment the callback is `add!`ed after `initialize!`; in the control run it's before. The experiment's `save_interval` also uses `model_control.time_stepping.Δt_sec`.
4. **The plotting code won't run.** `grid!(ax, alpha=0.3)` isn't a Makie function (grid lines are an `Axis` attribute), and `using Makie` alone has no backend to `save` a PNG. Use `CairoMakie`.
5. **The header anchor syntax is wrong.** `## William's ... (@id radiation-ssm)` should be `## [Williams' simple spectral model ...](@id radiation-ssm)`, otherwise `@ref radiation-ssm` won't resolve.

**Smaller issues**
- `trunc=31` is deprecated; use `truncation=31` in both files.
- `mean(temp_grid)` isn't area-weighted, so on a Gaussian grid it's a biased global mean. Also, `temperature[:, nlayers]` is lowest-layer air temperature, not surface temperature, so the axis label and text should say so.
- Pre-allocating `365 * 10` slots and then filtering out `.!= 0` is fragile. Just `push!` in the callback.
- Typos: "William's" should be "Williams'"; "the[documentation" is missing a space; in the warning box, "Only the `AnalyticBandRadiation` uses it" reads oddly; `co2_forcing.md` and `radiation.md` are missing trailing newlines; `make.jl` is missing a trailing comma after the new entry.

On the interface question: I've opened #1261 with the `Parameterization` stub you suggested. Once it and a one-line method in NumericalRadiation's extension are in, the `Base.get_extension` lines here can become `Parameterization(spectral_grid, AnalyticBandLongwave())`.

File: docs/src/co2_forcing.md (hunk 1 above)
Comment (milankl-claude) [reply docs/src/co2_forcing.md:14]: @maximilian-gelbrecht Good idea, and we already do the same thing for `TerrariumOutput` and `animate`. It's in #1261: `Parameterization` is exported as a function stub, with a pass-through method for things that are already an `AbstractParameterization`. The NumericalRadiation extension would then only need

```julia
SpeedyWeather.Parameterization(SG::SpectralGrid, scheme::AnalyticBandLongwave; kwargs...) =
    SpeedyAnalyticBandLongwave(SG, scheme; kwargs...)
```

and users write `Parameterization(spectral_grid, AnalyticBandLongwave())` without `Base.get_extension`. A possible follow-up is for the model constructors to call `Parameterization` on every parameterization keyword automatically, so that `PrimitiveWetModel(spectral_grid; longwave_radiation = AnalyticBandLongwave())` just works.

milankl-claude:
One more thing I missed in the review: since #1239, references live centrally in `docs/src/references.bib` and are rendered with DocumenterCitations on the [References](https://speedyweather.github.io/SpeedyWeather.jl/dev/references) page. There are no more footnotes at the end of each markdown file. `radiation.md` on `main` already cites everything this way, e.g. `[Frierson2006](@citet)`. So after rebasing onto `main`:

- drop the `[^W26]: Williams, A.I.L. ...` footnote and add a bibtex entry to `docs/src/references.bib` instead, e.g.

  ```bibtex
  @article{Williams2026,
    author = {Williams, A. I. L.},
    title = {Bridging clarity and accuracy: A simple spectral longwave radiation scheme for idealized climate modeling},
    journal = {Journal of Advances in Modeling Earth Systems},
    year = {2026},
    volume = {18},
    pages = {e2025MS005405},
    doi = {10.1029/2025MS005405}
  }
  ```
  (`Williams2009`/`Williams2011` are already taken by P. D. Williams, so the key needs to stay distinct)
- replace the `[^W26]` citations with `[Williams2026](@citet)` (or `@citep`)
- the diff context still shows the old `[^KMB06]`/`[^FH06]` footnotes, so this will probably also conflict in `radiation.md` until it's rebased

maximilian-gelbrecht:
This PR will be superseded soon by another PR that actually implements an extension for NumericalRadiation here. Until then this can stay open though.

PR #1090, opened 2026-05-12
Coupling to Terrarium via TerrariumExtension 

maximilian-gelbrecht:
This takes the existing script from @bgroenks96 over at Terrarium.jl and turns it in an extension for SpeedyWeather. 

In this approach Terrarium just lives within SpeedyWeather as its land model. It is kind of Frankensteining together Speedy and Terrarium until we have a more stable coupling via NumericalEarth.jl

- [x]  I have to critically look at the code again as part of it was just Claude generated 
- [x] We'll discuss it together
- [x] Once we are happy with it we define proper bindings (probably with some function stubs in the main module), so that users don't have to load the extension via `Base.get_extension` anymore


SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -0,0 +1,302 @@
...
+        input_variables::Tuple = (),
+        initializers::NamedTuple = (;),
+        fields::NamedTuple = (;),
+        Δt::Real = 300,
+    ) where {NF}
+    field_grid = Terrarium.get_field_grid(model.grid)
+    Δz_arr = on_architecture(Terrarium.CPU(), field_grid.z.Δᵃᵃᶜ)
+    # The Oceananigans vertical spacing is an OffsetArray; take the last entry
+    # as the bottom-most layer thickness for the SpeedyWeather LandGeometry.
+    geometry = LandGeometry(1, NF[Δz_arr[end]])
```
milankl, line 122: shouldn't this be `on_architecture`d back to SpeedyWeather's architecture?

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -0,0 +1,302 @@
...
+        PrognosticVariable(
+            name = :terrarium, dims = TerrariumVars(),
+            namespace = :land, desc = "Terrarium land state",
+        ),
+        # Thin grid-side mirrors of surface soil temperature / moisture so the
+        # rest of SpeedyWeather (longwave/shortwave radiation, surface fluxes,
+        # output writers) can read them. They are kept in sync from the
+        # Terrarium state inside `initialize!` and `timestep!`.
+        PrognosticVariable(
+            name = :soil_temperature, dims = Land3D(),
```
milankl: If "thin" why not `Land2D()` ?

maximilian-gelbrecht:
Before this works https://github.com/NumericalEarth/Terrarium.jl/pull/117 has to be merged

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -0,0 +1,296 @@
...
+* [`terrarium_fields`](@ref)`(land)` → `NamedTuple`
+* [`terrarium_substep`](@ref)`(land)` → seconds (`Real`)
+* [`set_terrarium_timestepper!`](@ref)`(land, ts)` (only needed if the
+  timestepper requires state-dependent initialization)
+"""
+abstract type AbstractTerrariumLandModel <: AbstractLand end
+
+@inline get_nlayers(land::AbstractTerrariumLandModel) = land.geometry.nlayers
+
+"""$(TYPEDSIGNATURES) Return the underlying Terrarium `AbstractModel`."""
```
maximilian-gelbrecht: I am a bit indifferent to getter functions like this. We can also just get rid of them.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -0,0 +1,425 @@
...
+    set!(inputs.snowfall, snow)
+    set!(inputs.windspeed, wind)
+    set!(inputs.surface_shortwave_down, Rsd)
+    set!(inputs.surface_longwave_down, Rld)
+
+    # Constructing ModelIntegrator is allocation-free: it is an immutable struct
+    # of references so no data is copied.  InputSources() is empty intentionally:
+    # we own the input-update cycle above (via set!) and do not want Terrarium's
+    # update_inputs! to overwrite those values during the substeps.
+    integrator = ModelIntegrator(
```
maximilian-gelbrecht, line 235: This is a crucial point. 

Constructing the `ModelIntegrator` just here on the fly is a consequence of the `StateVariables` being initialised by SpeedyWeather's variable system in the coupling here above in the ~~`Base.zero` extension~~ `allocate` and not by the `ModelIntegrator` call as usual in Terrarium. However, this construction here should just assemble the existing references and should be basically free. So I think it's okay? I am not 100% certain though this is really what we want to do.

maximilian-gelbrecht: Probably best to just confirm this via some benchmarks

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
milankl: I think this depends on how much Brian expects to change these locations. I think getter functions have a point of creating a unified interface and can also bypass an otherwise breaking change -- but they can also just create clutter if not actually needed

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -0,0 +1,425 @@
...
+        # The full Terrarium state, owned by SpeedyWeather's Variables tree.
+        PrognosticVariable(
+            name = :terrarium, dims = TerrariumVars(),
+            namespace = :land, desc = "Terrarium land state",
+        ),
+        # Thin grid-side mirrors of surface soil temperature / moisture so the
+        # rest of SpeedyWeather (longwave/shortwave radiation, surface fluxes,
+        # output writers) can read them. They are kept in sync from the
+        # Terrarium state inside `initialize!` and `timestep!`.
+        PrognosticVariable(
```
maximilian-gelbrecht: Fundamentally this is probably hard to avoid to have these double as one is a `SpeedyWeather.Field` and the other within the `Terraium.StateVariables` are `Oceananigans.Field`s.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
milankl: Is `ModelIntegrator` like `Simulation`? ("-or" sounds like an operator to me but then it also contains the state variables, right?) I think generally mutable objects will be reused as a flat copy then in `ModelIntegrator` but immutable objects may be copied

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -0,0 +1,425 @@
...
+        state, terrarium_initializers(land), terrarium_timestepper(land),
+    )
+    Terrarium.run!(integrator; period = vars.prognostic.clock.Δt, Δt = terrarium_substep(land))
+
+    # Push Terrarium output back into SpeedyWeather coupling variables.
+    # The Terrarium state itself is mutated in place above and remains in
+    # `vars.prognostic.land.terrarium`; the soil mirrors are refreshed so
+    # SpeedyWeather radiation / surface flux components see current values.
+    vars.prognostic.land.soil_temperature .= state.skin_temperature .+ NF(273.15)
+    vars.prognostic.land.soil_moisture .= interior(state.saturation_water_ice)[:, 1, end]
```
milankl: The slicing here triggers allocations no?

milankl: Use a view instead?

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -0,0 +1,425 @@
...
+    fields::FL
+    "Terrarium-internal sub-step (seconds) used to integrate within each SpeedyWeather step"
+    Δt::Float64
+end
+
+"""$(TYPEDSIGNATURES)
+
+Construct a [`TerrariumDryLand`](@ref) from a Terrarium soil model and a
+SpeedyWeather spectral grid."""
+function TerrariumDryLand(
```
milankl: The information that this is a dry land model isn't actually passed on to Terrarium, is it? It's not necessarily needed but the idea is indeed that without the presence of water some computations can be skipped

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -0,0 +1,425 @@
...
+    # Same reasoning as in TerrariumWetLand.timestep!: free to construct, empty
+    # InputSources() so SpeedyWeather owns the input-update cycle.
+    integrator = ModelIntegrator(
+        state.clock, tmodel, InputSources(),
+        state, terrarium_initializers(land), terrarium_timestepper(land),
+    )
+    Terrarium.run!(integrator; period = vars.prognostic.clock.Δt, Δt = terrarium_substep(land))
+
+    # Surface soil temperature from the bottom of the column (last z-index)
+    vars.prognostic.land.soil_temperature .= interior(state.temperature)[:, 1, end] .+ NF(273.15)
```
milankl: Ah okay I see here whatever Terrarium does but soil moisture isn't communicated back, so technically one could have it moist on the Terrarium side but it behaves like dry for SpeedyWeather. Fine with me!

milankl (commented):
I'm unsure about the logic of that ModelIntegrator which is created on the fly, could that not live in `model.land.integrator` or similar?

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: The `Oceananigans.Field` has to be transferred to CPU by `Oceananigans.on_architecture` routine (which is loaded in the beginning).

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: Yes, `ModelIntegrator` is like `Simulation`. 

I am fairly certain this isn't allocating. However, I just noticed there is another problem here. The Terrarium `ModelIntegrator` was never initialised. I do that now directly in the `SpeedyWeather.initialize!(vars, ::AbstractTerrariumLandModel, ::Model)`. Does that work? 

The fundamental problem here is that we would like to keep the separation of variables and model structures. So `Terrarium.StateVariables` should live within `SpeedyWeather.Variables` in some way, and `Terrarium.Model` within the `SpeedyWeather.Model` and both together form a `SpeedyWeather.Simulation`. This leaves the `ModelIntegrator` as the odd one out. 

The `TerrariumWetLand` practically hold all information that `ModelIntegrator` would hold except for the `StateVariables`, so it is in some sense its replacement here. 

We could also get rid of the `ModelIntegrator` completely here and just call the more low level functions directly. But as long as the direct `ModelIntegrator` construction is almost for free I don't really think it's that bad. It just makes calling the Terrarium time stepping easier. The Terrarium model still has to be correctly initialised though. 

@bgroenks96 probably has thoughts on this too.

> I'm unsure about the logic of that ModelIntegrator which is created on the fly, could that not live in model.land.integrator or similar?

This is actually what we do in the original coupling script https://github.com/NumericalEarth/Terrarium.jl/blob/main/examples/simulations/speedy_wet_land.jl but then you also have the `Terrarium.StateVariables` (as part of the `ModelIntegrator`) within the `PrimitiveWetEquation`. Not really ideal.

maximilian-gelbrecht:
Locally the tests (that just check for no NaNs) were actually passing, I think to resolve the package depedency conflicts, we need a new Terrarium version. For the time being I use `main` directly here

Edit: doesn't work in Julia 1.10 so the tests are still failing.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: Good point. In the original this was `Grid3D()` and I changed it to `Land3D()`, but actually we just need the surface layer, so `Land2D()` should be fine as well. Right @bgroenks96 ?

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -0,0 +1,433 @@
+module SpeedyWeatherTerrariumExt
+
+using SpeedyWeather
+using Terrarium
+using DocStringExtensions
+
+import SpeedyWeather: AbstractVariableDim, AbstractVariable, AbstractModel,
```
bgroenks96: As is often suggested by @giordano, explicit `import`s of individual names should generally be avoided to prevent accidental shadowing. I guess this is less of a problem in an extension module, but still, there's no downside to just `using` instead. For things where you actually define dispatches, use the explicit qualifier `SpeedyWeather.variables(...)`.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -0,0 +1,433 @@
...
+    PrimitiveEquation, get_nlayers,
+    variables, initialize!, timestep!, allocate
+import SpeedyWeather.RingGrids
+
+import Terrarium: Clock, ColumnRingGrid, ForwardEuler, InputSources,
+    ModelIntegrator, interior, on_architecture, set!
+
+"""$(TYPEDEF)
+
+Variable dimension for the full Terrarium land state. `Base.zero` for an
```
bgroenks96: outdated docstring

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
bgroenks96: Since this *should* be an internal type and corresponding API for the extension module, I don't see much purpose in having these getter functions unless there is a specific need for dispatch.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -0,0 +1,433 @@
...
+terrarium_input_variables(land::AbstractTerrariumLandModel) = land.input_variables
+
+"""$(TYPEDSIGNATURES) Return the Terrarium field initializers `NamedTuple`."""
+terrarium_initializers(land::AbstractTerrariumLandModel) = land.initializers
+
+"""$(TYPEDSIGNATURES) Return the preconstructed Terrarium fields `NamedTuple`."""
+terrarium_fields(land::AbstractTerrariumLandModel) = land.fields
+
+"""$(TYPEDSIGNATURES) Return the Terrarium-internal sub-step in seconds."""
+terrarium_substep(land::AbstractTerrariumLandModel) = land.Δt
```
bgroenks96: I wouldn't call this a `substep` since that has other connotations in multi-step and multi-stage timesteppers. It's just the Terrarium time step. In principle, it could also be larger than the Speedy time step, though that's not really supported by the current implementation.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
bgroenks96: Related to my previous comment, I would *very strongly* suggest that you use explicit imports sparingly here; and especially for functions like `on_architecture` that are separately defined in multiple packages, you should always be explicit in which one you're using in a context where both are loaded, e.g. `Oceananigans.on_architecture` and `SpeedyWeatherInternals.on_architecture`.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -0,0 +1,433 @@
...
+        Δt::Real = 300,
+    ) where {NF}
+    field_grid = Terrarium.get_field_grid(model.grid)
+    Δz_arr = on_architecture(Terrarium.CPU(), field_grid.z.Δᵃᵃᶜ)
+    # The Oceananigans vertical spacing is an OffsetArray; take the last entry
+    # as the bottom-most layer thickness for the SpeedyWeather LandGeometry.
+    geometry = LandGeometry(1, NF[Δz_arr[end]])
+    return TerrariumWetLand(
+        spectral_grid, geometry, model, timestepper, clock,
+        boundary_conditions, input_variables, initializers, fields, Float64(Δt),
```
bgroenks96: Why is `Float64` hardcoded here?

bgroenks96: should be `NF` I think. I guess this is probably a Claude-ism?

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -0,0 +1,433 @@
...
+        clock = integrator.clock,
+        initializers = integrator.initializers,
+    )
+end
+
+function variables(::AbstractTerrariumLandModel)
+    return (
+        # The full Terrarium state, owned by SpeedyWeather's Variables tree.
+        PrognosticVariable(
+            name = :terrarium, dims = TerrariumVars(),
```
bgroenks96, line 151: I still think it's very weird (and potentially wrong) to have the full state represented as a single prognostic variable. Maybe it's beyond the scope of this PR, but I think we really need a fully defined shared state variables type to resolve this.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
bgroenks96: They could be initialized with the same `Array` storage though.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
bgroenks96: Yes

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
bgroenks96: It is allocation-free, that's not an issue.

> So Terrarium.StateVariables should live within SpeedyWeather.Variables in some way, and Terrarium.Model within the SpeedyWeather.Model and both together form a SpeedyWeather.Simulation. This leaves the ModelIntegrator as the odd one out.

The purpose of `ModelIntegrator` is to implement the Oceananigans model interface. It serves as the bridge between an Oceananigans "model" and a Terrarium "model". For the purposes of a direct Speedy-Terrarium coupling, it's largely irrelevant.

I think the current solution of constructing it on-the-fly is the correct solution. I wouldn't get rid of it since several of the timestepping related methods are defined on `ModelIntegrator` and trying to bypass it would result in a lot of code duplication.

bgroenks96: Regarding the more fundamental issue of how to combine state across models, I think the current solution is sub-optimal; e.g. `TerrariumWetLand` currently contains `clock`, `timestepper`, and `fields`, all of which violate the model-state separation principle.

I think what is really needed here in the end is a shared state variables type for both SpeedyWeather and Terrarium that merges all state variables into a single data structure. We also should have a harmonized `initialize!` implementation that allocates shared memory for all `Field`s at the interface.

It might be worth mentioning that this is starting to feel a bit redundant with the coupling infrastructure in NumericalEarth.jl. Maybe we should take a step back and try to think clearly about what the scope of this extension module is relative to that.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -0,0 +1,433 @@
...
+"""$(TYPEDEF)
+
+[`AbstractTerrariumLandModel`](@ref) for dry atmospheric models: wraps a
+Terrarium soil model (heat conduction only, no surface energy balance or
+moisture). The only atmospheric forcing passed in is near-surface air
+temperature; soil moisture is not tracked. Pass as `land=` to
+`PrimitiveDryModel`.
+
+$(TYPEDFIELDS)"""
+struct TerrariumDryLand{
```
bgroenks96: Do we actually want to support this? Is there a clear use case?

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
bgroenks96: I would also support removing the wet/dry distinction and instead just conditionally updating variables based on whether or not the Speedy model declares them, I think. The current implementation is very redundant.

Maybe even better would be to add dispatches for a `TerrariumLand` on `PrimitiveWet` and `PrimitiveDry`. I see no need for the types themselves to be different.

SpeedyWeather/test/parameterizations/terrarium_coupling.jl
```diff
@@ -0,0 +1,142 @@
+using Terrarium
+using Test
+using Dates
+
+# The SpeedyWeatherTerrariumExt extension activates once both packages are
+# loaded. Pull the public names out of the extension module.
+const SWTerrarium = Base.get_extension(SpeedyWeather, :SpeedyWeatherTerrariumExt)
```
bgroenks96, line 7: This is very bad practice. `Base.get_extension` is [not a public API](https://github.com/JuliaLang/julia/pull/59593) and its use should generally be avoided outside of interactive code, and it *definitely* should never be used within package modules because you can very quickly end up in code loading order hell.

This is a test script, so maybe it's a bit of a gray zone, but I would argue that tests should exemplify the intended use of the public API. This definitely should not be it.

I would again suggest defining a method/constructor in `SpeedyWeather` which can be extended in the Terrarium extension. There should be no need to reference the extension module types here.

bgroenks96 (changes requested):

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: > Regarding the more fundamental issue of how to combine state across models, I think the current solution is sub-optimal; e.g. TerrariumWetLand currently contains clock, timestepper, and fields, all of which violate the model-state separation principle.

With regards to `clock` and `fields` it might also a bit of an unfortunate naming because what it does here is just setting the initial condition. We could just rename those to `initial_fields` and `initial_clock` or better just set the Terrarium `clock` with the initial date set within the Speedy model. I'll try that right away for the clock.

In principle, I think having information about the initial conditions within the model type isn't bad. It's something that we also do in SpeedyWeather. It's allocated anyway, might as well have it in the model type. It makes the `initialize!` functions easier, as they can just get the information from the model. 

>  I think what is really needed here in the end is a shared state variables type for both SpeedyWeather and Terrarium that merges all state variables into a single data structure. We also should have a harmonized initialize! implementation that allocates shared memory for all Fields at the interface.
> 
> It might be worth mentioning that this is starting to feel a bit redundant with the coupling infrastructure in NumericalEarth.jl. Maybe we should take a step back and try to think clearly about what the scope of this extension module is relative to that.

I also think that doing the former would move this already in the direction of NumericalEarth. 

I was thinking that this extension here should serve a direct and relatively light-weight way to couple Speedy and Terrarium by just defining Terrarium as a `SpeedyWeather.Land` model. It's just quite hacky at the moment admittedly. 

The `timestepper` is probably the biggest offender, but we'll talk about that separately.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: fixed

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: Removed all of them

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: Removed

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: Yeah you are right, there's now an explicit call here

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: fixed

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: Yeah, I agree with you that it's weird, but I think that's really a bit out of scope, because it's not easy to do it differently. 

I think we should rather put that energy into making the coupling work via NumericalEarth

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: Adjusted to `Land2D`

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: Yes, I agree. I'll do this a bit later, early next week.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: I struggle to a super easy way to do that though that fits in the variable systems of SpeedyWeather and Terrarium that we have. 

I think that's one of the points for which it's okay to accept a sub-optimal solution here. With just two 2D variables, it's also really not that bad.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: The `clock` is now removed.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: Ok, addressed in the last commit

SpeedyWeather/test/parameterizations/terrarium_coupling.jl
maximilian-gelbrecht: Sure, this is temporary, as stated in the post in the beginning. 

But just saying, there are a lot of packages that use `Base.get_extension` quite extensively (e.g. also NumericalEarth). Personally, I'd also prefer to just define function stubs or other kind of constructors in the main package, but I really dislike the Julia decision to not be able to export symbols from extensions. 

Here, I just wanted to have something temporary, until we know that this extension is roughly what we want. 

I think we might just do a 
```Julia 
LandModel(spectral_grid::SpectralGrid, varargs...) = ....
LandModel(terrarium_model::TerrariumModel, varargs.... ) = TerrariumWetLandModel()
```
in the main code in the end.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: ok done

SpeedyWeather/test/parameterizations/terrarium_coupling.jl
bgroenks96: > But just saying, there are a lot of packages that use Base.get_extension quite extensively (e.g. also NumericalEarth).

I don't think they should. I'm honestly surprised that Julia even allows this. It should probably be in `InteractiveUtils` or `Core`.

It runs fundamentally against the notion of an *extension* to import the types into the parent module, and importing extensions from other packages is very buggy and causes all kind of weird precompilation issues.

> I really dislike the Julia decision to not be able to export symbols from extensions.

Their decision makes perfect sense though for the way extensions are implemented. How would you expect this to work? Should `using SpeedyWeather, Terrarium` import names into the local scope that are not defined in either of the two namespaces? That doesn't make much sense. Should the user have to explicitly  `import SpeedyWeatherTerrariumExt`? That would also be weird because they didn't explicitly add it as a dependency. This would be a clear breach of encapsulation between packages/modules.

And that's not even to mention the extraordinary complexity that conditionally loading types/symbols adds to the code loading procedure. It's been years since extensions were added, and I still don't really understand how Julia decides what order to load them.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: I resolve this, let's have a discussion about the use of the dry model in the other comment.

SpeedyWeather/test/parameterizations/terrarium_coupling.jl
maximilian-gelbrecht: So I defined 

```
LandModel(spectral_grid::SpectralGrid, varargs...) = ....
LandModel(terrarium_model::TerrariumModel, varargs.... ) = TerrariumWetLandModel()
```
now. 

Regular users don't need to use the `Base.get_extension` anymore as planned. 

However, in the unit tests this is still might be good to have to test everything we define in the extension. We could also remove tests e.g. on the type hierarchy, but it's also not bad to have them.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: The `timestepper` remains an offender here of course. But I would say it's pragmatically okay for now, at least until we are sure of how we'll revise it on the Terrarium side.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: Added dispatches now for `initialize!` and `timestep!` based on the SpeedyWether model type.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: I am also not sure, but let's discuss this in the other comment chain.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -0,0 +1,307 @@
...
+    fields::FL
+    "Terrarium-internal sub-step (seconds) used to integrate within each SpeedyWeather step"
+    Δt::NF
+end
+
+"""$(TYPEDSIGNATURES)
+
+Construct a [`TerrariumLand`](@ref) from a Terrarium land model and a
+SpeedyWeather spectral grid. The single effective land layer thickness is
+derived from the bottom-most layer of the Terrarium grid."""
```
bgroenks96: topmost

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -0,0 +1,307 @@
...
+        boundary_conditions::NamedTuple = (;),
+        input_variables::Tuple = (),
+        initializers::NamedTuple = (;),
+        fields::NamedTuple = (;),
+        Δt::Real = 300,
+    ) where {NF}
+    field_grid = Terrarium.get_field_grid(model.grid)
+    Δz_arr = Terrarium.on_architecture(Terrarium.CPU(), field_grid.z.Δᵃᵃᶜ)
+    # The Oceananigans vertical spacing is an OffsetArray; take the last entry
+    # as the bottom-most layer thickness for the SpeedyWeather LandGeometry.
```
bgroenks96: it's just a single layer right? so "bottom-most" doesn't really mean anything here?

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -0,0 +1,307 @@
...
+        ::PrimitiveDryModel,
+    )
+    state = vars.prognostic.land.terrarium
+    NF = eltype(vars.prognostic.land.soil_temperature)
+
+    # Sync the Terrarium clock's initial time with the SpeedyWeather clock,
+    # which was set from the `time` kwarg of `initialize!(model; time=...)`.
+    state.clock.time = vars.prognostic.clock.time
+
+    vars.prognostic.land.soil_temperature .= interior(state.temperature)[:, 1, end] .+ NF(273.15)
```
bgroenks96:
```suggestion
    vars.prognostic.land.soil_temperature .= @view(interior(state.temperature)[:, 1, end]) .+ NF(273.15)
```

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -0,0 +1,307 @@
...
+    # Same reasoning as in TerrariumLand.timestep!: free to construct, empty
+    # InputSources() so SpeedyWeather owns the input-update cycle.
+    integrator = ModelIntegrator(
+        state.clock, tmodel, InputSources(),
+        state, land.initializers, land.timestepper,
+    )
+    Terrarium.run!(integrator; period = vars.prognostic.clock.Δt, Δt = land.Δt)
+
+    # Surface soil temperature from the bottom of the column (last z-index)
+    vars.prognostic.land.soil_temperature .= interior(state.temperature)[:, 1, end] .+ NF(273.15)
```
bgroenks96:
```suggestion
    vars.prognostic.land.soil_temperature .= @view(interior(state.temperature)[:, 1, end]) .+ NF(273.15)
```

SpeedyWeather/test/parameterizations/terrarium_coupling.jl
bgroenks96: Ok

bgroenks96 (changes requested):

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: Yes you are right. The whole layer_thickness also means nothing at all for Speedy when it doesn't use one of its own land models. As we already have it in `LandGeometry` anyway, I would still set it to the correct value from Terrarium though just for consistency.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: I just deleted the whole sentence because it doesn't make any sense. We just set the thickness for consistency (see comment below)

docs/src/land.md
```diff
@@ -410,3 +412,63 @@ save("ocean_land_albedo.png", ans) # hide
 nothing # hide
 ```
 ![Ocean-land albedo](ocean_land_albedo.png)
+
+## Terrarium coupling
+
+We can also use a [Terrarium](https://github.com/NumericalEarth/Terrarium.jl) model within SpeedyWeather! For a detailed introduction to Terrarium please consult its [own documentation](https://numericalearth.github.io/Terrarium.jl/dev/). Here, we will only demonstrate the coupling.
+
```
maximilian-gelbrecht, line 419: We should probably add a bit more to explain the coupling a least a little bit.

bgroenks96 (approved):

PR #1093, opened 2026-05-14
Browzarr extension

milankl:
Just so that one can conveniently do

```julia
using SpeedyWeather
using Browzarr

# define and run a simulation (just a brief example)
spectral_grid = SpectralGrid()
model = PrimitiveWetModel(spectral_grid)
simulation = initialize!(model)
run!(simulation, output=true)

# open a Browzarr window pointing to the output store (zarr/netcdf) that was just written
browzarr(simulation) 
```

@lazarusA okay to extend `browzarr` for `SpeedyWeather.AbstractSimulaiton`? Not sure how to write tests for this, any suggestions?

PR #1096, opened 2026-05-15
allow also zarr 0.9

lazarusA:
expand zarr bounds, most packages are still using 0.9


maximilian-gelbrecht (approved):
I guess it's unlikely we use some features exclusive to `0.10` in the Zarr extension? 

Let me know if you want a new minor release for this. I think we could do that, but maybe we bundle it together with #1093

milankl:
I vote to bundle it!

PR #1097, opened 2026-05-15
mv Enzyme to weakdeps

### LINKED ISSUE #1095: Enzyme in deps and weekdeps
as it is,  Enzyme is set in `deps` as well in `weekdeps`, which is not triggered on using Enzyme since is already in `deps`.  Is this intended?

This takes a toll on pre-recompilation times. Since doing `add SpeedyWeather`, and then `using SpeedWeather` will always pull `Enzyme`.

lazarusA:
be speedy again. (Testing all CI works)

closes https://github.com/SpeedyWeather/SpeedyWeather.jl/issues/1095


maximilian-gelbrecht:
Oh thanks! Well caught, I think this maybe slipped through accidentally.

maximilian-gelbrecht:
Yeah, it was me in #971

milankl (approved):

PR #1098, opened 2026-05-15
use @docs instead of @doc

lazarusA:
the layout is broken due to `@doc` expanding beyond the `content` container:

Screenshot 2026-05-15 at 18 27 46

and in addition, I also change another 2 places. Let's see how it goes. There is also `animate`, but that one is only defined after the extension is loaded, hence doing `@docs` does not work, so I left it as is. 

And now:
-  https://speedyweather.github.io/SpeedyWeatherDocumentation/previews/PR1098/output#Further-options

Screenshot 2026-05-16 at 09 11 43


maximilian-gelbrecht:
Thanks a lot! Looks good to me. Do you want to add something else or shall we merge this? 

Stupid question but what the difference between ` ``` ` and ` ```` ` blocks?

maximilian-gelbrecht (approved):

lazarusA:
> Thanks a lot! Looks good to me. Do you want to add something else or shall we merge this?

I'm done with this one. So, it will be good to merge.

> Stupid question but what the difference between ``` and ```` blocks?

Not much, in most cases you will see the same behaviour,  but I like to use the second one when I know that I could have block-codes inside ( a nested case). Personally, I use them to have a mental model of what is involved. 


Edit: Also, latest commit adds the option to manually trigger this workflow via the actions button, no need to do dummy commits.

PR #1099, opened 2026-05-18
Mega transform! GPU-optimized batched transforms (Part 1: `transform!(vars, model)`) 

maximilian-gelbrecht:


milankl:
Have you every checked how much faster a fused transform could be? E.g.

- 4x T31L8 vs 1x T31L32
- 4x T127L16 vs 1x T127L64

I reckon it makes more sense to compare at low resolution as I expect the benefit to decrease at high resolution

@gregordecristoforo FYI

maximilian-gelbrecht:
Jack made those benchmarks that should still be up to date https://github.com/SpeedyWeather/SpeedyWeather.jl/tree/main/SpeedyWeather/benchmark/CUDA

Just not for `MatrixSpectralTransform`.

milankl:
Also I guess the Legendre transform is written such that the kernels just get bigger, but for the FFTs do we need a specific plan for every fuse size? I've also been wondering whether we can simply drop the loop here

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/bd674a4d1b0765847bf155df72916c04052d041c/SpeedyTransforms/src/fourier.jl#L215-L230

in case it's a `field <: AbstractFullField` and then apply an FFT planned for batch in latitude and vertical. But I reckon that also means that the transform scratch memory has to be bigger?

maximilian-gelbrecht:
The naive way would be to just initialise `SpectralTransform` with `4 * nlayers + 1` as that should be the maximum amount of layers that are fused together (I guess). Smaller transforms with less layers always work. 

No matter how exactly we do it, the scratch memory `g` for the Legendre transformed not yet Fourier transformed is getting larger, yes.

maximilian-gelbrecht:
I think dropping the loop for full grids is an orthogonal thing. We could definitely do that though, just not in this PR.

maximilian-gelbrecht:
I made a new benchmark comparing the sequential to batched transform on GPU. Even at T511 we still get a 5x speed up by batching the transforms. That's of course the biggest of these batched transforms. But we have 3(?) if them in the dynamical core. And can then also merge a bunch of smaller ones together as well. 

### Results 

T31: batched 1.178681e6 ns - sequential 2.3084332e7 ns; speed-up: 19.58488513855742
T63: batched 2.516718e6 ns- sequential 4.7592253e7 ns; speed-up: 18.91044328367342
T125: batched 5.836238e6 ns- sequential 1.00177221e8 ns; speed-up: 17.16469085051021
T255: batched  1.8021913e7 ns- sequential 2.06443085e8 ns; speed-up: 11.455114948118993
T511: batched 8.71955635e7 ns - sequential 4.72406908e8 ns; speed-up: 5.417786055135706

### Code
 
```Julia 
using SpeedyWeather, CUDA, BenchmarkTools

NLAYERS = 8 
NSLOTS = 4*NLAYERS + 1

function sequential_transform!(s1, s2, g1, g2, S)
    transform!(s1, g1, S)
    transform!(s1, g1, S)
    transform!(s1, g1, S)
    transform!(s1, g1, S)
    transform!(s2, g2, S)
    return nothing
end 

for trunc in [31, 63, 125, 255, 511]
    
    spectral_grid = SpectralGrid(trunc=trunc, nlayers=NLAYERS, architecture=SpeedyWeather.GPU())
    NF = spectral_grid.NF
    spectrum = spectral_grid.spectrum
    grid = spectral_grid.grid 
    spectral_transform = SpectralTransform(spectrum, grid, NF=NF, nlayers=NSLOTS)

    S_1 = rand(Complex{NF}, spectrum, NLAYERS)
    S_2 = rand(Complex{NF}, spectrum)

    S_C = rand(Complex{NF}, spectrum, NSLOTS)

    G_1 = rand(NF, grid, NLAYERS)
    G_2 = rand(NF, grid)

    G_C = rand(NF, grid, NSLOTS)

    bench_seq = @benchmark CUDA.@sync sequential_transform!($S_1, $S_2, $G_1, $G_2, $spectral_transform)
    bench_batched = @benchmark CUDA.@sync transform!($S_C, $G_C, $spectral_transform)
    time_batched =  median(bench_batched).time
    time_seq = median(bench_seq).time 
    speed_up = time_seq / time_batched
    println("T$(trunc): $(time_batched) ns - $(time_seq) ns; speed-up: $(speed_up)")
end 
```

milankl:
So on CPU batching vs not doesn't make a difference? On GPU a batched transform is 20x faster at T31 and still 5x faster at T511? How's that possible? I would have assumed that at most we can do 4 variables at the cost of 1, that would be 4x theoretical cap at all resolutions

maximilian-gelbrecht:
About CPU: The test above is GPU only. I've just rerun the test on an x86-64 Intel CPU. At low resolutions there also seems to be speed up probably due to better vectorizaotins. But nothing near the results for GPU 

T31: 616524.0 ns - 810224.0 ns; speed-up: 1.3141807942594286
T63: 3.280886e6 ns - 3.73976e6 ns; speed-up: 1.1398628297356264
T125: 2.0873502e7 ns - 2.0314934e7 ns; speed-up: 0.9732403312103547

Note that T125 on CPU is as fast as T255 on GPU btw. 

About the GPU performance: to be honest, I only have half-knowledge of this. We merge together 5 transforms (on GPU the 2D transform takes almost the same time as a 3D transform). Until we really fully saturate the GPU memory, we'll gain massive speed up. And we also avoid some overhead. The overhead might actually be quite significant at low resolutions. But yeah, to be honest I am also surprised that it's really that much faster to batch the transform. At least in isolation.

maximilian-gelbrecht:
I mean logically, the conclusion kind of has to be that we have massive overheads for low to mid resolution on GPU? Or not?

milankl:
> I mean logically, the conclusion kind of has to be that we have massive overheads for low to mid resolution on GPU? Or not?

Yeah no I'm not surprised that we'd save on many overheads, but in a mental model of TIME = 5*OVERHEAD + 5*EXECUTION, for both OVERHEAD >> EXECUTION as well as EXECUTION >> OVERHEAD cases I would have expected the BATCHED_TIME >= OVERHEAD + EXECUTION and so TIME/BATCHED_TIME <= 5x hence I was never expecting a speedup higher than the number of variables we're batching

maximilian-gelbrecht:
> > I mean logically, the conclusion kind of has to be that we have massive overheads for low to mid resolution on GPU? Or not?
> 
> Yeah no I'm not surprised that we'd save on many overheads, but in a mental model of TIME = 5_OVERHEAD + 5_EXECUTION, for both OVERHEAD >> EXECUTION as well as EXECUTION >> OVERHEAD cases I would have expected the BATCHED_TIME >= OVERHEAD + EXECUTION and so TIME/BATCHED_TIME <= 5x hence I was never expecting a speedup higher than the number of variables we're batching

With execution and overhead in a similar magnitude you could argue for 10x, right? But it's still more than that in low resolutions, which is quite something. But I also don't see a mistake in the script above.

maximilian-gelbrecht:
Oh gosh, I forgot about the limitations FFT have in Julia with really _all_ dimensions having to be pre-planned, even the batched ones. I am really annoyed by AbstractFFTs enforcing this limitation. Years ago somebody already has implemented a much more flexible FFT for CUDA.jl and it was shot down because of AbstractFFTs. 

Edit: I realise now, at least the past implementation for CUDA.jl also just achieved this by sequentially executing the plans, not by actually batching them. So this is also not what we would want.

maximilian-gelbrecht:
I am starting to think we should write our own FFT implementations...

maximilian-gelbrecht:
I mean, the speed up is so massive that I it's still worth it for the GPU version. But working around this limitation is a headache for sure. 

So for CPU, just doing serial FFTs instead of batched FFTs is totally fine. There's little to no speed up anyway. 

On GPU, I guess we just have to allocate a bunch of plans with different sizes. Thissome a bunch of memory, but it's the safest option. FFT plans don't really take up that much space and the trailing batch dimension don't take up extra memory. We'd have to think a bit what's actually worth it and what isn't: 

* All prognostic variables together. This one is used actually a few times as well: grid tendencies -> spectral tendencies, and spectral state -> grid state (and inverse). The speed up is massive as shown above. 
* A single prognostic variable with `nlayers`: Definitely needed, no questions asked
* Then, we have to see what else. 

I also noticed there are apparently differences between the Apple ARM and the Intel x86 implementations of FFTW and FFTW.jl, all unit tests pass on my laptop but fail on the CI.

maximilian-gelbrecht:
Ok, so a few considerations what's needed and what's possible wrt the fused variables transform: 

* `MatrixSpectralTransform` 
  * CPU: no issues at all, minor or no speedup
  * GPU: no issues at all, huge speedup 

* `SpectralTransform` (FFT+LT)
  * CPU: minor or no speedup possible; therefore just make sure there's a proper fallback that executes the batched transforms sequentially 
    * CPU-ARM: Everything easy, fallbacks work well even if done naively 
    * CPU-x86: Fallbacks have to be defined with proper memory alignment
   * GPU: huge speedup possible; but limitations with the plans. We definitely need separate plans for the biggest batched transforms, and the most common other ones. For the rest consider whether we either just have sequential fallback or maybe do unplanned, direct FFT (`rfft(x)`)

maximilian-gelbrecht:
While I mostly implemented a solution to the problem above, a new problem emerged: Reactant compatibility. There's currently some mul! functions on SubArray missing https://github.com/EnzymeAD/Reactant.jl/issues/2912

maximilian-gelbrecht:
So, this PR I want to keep limited to optimising the big `transform!(::Variables, ;:PrimitiveEquation)` and creating the infrastructure for that. I think I've done basically everything there is to that. We'll have to wait for #1035 and https://github.com/EnzymeAD/Reactant.jl/issues/2912 before merging this. 

Here are some benchmarks on `transform!(::Variables, ;:PrimitiveEquation)` now: 

### CPU — median time (ms), nlayers=8

| Transform | Branch | T31 | T63 | T127 | T197 |
|---|---|---:|---:|---:|---:|
| SpectralTransform | main | 1.42 | 6.22 | 34.2 | 106 |
| SpectralTransform | feature | 1.35 | 5.98 | 33.5 | 103 |
| MatrixSpectralTransform | main | 2.35 | 10.0 | 164 | 860 |
| MatrixSpectralTransform | feature | 2.43 | 6.51 | 92.0 | 475 |

### GPU — median time (ms), nlayers=8

| Transform | Branch | T31 | T63 | T127 | T197 |
|---|---|---:|---:|---:|---:|
| SpectralTransform | main | 9.19 | 18.5 | 37.9 | 61.3 |
| SpectralTransform | feature | 2.91 | 5.76 | 12.0 | 19.5 |
| MatrixSpectralTransform | main | 0.89 | 2.69 | 32.0 | 183 |
| MatrixSpectralTransform | feature | 0.26 | 0.95 | 9.63 | 58.8 |

There still might actually be some bottleneck in there but it's already a nice improvement.

maximilian-gelbrecht:
Profiling (the script is a bit flawed in the execution, that's why it doesn't sum up to 100%, but the ms should be correct)

### T127, SpectralTransform, GPU, nlayers=8

| Step | ms | % of FULL |
|---|---:|---:|
| save_prev! | 0.030 | 0.3% |
| UV_from_vordiv! | 0.015 | 0.1% |
| temperature_average! | 0.020 | 0.2% |
| transform! prognostic (spec→grid) | 5.68 | 50.3% |
| ↳ _legendre! | 0.31 | 2.8% |
| ↳ _fourier! | 5.61 | 49.7% |
| hole_filling! | 0.014 | 0.1% |
| transform! uv (spec→grid, unscale_coslat) | 5.61 | 49.7% |
| ↳ _legendre! | 0.24 | 2.1% |
| ↳ _fourier! | 5.77 | 51.1% |
| geopotential! | 0.020 | 0.2% |
| FULL | 11.28 | 100% |

### T197, SpectralTransform, GPU, nlayers=8

| Step | ms | % of FULL |
|---|---:|---:|
| save_prev! | 0.060 | 0.3% |
| UV_from_vordiv! | 0.020 | 0.1% |
| temperature_average! | 0.023 | 0.1% |
| transform! prognostic | 9.85 | 48.4% |
| ↳ _legendre! | 1.56 | 7.6% |
| ↳ _fourier! | 9.46 | 46.5% |
| hole_filling! | 0.019 | 0.1% |
| transform! uv | 9.42 | 46.3% |
| ↳ _legendre! | 0.95 | 4.7% |
| ↳ _fourier! | 9.30 | 45.7% |
| geopotential! | 0.038 | 0.2% |
| FULL | 20.36 | 100% |

So, actually we should probably still merge the `transform! uv` with the big `transform!` prognostic, we just have to unscale coslat seperately then.

maximilian-gelbrecht:
In the benchmark above we also clearly see that it's the Fourier transform itself that is the major bottleneck (@gregordecristoforo)

maximilian-gelbrecht:
The Reactant fix is merged, with the next released hopefully all tests pass

maximilian-gelbrecht:
# `transform!(::Variables, lf, model)` on GPU — current branch (graphs + fusion) vs released v0.20.3

PrimitiveWet, OctahedralGaussianGrid, NVIDIA A40. Wall time = host enqueue + GPU, median of reps,
per `transform!(vars, lf, model)` call (the spectral→grid propagation of the prognostic state).

- **v0.20.3 (released)**: per-variable transforms, generic allocating GPU Fourier path (no graphs, no fusion)
- **current branch**: fused mega-batch transform + CUDA graphs

Speedup = v0.20.3 / current. The win combines **both** fusion (6 transforms → 2) and CUDA graphs
(collapsing the many small FFT launches into one replay).

| Resolution | v0.20.3 (µs) | current (µs) | speedup |
|-----------|-------------:|-------------:|--------:|
| T31  L8 |   11425.0 |    468.4 | 24.4× |
| T63  L8 |   21909.3 |   1098.4 | 20.0× |
| T127 L8 |   45231.0 |   2972.5 | 15.2× |
| T255 L8 |   92396.1 |  17296.7 |  5.3× |

## Notes

- The speedup shrinks with resolution (24× → 5×): at low truncation the routine is launch-bound,
  so collapsing launches dominates; at T255 the Legendre matmuls and large FFTs make it
  compute-bound, where graphs/fusion help less (but still 5×).
- This is the real-world, apples-to-apples comparison of the same high-level call in each release
  (identical signature). It deliberately conflates fusion + graphs; the graphs-alone win for the
  Fourier sub-step is ~6–10× (see results_graphs_comparison.md).
- Each version uses its own resolved CUDA stack (v0.20.3 → CUDA 5.x, current → CUDA 6.x), which is
  the honest "what each release delivers" comparison rather than a controlled single-variable test.

milankl:
24x faster!! This is so exciting 😍

maximilian-gelbrecht:
Okay, the merge wasn't that terrible. I need to confirm the GPU performance tomorrow, and there are no regressions through the merge, but outside of that this is ready I think.

SpeedyWeather/src/output/restart.jl
```diff
@@ -44,7 +44,7 @@ function finalize!(
     restart.path = path     # update path in case it was empty before
 
     jldopen(joinpath(path, filename), "w"; compress) do f
-        f["variables.prognostic"] = vars.prognostic
+        f["variables.prognostic"] = materialize_views(vars.prognostic)
```
milankl, line 47: If we materialize the views for output, should the fused array be removed to not save duplicated arrays?

SpeedyTransforms/src/fourier.jl
```diff
@@ -1,21 +1,22 @@
-# function barrier for batched or serial transforms as FFTW plans cannot be reused for fewer vertical layers
+# Function barrier for batched or serial transforms. FFTW/cuFFT plans bake the batch dim K
```
milankl, line 1: Is it wise to talk about vertical index `k` and batch dim `K`? They are the same no? Whether they are the same or not, should this be better explained?

SpeedyWeather/benchmark/README.md
```diff
@@ -386,102 +386,102 @@ Toolchain:
...
 
 ### Models, default setups
 
 | Model | T | L | Physics | Δt | SYPD | Memory|
 | --- | --- | --- | --- | --- | --- | --- |
-| BarotropicModel | 31 | 1 | false | 2400 | 1701 | 432.95 KB |
-| ShallowWaterModel | 31 | 1 | false | 2400 | 850 | 436.08 KB |
-| PrimitiveDryModel | 31 | 8 | true | 2400 | 1505 | 579.99 KB |
-| PrimitiveWetModel | 31 | 8 | true | 2400 | 1235 | 585.21 KB |
+| BarotropicModel | 31 | 1 | false | 2400 | 1644 | 446.12 KB |
```
milankl: This now runs by default with `NCycleLorenz` and on a shorter time step, 1800s. Maybe this is hardcoded in the benchmark suite somewhere ... ?

SpeedyWeather/src/variables/variables.jl
```diff
@@ -250,10 +250,101 @@ function Variables(model::AbstractModel)
...
     fused = build_fused_namespace(fuse_parents)
 
+    # If both the spectral prognostic parent and its grid-snapshot parent exist, they
+    # must declare members in matching order so a mega-batched spec→grid transform can
+    # map slot k of `prognostic` to slot k of `grid`.
+    # TODO: Maybe just move this to unit tests? Do we need to test this every time? 
+    if haskey(fused, :prognostic) && haskey(fused, :grid)
+        _assert_fuse_alignment(fused.prognostic, fused.grid;
+                               name_a = :prognostic, name_b = :grid)
+    end
```
milankl, line 260: I'd leave this for now including the TODO, if after a release or 2 we realise that this always passes anyway then we can move it to a unit test?

SpeedyWeather/src/time_stepping/transform.jl
```diff
@@ -146,31 +139,34 @@ function SpeedyTransforms.transform!(
...
     UV_from_vordiv!(U, V, vor, div, S)
 
-    # transform from U, V in spectral to u, v on grid (U, V = u, v*coslat)
-    transform!(u_grid, U, scratch_memory, S, unscale_coslat = true)
-    transform!(v_grid, V, scratch_memory, S, unscale_coslat = true)
+    # Batched spec→grid for the prognostic state: one call covers vorticity, divergence,
+    # temperature, pressure (and humidity for PrimitiveWet).
+    prog_parent = parent(vars.fused.prognostic)
+    grid_parent = parent(vars.fused.grid)
+    transform!(get_prognostic_step(grid_parent, time_stepping, S), get_prognostic_step(prog_parent, time_stepping, S), scratch_memory, S)
```
milankl: I like it that his can actually be written down in quite a neat way!

SpeedyWeather/src/time_stepping/transform.jl
```diff
@@ -146,31 +139,34 @@ function SpeedyTransforms.transform!(
...
+
+    if model isa PrimitiveWet
+        humid_grid = get_prognostic_step(vars.grid.humidity, time_stepping, S)
+        hole_filling!(humid_grid, model.hole_filling, model)  # remove negative humidity
+    end
+
+    # Batched spec→grid for the velocities: the general-purpose `:spectral_scratch` fuse packs
+    # `(:a, :b)` (here holding U, V) into one Spectral3D parent, and `:uv_grid` packs `(:u, :v)`
+    # TODO: theoretically we could merge this with the other big transform and then unscale coslat
+    # seperately, but the dimensions don't quite align, shall we still do that in a hacky way?
```
milankl, line 161: That would require to batch the prognostic spectral variables with the `a, b` scratch arrays? Doesn't sounds too bad to be honest though, but the `fuse = :prognostic` wouldn't be quite adequate anymore ...

SpeedyWeather/Project.toml
```diff
@@ -91,7 +91,7 @@ Random = "1.10"
 Reactant = "0.2.264"
 ReactantCore = "0.1.16"
 RingGrids = "0.1"
-SpeedyTransforms = "0.1"
+SpeedyTransforms = "0.2"
```
milankl, line 94: `-DEV` ?

milankl (approved):
Looks great to me, I like how little has to be changed outside the transforms.

SpeedyWeather/src/output/restart.jl
maximilian-gelbrecht: The restart only writes the prognostic variables, and those do not include the fused array as its is in its own category `.fused`, so this should be fine and already avoid duplication.

SpeedyTransforms/src/fourier.jl
maximilian-gelbrecht: Yes good point. From a mathematical notation point of view I kind of like `K` and `k`, as the batch dim `K` is a sum of the numbers of vertical levels (indexed by `k`), but from a coding notation I can also see that depending on capitalisation can be confusing. Do you have a preference for that? 

Either way I would add some more explanation here.

SpeedyWeather/benchmark/README.md
maximilian-gelbrecht: I know we changed the default timestepper, but what do you mean exactly here?

SpeedyWeather/src/variables/variables.jl
maximilian-gelbrecht: Agreed

SpeedyWeather/src/time_stepping/transform.jl
maximilian-gelbrecht: We could also define a `transform!(::FusedParent, ::FusedParent, ::AbstractTimeStepper, ::SpectralTransform)` then it would be even shorter. What do you think?

SpeedyWeather/Project.toml
maximilian-gelbrecht: I think for the compat/deps that's not needed, or not?

SpeedyWeather/src/time_stepping/transform.jl
maximilian-gelbrecht: The problem is that `u`,`v` / `U`,`V` are not time stepped variables, but the other fused variables here are. It's a kind of awkward fit, because it we would need to allocate them with a `step` dimension as well and then just only use one of them. Personally, I would just leave the TODO here for now and maybe review this later in the end of the whole rewrite.

SpeedyWeather/src/output/restart.jl
milankl: Correct, yeah nice!

SpeedyWeather/src/time_stepping/transform.jl
milankl: Ouh but then you have to pass on the `model` too to determine the step, I'd write

```julia
prognostic_variables = get_prognostic_step(parent(vars.fused.prognostic), time_stepping, S)
grid_variables = get_prognostic_step(parent(vars.fused.grid), time_stepping, S)
transform!(grid_variables, prognostic_variables, scratch_memory, S)
```

just to make it clear that the transform call is really the same as before

SpeedyWeather/src/time_stepping/transform.jl
milankl: Yeah no that seems like an ingredient for confusion ...

SpeedyWeather/Project.toml
milankl: Shoot yes, sorry!

SpeedyWeather/benchmark/README.md
milankl: I don't know why the markdown is writes 2400s for the time step into this table, it should be 1800s I believe

milankl: The benchmark suite does take the model time step and that does not seem to have been manually set to something else ... 

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/3852b6f3a35277a724e9cceab93dfee50bbaa94a/SpeedyWeather/benchmark/benchmark_suite.jl#L126

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/3852b6f3a35277a724e9cceab93dfee50bbaa94a/SpeedyWeather/benchmark/benchmark_suite.jl#L172-L175

SpeedyTransforms/src/fourier.jl
maximilian-gelbrecht: I added some explanation

SpeedyWeather/benchmark/README.md
maximilian-gelbrecht: Ah now I get it. I think this benchmark is just old and was done before the timestepper PR was merged. I'll update it

maximilian-gelbrecht: ok done

SpeedyWeather/benchmark/README.md
```diff
@@ -87,94 +87,94 @@ Threads: 1 default, 0 interactive, 1 GC (on 4 virtual cores)
...
+| PrimitiveWetModel | 85 | 24 | 128 | matrix | 900 | 4.1 | 1.79 GB |
+| PrimitiveWetModel | 127 | 24 | 192 | matrix | 600 | 0.4 | 8.26 GB |
 
 ### Primitive Equation, Float32 vs Float64
 
 | Model | NF | T | L | Δt | SYPD | Memory|
 | --- | --- | --- | --- | --- | --- | --- |
-| PrimitiveWetModel | Float32 | 31 | 8 | 2400 | 1145 | 4.84 MB |
-| PrimitiveWetModel | Float64 | 31 | 8 | 2400 | 1337 | 9.00 MB |
+| PrimitiveWetModel | Float32 | 31 | 8 | 2400 | 677 | 5.38 MB |
```
milankl, line 140: I think the benchmarking really fluctuated this time. The same simulation was 1300 SYPD above

maximilian-gelbrecht: Yeah, we probably have to think about a more stable way to do this. But by now the benchmark already takes a while, so just increasing rolll out length by a lot also isn't really an option.

milankl:
Nothing urgent but I just realised that already without this branch our variable size estimates are off
```julia
Simulation{...} (5.04 MB)
├ variables::Variables{...} (4.18 MB)
└ model::PrimitiveWetModel{...} (1.07 MB)

julia> simulation.variables
Variables{@NamedTuple{...}, ...} (6.41 MB)
```

One says variables are 4.2MB other one says its 6.4MB ... wondering whether this will get worse with the fused variables.

maximilian-gelbrecht:
> Nothing urgent but I just realised that already without this branch our variable size estimates are off
> 
> ```julia
> julia> simulation = initialize!(model)
> Simulation{...} (5.04 MB)
> ├ variables::Variables{...} (4.18 MB)
> └ model::PrimitiveWetModel{...} (1.07 MB)
> 
> julia> simulation.variables
> Variables{@NamedTuple{...}, ...} (6.41 MB)
> ```
> 
> One says variables are 4.2MB other one says its 6.4MB ... wondering whether this will get worse with the fused variables.

okay I fixed that. `Simulation` wasn't using our already adjusted memory estimation for `Variables` directly. We don't count SubArrays, they are already excluded. 

With that done, I'll merge once CI passes.

PR #1101, opened 2026-05-20
Mega transform! GPU-optimized batched transforms (Part 2: tendencies grid -> spectral)

maximilian-gelbrecht:
I want to keep these PRs relatively small so that they are easier to see through and review

Follow up from #1099 

Part 1: #1099 is limited to some general code infrastructure + optimising `transform!(::Variables, ::PrimitiveEquation)` 

Part 2: Here I want to tackle the next mega transform!: the grid tendencies -> spectral tendencies.


maximilian-gelbrecht:
Generally there's also some trade off decision to be made. 

Currently we have flexible scratch memory that's reused a lot, but this doesn't really work well with fusing together variables for these batched transforms. With 2D variables I wouldn't even think much about this and just allocate them separately, but all 3Ds as well?

maximilian-gelbrecht:
The main idea behind this PR is that we split most dycore computations in three parts: 
* compute grid tendencies 
* transform grid tendencies -> spectral tendencies: batched all together 
* compute what we need with the spectral tendencies

This affects: flux_divergence!, horizontal_advection!, bernoulli_potential! ,  vorticity_flux_curldiv! ,  vordiv_tendencies!, temperature_tendency!, humidity_tendency!  and surface_pressure_tendency! 

Then major part of the dycore basically reads 

```Julia 
gird_tendencies!(vars, model)
transform_tendencies!(vars, model)
spectral_tendencies!(vars, model)
```

From my point of view this and part 1 are the two major revisions, then there's some minor stuff we can still do as follow-ups in part 3.

maximilian-gelbrecht:
Ok I pushed the major refactor. Currently I also keep a sequential version here. I think we'll delete it eventually, but it also helps for testing. We could keep it in case we want actually a more memory optimised CPU version, but this isn't very important to me.

maximilian-gelbrecht:
Some early benchmarks 

## `dynamics_tendencies!` — PrimitiveWetModel, 8 layers

**Host:** login03 · **GPU:** NVIDIA A40 · **Julia:** 1.12.2 · **Metric:** `BenchmarkTools` median
**Feature branch:** `mg/gpu-mega-transform-part2` @ `b58fd05` vs **`main`**

### CPU — median time (ms)

| trunc | `main` | feature | Δ      | speedup |
|------:|-------:|--------:|-------:|--------:|
|    31 |   2.66 |    2.68 |  +0.02 |   0.99× |
|    63 |  12.34 |   11.76 |  −0.58 |   1.05× |
|   127 |  71.13 |   66.83 |  −4.30 |   1.06× |
|   255 | 435.07 |  388.63 | −46.44 |   1.12× |

### GPU (A40) — median time (ms)

| trunc | `main` | feature |     Δ  | speedup |
|------:|-------:|--------:|-------:|--------:|
|    31 |  13.61 |    3.60 | −10.01 |   3.78× |
|    63 |  29.77 |    7.08 | −22.69 |   4.21× |
|   127 |  56.00 |   15.57 | −40.43 |   3.60× |
|   255 | 119.85 |   46.74 | −73.11 |   2.56× |

maximilian-gelbrecht:
Correctness also looks good!

SpeedyWeather/src/time_stepping/leapfrog.jl
```diff
@@ -149,7 +149,7 @@ function leapfrog!(
     (; prognostic, tendencies) = vars
 
     for varname in keys(tendencies)
-        if !(tendencies[varname] isa NamedTuple)
+        if !(tendencies[varname] isa NamedTuple) && hasfield(typeof(prognostic), varname)
```
maximilian-gelbrecht: This is because `u` and `v` have tendency fields now. We could also just push those into `DynamicsVariabels` and get rid of this again.

milankl: ? u, v have tendencies in grid but not in spectral?

milankl:
2-4x speedup is really exciting!!! Very happy with this restructuring

SpeedyWeather/src/time_stepping/leapfrog.jl
maximilian-gelbrecht: Well, now we do. Happy to move those into `DynamicalVariables` though. But the `u` and `v` from `vordiv_tendencies` are the proper `u`, `v`, or not?

maximilian-gelbrecht:
This is also mostly adjusted now

RingGrids/ext/RingGridsReactantExt.jl
```diff
@@ -5,6 +5,6 @@ using Reactant
 
 const AnyReactantArray = Union{Reactant.AnyConcretePJRTArray, Reactant.AnyConcreteIFRTArray, Reactant.AnyTracedRArray}
 
-Base.Broadcast.BroadcastStyle(::Type{AbstractField{T, N, ArrayType, Grid}}) where {T, N, ArrayType <: AnyReactantArray, Grid} = Base.Broadcast.BroadcastStyle(ArrayType)
+Base.Broadcast.BroadcastStyle(::Type{F}) where {F <: AbstractField{T, N, ArrayType, Grid}} where {T, N, ArrayType <: AnyReactantArray, Grid} = Base.Broadcast.BroadcastStyle(ArrayType)
```
milankl: In #1150 Claude also flagged that if you dispatch over a type with constrained parameters you likely want to use `<:`. So, I think you can also write this as

```julia
Base.Broadcast.BroadcastStyle(::Type{<:AbstractField{T, N, ArrayType, Grid}}) where {T, N, ArrayType <: AnyReactantArray, Grid} = Base.Broadcast.BroadcastStyle(ArrayType)
```

SpeedyWeather/src/dynamics/spectral_grid.jl
```diff
@@ -300,11 +300,14 @@ Default `transform_batch` for a `SpectralGrid`. Architecture-dependent because:
...
 - On GPU, batched plans are essential, default:
   `[1, nlayers, 2*nlayers, 4*nlayers + 1]` — covers single-layer (`1`), one variable (`L`),
-  U/V together (`2L`), and the prognostic mega-batch (`4L + 1`).
+  U/V together (`2L`), and the prognostic batch (`4L + 1`),  6L+1 and 9L+1 (tendency batch)
 """
 default_transform_batch(arch::AbstractArchitecture, nlayers::Integer) = default_transform_batch(typeof(arch), nlayers)
 default_transform_batch(::Type{<:AbstractCPU}, nlayers::Integer) = Int[1, nlayers]
-default_transform_batch(::Type{<:AbstractArchitecture}, nlayers::Integer) = Int[1, nlayers, 2 * nlayers, 4 * nlayers + 1]
+
+# TODO: planning both 6L+1 and 9L+1 is slightly wasteful (we only need one of them for wet and dry model respectively)
```
milankl, line 338: Does that contribute significantly to creating a SpectralTransform?

SpeedyWeather/src/models/primitive_wet.jl
```diff
@@ -158,8 +158,13 @@ function variables(::Type{<:PrimitiveWet}, nsteps = DEFAULT_NSTEPS)
...
         GridVariable(:humidity, Grid4D(pg), desc = "Specific Humidity", units = "kg/kg", fuse=:grid),
-        TendencyVariable(:humidity, Spectral4D(ts), desc = "Tendency of specific humidity", units = "kg/kg/s"),
-        TendencyVariable(:humidity, Grid4D(ts), namespace = :grid, desc = "Tendency of specific humidity on the grid", units = "kg/kg/s"),
+        TendencyVariable(:humidity, Spectral4D(ts), desc = "Tendency of specific humidity", units = "kg/kg/s", fuse = :spectral_tendencies),
+        TendencyVariable(:humidity, Grid4D(ts), namespace = :grid, desc = "Tendency of specific humidity on the grid", units = "kg/kg/s", fuse = :grid_tendencies),
+
+        DynamicsVariable(:uq, Grid4D(tg), desc = "u*humidity intermediate on grid", namespace = :grid, fuse = :grid_tendencies),
+        DynamicsVariable(:vq, Grid4D(tg), desc = "v*humidity intermediate on grid", namespace = :grid, fuse = :grid_tendencies),
+        DynamicsVariable(:uq, Spectral4D(ts), desc = "u*humidity intermediate in spectral space", fuse = :spectral_tendencies),
+        DynamicsVariable(:vq, Spectral4D(ts), desc = "v*humidity intermediate in spectral space", fuse = :spectral_tendencies),
```
milankl: Ouf that's a lot of additional memory. Can that not be avoided?

SpeedyWeather/src/models/shallow_water.jl
```diff
@@ -81,6 +81,9 @@ function variables(model::ShallowWater)
         GridVariable(:η, Grid3D(pg), desc = "Interface displacement", units = "m", fuse = :grid),
         DynamicsVariable(:geopotential, Grid3D(), desc = "Geopotential", units = "m²/s²"),
 
+        DynamicsVariable(:kinetic_energy, Grid4D(tg), desc = "Kinetic energy intermediate, ½(u²+v²)+Φ", namespace = :grid, fuse = :grid_tendencies),
+        DynamicsVariable(:kinetic_energy, Spectral4D(ts), desc = "Kinetic energy intermediate in spectral space", fuse = :spectral_tendencies),
```
milankl: Same here, but I guess in shallow water we don't really care about the additional memory as much ...

SpeedyWeather/src/dynamics/spectral_grid.jl
milankl, line 338: Is it significant though? If it takes a little longer to create a `SpectralTransform` so be it?

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Dynamical core reordered and transformed batched for improved GPU performance. The majority of the dycore now follows the scheme: `grid_tendencies!` -> `transform!` -> `spectral_tendencies!` to batch together all grid -> spectral transforms of tendencies and auxiliary tendencies [#1101](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1101)
+- Prognostic and grid variables are now fused together and the `transform!(::Variables, ::PrimitveEquation)` is batched together to make it more efficient on GPU (uses the system introduced firsti in [#1083]). For this purposes new routines for flexible FFTs on CPU and GPU had to be introduced and the fused variables are now a new type `FusedParents` [#1099](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1099/)
```
milankl, line 6: Your Claude has a tendency to write really long changelog entries. I don't think any explanation is needed here just the PR title essentially?

SpeedyWeather/src/models/primitive_dry.jl
```diff
@@ -149,20 +149,30 @@ function variables(::Type{<:PrimitiveDry}, nsteps = DEFAULT_NSTEPS)
...
         
         GridVariable(:divergence, Grid4D(pg), desc = "Divergence", units = "1/s", fuse=:grid),
         GridVariable(:temperature, Grid4D(pg), desc = "Temperature", units = "K", fuse=:grid),
         GridVariable(:pressure, Grid3D(pg), desc = "Logarithm of surface pressure", units = "log(Pa)", fuse=:grid),
         ParameterizationVariable(:surface_pressure, Grid2D(), desc = "Surface pressure", units = "Pa"),
 
-        DynamicsVariable(:dpres_dx, Grid2D(), desc = "Zonal gradient of the logarithm of surface pressure"),
-        DynamicsVariable(:dpres_dy, Grid2D(), desc = "Meridional gradient of the logarithm of surface pressure"),
+        DynamicsVariable(:uT_anomaly, Grid4D(tg), desc = "u*T anomaly intermediate on grid", namespace = :grid, fuse = :grid_tendencies),
+        DynamicsVariable(:vT_anomaly, Grid4D(tg), desc = "v*T anomaly intermediate on grid", namespace = :grid, fuse = :grid_tendencies),
```
milankl: ouf not sure I approve of all of these new variables ...

SpeedyWeather/src/dynamics/tendencies.jl
```diff
@@ -960,13 +1353,69 @@ function bernoulli_potential!(
...
+
+"""$(TYPEDSIGNATURES)
+Gridded contribution of `bernoulli_potential!` for ShallowWater: writes `kinetic_energy_grid = ½(u²+v²)+Φ`
+(geopotential Φ is included because in ShallowWater geopotential lives only on the grid)."""
+function bernoulli_grid_potential!(vars::Variables, model::ShallowWater, time_stepping::AbstractTimeStepper)
+    u = get_prognostic_step(vars.grid.u, time_stepping, DynamicalCore(), model)
+    v = get_prognostic_step(vars.grid.v, time_stepping, DynamicalCore(), model)
+    Φ = vars.dynamics.geopotential
+    bernoulli_grid = get_step(vars.dynamics.grid.kinetic_energy)
+    half = convert(eltype(bernoulli_grid), 0.5)
```
milankl, line 1149: did the previous `1 // 2` not compile away?

milankl (commented):
I need to have an closer look at the dycore as you restructured a lot which is awesome! But to summarize we essentially allocate a bunch more intermediate variables so that things can be split and batched right? Just wondering whether it's still possible to avoid all those additional Fields now ...

maximilian-gelbrecht:
> I need to have an closer look at the dycore as you restructured a lot which is awesome! But to summarize we essentially allocate a bunch more intermediate variables so that things can be split and batched right? Just wondering whether it's still possible to avoid all those additional Fields now ...

In the end it's a trade off we can decide on. Right now I batched _all_ transforms of tendencies from grid to spectral. We can also try to just a batch a few and see what difference that makes. But in the end every small `transform!` is a bottleneck on GPU. We could also make it architecture dependent without too much trouble I think.

SpeedyWeather/src/dynamics/spectral_grid.jl
maximilian-gelbrecht: The plans are almost instantly created. It's just a bit wasteful in terms of memory, but it's also not that bad.

maximilian-gelbrecht:
Also look at the benchmark results, both in previous comments here and just redone and committed in the files. The GPU performance is really good. I don't thinking doing an architecture dependent dispatch here would be that bad.

milankl:
> Also look at the benchmark results, both in previous comments here and just redone and committed in the files. The GPU performance is really good. I don't thinking doing an architecture dependent dispatch here would be that bad.

🚀 Indeed, my favourite is probably the 160 SYPD for a T127 L24 1˚ model, sooo good! @pierluigividale But why is the primitive dry model always slower than the primitive wet model, and also the memory across matrix vs LT+FFT transforms does not seem to make sense?

RingGrids/ext/RingGridsReactantExt.jl
maximilian-gelbrecht: done

CHANGELOG.md
maximilian-gelbrecht: I disagree. If we only keep the title of the PR, then we could also abandon the changelog altogether because the title of the PRs are already automatically collected by GitHub when issuing a release.

maximilian-gelbrecht:
What would be your choice though? 

* Keep it as it is in this PR: fuse everything everywhere
* Architecture dependent: GPU fuse everything, CPU fuse next to nothing 
* Don't fuse everything, e.g. leave out `u*humidity` and `v*humidity`
* Just don't do it at all

maximilian-gelbrecht:
This is just the memory allocated by `Variables` (on CPU, but allocations on GPU should be similar) 

| trunc | nlayers | main | branch | increase |
|-------|---------|-----------:|-------------:|---------:|
| T31  | 8  | 4.0 MiB   | 4.7 MiB   | +18.5% |
| T42  | 8  | 6.6 MiB   | 7.9 MiB   | +18.7% |
| T63  | 8  | 13.9 MiB  | 16.5 MiB  | +18.8% |
| T85  | 8  | 23.9 MiB  | 28.5 MiB  | +19.0% |
| T127 | 8  | 51.8 MiB  | 61.7 MiB  | +19.0% |
| T170 | 8  | 90.5 MiB  | 107.8 MiB | +19.1% |
| T255 | 8  | 199.7 MiB | 237.9 MiB | +19.1% |
| T85  | 16 | 43.0 MiB  | 52.0 MiB  | +21.0% |
| T127 | 16 | 93.1 MiB  | 112.7 MiB | +21.0% |
| T170 | 16 | 162.7 MiB | 197.0 MiB | +21.1% |
| T255 | 16 | 359.0 MiB | 435.0 MiB | +21.2% |
| T85  | 24 | 62.1 MiB  | 75.6 MiB  | +21.8% |
| T127 | 24 | 134.4 MiB | 163.8 MiB | +21.8% |
| T170 | 24 | 234.9 MiB | 286.3 MiB | +21.9% |
| T255 | 24 | 518.4 MiB | 632.1 MiB | +21.9% |

I would say that's not so bad at all. A 20% increase and even at T255L24 `Variables` just takes 600MiB, that's totally okay still.

pierluigividale:
Hi. This seems quite modest indeed. For reference, is there a table that shows what kind of memory is normally available to GPUs, meaning on-node memory, not via MPI?

pierluigividale:
> > Also look at the benchmark results, both in previous comments here and just redone and committed in the files. The GPU performance is really good. I don't thinking doing an architecture dependent dispatch here would be that bad.
> 
> 🚀 Indeed, my favourite is probably the 160 SYPD for a T127 L24 1˚ model, sooo good! @pierluigividale But why is the primitive dry model always slower than the primitive wet model, and also the memory across matrix vs LT+FFT transforms does not seem to make sense?

Hi, 
1) I think that I need to repeat these benchmarks, because at the time JASMIN was all over the place, and a few jobs did sit on the machine forever, and had to be killed, but I also suspect that others were waiting on the scratch disk, which eventually had to be zapped (and we are now on an entirely new scratch disk). I can do that once you say that we have a meta-stable model version.
2) can you point me to the latest table with performance etc. at each resolution, so that I know where I am standing, before asking silly questions on "Issues"?

maximilian-gelbrecht:
> Hi. This seems quite modest indeed. For reference, is there a table that shows what kind of memory is normally available to GPUs, meaning on-node memory, not via MPI?

A A40 has 48GB, A H100 has 80GB, H200 141GB

Note though that what I posted there is just the `Variables`, the model itself also allocates memory, which should be less than that of the `Variables` but definitely also non-neglactable.

maximilian-gelbrecht:
> > > Also look at the benchmark results, both in previous comments here and just redone and committed in the files. The GPU performance is really good. I don't thinking doing an architecture dependent dispatch here would be that bad.
> > 
> > 
> > 🚀 Indeed, my favourite is probably the 160 SYPD for a T127 L24 1˚ model, sooo good! @pierluigividale But why is the primitive dry model always slower than the primitive wet model, and also the memory across matrix vs LT+FFT transforms does not seem to make sense?
> 
> Hi, can you point me to the latest table with performance etc. at each resolution?

Here are the results on `main` https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/benchmarks 
Here on this branch with the second part of the dycore revision, we didn't render the docs yet, you can find the raw results there: https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1101/changes#diff-cafe2dc6bc63f6b240606b135810f10321b700928999e0af45ea52dce8d2b1c6 (SpeedyWeather -> benchmark -> README.md) 
 
The benchmarking certainly isn't perfect though. Those are not paper-ready benchmarks. We run each configuriation for too short a time / rollout to have really well reliable results. The problem is that already with that short integrations the benchmarks take ~45 min. We definitely want to keep them under 1h so that we can already use them commonly during development. For publications we'd just integrate longer and repeat them to have a proper and more reliable  statistic on the performance.

pierluigividale:
Yes, for publication and for any HPC grant requests. BTW, I am not sure I copied you in on that, but Cambridge have a reasonably attractive GPU cluster, and they were using for relevant users, so I told them about Speedy and NumericalEarth. They already seemed aware of SpeedyWeather, so hopefully more people had entered that. Have not heard since.

CHANGELOG.md
milankl: Point taken! But I also see the changelog as important for us before a release is out to be more aware of the order in which we merge things or what was merged last

SpeedyWeather/src/dynamics/tendencies.jl
```diff
@@ -1,13 +1,78 @@
...
+    # compute tendencies in grid space: u, v
+    grid_tendencies!(vars, model)
+
+    # batched transform of grid tendencies to spectral space
+    transform!(get_tendency_step(parent(vars.fused.spectral_tendencies), time_stepping, DynamicalCore()),
+               get_tendency_step(parent(vars.fused.grid_tendencies), time_stepping, DynamicalCore()),
+               vars.scratch.transform_memory, model.spectral_transform)
+
+    # accumulates into final spectral tendencies: vorticity
+    spectral_tendencies!(vars, model)
```
milankl, line 76: generally really like this!

SpeedyWeather/src/dynamics/tendencies.jl
```diff
@@ -18,15 +83,24 @@ function dynamics_tendencies!(
...
+    # batched transform of grid tendencies to spectral space
+    transform!(get_tendency_step(parent(vars.fused.spectral_tendencies), time_stepping, DynamicalCore()),
+               get_tendency_step(parent(vars.fused.grid_tendencies), time_stepping, DynamicalCore()),
+               vars.scratch.transform_memory, spectral_transform)
+
+    # accumulates into final spectral tendencies: vorticity, divergence
+    spectral_tendencies!(vars, model)
 
     # = -∇⋅(uh, vh), tendency for interface displacement η
     volume_flux_divergence!(vars, model)
```
milankl: This can't be written as grid/spectral tendency?

SpeedyWeather/src/dynamics/tendencies.jl
```diff
@@ -453,9 +572,49 @@ function vordiv_tendencies!(
...
+    scratch_memory = vars.scratch.transform_memory
+    transform!(u_tend, u_tend_grid, scratch_memory, S)
+    transform!(v_tend, v_tend_grid, scratch_memory, S)
+
+    vordiv_spectral_tendencies!(vars, time_stepping, S)
+    return nothing
+end
+
+# TODO: Might rename this just to u, v spectral tendencies? Because that's what it does, but 
+# then it's also nice to always have the symmetric with *_grid_tendencies! _spectral_tendencies!
```
milankl, line 514: yeah I think that's fine! especially now that the transform is externalised

SpeedyWeather/src/dynamics/tendencies.jl
```diff
@@ -540,37 +711,39 @@ function parameterization_tendencies_only!(
...
-    transform!(u_tend, u_tend_grid, scratch_memory, S)
-    transform!(v_tend, v_tend_grid, scratch_memory, S)
-    transform!(temp_tend, temp_tend_grid, scratch_memory, S)
-
-    # humidity only for models that have humidity
-    if haskey(vars.tendencies, :humidity)
-        humid_tend_grid = get_tendency_step(vars.tendencies.grid.humidity, TS, DummyParameterization())
-        humid_tend = get_tendency_step(vars.tendencies.humidity, TS, DynamicalCore())
-        transform!(humid_tend, humid_tend_grid, scratch_memory, S)
-    end
```
milankl, line 566: Ah yeah great that this just disappears because humidity isn't part of the fuse!

SpeedyWeather/src/dynamics/tendencies.jl
```diff
@@ -582,11 +755,33 @@ function temperature_tendency!(
         vars::Variables,
         model::PrimitiveEquation,
     )
+    (; spectral_transform, time_stepping) = model
 
-    (; adiabatic_conversion, atmosphere, implicit, spectral_transform, time_stepping) = model
+    temperature_grid_tendency!(vars, model)
```
milankl: I guess this can be just generally part of `grid_tendencies!`?

milankl: Not necessarily now but in a future revision ...

SpeedyWeather/src/dynamics/tendencies.jl
```diff
@@ -611,13 +805,34 @@ function temperature_tendency!(
         pres_flux, temp_profile, adiabatic_conversion.σ_lnp_A, adiabatic_conversion.σ_lnp_B, atmosphere
     )
 
-    transform!(temp_tend, temp_tend_grid, scratch_memory, spectral_transform)
+    # write uT_anomaly_grid, vT_anomaly_grid (= u·T', v·T') for the flux divergence
+    flux_grid_divergence!(get_step(vars.dynamics.grid.uT_anomaly), get_step(vars.dynamics.grid.vT_anomaly), temp, vars, model)
```
milankl, line 698: why does a dynamics variable need a `get_step` method? This shouldn't be (?) and also the use of `get_step` without `::AbstractTimeStepper, ::ModelComponent` arguments should only happen for convenient user interface

SpeedyWeather/src/dynamics/tendencies.jl
```diff
@@ -611,13 +805,34 @@ function temperature_tendency!(
...
+`+=` because the tendencies already contain parameterizations and vertical advection.
+`T'` is the anomaly with respect to the reference/average temperature. Tᵥ is the virtual
+temperature used in the adiabatic term κTᵥ*Dlnp/Dt.
+
+Here, the previosuly computed gridded tendency contribution are accumulated and the spectral
+tendency computed with divergence is computed via the `flux_spectral_divergence!` function."""
+function temperature_spectral_tendency!(vars::Variables, S::AbstractSpectralTransform, time_stepping::AbstractTimeStepper)
+    temp_tend = get_tendency_step(vars.tendencies.temperature, time_stepping, DynamicalCore())
+    uT_spec = get_step(vars.dynamics.uT_anomaly)
+    vT_spec = get_step(vars.dynamics.vT_anomaly)
```
milankl, line 717: Also here, why do uT, vT have a step dimension?

SpeedyWeather/src/models/primitive_dry.jl
```diff
@@ -149,20 +149,30 @@ function variables(::Type{<:PrimitiveDry}, nsteps = DEFAULT_NSTEPS)
...
 
-        DynamicsVariable(:dpres_dx, Grid2D(), desc = "Zonal gradient of the logarithm of surface pressure"),
-        DynamicsVariable(:dpres_dy, Grid2D(), desc = "Meridional gradient of the logarithm of surface pressure"),
+        DynamicsVariable(:uT_anomaly, GridXYZT(tg), desc = "u*T anomaly intermediate on grid", namespace = :grid, fuse = :grid_tendencies),
+        DynamicsVariable(:vT_anomaly, GridXYZT(tg), desc = "v*T anomaly intermediate on grid", namespace = :grid, fuse = :grid_tendencies),
+        DynamicsVariable(:uT_anomaly, SpectralXYZT(ts), desc = "u*T anomaly intermediate in spectral space", fuse = :spectral_tendencies),
+        DynamicsVariable(:vT_anomaly, SpectralXYZT(ts), desc = "v*T anomaly intermediate in spectral space", fuse = :spectral_tendencies),
+
+        DynamicsVariable(:kinetic_energy, GridXYZT(tg), desc = "Kinetic energy intermediate, ½(u²+v²)", namespace = :grid, fuse = :grid_tendencies),
+        DynamicsVariable(:kinetic_energy, SpectralXYZT(ts), desc = "Kinetic energy intermediate in spectral space", fuse = :spectral_tendencies),
```
milankl, line 170: Why do these have a step dimension? They shouldn't need to?

SpeedyWeather/src/dynamics/tendencies.jl
```diff
@@ -688,20 +943,56 @@ function tracer_advection!(
...
+Writes `+A*div` to `A_tend_grid` and `(u*A, v*A)` to `(uA_grid, vA_grid)`
+"""
+function horizontal_grid_advection!(
+        A_tend_grid::AbstractField,         # Input/Output: A_tend on grid, accumulates +A*div term
+        A_grid::AbstractField,              # Input: grid field to be advected
+        vars::Variables,
+        model::AbstractModel;
+        add::Bool = true,                   # use muladd (true) or overwrite (false) for the +A*div term
+        uA_grid = get_step(vars.dynamics.grid.uT_anomaly),   # caller picks the correct named slot
+        vA_grid = get_step(vars.dynamics.grid.vT_anomaly),
```
milankl, line 823: Ouh is this type stable to pass on parts of the variables as key word arguments? Shouldn't this be moved inside the function?

SpeedyWeather/src/dynamics/tendencies.jl
```diff
@@ -749,32 +1056,61 @@ function flux_divergence!(
...
         flipsign::Bool = true,          # compute -∇⋅((u, v)*A) (true) or ∇⋅((u, v)*A)?
+        # Named slots for the spec/grid intermediates u*A and v*A. Default to the unfused
+        # :a/:b scratches (used by tracer_advection!, volume_flux_divergence! for η);
+        # named callers (temperature_tendency!, humidity_tendency!) pass slots from the
+        # :spectral_tendencies/:grid_tendencies fuse parents (uT_anomaly/vT_anomaly, uq/vq).
+        # When the slots are fused, dycore callers prefer the grid/spectral split below.
+        uA = vars.scratch.a,            # = u*A in spectral
+        vA = vars.scratch.b,            # = v*A in spectral
+        uA_grid = vars.scratch.grid.a,  # = u*A on grid
+        vA_grid = vars.scratch.grid.b,  # = v*A on grid
```
milankl, line 900: Is this necessary to maintain backwards compatibility to non-fused variables?

milankl (commented):
Generally really great and I approve but 

1. Can we check why the new dynamics variables have a step dimension? With the generalised time stepping I tried to introduce a logic whereby all prognostic/tendency variables need a step dimension but no other variable should have one. Otherwise we should have gone straight to *all* variables having a step dimension which seems an overkill to implement now though I see the appeal for consistency. Anyway having some dynamics variables having a step dimension is neither of these two logics

2. A user-facing `get_step` can drop any arguments, that's okay but internally I would always want the "var, time stepper, component, and optionally model" arguments as otherwise the time stepper has no control over the step dimension

3. Is it really okay to pass on some arrays as keyword arguments? How does that relate to the compatibilty with/without fused variables?

SpeedyWeather/src/dynamics/tendencies.jl
maximilian-gelbrecht: Ah yes, well spotted. To be honest, I mostly looked at the primitive equation model, so this got kind of lost. But it also already written like this, just not used like that here.

SpeedyWeather/src/dynamics/tendencies.jl
maximilian-gelbrecht: Which variant is fine? Keep it as it is or rename it?

SpeedyWeather/src/dynamics/tendencies.jl
maximilian-gelbrecht: It is part of the `grid_tendencies!` 

This here is just the old / sequential version calling it. 

Mh... maybe we actually should move those to a different file? Or just delete them alltogether?

SpeedyWeather/src/dynamics/tendencies.jl
maximilian-gelbrecht: Yes, that's not ideal, but from what I can see unfortunately necessary. 

These variables are fused together with tendencies that have a step dimension. Currently we only fuse along the vertical layer dimension, so here the additional tendency - dynamics variables get a superfluous step. 

I don't really see an alternative at the moment.

SpeedyWeather/src/dynamics/tendencies.jl
maximilian-gelbrecht: See above

SpeedyWeather/src/models/primitive_dry.jl
maximilian-gelbrecht: See above

SpeedyWeather/src/dynamics/tendencies.jl
maximilian-gelbrecht: Mh, I think it should?

But this is also just here to have still compatibility with the old / sequential un-fused way of calling it without code duplication.

SpeedyWeather/src/dynamics/tendencies.jl
maximilian-gelbrecht: yes exactly. also see above

milankl: great sounds good, maybe at some point we decide to only support fused variables but that's not for now, thanks for the clarification

SpeedyWeather/src/dynamics/tendencies.jl
milankl: Yep that sounds fine with me, we could move that in case we decide to drop the compatibility

SpeedyWeather/src/dynamics/tendencies.jl
maximilian-gelbrecht: done

maximilian-gelbrecht:
I reorganised the sequential tendency functions into a new file and also included the `volume_flux_divergence!` into the fused approach for the `ShallowWaterModel`.

SpeedyWeather/src/dynamics/tendencies.jl
milankl: Great we can do that in a follow up just so I know, as you said in the end splitting everything into

- grid
- transforms
- spectral

makes absolute sense, and allows for batched transforms and fused kernels whether implemented now or later. I don't think a consistent logic is necessary for this PR though, this is about batched transforms after all

SpeedyWeather/src/dynamics/tendencies.jl
milankl: Yeah eventually I'd be keen to remove old code for clarity but also this is dycore so not very user-facing or concerns about breaking changes, so all good for now, let's clean that up another time. This is about batched transforms in the first place

maximilian-gelbrecht:
@milankl I would merge this now. But in Julia 1.12 the vertical coordinates test took really forever. Do you have any idea what that could be?

SpeedyWeather/src/models/shallow_water.jl
```diff
@@ -81,6 +81,14 @@ function variables(model::ShallowWater)
         GridVariable(:η, GridXYT(pg), desc = "Interface displacement", units = "m", fuse = :grid),
         DynamicsVariable(:geopotential, GridXYZ(), desc = "Geopotential", units = "m²/s²"),
 
+        DynamicsVariable(:kinetic_energy, GridXYZT(tg), desc = "Kinetic energy intermediate, ½(u²+v²)+Φ", namespace = :grid, fuse = :grid_tendencies),
+        DynamicsVariable(:kinetic_energy, SpectralXYZT(ts), desc = "Kinetic energy intermediate in spectral space", fuse = :spectral_tendencies),
+
+        DynamicsVariable(:uh, GridXYT(tg), desc = "u*h volume flux intermediate on grid", namespace = :grid, fuse = :grid_tendencies),
+        DynamicsVariable(:vh, GridXYT(tg), desc = "v*h volume flux intermediate on grid", namespace = :grid, fuse = :grid_tendencies),
+        DynamicsVariable(:uh, SpectralXYT(ts), desc = "u*h volume flux intermediate in spectral space", fuse = :spectral_tendencies),
+        DynamicsVariable(:vh, SpectralXYT(ts), desc = "v*h volume flux intermediate in spectral space", fuse = :spectral_tendencies),
```
milankl, line 90: shouldn't these be `GridXYT(1)` always? Like I understand you need the step dimension but for NCycleLorenz this would be 2 steps not one, which isn't needed to fuse with other spectral tendencies is it?

SpeedyWeather/src/models/primitive_dry.jl
```diff
@@ -149,20 +149,30 @@ function variables(::Type{<:PrimitiveDry}, nsteps = DEFAULT_NSTEPS)
...
         GridVariable(:temperature, GridXYZT(pg), desc = "Temperature", units = "K", fuse=:grid),
         GridVariable(:pressure, GridXYT(pg), desc = "Logarithm of surface pressure", units = "log(Pa)", fuse=:grid),
         ParameterizationVariable(:surface_pressure, Grid2D(), desc = "Surface pressure", units = "Pa"),
 
-        DynamicsVariable(:dpres_dx, Grid2D(), desc = "Zonal gradient of the logarithm of surface pressure"),
-        DynamicsVariable(:dpres_dy, Grid2D(), desc = "Meridional gradient of the logarithm of surface pressure"),
+        DynamicsVariable(:uT_anomaly, GridXYZT(tg), desc = "u*T anomaly intermediate on grid", namespace = :grid, fuse = :grid_tendencies),
+        DynamicsVariable(:vT_anomaly, GridXYZT(tg), desc = "v*T anomaly intermediate on grid", namespace = :grid, fuse = :grid_tendencies),
+        DynamicsVariable(:uT_anomaly, SpectralXYZT(ts), desc = "u*T anomaly intermediate in spectral space", fuse = :spectral_tendencies),
+        DynamicsVariable(:vT_anomaly, SpectralXYZT(ts), desc = "v*T anomaly intermediate in spectral space", fuse = :spectral_tendencies),
```
milankl, line 167: Also here, would we ever need 2 steps or always just a singleton step dimenson, then use `GridXYZT()` and `SpectralXYZT()`?

milankl (approved):

milankl:
> @milankl I would merge this now. But in Julia 1.12 the vertical coordinates test took really forever. Do you have any idea what that could be?

They are more extensive now, testing up to 64 layers, but we could just not test that many!!! Feel free to shorten these tests, I'm happy with these being tested one-off and we can make them more extensive when overhauling the vertical coordinates completely

maximilian-gelbrecht:
> > @milankl I would merge this now. But in Julia 1.12 the vertical coordinates test took really forever. Do you have any idea what that could be?
> 
> They are more extensive now, testing up to 64 layers, but we could just not test that many!!! Feel free to shorten these tests, I'm happy with these being tested one-off and we can make them more extensive when overhauling the vertical coordinates completely

They took 43 minutes here just now, this just sounds more like something went wrong maybe.

SpeedyWeather/src/models/shallow_water.jl
maximilian-gelbrecht: If you want to concat the arrays together along the vertical layer dimension the other dimensions have to identical.

maximilian-gelbrecht: It's not great that this allocates superfluous memory, but I currently don't see a way around it.

milankl: then let's keep it for now! As we decided memory is not a concern here

milankl:
A shorter version of the vertical coordinates tests should be

```julia
@testset "Held-Suarez with varying layers and sigma spacing" begin
    for nlayers in (16, 24)     # don't test more layers regularly on CI but 64 should work too
        spectral_grid = SpectralGrid(trunc = 31; nlayers)

        σ_half_configs = (
            ("default",   SpeedyWeather.sigma_half_spacing(nlayers)),
            ("Frierson",  SpeedyWeather.sigma_half_spacing(nlayers, SpeedyWeather.frierson_profile)),
        )

        for (label, σ_half) in σ_half_configs
            coords = (
                ("SigmaCoordinates/$label",          SigmaCoordinates(spectral_grid, σ_half)),
                # like sigma only for now with A=0
                ("SigmaPressureCoordinates/$label",  SigmaPressureCoordinates(spectral_grid, σ_half; transition = _ -> 1.0)),
            )

            for (name, coord) in coords
                @testset "$name (nlayers=$nlayers)" begin
                    @info "$name (nlayers=$nlayers)"
                    model = PrimitiveDryModel(
                        spectral_grid;
                        geometry  = Geometry(spectral_grid; vertical_coordinates = coord),
                        forcing   = HeldSuarez(spectral_grid),
                        drag      = LinearDrag(spectral_grid),
                        dynamics_only = true,
                        orography = EarthOrography(spectral_grid),
                    )
                    run!(simulation; period = Day(5))               # don't test longer on CI but otherwise could be Day(20) too
                    if label == "default" && nlayers > 32           # current this blows up but that might be numerically expected
                        @test_broken model.feedback.nans_detected == false
                    else
                        @test model.feedback.nans_detected == false
                    end
                end
            end
        end
    end
end
```

don't want to commit this now here as this would retrigger all CI again, we can do a separate PR afterwards?

PR #1103, opened 2026-05-22
Benchmark across architectures

maximilian-gelbrecht:
This takes the existing benchmark suite and makes it possible to run it across different architectures. It only overwrites results from the same architecture, but keeps the others. It also keeps results from architectures as JSON files and also arranges a comparison table and figure. 

It also deletes the old gpu model benchmarks as they aren't needed anymore. I kept the CUDA transforms benchmarks, as these are not covered by our general benchmark suite


maximilian-gelbrecht:
I think the Reactant execution isn't completely right yet here, but maybe I also just adjust this later.

milankl:
Link to the respective readme in this branch https://github.com/SpeedyWeather/SpeedyWeather.jl/tree/mg/benchmark-revision/SpeedyWeather/benchmark

maximilian-gelbrecht:
I also added a lot of extra flexibility here so that we can easily modify the model setup and run with different configuration if we want to. 

The Reactant compat here is still just a stub, but this is also because Reactant for Speedy as a whole isn't fully working yet anyway.

PR #1104, opened 2026-05-22
Run the `ShallowWaterModel` on GPU

maximilian-gelbrecht:
This is a collection of bug fixes that we overlooked in previous PRs and genuine GPU porting. 

I noticed while working on #1103 that the `ShallowWater` doesn't actually run on GPU, I kind of forgot about that.

### DIFF
--- a/SpeedyWeather/src/variables/set.jl
+++ b/SpeedyWeather/src/variables/set.jl
@@ -145,8 +145,14 @@ function set!(
     # all elements are zero except for the 0,0 one
     var_new = zero(var)
 
-    for k in eachmatrix(var_new)
-        var_new[1, k] = norm_sphere * s
+    # set the [1, :] row in a GPU-safe way via a view on the underlying data
+    data = var_new.data
+    norm_s = convert(eltype(data), norm_sphere * s)
+    if ndims(data) == 1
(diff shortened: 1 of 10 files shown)


SpeedyWeather/src/variables/set.jl
```diff
@@ -145,8 +145,14 @@ function set!(
     # all elements are zero except for the 0,0 one
     var_new = zero(var)
 
-    for k in eachmatrix(var_new)
-        var_new[1, k] = norm_sphere * s
+    # set the [1, :] row in a GPU-safe way via a view on the underlying data
+    data = var_new.data
+    norm_s = convert(eltype(data), norm_sphere * s)
+    if ndims(data) == 1
```
maximilian-gelbrecht, line 151: this is some claude generated weird way to do it, but in the Reactant PR, we have a `set_scalar!` we can just reuse here then as well

maximilian-gelbrecht:
It runs now 🚀

SpeedyWeather/src/variables/set.jl
maximilian-gelbrecht: Actually, I think it's okay. Because this is also more than just `set_scalar!`

PR #1106, opened 2026-05-26
Handle heterogeneous parameter attributes

bgroenks96:
The existing implementation of `ParameterTable` does not properly handle the case when some parameters define an attribute and others do not. It simply takes the keys/attributes defined on the `first` parameter as being representative of the whole `ParameterTable`.

This PR adds a quick and dirty solution to handle this within `getindex`. A more elegant solution might be to use the functionality in `ModelParameters` to harmonize all `Param` fields when constructing the `ParameterTable`, but I don't see any concrete reason to prefer this at the moment.

Also adds a workaround for rafaqz/ModelParameters.jl#70

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Handle heterogeneous parameter attributes [#1106](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1106)
 - Coupling to the Terrarium.jl land model via an SpeedyWeatherTerrariumExt [#1090](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1090)
 - The `ShallowWaterModel` is now GPU compatible [#1104](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1104)
 - Updated citations and funding information [#1100](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1100)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Handle heterogeneous parameter attributes [#1106](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1106)
+- Handle heterogeneous parameter attributes in `ParameterEditing` [#1106](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1106)
 - Coupling to the Terrarium.jl land model via an SpeedyWeatherTerrariumExt [#1090](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1090)
 - The `ShallowWaterModel` is now GPU compatible [#1104](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1104)
 - Updated citations and funding information [#1100](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1100)
--- a/SpeedyWeatherInternals/src/ParameterEditing/parameter_table.jl
+++ b/SpeedyWeatherInternals/src/ParameterEditing/parameter_table.jl
@@ -87,11 +87,13 @@ Base.setindex!(ps::ParameterTable, x, nm::Symbol) = setindex!(ps, x, :, nm)
     end
 end
 
-Base.keys(params::ParameterTable) = (:idx, :fieldname, union(map(keys, params)...)...)
+# Workaround for https://github.com/rafaqz/ModelParameters.jl/issues/70
+# Need to iterate over params directly, not the ParameterTable itself
+Base.keys(ps::ParameterTable) = (:idx, :fieldname, union(map(keys, ModelParameters.params(parent(ps)))...)...)
 
-function Base.vec(params::ParameterTable)
+function Base.vec(ps::ParameterTable)
     # recursively unpack params and build component vector
-    return ComponentArray(unpack_params(params))
+    return ComponentArray(stripparams(ps))
 end
 
 # Override internal ModelParameters method _columntypes to condense type names in table schema (just to look nicer)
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Handle heterogeneous parameter attributes [#1106](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1106)
```
maximilian-gelbrecht:
```suggestion
- Handle heterogeneous parameter attributes in `ParameterEditing` [#1106](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1106)
```

SpeedyWeatherInternals/src/ParameterEditing/parameter_table.jl
```diff
@@ -59,8 +59,12 @@ Base.getindex(ps::ParameterTable, nm::Symbol) = getindex(ps, :, nm)
         1:length(ps)
```
maximilian-gelbrecht, line 59: Sorry, but I am a bit confused by the whole routine. 

Why does it return actually indices and that the elements of `ParameterTable` at those indices?

SpeedyWeatherInternals/src/ParameterEditing/parameter_table.jl
```diff
@@ -59,8 +59,12 @@ Base.getindex(ps::ParameterTable, nm::Symbol) = getindex(ps, :, nm)
...
     elseif nm == :fieldname
         ModelParameters.paramfieldnames(ps)
+    elseif any(map(p -> hasproperty(p, nm), ModelParameters.params(ps)))
+        map(ModelParameters.params(ps)) do p
+            hasproperty(p, nm) ? getproperty(p, nm) : nothing
```
maximilian-gelbrecht, line 64: Here it also actually returns the parameters, and not the indices, correct?

SpeedyWeatherInternals/src/ParameterEditing/parameter_table.jl
```diff
@@ -59,8 +59,12 @@ Base.getindex(ps::ParameterTable, nm::Symbol) = getindex(ps, :, nm)
         1:length(ps)
     elseif nm == :fieldname
         ModelParameters.paramfieldnames(ps)
+    elseif any(map(p -> hasproperty(p, nm), ModelParameters.params(ps)))
```
maximilian-gelbrecht, line 62: So this would cover a mix of the above or also something else?

SpeedyWeatherInternals/src/ParameterEditing/parameter_table.jl
bgroenks96: These are special cases for specific columns, in this case the index `idx` column.

SpeedyWeatherInternals/src/ParameterEditing/parameter_table.jl
bgroenks96: It always returns the specified column in the parameter table.

SpeedyWeatherInternals/src/ParameterEditing/parameter_table.jl
bgroenks96: This just covers the case where `nm` is any column other than previous two hardcoded ones, basically a property of the `Param` type.

SpeedyWeatherInternals/src/ParameterEditing/parameter_table.jl
maximilian-gelbrecht: Ok, got it

maximilian-gelbrecht (approved):

maximilian-gelbrecht:
You still want to reorganise the tests, correct? 

I think there should be also one new unit test added that tests the heterogeneous parameter attributes. 

Otherwise this could be merged from my side.

SpeedyWeatherInternals/src/ParameterEditing/parameterized.jl
```diff
@@ -119,14 +118,29 @@ macro parameterized(expr)
...
+    # for some reason, we need to use a dedicated block for the struct def otherwise it won't work
+    if !has_kwdef
+        push!(struct_block.args, Expr(:meta, :doc))
+    end
+    push!(struct_block.args, esc(new_expr))
+    push!(block.args, struct_block)
     ## 2. parameters method dispatch
-    push!(block.args, esc(:(ParameterEditing.parameters(obj::$(typename); kwargs...) = ParameterEditing.ParameterTable((; $(param_constructors...))))))
+    parameters_func = quote
+        function ParameterEditing.parameters(::Type{PT}, obj::$(typename); kwargs...) where {PT <: ParameterEditing.AbstractParam}
```
maximilian-gelbrecht, line 137: I don't quite understand the change here, could you just quickly explain it?

maximilian-gelbrecht (commented):
Looks good to me. Is there anything more that you want to do here or is this it?

maximilian-gelbrecht:
Also after merging with a new release in #1106 , I don't think there's any dependency/compat trouble right? We just need to also release a new minor version of SpeedyWeatherInternals as well.

SpeedyWeatherInternals/src/ParameterEditing/parameterized.jl
bgroenks96: In Terrarium, I have some instances of `@component`s that are empty when a particular parameterizaton/process has no parameters. These show up as empty NamedTuples in the resulting `ParameterTable`. This change just automatically removes empty subcomponents from the parameter `NamedTuple`.

bgroenks96:
I have no further changes to make here, I think.

SpeedyWeatherInternals/src/ParameterEditing/parameterized.jl
maximilian-gelbrecht: Ok sounds good.

PR #1107, opened 2026-05-27
Output in Simulation 

maximilian-gelbrecht:
Draft made with Claude. 

I am not super convinced as this is really a massive breaking change. Let's discuss it, but I think other things have priority. The time stepping would need a few further changes to clean this up. 

In principle though I do think that the model types should hold the physical model specifications only and output related functions should rather fit in the simulation. But this isn't something very urgent. 

Unit tests all pass locally.


maximilian-gelbrecht:
This branch is completely outdated by now. Just redoing it with Claude is more efficient, if we really want it

PR #1109, opened 2026-05-29
CUDA Graphs for fourier!

maximilian-gelbrecht:
I just gave this to Claude with specific goals for the correctness and speedup. Otherwise not much interaction from me. 

Correctness tests pass and there's a 4-5x speed up at T127L8. What's left to here is basically clean-up, further robustness tests and benchmarks.


maximilian-gelbrecht:
Note, there is also definitely some room for optimisation here because of the individual `copyto!`s that take up a significant part of the compute.

milankl:
> Note, there is also definitely some room for optimisation here because of the individual `copyto!`s that take up a significant part of the compute.

But they are only necessary in FFTW plans as they expect a certain stride, maybe this is different for cuFFT and we should specialize here?

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
```diff
@@ -1,14 +1,246 @@
...
+# The captured graph bakes in the device pointers of `field.data`, the scratch buffers
+# (`S.scratch_memory.north/.south`, stable for the lifetime of `S`) and the per-ring
+# work buffers. In the SpeedyWeather time loop the same variable buffers are reused every
+# timestep, so a graph captured for a given `field` is replayed on all subsequent steps.
+# Graphs are cached per `SpectralTransform` and keyed by the `field.data` object.
+# =====================================================================================
+
+"""Toggle for the CUDA-Graphs accelerated batched Fourier transform. Set to `false` to
+fall back to the generic (allocating) per-ring GPU path, e.g. for benchmarking."""
+const FOURIER_GRAPHS_ENABLED = Ref(true)
```
milankl: Can this be a field in the immutable `SpectalTransform`?

maximilian-gelbrecht:
> > Note, there is also definitely some room for optimisation here because of the individual `copyto!`s that take up a significant part of the compute.
> 
> But they are only necessary in FFTW plans as they expect a certain stride, maybe this is different for cuFFT and we should specialize here?

Yeah Claude just optimised the buffer here now as well. I look into this in detail on Monday. 

Now we are at 
```
========== CORRECTNESS (full transform!, ON vs OFF) ==========
graphs created: 2
forward (grid->spec) max abs err = 3.5762787e-7  (rel 2.0118368e-7)
inverse (spec->grid) max abs err = 0.0  (rel 0.0)
match within fp ✓
repeated-call forward err vs OFF = 2.3841858e-7

========== BENCHMARK _fourier_batched! (per call) ==========
forward  OFF=5972.3 µs   ON=767.6 µs   speedup=7.78x
inverse  OFF=5767.1 µs   ON=1120.9 µs   speedup=5.15x

========== BENCHMARK full transform! (per call) ==========
forward  OFF=6286.7 µs   ON=1085.1 µs   speedup=5.79x
inverse  OFF=7289.9 µs   ON=1205.6 µs   speedup=6.05x

========== INTEGRATION: PrimitiveWet, 6 steps, ON vs OFF ==========
graphs used in run: 17
PrimitiveWet vorticity max abs diff (ON vs OFF) = 3.5458925e-10  rel=1.393176e-5
```

maximilian-gelbrecht:
A main thing to test is how this interacts with the batching in the other PRs with different layer numbers, there might some things we need to treat slightly different. We already need different plans for the different layer numbers anyway, and with this we also might need different caches for the different layer numbers.

maximilian-gelbrecht:
This is now adjusted in a way that should make it relatively easy to extend it to the case of transforms with different batch sizes that introduced in #1099.

maximilian-gelbrecht:
In principle this is orthogonal to all the other changes and could be merged before the other PRs. @milankl what do you think? 

To summarise very briefly the logic and what's implemented here: 

* CUDA Graphs get rid of the kernel launch overhead by "recording" the kernel launching and then "replaying" them. 
* This only works if there are zero allocations in the recorded loop and the memory addresses of each graph stay the same 
* Therefore we define here zero-allocation variants of our ring Fourier transform and a buffer in which we collect and write the data 
* So every FT is: write into buffer -> CUDA graph FT -> read from buffer 
* The first time a transform is executed the graph is recorded, all consecutive ones it is just executed. 
* The graphs are saved in an `IDDict` in the extension and are keyed with the FFT plans and arrays they are defined for
* There a maximum number of graphs set so that we that we prevent leakage/runaway memory, in the unit tests we also check that this doesn't happen und the number of recorded graphs isn't increasing during model integration.

milankl:
> In principle this is orthogonal to all the other changes and could be merged before the other PRs. @milankl what do you think?

Yes absolutely keen to reduce the backlog of PRs and starting merging things whenever we can!

> There a maximum number of graphs set so that we that we prevent leakage/runaway memory, in the unit tests we also check that this doesn't happen und the number of recorded graphs isn't increasing during model integration.

What does number of graphs refer to algorithmically? is that the number of FFTs or generally the worksize over which the kernel is launched? So for 128 max graphs what we limited us here too.

I suggested to have a quick performance check not just with the `OctahedralGaussianGrid` but also `OctaminimalGaussianGrid` (because smallest FFT is 4, not 20 points), `OctaHEALPix` (also smallest FFT 4 but an uneven number of latitude rings, does that make a difference odd/even?) and `HEALPix` because the max FFT length is half as much as on other grids in the limit.

I'm basically just curious what the relationship between FFT length, graphs and performance is. Maybe helps us to understand what CUDA graphs are good or not so good at.

Just a quick overview of FFT lengths on different grids, all with `nlat_half = 16`. By column, FullGaussian, OctahedralGaussian, OctaminimalGaussian, OctaHEALPix, HEALPix

```
 F    O   M  OH   H    # Grid type full Gaussian, octahedral Gaussian, octaminial Gaussian, OctaHEALPix, HEALPix
 64  20   4   4   4    # <- FFT length around North pole
 64  24   8   8   8
 64  28  12  12  12
 64  32  16  16  16
 64  36  20  20  20
 64  40  24  24  24
 64  44  28  28  28
 64  48  32  32  32
 64  52  36  36  32
 64  56  40  40  32
 64  60  44  44  32
 64  64  48  48  32
 64  68  52  52  32
 64  72  56  56  32
 64  76  60  60  32
 64  80  64  64  32    # FFT length at equator
 64  80  64  60  32
 64  76  60  56  32
 64  72  56  52  32
 64  68  52  48  32
 64  64  48  44  32
 64  60  44  40  32
 64  56  40  36  32
 64  52  36  32  32
 64  48  32  28  28
 64  44  28  24  24
 64  40  24  20  20
 64  36  20  16  16
 64  32  16  12  12
 64  28  12   8   8
 64  24   8   4   4
 64  20   4            # South pole
```

maximilian-gelbrecht:
> What does number of graphs refer to algorithmically? is that the number of FFTs or generally the worksize over which the kernel is launched? So for 128 max graphs what we limited us here too.

A graph is the recording of the kernel executions und a graph is recording the memory addresses / pointers as well. So you get a graph for every array you record. While we do use buffers, we also need to record the reading and scattering from/to the buffer. In practice for our model, we currently have one graph per prognostic variable. As the array's addresses don't change during the integration of the model, the graphs can be always reused. In the GPU CI we now test as well that the number of graphs stays constant during integration. The number of graphs will change with #1099, but it will still stay at a low number (probably still single digit). So 64 or 128 is actually way more than we need. Each graph just seems to take a few megabytes here for us. 

## Benchmarks 

Device: NVIDIA A40.  Times are µs per call (wall = host enqueue + GPU, median of reps).  Speedup = generic (graphs **off**) / graphs **on**.

| Grid | Resolution | Fwd off (µs) | Fwd on (µs) | Fwd speedup | Inv off (µs) | Inv on (µs) | Inv speedup |
|------|------------|-------------:|------------:|------------:|-------------:|------------:|------------:|
| FullGaussianGrid | T63 L8 | 2488.1 | 280.0 | 8.9× | 2671.4 | 398.6 | 6.7× |
| FullGaussianGrid | T63 L16 | 2524.6 | 284.2 | 8.9× | 2896.9 | 407.6 | 7.1× |
| FullGaussianGrid | T127 L8 | 5270.3 | 593.8 | 8.9× | 6247.8 | 916.8 | 6.8× |
| FullGaussianGrid | T127 L16 | 5091.7 | 635.7 | 8.0× | 6181.2 | 970.0 | 6.4× |
| OctahedralGaussianGrid | T63 L8 | 2519.9 | 310.1 | 8.1× | 2961.3 | 439.0 | 6.7× |
| OctahedralGaussianGrid | T63 L16 | 2696.0 | 317.6 | 8.5× | 3003.1 | 447.6 | 6.7× |
| OctahedralGaussianGrid | T127 L8 | 5245.9 | 762.7 | 6.9× | 6167.5 | 1119.0 | 5.5× |
| OctahedralGaussianGrid | T127 L16 | 5458.3 | 838.1 | 6.5× | 6431.9 | 1158.6 | 5.6× |
| OctaminimalGaussianGrid | T63 L8 | 2634.0 | 292.9 | 9.0× | 3163.0 | 422.0 | 7.5× |
| OctaminimalGaussianGrid | T63 L16 | 2657.2 | 300.4 | 8.8× | 3171.7 | 431.0 | 7.4× |
| OctaminimalGaussianGrid | T127 L8 | 5268.5 | 736.5 | 7.2× | 6160.5 | 1082.5 | 5.7× |
| OctaminimalGaussianGrid | T127 L16 | 5179.5 | 813.5 | 6.4× | 6189.1 | 1135.1 | 5.5× |
| OctaHEALPixGrid | T63 L8 | 2623.6 | 291.6 | 9.0× | 3137.8 | 416.1 | 7.5× |
| OctaHEALPixGrid | T63 L16 | 2514.7 | 298.8 | 8.4× | 3110.0 | 425.2 | 7.3× |
| OctaHEALPixGrid | T127 L8 | 5443.7 | 739.0 | 7.4× | 6515.9 | 1082.2 | 6.0× |
| OctaHEALPixGrid | T127 L16 | 5168.5 | 814.1 | 6.3× | 6137.4 | 1131.3 | 5.4× |
| HEALPixGrid | T63 L8 | 2649.4 | 249.6 | 10.6× | 3055.2 | 379.2 | 8.1× |
| HEALPixGrid | T63 L16 | 2645.1 | 254.3 | 10.4× | 3116.3 | 384.2 | 8.1× |
| HEALPixGrid | T127 L8 | 5401.2 | 567.0 | 9.5× | 6294.8 | 861.3 | 7.3× |
| HEALPixGrid | T127 L16 | 5402.3 | 610.2 | 8.9× | 6258.8 | 924.3 | 6.8× |

So while there are differences, overall it doesn't matter so much which grid, all get pretty large speedups that are roughly in the same ballpark. 

<details> 
<summary>Benchmark script </summary>

```Julia
using CUDA
using SpeedyWeather
using Printf
using Statistics: median

import SpeedyWeather.SpeedyTransforms: SpectralTransform, _fourier_batched!

const EXT = Base.get_extension(SpeedyWeather.SpeedyTransforms, :SpeedyTransformsCUDAExt)
@assert EXT !== nothing "SpeedyTransformsCUDAExt not loaded — is CUDA functional?"

"""Wall time (host enqueue + GPU) per call in ns, median of `reps` measurements of `n` calls
each (one sync per measurement). This is the realistic per-call cost: graph replay collapses
the many small launches of the generic path into one."""
function bench_wall(f; n, reps, warmup = 10)
    for _ in 1:warmup
        f()
    end
    CUDA.synchronize()
    times = Float64[]
    for _ in 1:reps
        t0 = time_ns()
        for _ in 1:n
            f()
        end
        CUDA.synchronize()
        push!(times, (time_ns() - t0) / n)
    end
    return median(times)
end

const GRIDS = [
    FullGaussianGrid,
    OctahedralGaussianGrid,
    OctaminimalGaussianGrid,
    OctaHEALPixGrid,
    HEALPixGrid,
]
const RESOLUTIONS = [(63, 8), (63, 16), (127, 8), (127, 16)]

# (n, reps) for the graphs-on (fast replay) and graphs-off (slow allocating) paths
const N_ON, REPS_ON = 200, 3
const N_OFF, REPS_OFF = 50, 3

rows = NamedTuple[]

for Grid in GRIDS, (trunc, L) in RESOLUTIONS
    label = @sprintf("%-24s T%-3d L%-2d", string(nameof(Grid)), trunc, L)
    try
        spectral_grid = SpectralGrid(; trunc, nlayers = L, Grid, architecture = SpeedyWeather.GPU())
        field = rand(Float32, spectral_grid.grid, L)
        fn = S.scratch_memory.north
        fs = S.scratch_memory.south
        fwd!() = _fourier_batched!(fn, fs, field, S)
        inv!() = _fourier_batched!(field, fn, fs, S)

        # graphs ON
        EXT.clear_fourier_graph_cache!()
        EXT.FOURIER_GRAPHS_ENABLED[] = true
        fwd_on = bench_wall(fwd!; n = N_ON, reps = REPS_ON)
        inv_on = bench_wall(inv!; n = N_ON, reps = REPS_ON)

        # graphs OFF (generic allocating per-ring path)
        EXT.clear_fourier_graph_cache!()
        EXT.FOURIER_GRAPHS_ENABLED[] = false
        fwd_off = bench_wall(fwd!; n = N_OFF, reps = REPS_OFF)
        inv_off = bench_wall(inv!; n = N_OFF, reps = REPS_OFF)
        EXT.FOURIER_GRAPHS_ENABLED[] = true
        EXT.clear_fourier_graph_cache!()

        push!(rows, (; grid = string(nameof(Grid)), trunc, L, ok = true, fwd_off, fwd_on, inv_off, inv_on))
        @printf("%s | fwd %8.1f -> %7.1f us (%4.1fx) | inv %8.1f -> %7.1f us (%4.1fx)\n",
            label, fwd_off / 1.0e3, fwd_on / 1.0e3, fwd_off / fwd_on,
            inv_off / 1.0e3, inv_on / 1.0e3, inv_off / inv_on)
    catch e
        msg = sprint(showerror, e)
        push!(rows, (; grid = string(nameof(Grid)), trunc, L, ok = false, err = msg))
        println(label, " | ERROR: ", first(split(msg, '\n')))
    end
    flush(stdout)
end

# ---------------------------------------------------------------------------------- markdown
io = IOBuffer()
println(io, "# CUDA Graphs batched Fourier transform — speedup vs generic GPU path\n")
println(io, "Device: ", CUDA.name(CUDA.device()), ".  ",
    "Times are µs per call (wall = host enqueue + GPU, median of reps).  ",
    "Speedup = generic (graphs **off**) / graphs **on**.\n")
println(io, "| Grid | Resolution | Fwd off (µs) | Fwd on (µs) | Fwd speedup | Inv off (µs) | Inv on (µs) | Inv speedup |")
println(io, "|------|------------|-------------:|------------:|------------:|-------------:|------------:|------------:|")
for r in rows
    res = "T$(r.trunc) L$(r.L)"
    if r.ok
        @printf(io, "| %s | %s | %.1f | %.1f | %.1f× | %.1f | %.1f | %.1f× |\n",
            r.grid, res, r.fwd_off / 1.0e3, r.fwd_on / 1.0e3, r.fwd_off / r.fwd_on,
            r.inv_off / 1.0e3, r.inv_on / 1.0e3, r.inv_off / r.inv_on)
    else
        @printf(io, "| %s | %s | — | — | ERROR | — | — | ERROR |\n", r.grid, res)
    end
end
md = String(take!(io))

println("\n", md)
outfile = joinpath(@__DIR__, "results_graphs_comparison.md")
write(outfile, md)
println("wrote $outfile")
```

</details>

## Healpix 

Looking at the Healpix grid again, we should absolutely write a specialised transform at some point. This could be very fast.

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
```diff
@@ -1,14 +1,372 @@
...
+"""$(TYPEDSIGNATURES)
+Cache (one per transform SIZE, i.e. per FFT plan) holding the pre-allocated packed
+work buffers, the per-ring reshaped views and FFT plans used by the transforms, the device
+gather/scatter index metadata, and the instantiated CUDA graphs (one per distinct `field`
+buffer and direction). A graph value of `nothing` marks a buffer for which capture failed
+(fall back to direct loop)."""
+mutable struct GPUFourierGraphCache
+    packed_real                 # CuVector{NF}        — all rings' dense real blocks
+    packed_cplx                 # CuVector{Complex{NF}} — all rings' dense complex blocks
+    real_view::Vector           # per-ring reshaped (nlon_j × nlayers) view into packed_real
```
milankl: Is it not a problem that this struct has type unstable fields?

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
```diff
@@ -1,14 +1,372 @@
...
+    cplx_view::Vector           # per-ring reshaped (nfreq_j × nlayers) view into packed_cplx
+    rfft_plans                  # forward FFT plans for THIS size (nlayers batches), per ring
+    brfft_plans                 # inverse FFT plans for THIS size, per ring
+    roff::CuVector{Int}         # 0-based real-block offset per ring
+    coff::CuVector{Int}         # 0-based complex-block offset per ring
+    nlon::CuVector{Int}         # longitudes per ring
```
milankl: This is called `RingGrids.get_nlons` because per ring I named it plural, in contrast to `get_nlon_per_ring` which would be `nlon` a scalar. Should we rename those here? Always plural per ring, singular for a given ring?

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
```diff
@@ -1,14 +1,372 @@
...
+# the plan set is the canonical per-size resource (one-to-one with a size, stable for the lifetime
+# of `S`), so keying on it yields one independent cache (packed buffers, views, metadata,
+# plans, graphs) per size. 
+const GRAPH_CACHES = IdDict{Any, GPUFourierGraphCache}()
+
+# EXTENSION POINT for batching: return the
+# (forward, inverse) per-ring FFT plan sets for a transform of `nlayers` layers. The default
+# assumes a single size; override this for an `S` that stores several plan sets (e.g. keyed
+# by layer count) and the rest — caches, packing, capture, replay — follows automatically.
+fft_plans(S::SpectralTransform, nlayers::Integer) = (S.rfft_plans, S.brfft_plans)
```
milankl, line 143: Do we plan to do this generalization to have different plans per layer? I hope not!

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
```diff
@@ -1,14 +1,372 @@
...
+    packed_real = CUDA.zeros(NF, sum(block_r))
+    packed_cplx = CUDA.zeros(Complex{NF}, sum(block_c))
+    real_view = [reshape(view(packed_real, roff[j] + 1:roff[j] + block_r[j]), nlons[j], nlayers) for j in 1:nlat_half]
+    cplx_view = [reshape(view(packed_cplx, coff[j] + 1:coff[j] + block_c[j]), nfreqs[j], nlayers) for j in 1:nlat_half]
+
+    dev(x) = CuArray(x)
+    return GPUFourierGraphCache(
+        packed_real, packed_cplx, real_view, cplx_view,
+        rfft_plans, brfft_plans,
+        dev(roff), dev(coff), dev(nlons), dev(nfreqs),
```
milankl: called `nlons` here but `nlon` in the struct itself ... make `nlons` there too?

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
```diff
@@ -1,14 +1,372 @@
...
+        IdDict{Any, Union{Nothing, CuGraphExec}}(),
+    )
+end
+
+"""$(TYPEDSIGNATURES)
+The per-size resource that identifies a graph cache. The forward FFT plan set is unique per
+transform size and stable for the lifetime of `S`, so it remains a valid key even when a
+single `scratch_memory` is shared across sizes (variable-batching). Must return *the stored*
+plan object (not a fresh allocation) so the identity is stable across calls."""
+cache_key(S::SpectralTransform, nlayers::Integer) = fft_plans(S, nlayers)[1]
```
milankl, line 191: this feels hacky ... but okay?

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
```diff
@@ -1,14 +1,372 @@
...
+    # southern rings (the equator ring, if any, is zeroed rather than transformed)
+    launch!(arch, Array3DWorkOrder, (cache.nlon_max, nlh, L), gather_real_kernel!, cache.packed_real, field.data, cache.roff, cache.nlon, cache.rstart_s)
+    @inbounds for j in 1:nlh
+        if cache.has_equator && j == cache.jeq
+            fill!(cplx_view[j], 0)
+        else
+            mul!(cplx_view[j], rfft_plans[j], real_view[j])
+        end
+    end
+    launch!(arch, Array3DWorkOrder, (cache.nfreq_max, nlh, L), scatter_cplx_kernel!, f_south, cache.packed_cplx, cache.coff, cache.nfreq)
```
milankl: this is a follow-up kernel launch that writes the data back from the buffers into the the actual f_south array correct?

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
```diff
@@ -1,14 +1,372 @@
...
+    return nothing
+end
+
+"""$(TYPEDSIGNATURES)
+Allocation-free, fused inverse (spectral → grid) batched Fourier loop: one gather kernel
+packs all rings from the scratch, per-ring in-place inverse cuFFTs run on reshaped views,
+one scatter kernel writes the grid field. Suitable for CUDA-graph capture."""
+function inverse_loop!(cache::GPUFourierGraphCache, field::AbstractField, g_north, g_south, S::SpectralTransform)
+    arch = cache.arch
+    nlh = cache.nlat_half
```
milankl: avoid this renaming and stick to `nlat_half`? If it's about brevity then unpack fields from `cache` for the kernel launch?

milankl (commented):
This looks good to me! Major question I have is whether with the graphs, does it actually make a difference if we split the rings as we currently do 

```
(forward loop)
(1) gather northern rings from `field` to `packed_real`
(2) for j in north pole to equator do the transform
(3) scatter from `packed_complex` to `f_north`
(4) gather southern rings
(5) for j in south pole to equator do the transform
(6) scatter from packed to `f_south`
```

which I believe is just legacy from CPU days. Could we not do

```
(1) + (4) gather all rings
(2) + (5) loop north pole to south pole
(3) + (6) scatter all packed data
```

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
```diff
@@ -1,14 +1,377 @@
...
+Replays a cached CUDA graph of the fused gather + per-ring cuFFTs + scatter; see
+[`run_graph!`](@ref)."""
+function _fourier_batched!(
+        f_north::CuArray{<:Complex, 3},
+        f_south::CuArray{<:Complex, 3},
+        field::AbstractField,
+        S::SpectralTransform,
+    )
+    @assert eltype(field) == eltype(S) "Number format of grid $(eltype(field)) and SpectralTransform $(eltype(S)) need to match."
+    if !FOURIER_GRAPHS_ENABLED[]
```
milankl: This could be `if S.fourier_graphs_enabled` to not make it a global constant?

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
maximilian-gelbrecht: No, apparently not. As it's also always unpacked and never by itself in a hot loop, this seems to be fine. With it being mutable, it's also actually isn't that easy to make it concrete.

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
maximilian-gelbrecht: Yeah, sure. Makes testing need two different plans, but I guess that's just a very minor annoyance.

maximilian-gelbrecht:
> This looks good to me! Major question I have is whether with the graphs, does it actually make a difference if we split the rings as we currently do
> 
> ```
> (forward loop)
> (1) gather northern rings from `field` to `packed_real`
> (2) for j in north pole to equator do the transform
> (3) scatter from `packed_complex` to `f_north`
> (4) gather southern rings
> (5) for j in south pole to equator do the transform
> (6) scatter from packed to `f_south`
> ```
> 
> which I believe is just legacy from CPU days. Could we not do
> 
> ```
> (1) + (4) gather all rings
> (2) + (5) loop north pole to south pole
> (3) + (6) scatter all packed data
> ```

Yeah very good point. This was just a direct translation of our existing transform so far. 

Would the equator ring still need special treatment in this case for grids that have one?

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
maximilian-gelbrecht: The main reason the graph cache is here and not in `S` is to keep `S` immutable, which would be kind of hard otherwise. 

But I do agree this is quite hacky. I am not super happy with it either, but it works 🤷

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
maximilian-gelbrecht: Maybe that's not formulated well. But due to the limitations of FFT plans we do need several different plans when batching in #1099, similar to how we already have 2 plans at the moment 1D and 2D. 

With #1099 it's actually 4 or 5 plans that we need.

maximilian-gelbrecht:
After some more thoughts on it. I think it would prefer to keep this PR as a direct adaption of the existing fourier transform scheme. 

The problem is that with doing all latitudes in one go we would also need to change and adjust the Legendre transform as well. 

I would prefer this PR to just do the basic direct implementation of CUDA Graphs, and then in a follow up we can actually revise the scheme itself for both Fourier and Legendre to be more GPU friendly, because both would need to be revised. It's definitely worth looking into this though!

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
maximilian-gelbrecht: done

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
maximilian-gelbrecht: done

milankl:
Yes happy to leave it for now. Maybe add a TODO comment in the code or we just keep it in mind that there's a different ordering possible or at least should be looked at!

Yeah no if that requires changes to Legendre then let's not do that now. And I am not even sure it'll be beneficial. It's already fast so amazing happy to get it merged as is!

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
maximilian-gelbrecht: Done

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
maximilian-gelbrecht: You mean just renaming `nlon` to `nlons` here? I've just done that.

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
maximilian-gelbrecht: yes correct

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
maximilian-gelbrecht: done

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
milankl: mhh interesting, hope that doesn't cause a problem with Reactant either ... ?

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
milankl: Great, yeah just to have consistent naming!

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
maximilian-gelbrecht: Reactant will never see this. While I didn't ask, I am also just assuming that CUDA Graphs won't work with Reactant either way. The code gets compiled into StableHLO which is vendor agnostic, so I am sure CUDA Graphs are not a part of that.

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
maximilian-gelbrecht: Btw: I previously thought that cuFFT in C actually would support a more flexible transform with variable batch dimensions, but it doesn't. So this limitation isn't just a Julia thing.

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
maximilian-gelbrecht: In principle the reason why the IDDict is chosen here (Claude was that automatically) makes a lot of sense to me in this situation because everything about CUDA graphs is bound to exact memory addresses. It's not only supposed to be an identical object, it has to be literally the same (just maybe with different values in the arrays). So having an IDDict that's keyed by the FFT plans which define the transform, seems like a decent choice to me.

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
maximilian-gelbrecht: Ok, nonetheless I actually made it parametric now, to also satisfy my own code-OCD as well ;) 

But it doesn't change anything about the performance. I think another reason for it is because the whole code surrounding it actually only gets seen during the graph recording, when the graph is replayed it doesn't matter anymore what type this struct has, it just operates on the baked in pointer addresses anyway. At least this is my understanding of it.

maximilian-gelbrecht:
I think I addressed everything now. 

We take two interesting To-Dos from this 

* Look into revising the transform to do it without the north-south split on GPU 
* Look into a batched fourier for HEALPix along its central equal-length latitudes

SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl
milankl: okay great yeah that's a good justification!

milankl (approved):
Looks great to me! I just made some renamings because claude somehow decided that the in-ring index i is here now r which I thought was confusing. Summary of renamings from the last commit

Old name | New name
-- | --
packed_cplx | packed_complex
cplx_view | complex_view
roff | real_offset
coff | complex_offset
rstart_n | istart_n
rstart_s | istart_s
jeq | j_equator
block_r / block_c | block_real / block_complex
scatter_cplx_kernel! | scatter_complex_kernel!
gather_cplx_kernel! | gather_complex_kernel!
kernel locals r, nl | i, nlon
loop variable L | nlayers

<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(204, 204, 204); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(33, 37, 43); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">Every call site, struct field, local variable, and kernel parameter that uses these names was updated consistently. The logic — index arithmetic, loop structure, conditionals — is identical in all cases.</p>

maximilian-gelbrecht:
Thanks! In the PR that I just merged I also added to the Claude.md a bullet point about naming variables more explicit and not to use too many abbreviations. 

I'll merge this one here first thing on Monday.

PR #985, opened 2026-02-09
Reactant: primitive equation models 

maximilian-gelbrecht:
I don't want one monster PR, so I split the Reactant work a little. 

#970 is about all the basics and getting the barotropic model going 

Here, I continue with the primitive wet model

PR #1113, opened 2026-06-03
Fix SolarZenithSeason kernel: uninitialized typed variables and acos domain error

nviebig:
## Problem

`daily_cycle=false` uses `SolarZenithSeason`, which fails inside the `@kernel` implementation.

**1. `UndefVarError: h₀ not defined`**

The kernel contains

```
h₀::NF
cos_zenith_j::NF
```

which are type assertions and result in an `UndefVarError` before assignment.

**2. `DomainError` in `acos`**

After removing the type assertions, the next failure is

```
acos(-tan(ϕ) * tan(δ))
```

inside an `ifelse`. Since `ifelse` evaluates both branches, `acos` is called even in the polar day/night case, where the argument can lie outside `[-1,1]`.

## Fix

Remove the uninitialized type assertions and clamp the `acos` argument:

```
ϕ = lat[j]

h₀ = NF(ifelse(
    2 * (abs(δ) + abs(ϕ)) < π,
    acos(clamp(-tan(ϕ) * tan(δ), -one(NF), one(NF))),
    ifelse(ϕ * δ > 0, π, 0)
))

sinϕ, cosϕ = sinlat[j], coslat[j]

cos_zenith_j = NF(
    h₀ * sinδ * sinϕ +
    cosδ * cosϕ * sin(h₀)
) / NF(π)
```

## Reproducer

```
using SpeedyWeather

sg = SpectralGrid()
model = PrimitiveWetModel(
    sg,
    planet = Earth(sg; daily_cycle=false, seasonal_cycle=false)
)

sim = initialize!(model)
SpeedyWeather.timestep!(sim)
```

Before the fix:

```
ERROR: UndefVarError: `h₀` not defined
```

After removing the type assertions:

```
ERROR: DomainError with 8.502875:
acos(x) not defined for |x| > 1
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fixed `SolarZenithSeason` kernel: uninitialized typed variables and `acos` domain error [#1113](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1113)
 - Fixed punctuation in spectral transform docs [#1112](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1112)
 
 ## v0.20.3
--- a/SpeedyWeather/src/parameterizations/radiation/zenith.jl
+++ b/SpeedyWeather/src/parameterizations/radiation/zenith.jl
@@ -299,19 +299,16 @@ end
     j = whichring[ij]
 
     NF = eltype(cos_zenith)     # force type stability
-    h₀::NF                      # hour angle sunrise to sunset
-    cos_zenith_j::NF            # at latitude j
 
     ϕ = lat[j]
-    h₀ = ifelse(
+    h₀ = NF(ifelse(
         2 * (abs(δ) + abs(ϕ)) < π,    # polar day/night?
-        acos(-tan(ϕ) * tan(δ)),             # if not: calculate length of day
+        acos(clamp(-tan(ϕ) * tan(δ), -one(NF), one(NF))),  # length of day
         ifelse(ϕ * δ > 0, π, 0)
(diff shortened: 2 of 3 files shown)


milankl:
Awesome, yeah happy for some type-stability related changes here to be made. There's some overlap with #985 though maybe check that this goes in the same direction!

maximilian-gelbrecht:
Don't worry about #985, this here is a fairly minor thing, I will handle the merge into #985. Even though we can ideally merge that one in the next weeks as well finally.

maximilian-gelbrecht (approved):

maximilian-gelbrecht:
I guess this slipped through because it's not a problem on CPU without the kernels, and on GPU we only test the default config? 

Or is this not tested at all? 

Maybe this should still be looked into, so that we won't have a problem here again. Otherwise this can be merged. 

Edit: Ah I see, it's initialised but not run, probably to save some CI time. But those can actually be run without full model init very easily just 

```
model = PrimitiveWetModel(....) 
field = zeros(spectral_grid.grid) 
SpeedyWeather.cos_zenith!(field, model.solar_zenith, DateTime(2000,12,1), model.geometry
```

something like this should be added to the tests. Some basic check based on the date time is enough.

nviebig:
Added tests in the new "cos_zenith!" testset, calls cos_zenith! directly for both SolarZenith and SolarZenithSeason at June and December solstice, checks values are in [0, 1], and verifies the two dates produce different spatial patterns. Is this along the lines of what you had in mind?

Edit: No full model init

maximilian-gelbrecht:
Okay great, yeah that's totally fine! If all tests pass feel free to merge this. Very nice!

PR #1114, opened 2026-06-05
Surface roughness as a component

maximilian-gelbrecht:
Based on #959


SpeedyWeather/src/output/variables/boundary_layer.jl
```diff
@@ -149,10 +149,61 @@ end
...
+
 # collect all in one for convenience
 BoundaryLayerOutput() = (
     ZonalVelocity10mOutput(),
     MeridionalVelocity10mOutput(),
     SurfaceTemperatureOutput(),
     BoundaryLayerDragOutput(),
+    SurfaceRoughnessOutput(),
+    LandSurfaceRoughnessOutput(),
+    OceanSurfaceRoughnessOutput(),
```
milankl: Can we remove these here? I'm happy to have them defined as output but adding them here means that we'll write 3 arrays that essentially contain the information of one when using `SpeedyWeather.AllOutputVariables()`. With all the surface fluxes I'd left out the ocean/land only versions from output and only the land-sea-mask-averaged version is written

milankl (commented):
Generally great, but I suggest to nest them so

```julia
model.boundary_layer isa BoundaryLayer
model.boundary_layer.drag
model.boundary_layer.surface_roughness
```

like we already do for radiation and land and we technically also could do for surface fluxes. This also means we'd formulate

```julia
function paramterization!(ij, vars, ::BoundaryLayer, ...)
    surface_roughness!(...)
    boundary_layer_drag!(...)
end
```

and therefore have more options to also pass data directly between these two schemes rather than just write to `simulation.variables` and read from there again. This would mean the user interface would be different too as the boundary layer scheme would need to be assembled consisting of the surface roughness computation and then the associated drag.

```julia
boundary_layer = BoundaryLayer(drag=BulkRichardsonDrag(...), surface_roughness=ConstantSurfaceRoughness(...))
model = PrimitiveWetModel( ...; boundary_layer)
```

EDIT: Did this in [e7c9417](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1114/commits/e7c9417e4160f5ba86338a15792b717628d56647)

SpeedyWeather/src/output/variables/boundary_layer.jl
milankl: done in [e7c9417](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1114/commits/e7c9417e4160f5ba86338a15792b717628d56647)

milankl (approved):

maximilian-gelbrecht:
Nice! Do you want to change anything more? Or can this be merged? 

I would also directly issue a release. I think it's good to have one right before we merge the other breaking changes related to the dycore as well.

milankl:
Sounds good to me!!!

PR #1116, opened 2026-06-08
Restrict Enzyme compat

maximilian-gelbrecht:
Since the release of Enzyme 0.13.154 this night, there seems to be an issue, the CI is just timing out. Let's first confirm this here by restricting and watching the tests.


maximilian-gelbrecht:
I am investigating the actual cause of this separately, but while the culprit isn't found and fixed, I'll merge this and hopefully loosen the compat restrictions later again.

PR #1127, opened 2026-06-16
Allocation-free masked copy between RingGrids Field and subet array

milankl:
# Issue

At the moment these 

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/5b9dafbde3ace449488900067fc47549eb335a9a/SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl#L224-L232

are allocating.

# Solution

# `unmasked_indices` and `copy_unmasked!`

These two functions together let you extract the unmasked subset of a ring-grid field into a plain array, work on it, and write results back.

**Convention:** `mask` is a `Bool` field where `true` = masked (excluded), `false` = unmasked (included).

---

## 1. Build the index vector once

```julia
using RingGrids
mask = rand(Bool, grid)                   # or build your own Bool field
indices = unmasked_indices(mask)          # Vector of grid-point indices where mask == false
n = length(indices)                       # number of unmasked points
```

`indices` lives on the same device as `mask` (CPU or GPU) and is sorted.

---

## 2. Copy field → plain array (gather)

```julia
# 2D (single layer)
field = rand(Float32, grid)
array = zeros(Float32, n)
copy_unmasked!(array, field, indices)     # array[i] == field[indices[i]]

# 3D (multiple layers)
nlayers = 3
field = rand(Float32, grid, nlayers)
array = zeros(Float32, n, nlayers)
copy_unmasked!(array, field, indices)     # array[i, k] == field[indices[i], k]
```

---

## 3. Copy plain array → field (scatter)

```julia
copy_unmasked!(field, array, indices)     # field[indices[i], k] = array[i, k]
```

Grid points not referenced by `indices` (the masked ones) are **left unchanged**.


maximilian-gelbrecht:
Oh, yeah, that's a good idea to just do it like this

maximilian-gelbrecht (approved):
Fine with me to merge once all tests pass!

maximilian-gelbrecht:
It's not applied to the Terrarium extension yet though, but this could also be a separate PR.

milankl:
@bgroenks96 this should allow for allocation-free SpeedyWeather-Terrarium coupling. Let me know what you think about the interface. 

1. You'd need to precompute and store the `indices = unmasked_indices(mask)` but it's just one 2D integer array
2. In the case of land, mask = false is unmasked so land, mask = true is masked so ocean. This is in contrast to our `land-sea-mask` where 1 is land, 0 is ocean, see https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/land_sea_mask#The-land-sea-mask
3. I hope `copy_unmasked!` is good naming as we're copying the data that's not masked

If all passes here do you want to try and update the extension accordingly in another PR?

milankl:
Maybe we should just rename "land-sea mask" to land fraction because that's so much clearer, no?

milankl:
Quick check that this also works on the GPU in 2D/ND (me doughnut forgot that you can't dispatch `::AbstractField` inside a kernel because it's adapted to `.data`, fixed now) and causes no allocations other than the kernel launch

```julia
julia> @btime copy_unmasked!($field, $array, $indices)
  15.042 μs (92 allocations: 3.02 KiB)
49152-element, 255-ring HEALPixField{Float32, 1} as MtlArray on GPU(MetalBackend())
 ...

julia> @btime copy_unmasked!($array, $field, $indices)
  17.458 μs (92 allocations: 3.02 KiB)
24636-element MtlVector{Float32, Metal.PrivateStorage}:
...
```

(The 3KB are independent or resolution)

RingGrids/src/copy_unmasked.jl
```diff
@@ -0,0 +1,63 @@
...
+function copy_unmasked!(dst::AbstractArray, src::AbstractField, indices)
+    arch = architecture(dst)
+    @boundscheck size(indices, 1) == size(dst, 1) || throw(BoundsError(indices, dst))
+    @boundscheck size(dst, 1) <= size(src, 1) || throw(BoundsError(indices, dst))
+
+    # technically without this check the kernel is unsafe but `maximum` requires reading in the array ...
+    # @boundscheck 1 <= maximum(indices) <= size(src, 1) || throw(BoundsError(indices, src))
+
+    # call the gather kernel as dst is expected to be smaller as elements in src are masked
+    launch!(arch, ArrayWorkOrder, size(dst), copy_unmasked_gather_kernel!, dst, src, indices)
```
bgroenks96, line 31: Yeah I like this a lot, very elegant! Speedy's kernel launching system really makes this much easier, I think.

bgroenks96 (approved):
Looks great! I have no objections!

And yes, I would take care of the Terrarium extension in a separate PR.

milankl:
Great then this is going in as is. Didn't bother writing a

```julia
array = copy_unmasked(field, mask)
```

allocating version of it because that's equivalent to `field[.~mask, ...]`

PR #1130, opened 2026-06-18
Fix wrong dynamics setting in ic-example

SieglStefan:
Hi,

there was a  relict in the Rossby-Haurwitz wave example in the [initial condition](https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/initial_conditions#Rossby-Haurwitz-wave-in-primitive-equations) section leading to a wrong plot.

The text mentioned (the old) setting `physics=false` but then set `dynamics_only=false` as the model argument.


milankl (approved):
Awesome, thanks Stefan!!!

milankl:
(update changelog? happy to bypass that too)

milankl (approved):

PR #1131, opened 2026-06-18
Vertical advection with step pass thorugh

maximilian-gelbrecht:
This is a possible fix to the performance regression of the vertical advection

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fix vertical advection performance regression by indexing the contiguous time step directly in the stencil kernel instead of a `get_*_step` view [#1131](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1131)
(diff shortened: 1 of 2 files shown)


milankl (approved):
I'm wondering whether we should establish a letter for the step dimension, at the moment we have

- `ij, k` on the grid
- `lm, k` in spectral space

the next one would be `n` but that's typically an integer used for length/size not an index. `o, p, q, r` could also be used but maybe we should just use `s` given it's the first letter in `step`, so suggestion

- `ij, k, s` on the grid
- `lm, k, s` in spectral

what do you think?

maximilian-gelbrecht:
> I'm wondering whether we should establish a letter for the step dimension, at the moment we have
> 
> * `ij, k` on the grid
> * `lm, k` in spectral space
> 
> the next one would be `n` but that's typically an integer used for length/size not an index. `o, p, q, r` could also be used but maybe we should just use `s` given it's the first letter in `step`, so suggestion
> 
> * `ij, k, s` on the grid
> * `lm, k, s` in spectral
> 
> what do you think?

Yeah agreed, I adapted it to `s` now here in the vertical advection.

PR #1132, opened 2026-06-18
`PrimitiveWetModel` `NcycleLorenz` GPU test

maximilian-gelbrecht:
Adds a basic test to the GPU CI just to make sure everything compiles and there's no syntax error, stability isn't tested, it's just 3 steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


milankl:
Note `NCycleLorenz` and `PrimitiveEquation` models are currently not compatible as I haven't translated the implicit solver yet in this case. It shouldn't be much work but just haven't done that yet

maximilian-gelbrecht:
Oh okay, I thought it's just not stable. 

My intention was primarily to also have a gpu test for the fused variables that uses two tendency steps. The majority of those changes only affect the primitive eq model. But then I wait or think about something different to test those

PR #1134, opened 2026-06-19
Rename land-sea mask array `land_fraction`

milankl:
For clarity

### DIFF
--- a/docs/src/land.md
+++ b/docs/src/land.md
@@ -441,7 +441,7 @@ column_grid = Terrarium.ColumnRingGrid(
     Terrarium.CPU(), Float32,
     Terrarium.ExponentialSpacing(; N = Nz, Δz_min),
     ring_grid,
-    land_sea_mask.mask .> 0
+    land_sea_mask.land_fraction .> 0
 )
 
 # Soil column + initial state, matching `LandModel: Soil, no vegetation`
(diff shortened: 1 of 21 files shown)


bgroenks96:
As long as the intention of the land-sea mask in Speedy is to *always* be fractional, then I agree with this change.

docs/src/land.md
```diff
@@ -441,7 +441,7 @@ column_grid = Terrarium.ColumnRingGrid(
     Terrarium.CPU(), Float32,
     Terrarium.ExponentialSpacing(; N = Nz, Δz_min),
     ring_grid,
-    land_sea_mask.mask .> 0
+    land_sea_mask.land_fraction .> 0
```
bgroenks96, line 444: I wholeheartedly endorse the choice to call this `land_fraction` instead of `ocean_fraction` which reflects the inherent superiority of land over the ocean 💯 😆

bgroenks96 (approved):

milankl:
> As long as the intention of the land-sea mask in Speedy is to _always_ be fractional, then I agree with this change.

Fractional is backwards compatible 😉 You can always go back to binary ...

docs/src/land.md
milankl: Blame whoever established "land-sea mask" over "sea-land mask". A land (-sea) mask, masks land so mask = 1 means land, that's a land fraction generalising from binary ...

PR #932, opened 2025-11-28
Switch off dynamics for BarotropicModel and backwards particles

milankl:
To be used like

Version 1: Have some existing u,v imposed
```julia
spectral_grid = SpectralGrid(nlayers=1, nparticles=100)
particle_advection = ParticleAdvection2D(spectral_grid, backwards=true)
model = BarotropicModel(spectral_grid, dynamics=false; particle_advection)
simulation = initialize!(model)

# some custom data
u = randn(spectral_grid.grid)
v = randn(spectral_grid.grid)

simulation.diagnostic_variables.grid.u_grid .= u
simulation.diagnostic_variables.grid.v_grid .= v

# run particle advection without changing velocity field, tracing it backwards in time
run!(simulation)
```

Version 2: Freeze initial velocity field
```julia
spectral_grid = SpectralGrid(nlayers=1, nparticles=100)
particle_advection = ParticleAdvection2D(spectral_grid, backwards=true)
model = BarotropicModel(spectral_grid, dynamics=false; particle_advection)
simulation = initialize!(model)

# do initial transform manually
progn, diagn, model = SpeedyWeather.unpack(simulation)
SpeedyWeather.scale!(progn, diagn, model.planet.radius)
lf = 1
transform!(diagn, progn, lf, model)

# run particle advection with initial velocity field frozen in time, tracing it backwards in time
run!(simulation)
```

PR #1136, opened 2026-06-19
Backwards particle advection

milankl:
supersedes #932


milankl:
@Jullcifer also note this related PR here that just attempted to implement backwards particle advection (useful for finite time lyapunov exponents, FTLE) as well as setting a constant velocity field and then only doing particle advection, this might be very useful for testing too

#1209

PR #1139, opened 2026-06-25
Scale time step on the fly and forcing/drag/parameterizations inside dycore

milankl:
Why?

- All atmospheric variables use a scaled time step, ocean and land do not, so the time stepper needs to run some `update_prognostic!` calls with scale some without.
- The scaled time step in `model.time_stepping` is user-facing and can cause confusion, we therefore want to hide the scaled time step ideally

This now
- treats every atmospheric + tracer variable with a scaled time step, ocean, land and sea ice without
- removes `model.time_stepping.Δt_sec` because `.Δt` is in seconds now and not s/m because /radius

# Compatibility summary

- `model.time_stepping.Δt_sec` has been removed, use `.Δt` instead, it's the unscaled time step now in seconds.
- In most cases, custom forcings should not include a scaling with the radius, see docs "Forcing scaling"

@pierluigividale You suggested this PR a while ago, here you go. Possible places where your code might be breaking ☝🏼 

# Summary

Now all parameterizations + forcing + drag are being scaled automatically within the dynamical core, so for most forcing/drag/parameterizations a user/developer doesn't need to worry about it. The one remaining place where scaling has to be done manually is when forcing vorticity/divergence directly (not u, v, or any other variable) and when that forcing is not proportional to these variables respectively, i.e. compare

```math
\begin{aligned}
\frac{\partial \zeta}{\partial t} &= F \qquad &(1)\\
\frac{\partial \zeta}{\partial t} &= a\zeta \qquad &(2)
\end{aligned}
```

now scale both with $R^2$ (all other equations are scaled with $R$ meaning for temperature, humidity, pressure, the following does not apply ...), one radius scaling is going into the time step $t \to t/R = t^\star$, the other one into the variable itself $\zeta \to \zeta R = \zeta^\star$, then we have

```math
\begin{aligned}
\frac{\partial \zeta^\star}{\partial t^\star} &= R^2F = RF^\star \qquad &(1)\\
\frac{\partial \zeta^\star}{\partial t^\star} &= (a\zeta^\star)^\star \qquad &(2)
\end{aligned}
```

So the in the (1) case, one $R$ is the now automatically applied scaling of the tendency but the 2nd one is necessary to do by hand. In the (2) one R-scaling goes into the variable $\zeta^\star$ one is automatically applied now, so a user/developer can just write their function without worrying about scaling, which is the philosophy I would like to have, only exception is

- if you force vorticity or divergence with a constant/non-linear term $\propto \zeta^n$ where $n \neq 1$ then manual scaling with $R^{1-n}$ has to be applied
- this only applies for forcing in spectral space (you can't force those in grid space, see docs @ "Allowed forcings") with a constant ($n=0$, so scale with $R$ manually), or a quadratic ($n=2$, so scale with $1/R$), and higher order terms similarly ...

### DIFF
--- a/SpeedyWeather/src/dynamics/forcing.jl
+++ b/SpeedyWeather/src/dynamics/forcing.jl
@@ -53,14 +53,12 @@ function initialize!(
     )
 
     (; latitude, width, speed, time_scale, amplitude) = forcing
-    (; radius) = model.planet
 
     # Some constants similar to Galewsky 2004
     θ₀ = (latitude - width) / 360 * 2π          # southern boundary of jet [radians]
     θ₁ = (latitude + width) / 360 * 2π          # northern boundary of jet
     eₙ = exp(-4 / (θ₁ - θ₀)^2)                  # normalisation, so that speed is at max
     A₀ = speed / eₙ / time_scale.value          # amplitude [m/s²] without lat dependency
-    A₀ *= radius                                # scale by radius as are the momentum equations
 
     (; colat) = model.geometry
     # latitude in radians, abs for north/south symmetry
@@ -182,8 +180,8 @@ function forcing!(
     S_masked = vars.scratch.a_2D
     transform!(S_masked, S_grid, vars.scratch.transform_memory, spectral_transform)
 
-    # scale by radius^2 as is the vorticity equation, and scale to forcing strength
-    S_masked .*= (vars.prognostic.scale[]^2 * forcing.strength)
+    # scale by radius as is vorticity, and scale to forcing strength
+    S_masked .*= (vars.prognostic.scale[] * forcing.strength)
 
     # force every layer
     vor_tend = get_tendency_step(vars.tendencies.vorticity, time_stepping, forcing)
@@ -231,8 +229,7 @@ function forcing!(
         model::AbstractModel,
     )
     (; latds) = model.geometry
-    # scale by radius as is the vorticity equation
-    s = forcing.strength * vars.prognostic.scale[]
+    s = forcing.strength
     k = forcing.wavenumber
 
     Fu = get_tendency_step(vars.tendencies.grid.u, model.time_stepping, forcing)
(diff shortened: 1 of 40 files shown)


SpeedyWeather/src/dynamics/forcing.jl
```diff
@@ -329,7 +325,6 @@ function initialize!(
 
     # Held and Suarez equation 4
     temp_relax_freq .= kₐ .+ (kₛ - kₐ) * max.(0, (σ .- σb) ./ (1 - σb)) .* (coslat') .^ 4
-    temp_relax_freq .*= radius  # scale by radius as is the temperature equation
```
maximilian-gelbrecht, line 332: I am confused why this was needed before and not anymore.

maximilian-gelbrecht (commented):
I like this! I think this does make things clearer with the scaling! 

While checking this, I also briefly looked at the documentation and noticed that the radius scaling is just explained within the Barotropic model section. So it's not explained for the primitive model. Maybe it's also fine to just add it to the short paragraph that is already in the barotropic section: 

> Many concepts of the [Shallow water model](https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/shallowwater#shallow_water_model) and the [Primitive equation model](https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/primitiveequation#primitive_equation_model) are similar, such that for example [horizontal diffusion](https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/barotropic#diffusion) and the [Time integration](https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/time_integration#leapfrog) are only explained here.

SpeedyWeather/src/dynamics/forcing.jl
milankl: The radius scaling here was just baked here into the Held-Suarez temperature forcing as we didn't scale this forcing before accumulating other terms in the dycore ... see above

maximilian-gelbrecht (approved):

pierluigividale:
Christmas came early this year!!

PR #1141, opened 2026-06-26
Fix NetCDFOutput for non-standard AbstractFloat types (convert σ levels to Float32)

jamesquinlan:
σ variable changed to Float32.(σ) to work with non-standard AbstractFloat types (e.g., BFloat16 or Post16)

σ variable is stored in the native NF, so when NF is e.g., BFloat16/Posit16, NCDatasets rejects it. Converting to Float32 before writing is a fix since NetCDF sigma coordinates don't need sub-float32 precision.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- NetCDF output coordinates always in Float64 as determined by RingGrids.get_lond [#1141](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1141)
 - ArrayWorkOrder instead of Array3DWorkOrder as both preserve dimensions [#1127](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1127)
 - Allocation-free masked copy between RingGrids Field and subset array via mask [#1127](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1127)
 - Rename land-sea mask array land_fraction [#1219](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1134)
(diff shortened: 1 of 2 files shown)


milankl:
Thanks yeah that's good and we should obviously fix that! Wondering what the best general approach is though

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/3a4b706f3e84385d9f0d8b6688ac69b5eccf9ccc/SpeedyWeather/src/output/writers/netcdf_output.jl#L179-L182

So `lond`, `latd` are computed on the fly from the output fields. The fields are created with `eltype` as determined by the `output_NF` kwarg

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/3a4b706f3e84385d9f0d8b6688ac69b5eccf9ccc/SpeedyWeather/src/output/writers/netcdf_output.jl#L84

but `get_lond` and `get_latd` by default always return coordinates in Float64

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/3a4b706f3e84385d9f0d8b6688ac69b5eccf9ccc/RingGrids/src/grids/full_gaussian.jl#L47-L56

So I'm wondering whether we should actually write the sigma coordinate with the same number format as determined by the other coordinate vectors

```julia
lond = get_lond(output.field2D)
latd = get_latd(output.field2D)
σ = convert.(eltype(lond), on_architecture(CPU(), model.geometry.σ_levels_full))
```

So in case the model lives on the GPU we first convert to the CPU and then change the number format to whatever other coordinates are too (Float64 by default). This order is important because Metal e.g. does not support Float64 so we can't have a `MtlVector{Float64}`.

What do you think?

milankl (approved):

jamesquinlan:
Yes.  That makes sense to me.  Thanks for your support.

PR #1137, opened 2026-06-21
Sigma-pressure coordinates part 1

milankl:
Just started drafting some functionality for this #329, TODO

- [x] Define `SigmaPressureCoordinates` with A/B hybrid coefficients and a user-supplied `transition` function
- [x] Define `CubicSigmaPressureCoordinates` with cubic smoothstep transition (`pressure_only_above`, `σ_only_below`)
- [x] Add `sigma(k, coord)` alongside `pressure(k, pₛ, coord)` and `pressure_thickness(k, pₛ, coord)` for a coordinate-agnostic interface
- [x] Construct `Geometry` with either coordinate type; warn in `initialize!` when `SigmaPressureCoordinates.reference_pressure` differs from `atmosphere.reference_pressure`
- [x] Fix bug in `pressure_thickness` for `SigmaPressureCoordinates` (used `+` instead of `-` for layer thickness)
- [x] Fix reference pressure warning in `initialize!(geometry, model)`: `haskey` → `hasproperty`, `model.planet` → `model.atmosphere`
- [x] Adapt `flux_to_tendency` / `surface_flux_to_tendency` to use `pressure_thickness(k, pₛ, coord)`
- [x] Adapt SPPT taper initialization to use `sigma(k, coord)`
- [x] Styled `show` with visual A/B bar for both coordinate types
- [x] Comprehensive unit tests for all of the above
- [x] Documentation page under "Advanced" with live Documenter examples
- [ ] Adapt dynamical core to hybrid coordinates: `vertical_integration!`, `vertical_velocity!` — the Δσ-weighted integrals are baked into the sigma-coordinate form of the continuity equation (Simmons & Burridge 1981)
- [x] Adapt geopotential: `log(σ_full / σ_half)` ratios assume `p ∝ σ` and need replacing with `log(p_full / p_half)` evaluated at reference surface pressure
- [ ] Adapt adiabatic conversion: precomputed `σ_lnp_A`, `σ_lnp_B` coefficients are sigma-specific and become surface-pressure-dependent for hybrid coordinates
- [ ] Adapt the implicit solver (`ImplicitPrimitiveEquation`): matrix operators **R**, **L**, **W** depend on `Δσ` level spacing
- [ ] Adapt remaining parameterizations with TODO comments: `LinearDrag`, surface condition (`σ^(-κ)` extrapolation), convection, large-scale condensation, vertical diffusion, shortwave/longwave transmissivity

### Constructors

[...]

PR #1142, opened 2026-06-29
Fix scalar indexing in sigma coordinate construction

maximilian-gelbrecht:
fixes an error introduced by #1137 on GPU

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


maximilian-gelbrecht:
@milankl I just fixed urgent stuff here that broke `main`. For full GPU compatibility of the hybrid coordinates, there might be further fixes needed, but I'll leave this to your follow up PR.

milankl (commented):
Apologies I missed this. Interesting this was needed because previously the inner constructor used scalar indexing too and the 
```julia
SigmaCoordinates(SG::SpectralGrid, σ_half::AbstractVector)
```
should take `σ_half` as CPU vector given it's the user facing interface to construct a `SigmaCoordinates` ...

PR #1143, opened 2026-06-29
Partial Enzyme with Julia 1.12 compatibility

maximilian-gelbrecht:
One issue was fixed (https://github.com/EnzymeAD/Enzyme.jl/issues/3038), let's see if this is already sufficient to also run Enzyme with Julia 1.12


milankl:
Given that compile times are much longer with 1.12 compared to 1.10 (see e.g. https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1137#issuecomment-4808297426) do you expect Enzyme compilation also to take longer?

maximilian-gelbrecht:
No idea to be honest, but it seems we run into other issues here still, anyway.

SpeedyWeather/src/time_stepping/steppers/leapfrog.jl
```diff
@@ -215,7 +215,8 @@ function update_prognostic!(
     Δt = time_step(time_stepping, clock)
     Δt /= oftype(Δt, scale)                         # scale time step on the fly *1/radius for atmospheric variables
     lf = prognostic_step(time_stepping, clock)      # leapfrog prognostic step index
-    var_old, var_new = get_steps(var)
```
maximilian-gelbrecht, line 221: @milankl There's a problem with the `get_steps` that creates runtime length NTuple with Enzyme with Julia 1.12 (see https://github.com/EnzymeAD/Enzyme.jl/issues/3275). It's very easy to workaround this issue though, so that's what I do here

PR #1144, opened 2026-06-29
Remove Enzyme compat restriction in SpeedyTransforms

maximilian-gelbrecht:
I think I just missed this before


maximilian-gelbrecht:
Superseeded by #1151

PR #1147, opened 2026-06-30
towards HIP graphs 

gregordecristoforo:
This draft pull request introduces HIP graphs acceleration for batched Fourier transforms on AMDGPU, mirroring the existing CUDA graphs acceleration. There is definitely a lot of code repetition from the CUDA graphs implementation that can be refactored more elegantly but I wanted to ask first whether this feature is of interest before spending too much time on cleaning up the code. 

I benchmarked this feature on LUMI's MI250x GPUs with ROCm 6.3 with the following code:

```Julia
using SpeedyWeather, AMDGPU, BenchmarkTools

spectral_grid = SpectralGrid(trunc=42, nlayers=8, architecture=GPU())

# create some data

# transforms in-place
transform!(field, L, S)   # spectral to grid space
@benchmark transform!(L, field, S)   # grid to spectral space`
```

Without the HIP graph implementation I get the following results:

```Julia
julia> @benchmark transform!(L, field, S)   # grid to spectral space
BenchmarkTools.Trial: 1061 samples with 1 evaluation per sample.
 Range (min … max):  3.284 ms … 894.017 ms  ┊ GC (min … max): 0.00% … 7.41%
 Time  (median):     3.764 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   4.698 ms ±  27.329 ms  ┊ GC (mean ± σ):  1.33% ± 0.23%

                   ▁█▆▅▄▃▄                                     
  ▂▁▁▁▂▁▁▂▁▁▁▂▂▂▃▅▇████████▇▆▆▇▅▅▅▅▄▄▄▅▅▄▅▄▄▃▄▄▄▅▅▅▄▅▄▃▃▄▃▃▂▃ ▄
  3.28 ms         Histogram: frequency by time        4.45 ms <

 Memory estimate: 529.98 KiB, allocs estimate: 13059.
```

with HIP graphs I get:

```Julia
julia> @benchmark transform!(L, field, S)   # grid to spectral space`
BenchmarkTools.Trial: 9126 samples with 1 evaluation per sample.
 Range (min … max):  290.832 μs …   3.398 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     395.760 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   543.181 μs ± 512.100 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▆█▅▃▂▂▂▂▂▁                                                   ▁
  ▇██████████▇▄▃▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▁▅▃▃▄▃▅▆▆▇▇▇██████▇███▇█ █
  291 μs        Histogram: log(frequency) by time       2.73 ms <

 Memory estimate: 13.64 KiB, allocs estimate: 321.
```

Compared to the CUDA graphs benchmakrs on my laptop NVIDIA GPU these results are anything but stellar, but there is a clear performance upgrade compared to the no GPU graphs on AMD GPUs.

I would be happy to clean up and refactor the code if you think this is worth merging eventually.


SpeedyTransforms/Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "SpeedyTransforms"
 uuid = "ed023a3f-7a3c-42dd-bab6-68c21c8c3105"
-version = "0.1.5+DEV"
+version = "0.2.0-DEV"
```
milankl: why would this be a breaking change?

SpeedyTransforms/ext/SpeedyTransformsAMDGPUExt.jl
```diff
@@ -1,14 +1,373 @@
...
+# =====================================================================================
+
+"""$(TYPEDSIGNATURES)
+HIP-graphs accelerated forward (grid → spectral) batched Fourier transform.
+Replays a cached HIP graph of the fused gather + per-ring rocFFTs + scatter when the HIP
+graph API is available; otherwise runs the allocation-free fused loop directly. See
+[`run_graph!`](@ref)."""
+function _fourier_batched!(
+        f_north::ROCArray{<:Complex, 3},
+        f_south::ROCArray{<:Complex, 3},
```
milankl, line 81: Can this not be defined for `GPUArray` and moved into `src/fourier.jl` if it's identical for CUDA/AMD?

SpeedyTransforms/ext/SpeedyTransformsAMDGPUExt.jl
```diff
@@ -1,14 +1,373 @@
...
+# Capture / replay management
+# =====================================================================================
+
+"""$(TYPEDSIGNATURES)
+Run `loop!` (a `() -> ...` closure over the allocation-free batched FFT) either by
+replaying a cached HIP graph keyed by `key`, or — on first use — by warming up, capturing,
+instantiating and caching a graph. Falls back to running `loop!` directly if the HIP graph
+API is unavailable (older AMDGPU.jl), if capture fails, or if the per-direction cache is
+full. Even without graph capture the allocation-free fused loop avoids per-ring allocations."""
+function run_graph!(execs::AbstractDict, key, loop!::F) where {F}
```
milankl: Maybe here we can dispatch over `AMDGPU` or `CUDAGPU` somehow?

SpeedyTransforms/src/spectral_transform.jl
```diff
@@ -76,11 +76,11 @@ struct SpectralTransform{
...
-    # CUDA GRAPHS
-    # toggle for the CUDA-Graphs accelerated batched Fourier transform
-    # Set to `false` to fall back to the generic (allocating) per-ring GPU
-    # Meaningless for non-CUDA architectures.
-    cuda_graphs::B
+    # GPU GRAPHS (CUDA graphs / HIP graphs)
+    # toggle for the GPU-graphs accelerated batched Fourier transform
+    # Set to `false` to fall back to the generic (allocating) per-ring GPU path.
+    # Only effective when a CUDA or AMDGPU extension is loaded; ignored on CPU.
+    gpu_graphs::B
```
milankl: yep I like!

milankl (commented):
Looks great already! Yeah, as you say, if there's code now that's largely the same between CUDA/AMD then I'd suggest to move that code into e.g. `src/graphs.jl` or directly into `src/fourier.jl` if it can be written generically so that both Ext can use it. Defining a new type and some functions isn't going to hit precompile/load time massively so I think that's fine.

maximilian-gelbrecht (commented):
Nice! This is great. I don't have time for a super detailed review right now, but just my initial impression and also answering your question. 

So first of all, it's great to have this feature for AMD as well! So yes, let's make it work. 

In terms of code repetition I would like to strike a balance here. So, one the one side I don't think it's our job here to create abstractions for the Graph interfaces in general, but we could easily move some of common utilities inside of the main package so that they can be used by both. So for example the gather and scatter kernels could be easily just moved in the main package. Some repetition is probably unavoidable and also completely fine. 

Also in general, I mostly just vibe coded the initial CUDA Graph implementation very quickly, so if you find any of it questionable or think you have a better idea to do it and unify it with the AMD implementation, please do so. Don't regard the CUDA implementation as completely set in stone.

milankl:
Also can you create the next PR from a branch within this repo given you're a collaborator now? PRs that come from external don't deploy the docs so we know if they build successfully but we can't see the docs before merging.

SpeedyTransforms/ext/SpeedyTransformsAMDGPUExt.jl
```diff
@@ -1,14 +1,373 @@
...
+# =====================================================================================
+
+"""$(TYPEDSIGNATURES)
+Cache (one per transform SIZE, i.e. per FFT plan) holding the pre-allocated packed
+work buffers, the per-ring reshaped views and FFT plans used by the transforms, the device
+gather/scatter index metadata, and (when the HIP graph API is available) instantiated HIP
+graphs keyed by field buffer. A graph value of `nothing` marks a buffer for which capture
+failed (fall back to direct loop). Uses `Dict{UInt, Any}` for the exec dicts so the struct
+is well-defined regardless of whether the HIPGraphExec type exists."""
+struct GPUFourierGraphCache{PR, PC, RV, CV, IV, A}
```
milankl: Maybe the `A` parameters moves forward we should be able to dispatch over it no?

```julia
foo(::GPUFourierGraphCache, args...)
foo(::GPUFourierGraphCache{<:AMDGPU}, args...)
```

SpeedyTransforms/Project.toml
gregordecristoforo: I'll go back to 0.1.5

gregordecristoforo:
**Status update**: The code is now refactored in a way that should reduce code duplication. However, there is a nasty bug in the AMDGPU path that still isn't fixed. It appears there is an invalid memory access in the full-model HIP graphs integration test. 

I will need to dig deeper when I am back from my holiday at the end of July. 

Note to @gregordecristoforo:
Throwing this problem at Claude gives this hypothesis: Current leading hypothesis, not yet confirmed: each ring's FFT reads/writes through an offset view into one shared packed device buffer rather than a standalone allocation, which may violate an alignment/addressing assumption in rocFFT's execution. Not yet ruled out: the bug may not be in the Fourier transform code at all — an isolated, repeated transform! call didn't fault in earlier benchmarking, only the full model run does, so the trigger could be something else in the timestepping loop.

maximilian-gelbrecht:
> Note to @gregordecristoforo: Throwing this problem at Claude gives this hypothesis: Current leading hypothesis, not yet confirmed: each ring's FFT reads/writes through an offset view into one shared packed device buffer rather than a standalone allocation, which may violate an alignment/addressing assumption in rocFFT's execution.

It's definitely true that they write into a shared buffer that's core to the design and also quite hard to avoid from what I can see, at least for CUDA Graphs as they need to bake in the memory addresses in the graphs. But I suspect that's similar for AMD

maximilian-gelbrecht:
@gregordecristoforo note that I added an `add` kwarg to the `_fourier!` routine and therefore also to the CUDA graph implementation that makes the Fourier transform just add to the existing array and not overwrite it (we needed this for a AD custom rule). You'll see this when you merge `main` again into your branch

gregordecristoforo:
I spent some more time digging deeper into the memory bug on LUMI (MI250X, ROCm 6.3.4, AMDGPU.jl 2.1.2). Even after merging the latest `main` branch, running a real model which captures several graphs per timestep still triggers the same issue, with the identical crash signature (GPU memory access fault, aborted process).

I did some live debugging with `rocgdb` (`set amdgpu precise-memory on`) on this and the failure showed that the SIGSEGV occurs in code entirely unrelated to the Fourier transforms, later in the same timestep, reading a buffer whose GPU memory had become unmapped. That points to an earlier graph replay corrupting the memory allocator's own bookkeeping, so the crashing code is just the first innocent thing to touch the broken mapping, not the actual bug.

If I understood this correctly this means that unlike CUDA, HIP's stream-capture validator doesn't reliably reject illegal-to-capture operations at capture time, so a bad capture can silently "succeed" and only manifest as corruption later (similar issues have been reported here I think: pytorch/pytorch#155684, pytorch/pytorch#155720). So if I understand this issue correctly, this is not a bug in the Julia code but a much deeper issue in the ROCm ecosystem.

I am a bit disappointed after spending quite some time on this but I don't see how this can be made to work anytime soon with the current state. LUMI will have a maintenance break in October in which hopefully a newer ROCm version will be installed but whether this actually fixes this issue is pure speculation. 

Is some of the code (maybe `SpeedyWeather/test/GPU/gpu_graphs_shared.jl`) still helpful as it might make it easier in the future to port the code to non NVidia hardware or should this PR just be closed?

maximilian-gelbrecht:
> If I understood this correctly this means that unlike CUDA, HIP's stream-capture validator doesn't reliably reject illegal-to-capture operations at capture time, so a bad capture can silently "succeed" and only manifest as corruption later (similar issues have been reported here I think: [pytorch/pytorch#155684](https://github.com/pytorch/pytorch/issues/155684), [pytorch/pytorch#155720](https://github.com/pytorch/pytorch/issues/155720)). So if I understand this issue correctly, this is not a bug in the Julia code but a much deeper issue in the ROCm ecosystem.

Am I understanding it correctly that the ROCm library is just not flagging an error correctly? This sounds a little bit like it's just very painful to debug or is it also corrupting otherwise correct graphs? Anyway of course I understand, without a proper way to debug it's almost impossible to develop it. 
 
> Is some of the code (maybe `SpeedyWeather/test/GPU/gpu_graphs_shared.jl`) still helpful as it might make it easier in the future to port the code to non NVidia hardware or should this PR just be closed?

Yes absolutely, I think having it in a state where it where adding a further graph implmementation is easier, is great!

milankl:
I agree, if you feel this is now a problem that's beyond your powers or capacity to fix then I'd make this PR a "towards HI graphs" PR, so putting everything in place to pick the actual problem up in the future would be great. And as you say, then maybe focus on other problems we can fix this month!

gregordecristoforo:
> Am I understanding it correctly that the ROCm library is just not flagging an error correctly?

Yes, the problem seems to be that ROCm flags the illegal captures inconsistently.

I kept the shared structure (`SpeedyWeather/test/GPU/gpu_graphs_shared.jl`) but removed now the faulty hip graph implementation and left a warning that it is currently not implemented. I would be happy to give it another try once LUMI has newer ROCm installed. 

A lot of the comments are AI written and therefore quite verbose. Let me know if you want me to make any changes.  

Do you think this is worth reviewing now already?

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Lower free convection gust constant [#1178](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1178)
```
maximilian-gelbrecht, line 20: This is probably just accidentally deleted

SpeedyWeather/test/GPU/gpu_graphs_shared.jl
```diff
@@ -0,0 +1,123 @@
+# Shared test logic for CUDA-graphs tests, called from cuda_graphs.jl.
+# HIP graphs are not implemented for AMDGPU (see SpeedyTransformsAMDGPUExt.jl and
+# hip_graphs.jl), so this is CUDA-only.
+# GRAPH_CACHES, MAX_GRAPHS, and clear_fourier_graph_cache! live inside the CUDA extension
+# module (SpeedyTransformsCUDAExt), not in the main SpeedyTransforms package — hence
+# accessed via `ext.` below, not `SpeedyTransforms.`. Keeping this code extension-local
+# (rather than shared via the main package) is deliberate: it broke Enzyme's CPU-only AD
```
maximilian-gelbrecht: what's this Enzyme problem?

maximilian-gelbrecht (commented):
So in the end you didn't set up the graphs with a common interface in SpeedyTransform. The tests say something about Enzyme compatibility, what was this problem about?

CHANGELOG.md
gregordecristoforo: Yes of course

gregordecristoforo:
My apologies, I reverted the changes too far back. Let me enable the common interface again. I'll let you know when it is ready to be reviewed.

SpeedyWeather/test/GPU/gpu_graphs_shared.jl
gregordecristoforo: In a previous commit ([3d49207](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1147/commits/3d49207c0d49bdf90e57be0ff07402365de60e9c)) the kernels/cache/loop machinery were in moved into `src` which interfered with the CPU Enzyme tests. Everything is moved into `ext` now to circumvent this problem. I still need to clean up the comments.

gregordecristoforo:
I enabled the common interface for the graphs again and cleaned up some smaller issues. 

@maximilian-gelbrecht, may I ask you to review the PR again?

maximilian-gelbrecht:
> I enabled the common interface for the graphs again and cleaned up some smaller issues.
> 
> @maximilian-gelbrecht, may I ask you to review the PR again?

Sure. Will do tomorrow afternoon on my way home from JuliaCon.

gregordecristoforo:
I am a bit puzzled that the ` parameterizations/terrarium_coupling` tests fail for Julia v1.10

I cannot reproduce this issue on my machine and the tests passed for the previous commit, [82fbd0a](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1147/commits/82fbd0ad3052bd1b44807464603df55e3efc7f7f). The only things that changed since then is removing a redundant dependency and updating the changelog.

SpeedyTransforms/ext/gpu_graphs_common.jl
```diff
@@ -0,0 +1,344 @@
...
+# `include()`-d by each graph-capturing backend extension (currently SpeedyTransformsCUDAExt.jl
+# and SpeedyTransformsAMDGPUExt.jl). See SpeedyTransformsCUDAExt.jl for the rationale of
+# GPU-graphs acceleration itself.
+#
+# This file must stay under SpeedyTransforms/ext/ and be `include()`-d from within an
+# extension module, NOT moved to SpeedyTransforms/src/: the `_fourier_batched!` methods that
+# actually use this machinery live in each extension, and the main package loads
+# unconditionally in every Julia session — so if any of this were in src/ instead, those
+# methods would exist even in CPU-only sessions with no GPU backend loaded, which breaks
+# Enzyme's static activity analysis (see docs/dev/2026-08/gpu-graphs-common-interface.md).
```
maximilian-gelbrecht: Generally it's fine and good to include those files for more transparency and reproducibility with LLM-based development (this is also in our Claude.md file, that's why Claude tends to even do it automatically). 
For minor changes that's not really necessary, but the comments should definitely not refer to a non-existent file in the repo.  

I also still wonder what this Enzyme issue actually is. Because this code shouldn't be in the hot-loop of the configuration that is tested with Enzyme (which is done only CPU). Even on GPU Enzyme shouldn't actually see the graphs because we have custom rules for the complete transform.

maximilian-gelbrecht: I think I would prefer to really get to the bottom of this. I am also happy to take a look at the myself some time in the upcoming weeks when it's Enzyme related.

SpeedyTransforms/ext/gpu_graphs_common.jl
```diff
@@ -0,0 +1,344 @@
...
+    # southern rings (the equator ring, if any, is skipped: north already wrote those rows)
+    launch!(arch, ArrayWorkOrder, complex_size, gather_complex_kernel!, packed_complex, g_south, complex_offset, nfreqs)
+    @inbounds for j in 1:nlat_half
+        (cache.has_equator && j == j_equator) && continue
+        mul!(real_view[j], brfft_plans[j], complex_view[j])
+    end
+    launch!(arch, ArrayWorkOrder, real_size, scatter_real_kernel!, field.data, packed_real, real_offset, nlons_s, istart_s, add)
+    return nothing
+end
+
```
maximilian-gelbrecht, line 261: It's a bit hard to see without seeing the difference. But up until here this is basically should the old CUDA Graph code with just very minor adjustments, correct?

SpeedyTransforms/ext/gpu_graphs_common.jl
```diff
@@ -0,0 +1,344 @@
...
+    launch::L
+    synchronize::Y
+end
+
+"""$(TYPEDSIGNATURES)
+Run `loop!` (a `() -> ...` closure over the allocation-free batched FFT) either by
+replaying a cached GPU graph keyed by `key`, or — on first use — by warming up, capturing,
+instantiating and caching a graph via the given `backend`. Falls back to running `loop!`
+directly if capture fails or the per-direction cache is full."""
+function run_graph!(execs::AbstractDict, key, loop!::F, backend::GraphBackend) where {F}
```
maximilian-gelbrecht, line 293: that's also mostly just the old one bad with the `GraphBackend` used, correct? I like it.

SpeedyTransforms/ext/gpu_graphs_common.jl
```diff
@@ -0,0 +1,344 @@
...
+
+"""$(TYPEDSIGNATURES)
+Bundles the four backend-specific primitives `run_graph!` needs to capture and replay a GPU
+graph, so the control flow in `run_graph!` itself can stay backend-agnostic:
+
+- `capture(loop!)`: run `loop!` under stream capture and return the recorded graph, or
+  `nothing` if capture failed/was invalidated (must not throw).
+- `instantiate(graph)`: turn a captured graph into an executable.
+- `launch(exec)`: replay an instantiated graph.
+- `synchronize()`: block until the device is idle (used once, before capture, so that
```
maximilian-gelbrecht: The `synchronize` should already just work with the regular `synchronize` for all devices. Maybe I would instead just hold our SpeedyWeather defined `device` here. And then you can do 
```
synchronize(backend.device)
```
as we do in the rest of the code base as well.

SpeedyTransforms/ext/gpu_graphs_common.jl
gregordecristoforo: Unfortunately, I already deleted the file `docs/dev/2026-08/gpu-graphs-common-interface.md`and never committed it. Also, sorry about the leftover comment reference. I can delete it this too.

As for the Enzyme issue: this comment was actually a remnant from my first attempt, where I accidentally put the HIP graphs implementation in `/src` instead of `/ext`, which triggered the problem I think. Since moving it to `/ext`, it hasn't popped up, though I can't completely rule out that I misconfigured something else during that initial try.

Because of that, I’m not entirely sure it’s worth spending too much time digging into. But if you'd still like to get to the bottom of it, would you prefer to investigate it in a separate issue/PR, or should we keep this PR open for it?

SpeedyTransforms/ext/gpu_graphs_common.jl
gregordecristoforo: Yes, this file holds everything that is backend-agnostic and is almost identical to the old CUDA implementation.

SpeedyTransforms/ext/gpu_graphs_common.jl
gregordecristoforo: yes exactly

SpeedyTransforms/ext/gpu_graphs_common.jl
gregordecristoforo: Thanks, good point. I'll fix this in the next commit.

gregordecristoforo: Addressed in [e8bfc4e](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1147/commits/e8bfc4eee255fffb5f7733ed710063269443815a)

gregordecristoforo:
Shall we go ahead and merge this PR (once fully reviewed of course) even though #1193 is already merged?

I would like to give the HIP graphs implementation another try but maybe a separate PR would be helpful if we run into further problems. What do you think @maximilian-gelbrecht

maximilian-gelbrecht:
Fine with me to merge this first, and then try the hip graphs again in a seperate PR. But please still address the comments above (e.g. the comment that points to a non-existent file).

SpeedyTransforms/ext/gpu_graphs_common.jl
maximilian-gelbrecht: Okay, yeah fine with me, but please edit the comment. 

Maybe put a TODO note there that implementing it in `/src` caused some Enzyme problem.

gregordecristoforo: I edited the comment and added a TODO note. I can try to reproduce this issue after attempting the HIP graphs.

gregordecristoforo:
Thanks @maximilian-gelbrecht! Is anyone still planing to review any other parts the PR or can I merge once all the tests passed? I am asking since the PR was never formally approved.

milankl (approved):
Looks good to me, but I'll let Max comment too. 

- I see that you left `gpu_graphs_common.jl` inside the `ext` folder and then `include` from the extensions. I guess alternatively this file could live in SpeedyTransforms directly, so is always loaded and only once. I guess now doing `using AMDGPU, CUDA` would load those functions twice, which may cause precompilation issues, but not sure if that's the same inside extensions. But at the same time, one wouldn't `using AMDGPU, CUDA` anyway and also then `GPU()` is overloaded, so maybe there's bigger problems then

SpeedyWeather/test/GPU/hip_graphs.jl
```diff
@@ -0,0 +1,20 @@
+# HIP graphs are not implemented for AMDGPU (see SpeedyTransformsAMDGPUExt.jl for why);
+# `gpu_graphs = true` must fall back to the generic Fourier path and warn exactly once.
+ext = Base.get_extension(SpeedyWeather.SpeedyTransforms, :SpeedyTransformsAMDGPUExt)
+
+@testset "HIP Graphs: not implemented, falls back to the generic path with a warning" begin
+    if ext !== nothing
```
milankl: shouldn't the else case throw something here? Otherwise looks like if the extension doesn't load for whatever reason this may come back as tests passed? Maybe not, just checking

maximilian-gelbrecht:
> But at the same time, one wouldn't `using AMDGPU, CUDA` anyway and also then `GPU()` is overloaded, so maybe there's bigger problems then

This currently breaks the model already anyway with how the architectures in `SpeedyWeatherInternals` are defined. 

I also think the regular `src` of SpeedyTransforms is maybe the slightly more proper place for it, but in the end I am also a bit indifferent towards it. It works like this and it seems there were some problems before when putting it into `src`.

maximilian-gelbrecht (commented):
As this directly affects a performance critical part, I just quickly run a benchmark on the H100 in the background. After that's one we can merge it.

maximilian-gelbrecht:
I currently get a pre compile error on Nvidia here.

```
WARNING: Imported binding Architectures.synchronize was undeclared at import time during import to SpeedyTransformsCUDAExt.
ERROR: LoadError: MethodError: no method matching CUDAGPU()
The function `CUDAGPU` exists, but no method is defined for this combination of argument types.
Stacktrace:
 [1] top-level scope
   @ /p/tmp/maxgelbr/code/SpeedyWeather.jl/SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl:59
 [2] include(mod::Module, _path::String)
   @ Base ./Base.jl:306
 ...
in expression starting at /p/tmp/maxgelbr/code/SpeedyWeather.jl/SpeedyTransforms/ext/SpeedyTransformsCUDAExt.jl:1
in expression starting at stdin:5
              ✗ SpeedyTransforms → SpeedyTransformsCUDAExt
  0 dependencies successfully precompiled in 10 seconds. 168 already precompiled.
┌ Error: Error during loading of extension SpeedyTransformsCUDAExt of SpeedyTransforms, use `Base.retry_load_extensions()` to retry.
│   exception =
│    1-element ExceptionStack:
│    Failed to precompile SpeedyTransformsCUDAExt [6ada7dab-cacc-5d93-946e-8c5c8729d369] to "/home/maxgelbr/.julia/compiled/v1.12/SpeedyTransformsCUDAExt/jl_AvZeXy".
└ @ Base loading.jl:1614
LOAD OK
``

gregordecristoforo:
This is strange, the tests didn't catch that. Which benchmark did you run? Let me fix this.

SpeedyWeather/test/GPU/hip_graphs.jl
gregordecristoforo: Good point! Addressed in [f2f0c3c](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1147/commits/f2f0c3cacefe2271b38ecaba22565cad8cc179db)

gregordecristoforo:
> I currently get a pre compile error on Nvidia here.

I am really surprised that the tests in [e8bfc4e](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1147/commits/e8bfc4eee255fffb5f7733ed710063269443815a) are shown as passing. When running them locally they fail as they did for you @maximilian-gelbrecht 

I quickly set up a PR that fixed it for me locally: https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1203
This is however vibe coded under time pressure so I would definitely wait for the test runner before going any further or asking for a review. 

I will look at it again with a fresh pair of eyes tomorrow morning. Thanks for catching this! I relied on the tests in  [e8bfc4e](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1147/commits/e8bfc4eee255fffb5f7733ed710063269443815a) being shown green. Is this normal that precompilation issues are not displayed as errors?

maximilian-gelbrecht:
I am also confused by this. So in principle, I think the Julia tests don't report an error to the CI if the actual tests don't error or the process doesn't terminate because of an error. I've definitely seen this before with other extensions failing. A failing loading of an extension is defintely not enough to kill the process. However, what confused me is why this didn't actually cause an error in the subsequent tests.

gregordecristoforo:
I ran the GPU testsuite for my last commit [a8414a0](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1147/commits/a8414a0d06ada623653027d0835ab382283a9f34) locally on my NVIDIA GPU and on LUMI's MI250x GPUs for both Julia v1.11 and 1.12 and all GPU tests passed. Also, I didn't observe any pre compile errors. 

The failing [Reactant Correctness / Reactant Correctness - Julia v1.11, test](https://github.com/SpeedyWeather/SpeedyWeather.jl/actions/runs/32231917185/job/96007084385?pr=1147) appears to be the same error as for the last commit on `main`: [b65590e7a5840323ce28d23360db035ec7a4373a](https://github.com/SpeedyWeather/SpeedyWeather.jl/commit/b65590e7a5840323ce28d23360db035ec7a4373a) so I don't think it is related to this PR. Is this reasonable?

Thanks again @maximilian-gelbrecht for manually testing the branch. Glad the pre compile bug wasn't merged into `main`. 

**Edit**: I saw that https://github.com/SpeedyWeather/SpeedyWeather.jl/commit/f28d90dc00852cda16342a615b1c125ca4270d4f appears to fix the Reactant test failure. I suspect merging `main` into `gd/hip-graphs` should fix this then.

maximilian-gelbrecht (approved):
I didn't commit them, but the benchmarks went through and performance looks fine. 

Please still wait for the GPU CI, but then this can be merged from my side.

gregordecristoforo:
Thanks for merging and your patience with this PR @maximilian-gelbrecht !

maximilian-gelbrecht:
Thank you for doing the PR!

PR #1140, opened 2026-06-25
Sea surface temperature and sea ice as time-stepped variables

milankl:
## Summary

This PR refines the leapfrog time-stepping treatment of sea surface temperature (SST) and sea
ice now that SST is a time-stepped prognostic variable, and introduces a general `filter!` API
for post-timestep corrections.

- **Leapfrog step selection for SST/sea ice**: `which_prognostic_step` now dispatches per
  component — parameterizations read the previous timestep (Euler forward), while `AbstractOcean`
  and `AbstractSeaIce` read the current timestep, since SST/sea ice concentration are now
  time-stepped themselves.
- **`filter!` API**: New `Base.filter!(vars, component, model)` hook, called once per timestep
  in `time_integration.jl` for `model.ocean`, `model.sea_ice`, and `model.land`, to apply
  corrections that don't fit naturally into a tendency (e.g. clamping).
- **Sea ice rewritten as tendency-based**: `ThermodynamicSeaIce` now declares
  `sea_ice_concentration` as a proper prognostic/tendency variable (via a new `variables(...)`
  method) instead of mutating it directly in `timestep!`; melting/freezing is computed as a
  tendency `dℵ`, and SST is nudged toward freezing via a tendency `dsst` rather than being
  clamped in place. Clamping `ℵ` to `[0, 1]` moved into the new `filter!` for `ThermodynamicSeaIce`.
- **Land also gets a `filter!`**: `AbstractLand` implements `filter!` mirroring its
  `initialize!`/`timestep!` structure, delegating to snow, soil moisture, vegetation, rivers,
  and temperature sub-components.
- **`SlabOcean` initial condition fix**: initial SST is now also capped at the sea ice freezing
  temperature (when a `sea_ice` model with `freezing_temperature` is present), in addition to
  the existing land masking.
- **Reorganized abstract types**: `AbstractParameterization`, `AbstractOcean`, and
  `AbstractSeaIce` moved out of `parameterizations.jl` (renamed `general.jl`) into a new
  `parameterizations/abstract_types.jl`, included earlier in `SpeedyWeather.jl` alongside the
  other abstract-type files.
- **Ocean/land fraction scaling in output**: `NetCDFOutput`/`ZarrOutput` now precompute a
  `land_fraction` field on the output grid; `output!` in `output/writers/general.jl` scales
  variables flagged with `ocean_fraction`/`land_fraction` (e.g. sea ice concentration) by the
  cell's ocean or land area fraction before writing.
[...]

PR #1149, opened 2026-07-01
Ensemble Zarr output

maximilian-gelbrecht:
@bgroenks96 wants to use SBI with Speedy and having ensembles output into the same store makes this more practical. Zarr  allows for parallel writes/reads so it's compatible with this without limiting it to purely sequential ensembles (as would be the case for NetCDF).


maximilian-gelbrecht:
### Ensemble output

Several runs of an ensemble can be written into a *single* Zarr store along an
additional `ensemble` dimension. This is disabled by default and controlled by two
options:

| Option | Meaning |
|--------|---------|
| `ensemble_index::Int` | This writer's ensemble member index. `0` (default) disables ensemble output and reproduces the layout described above. A value `> 0` adds an `ensemble` dimension and makes this writer store its data into slot `ensemble_index`. Members are indexed `1..n_ensemble`. |
| `n_ensemble::Int` | Total number of ensemble members. It sizes the `ensemble` dimension and must be passed up front; it has to satisfy `n_ensemble ≥ ensemble_index`. |
| `ensemble_timeout::Int` | Seconds a member waits for member 1 to create the shared store before erroring (default `600`). |

The design targets **parallel ensemble members running as separate processes**, all
writing into one store. This is safe because the `ensemble` axis is chunked with size 1,
so member `e` only ever writes chunk files carrying its own index and no two members
touch the same chunk file. Each process constructs a `ZarrOutput` with the *same*
`path`, `id`, `run_number`, `filename` and `n_ensemble`, and a distinct `ensemble_index`,
so all members resolve to the same run folder and store:

```julia
using SpeedyWeather, Zarr

# in each process, `member` is this process' ensemble index (1..n_ensemble)
output = ZarrOutput(spectral_grid, PrimitiveWet;
    ensemble_index = member,        # e.g. read from an environment variable / job array id
    n_ensemble = 10,
    interval = Hour(6),
)
```

The single shared artifacts (the group and coordinate metadata, and the `time` axis) are
written once by the *creator* — the member with `ensemble_index == 1`, which builds the
store schema and then signals readiness. The remaining *writer* members (`ensemble_index
> 1`) wait for that signal, open the existing store and write only their own ensemble
slice. This assumes that all members share the same time stepping so that one common
`time` axis is correct.

The resulting store gains an `ensemble` coordinate; the ensemble axis is the outermost
(slowest-varying) dimension, so an xarray-compatible reader reports e.g.

```python
import xarray as xr
ds = xr.open_zarr("run_0001/output.zarr", consolidated=False)
print(ds)
# Dimensions:  (ensemble: 10, time: 41, layer: 8, lat: 32, lon: 64)
# Data variables:
#     temp     (ensemble, time, layer, lat, lon) float32
ds["temp"].mean("ensemble")     # ensemble mean
```

Note that this ensemble layout is currently specific to `ZarrOutput`; the
[`NetCDFOutput`](@ref) does not support concurrent multi-process writes.

docs/src/other_output.md
```diff
@@ -141,6 +141,63 @@ work without extra work:
...
+### Ensemble output
+
+Several runs of an ensemble can be written into a *single* Zarr store along an
+additional `ensemble` dimension. This is disabled by default and controlled by two
+options:
+
+| Option | Meaning |
+|--------|---------|
+| `ensemble_index::Int` | This writer's ensemble member index. `0` (default) disables ensemble output and reproduces the layout described above. A value `> 0` adds an `ensemble` dimension and makes this writer store its data into slot `ensemble_index`. Members are indexed `1..n_ensemble`. |
+| `n_ensemble::Int` | Total number of ensemble members. It sizes the `ensemble` dimension and must be passed up front; it has to satisfy `n_ensemble ≥ ensemble_index`. |
```
milankl: Can we call this `ensemble_size`? Yes we use `n_steps`, `nlayers` but for those one doesn't say "step size" (difference between steps not total number of steps) or "layer size". But I think most people would say "how many members did you have in your ensemble" or "what is your ensemble size"?

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
```diff
@@ -103,25 +111,87 @@ function initialize!(
             " Please construct ZarrOutput with the same `nlayers_soil` as the model."
     end
 
-    # SHARED INITIALIZATION (run folder, output frequency, counters, callbacks)
-    initialize!(output.core, output, model)
+    ensemble = output.ensemble_index > 0
```
milankl, line 114: So for `ensemble_size = 1` nothing changes?

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
```diff
@@ -103,25 +111,87 @@ function initialize!(
...
+            "_ARRAY_DIMENSIONS" => ["time"]
+        ),
+    )
+    output!(output, vars.prognostic.clock.time)   # write initial time
+
+    # VARIABLES (pre-allocated to full length along the time axis)
+    for (key, var) in output.variables
+        define_variable!(g, output, var, n_outputs, eltype(output.field2D))
+        output!(output, var, Simulation(vars, model))
+    end
```
milankl, line 183: There's some changes here coming in #1140 with https://github.com/SpeedyWeather/SpeedyWeather.jl/commit/89ce01a5d52af8ce44693891a9754062ac7d3a08 so we'll have some conflicts but jut FYI

milankl: That's because there's currently some duplicated code between NetCDF and Zarr and updates in one didn't make it into the other

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
```diff
@@ -103,25 +111,87 @@ function initialize!(
...
+
+    # readiness marker so writer members (ensemble_index > 1) may proceed
+    ensemble && touch(ensemble_marker_path(output))
```
milankl, line 186: is this marker removed again? I don't think it has to but just checking

SpeedyWeather/test/output/zarr_output.jl
```diff
@@ -263,6 +263,57 @@ end
...
+    @test ndims(z_vor) == 5
+    @test size(z_vor) == (nlon, nlat, spectral_grid.nlayers, nt, n_ensemble)
+    @test z_vor.attrs["_ARRAY_DIMENSIONS"] == ["ensemble", "time", "layer", "lat", "lon"]
+    # ensemble axis is chunked with size 1 so members write disjoint chunk files
+    @test z_vor.metadata.chunks[end] == 1
+
+    # every member wrote finite data into its own ensemble slice
+    for e in 1:n_ensemble
+        @test all(isfinite, g["vor"][:, :, :, :, e])
+    end
```
milankl: given they all run the same simulation you can also test for equality across ensemble members

milankl (commented):
Yeah that's pretty cool! Just to check: For no ensemble the ensemble dimension does not exist in the zarr file, correct?

maximilian-gelbrecht:
> Yeah that's pretty cool! Just to check: For no ensemble the ensemble dimension does not exist in the zarr file, correct?

* For `n_ensemble = 0` (or `ensemble_size`) the dimension does not exist, everything as it was before.  
* For `n_ensemble = 1` the dimension does exist but is singleton obviously.

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
maximilian-gelbrecht: For `ensemble_size == 0` (or ensemble_size) the dimension does not exist, everything as it was before.

For `ensemble_size == 1` the dimension does exist but all of the actual logic should be identical, just with a singleton dimension 

For `ensemble_size > 1` the whole thing actually does the ensemble output

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
maximilian-gelbrecht: No, it's not but when we allow for distributed, independent processes that's also isn't even trivial to do. One meaningless file with a few bytes in the folder isn't that bad I think

docs/src/other_output.md
maximilian-gelbrecht: Done

SpeedyWeather/test/output/zarr_output.jl
maximilian-gelbrecht: Don't we have some random IC?

bgroenks96: I don't think so, not by default at least.

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
```diff
@@ -143,27 +213,42 @@ function initialize!(
...
+shared store schema, coordinates and time axis are in place."""
+ensemble_marker_path(output::ZarrOutput) = joinpath(output.run_path, ".zarr_ensemble_ready")
 
-    # VARIABLES (pre-allocated to full length along the time axis)
-    for (key, var) in output.variables
-        define_variable!(g, output, var, n_outputs, eltype(output.field2D))
-        output!(output, var, Simulation(vars, model))
-    end
+"""$(TYPEDSIGNATURES)
+Set a deterministic (non-auto-incrementing) run folder for an ensemble `output` so all
```
bgroenks96, line 225: Why is it not auto-increasing? Each separate ZARR dataset (each ensemble) should still auto-increase the run number right?

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
```diff
@@ -103,25 +111,87 @@ function initialize!(
...
     # Total number of output snapshots: IC + one per `output_every_n_steps`.
     n_outputs = vars.prognostic.clock.n_time_steps ÷ output.output_every_n_steps + 1
 
-    # CREATE ZARR GROUP (the Zarr store is a *directory*, not a single file)
+    # The Zarr store is a *directory*, not a single file.
     (; run_path, filename) = output
     store_path = joinpath(run_path, filename)
-    # remove a stale store at the same location when overwrite is on
+
+    # WRITER member (ensemble_index > 1): wait for the creator to build the store, then
```
bgroenks96, line 133: Hmm, what happens if something goes wrong in the first member? The rest would hang?

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
maximilian-gelbrecht: It's not auto incrementing at the moment, because that would be need some resizing. We could implement this obviously, but don't you always know the number of ensemble members already in the beginning?

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
maximilian-gelbrecht: The rest would error after the timeout. But building the store is already done in the `initialize!(simulation)` so before the model starts integrating. If something is going wrong at that point for the first member, then almost certainly the other members will also directly error for the same reason.

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
bgroenks96: Yes, I don't mean it should auto-increment for each ensemble member. I was more thinking about multiple successive ensemble runs.

SpeedyWeather/test/output/zarr_output.jl
maximilian-gelbrecht: Yeah you are right just for the Barotropic model the default is random.

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
maximilian-gelbrecht: Ah yes. That's actually kind of hard to do for potentially independent, parallel processes. Everything that I could quickly think of is kind of finicky and leaves some loophole for it to go wrong as well. It's just easier to have it non increasing, and set the name (including a number) of the folder manually.

SpeedyWeather/test/output/zarr_output.jl
maximilian-gelbrecht: ok, done. it tests approx equality now.

bgroenks96 (approved):

PR #1150, opened 2026-07-03
Fields and LowerTriangularArrays with dimensions

milankl:
The idea is to have an additional type for every field/LTA that contains information about what the dimensions mean, e.g. `XY` is horizontal 2D, `XYZ` would have another dimension for vertical, but `XYT` would use that for time. Needed in SpeedyWeather to actually not with `get_step` whether a dimension is time or vertical. All other dimensions remain unspecified.

A grid would continue to contain the horizontal information only

```julia
julia> using .RingGrids
julia> grid = HEALPixGrid(2)
3-ring HEALPixGrid{...}
├ nlat_half = 2 (12 points, ~59˚, reduced)
└ architecture = CPU
```

By default we follow the current logic that every other dimensions are unspecified. `XY` for 2D, `XY+` (just for the docstring) to day it's 2D + other unspecified dimensions, e.g.

```julia
julia> field = Field(grid)
12-element, 3-ring (XY) HEALPixField{Float32, 1} as Array on CPU
...

julia> field = Field(grid, 2)
12×2, 3-ring (XY+) HEALPixField{Float32, 2} as Array on CPU
...
```

The dimensions themselves are just implemented as empty `XY <: TwoDimensions <: AbstractArrayDimensions` in a new module (to be moved to SpeedyWeatherInternals) and are stored in `field.dims`

```julia
julia> field.dims
Main.RingGrids.ArrayDimensions.XY()
```

One can, when creating a field, specify those, e.g. here to denote that the additional dimensions is the vertical use `XYZ`

```julia
julia> field = Field(grid, ArrayDimensions.XYZ(), 2)
12×2, 3-ring (XYZ) HEALPixField{Float32, 2} as Array on CPU
...
```

or for time/step use `XYT`

```julia
julia> field = Field(grid, ArrayDimensions.XYT(), 2)
12×2, 3-ring (XYT) HEALPixField{Float32, 2} as Array on CPU
...
```

then we have `hastime` and `hasvertical` functions that I've checked are compile-time evaluated. So this can all be inlined and is a no-cost abstraction

```julia
julia> ArrayDimensions.hastime(field)
true
```

in constrast, this has no time dimension

```julia
julia> field = Field(grid, ArrayDimensions.XYZ(), 2)
12×2, 3-ring (XYZ) HEALPixField{Float32, 2} as Array on CPU
...

julia> ArrayDimensions.hastime(field)
false
```


milankl:
@maximilian-gelbrecht need to rebase, but could you have quick look at https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1150/commits/cecba6fe30082f2812654231f672351e8c486abe and see whether you see any problems with this design? I decided against the idea of having a tuple, e.g. `(XY, Z, T)` to define the dimensions of a N=3 i.e. 4D field which would have been more modular but that felt like an overkill given that at the moment we literally just need to distinhguish between is that dimensions a step dimension or a vertical. 

The idea is then to define more variable dimensions, while we currently have, e.g.

- `Grid2D`
- `Grid3D`
- `Grid4D`

which I believe should continue to mean XY + unspecified, we then would define

- `GridXY` (same as `Grid2D`)
- `GridXYZ` using `ArrayDimensions.XYZ`
- `GridXYT` using `ArrayDimensions.XYT`
- `GridXYZT` using `ArrayDimensions.XYZT`

and similar for `LowerTriangularArrays` of course.

milankl:
Other dimensions in the future could easily be added, e.g. `XYE`, `XYZE`, `XYTE` and `XYZTE` and `hasensemble` for an ensemble dimension for example. For now the largest work is to add all those methods to the field constructors to allow `dims::AbstratArrayDimensions` as an additional argument ... but it's not rocket science.

milankl:
I guess one more realization is that because within kernels we adapt only `field.data` all this functionality is only available outside kernels but I don't think that's an issue

RingGrids/src/field.jl
```diff
@@ -4,14 +4,16 @@ for compatibility across full and reduced grids), the other dimensions can be us
 time or other dimensions. The `grid` can be shared across multiple fields, e.g. a 2D and a 3D field
 can share the same grid which just defines the discretization and the architecture (CPU/GPU) the grid is on.
 $(TYPEDFIELDS)"""
-struct Field{T, N, ArrayType <: AbstractArray, Grid <: AbstractGrid} <: AbstractField{T, N, ArrayType, Grid}
+struct Field{T, N, ArrayType <: AbstractArray, Grid <: AbstractGrid, Dims <: AbstractArrayDimensions} <: AbstractField{T, N, ArrayType, Grid, Dims}
```
maximilian-gelbrecht, line 7: Maybe it's worth it to reorder it for `Dims` to be in third or forth position? Makes it easier to dispatch on.

maximilian-gelbrecht (commented):
I agree with the general approach. 

As you already stated, the main mechanic should be moved into SpeedyWeatherInternals. 
Before merging this, we also need to benchmark this to be sure it doesn't cause any problems in the numerics. 

Unavailablity within kernels also doesn't matter. The kernel launching handles the dimensionality of the arrays it's launched on, not the kernel body.

RingGrids/src/field.jl
milankl: Maybe but I did deliberately last position to not cause changes elsewhere because 3/4th position isn't arraytype or grid anymore. At the moment we don't need to dispatch over it? and

```julia
f(field) = hastime(field) ? A : B
```

would compile away anyway and so two methods would be compiled because field with different dimension would now also be different for the compiler

maximilian-gelbrecht: Okay, I am fine with it as well.

milankl:
@bgroenks96 this should be a non-breaking change for RingGrids but you could create fields now with additional information (just a type) on what the dimensions mean, e.g. a 2-layer horizontal field would be

```julia
field = Field(grid, ArrayDimensions.XYZ(), 2)
field = Field(grid, 2)   # this works as before with field.dims == XY() and other dimensions unspecified
```

maximilian-gelbrecht (commented):
Looks good to me. Do you want to keep this PR limited to the basic implementation or also apply it? 

If I see it correctly the only thing that would need to be changed in the model to set the dimensions correctly is changing the array type definitions in `SpectralGrid`. But then there also might be cases where we actually make use of it of course as well

milankl:
I thought of doing a second PR where we actually use it. But the problem now is that I see some performance regression here on my M4. Typical T31L8 speeds of `PrimitiveWetModel` I get on this branch are

```julia
julia> run!(simulation)
 100% Time: 0:00:01 (2000-02-10, 1572.92 years/day,  92 m/s, [ -81,   35] ˚C)
```

compare to `#main` 

```julia
julia> run!(simulation)
 100% Time: 0:00:01 (2000-02-20, 1723.80 years/day,  86 m/s, [ -80,   35] ˚C)
```

So this branch made things about 10% slower. The `dims` aren't actually used so I suspect that the compiler just gives up somewhere because e.g. `typeof(Variables)` has now increased in complexity.

maximilian-gelbrecht:
The timing includes the compilation of the first step as well, doesn't it? Is there also a different the second time you execute it / after a warm up?

milankl:
Warm up is excluded here! So I ran `run!` several times to make sure timings are stable. And copy pasted the last `run!` call

milankl:
I see a similar performance regression in both dycore and parameterizations, 

This PR dycore only
```julia
julia> run!(simulation, period=Day(50))
 100% Time: 0:00:03 (2000-03-31,  3.35 millenia/day,  69 m/s, [ -90,   33] ˚C)
```

vs main
```julia
julia> run!(simulation, period=Day(50))
 100% Time: 0:00:03 (2000-05-30,  3.50 millenia/day,  83 m/s, [ -88,   34] ˚C)
```

so  4% slower. And parameterizations only, this PR

```julia
julia> run!(simulation, period=Day(50))
 100% Time: 0:00:05 (2000-05-30,  2.05 millenia/day,  35 m/s, [ -74,   34] ˚C)
```

vs main

```julia
julia> run!(simulation, period=Day(50))
 100% Time: 0:00:05 (2000-07-19,  2.22 millenia/day,  35 m/s, [ -74,   34] ˚C)
```

so 9% slower

milankl:
Claude says the regression is from additional allocations 

Regression reproduced. Steady-state (Day(2), Octahedral), branch vs main:

T | branch time | main time | Δ time | branch allocs | main allocs | Δ allocs
-- | -- | -- | -- | -- | -- | --
31 | 0.330 s | 0.304 s | +8% | 645 k | 503 k | +28%
63 | 2.645 s | 2.518 s | +5% | 2.92 M | 1.89 M | +54%
127 | 23.27 s | 22.17 s | +5% | 15.0 M | 7.35 M | +104%

So the real regression is allocation count, which scales super-linearly — up to 2× more allocations at T127 — plus a modest ~5% wall-time hit. This is a per-array-operation extra allocation from the dims machinery, not a compile-time issue. Let me localize exactly where it allocates with the allocation profiler.

milankl:
More details:

# `mk/dimensions` performance regression — diagnosis & remediation

## TL;DR

The regression is **not** type instability, **not** compile-time/TTFX, and **not** the
dims `getindex`. It is a **steady-state allocation blow-up** caused by the extra `Dims`
type parameter on `LowerTriangularArray` defeating an optimizer elision inside the Legendre
transform. Since removing the type parameter is off the table, the fix should target the
**one hot `reshape`** that now heap-allocates.

## Hypotheses ruled out

- **`getindex` for dimensions type-unstable** — no. Every core op (dims `getindex`,
  `field[:,k]` / `L[:,k]`, `similar`, broadcast, `view`) is fully type-stable; no runtime
  dynamic dispatch.
- **Compile-time / "compiler struggles with the type parameter"** — first-`run!` TTFX is
  identical branch vs main (~22 s, 99% compilation). The ~20 s of "ArrayDimensions
  inference" merely *replaces* equivalent inference main already does (net zero).
- **Runtime type instability in primitives** — no. In-place hot-loop math (`@. dst = a + 2b`)
  is 0-allocation and same speed on both.

## What the regression actually is

Steady-state **allocation count**, scaling super-linearly with resolution:

| T   | branch  | main    | Δ      |
|-----|---------|---------|--------|
| 31  | 645 k   | 503 k   | +28%   |
| 63  | 2.92 M  | 1.89 M  | +54%   |
| 127 | 15.0 M  | 7.35 M  | +104%  |

Wall-time impact is smaller (~5% at high T) but real and grows with GC pressure.

## Root cause

Allocation profiling (branch vs main diff) attributes **~92% of the extra allocations to a
single site**: `reshape @ reshapedarray.jl:50`, reached from **`_legendre!`**
(`SpeedyTransforms/src/legendre.jl:76` and `:179`) via:

```julia
spec_view = view(specs.data, lm:lm_end, :)
```

Prognostic LTAs are slices of the fused variable buffer, so specs.data is a
non-contiguous 3-D SubArray (e.g. size (2144, 8, 2)). Applying 2 indices to a
3-D array forces _maybe_reshape_parent to reshape the SubArray → a heap-allocated
ReshapedArray. On main the optimizer elides this (escape analysis / SROA); the extra
Dims type parameter enlarges _legendre!'s specialization enough to break the elision,
so it heap-allocates once per wavenumber column per latitude.

Confirmed: de-parameterizing only LowerTriangularArray (Field untouched) restores
allocations to exactly main's numbers — so the regression is specifically the LTA type
parameter, exercised through the Legendre transform.

Remediation options (keeping the Dims type parameter)
Ordered by recommendation:

1. Hoist the reshape out of the hot loop in _legendre!.
view(specs.data, lm:lm_end, :) is called inside the m/lm wavenumber loop, so the
reshape happens O(nlat × mmax) times. Reshape specs.data to 2-D once at the top of
the function (single wrapper), then slice the 2-D result in the loop. Turns thousands of
allocations into one, and is independent of the type parameter.

2. Make the view match the parent's dimensionality.
Replace the 2-index view(specs.data, lm:lm_end, :) with a view that keeps all trailing
dims explicit (so no _maybe_reshape_parent reshape is triggered), or index the 3-D
SubArray directly.

3. Feed the transform contiguous / 2-D spectral data.
The reshape only arises because specs.data is a non-contiguous 3-D SubArray from the
fused buffer. If the transform receives a 2-D-data (single-level, contiguous) LTA — or a
reshaped-once view stored in scratch memory — the 2-index view is a no-op.

4. Restore the elision via a function barrier.
Wrap the view + _fused_oddeven_matvec! inner step in a small @inline'd helper so the
ReshapedArray provably doesn't escape, letting escape analysis stack-allocate it again
despite the larger outer type. Least predictable of the four; validate by re-profiling.

Options 1–3 are the robust fixes (they also benefit main). Option 1 is the smallest,
lowest-risk change with the biggest payoff.

Verify any fix by
Re-running the resolution sweep (T31/63/127, Octahedral, PrimitiveWet, no output) and
checking that steady-state allocation count returns to main's baseline
(≈504 k / 1.90 M / 7.35 M). Wall-time follows.

maximilian-gelbrecht:
### Confirmation 

One first thought about confirming these hypotheses. If this is really just about the view/slicing in the Legendre transform, then both the CPU parameterisation-only model and also the GPU model should not suffer from any regression. That's something easy to test before continuing. Both don't use the CPU Legendre transform. 

### Root Cause 

I don't think I really understand the compiler optimisation that's going wrong here with the additional type parameter. Could you ask to brake this down to a MWE? I am surprised the additional type parameter really matters here. 

### Suggestions

It's kind of hard to evaluate the different suggestions, because I don't really understand the main cause really well. 

1) Is basically suggesting that taking a view on one index first, and then another view on that view would help? This seems super unintuitive to me that this would help. But it's easy to test (especially with LLM help), so why not try it. 

2) I don't understand what exactly this would change / means 

3) Julia is column-major, The kind of indexing we do there cannot be contiguous easily. But maybe that's actually a good point and it could be even worth it performance-wise to once-per-transform reshape this? I don't know, but maybe worth trying 

4) `_fused_oddeven_matvec!` is already `@inline`d, could we inline more calls there? If yes then that's also just easy to try. But I wouldn't know what to inline.

milankl:
# Variable Dimension Type Updates Summary

## Overview
Updated all `variables(::ModelComponent, ...)` definitions throughout SpeedyWeather to use the new semantic variable dimension types (GridXYZ, GridXYT, GridXYZT, SpectralXYZ, SpectralXYT, SpectralXYZT, LandXYZ) instead of generic types (Grid3D, Grid4D, Spectral3D, Spectral4D, Land3D).

## Mapping Reference
| Old Type | New Type(s) | Use Case |
|----------|------------|----------|
| `Grid3D()` | `GridXYZ()` | 3D grid with vertical levels, no time cache |
| `Grid3D(n)` | `GridXYT(n=n)` | 3D grid with time dimension |
| `Grid4D(n)` | `GridXYZT(n=n)` | 4D grid with vertical levels + time dimension |
| `Spectral3D()` | `SpectralXYZ()` | 3D spectral with vertical levels, no time cache |
| `Spectral3D(n)` | `SpectralXYT(n=n)` | 3D spectral with time dimension |
| `Spectral4D(n)` | `SpectralXYZT(n=n)` | 4D spectral with vertical levels + time dimension |
| `Land3D()` | `LandXYZ()` | Land variables with vertical levels |

## Files Updated

### Model Files (4 files, 37 changes)
- **barotropic.jl**: 10 variable updates
  - Vorticity prognostic: Spectral4D→SpectralXYZT
  - Grid variables: Grid4D→GridXYZT
  - Scratch arrays: Spectral3D→SpectralXYZ
  
- **shallow_water.jl**: 11 variable updates
  - Divergence: Spectral4D→SpectralXYZT, Grid4D→GridXYZT
  - Interface displacement η: Spectral3D(ps)→SpectralXYT(n=ps)
  - Geopotential and scratch: Grid3D()→GridXYZ()
  
- **primitive_dry.jl**: 12 variable updates
  - Prognostic variables (divergence, temperature, pressure)
  - Tendency variables (spectral and grid)
  - Grid variables (all with proper vertical/time distinction)
  - Dynamics variables (pres_flux, w, div_sum_above): Grid3D()→GridXYZ()
  - Virtual temperature: Spectral3D()→SpectralXYZ()
  - Scratch arrays: Grid3D()→GridXYZ()
  
- **primitive_wet.jl**: 4 variable updates
  - Humidity prognostic: Spectral4D→SpectralXYZT
  - Humidity grid: Grid4D→GridXYZT
  - Humidity tendencies: Spectral4D/Grid4D→SpectralXYZT/GridXYZT

### Dynamics Files (2 files, 9 changes)
- **geopotential.jl**: 2 updates
  - Grid geopotential: Grid3D()→GridXYZ()
  - Spectral geopotential: Spectral3D()→SpectralXYZ()
  
- **tracers.jl**: 7 updates
  - Prognostic tracer: Spectral4D(ps)→SpectralXYZT(n=ps)
  - Grid tracer: Grid4D(pg)→GridXYZT(n=pg)
  - Tendency tracers: Spectral4D(ts)/Grid4D(tg)→SpectralXYZT(n=ts)/GridXYZT(n=tg)
  - Scratch arrays: Grid3D()→GridXYZ()

### Land Parameterization Files (4 files, 8 changes)
- **soil_moisture.jl**: 2 updates (Land3D→LandXYZ)
- **soil_temperature.jl**: 3 updates (Land3D→LandXYZ)
- **vegetation.jl**: 2 updates (Land3D→LandXYZ)
- **snow.jl**: 1 update (Land3D→LandXYZ)

### Support Code (1 file, 16 changes)
- **dimensions.jl**: Updated fusion helper functions
  - `fuse_family()`: Added recognition for GridXYZ, GridXYT, GridXYZT, SpectralXYZ, SpectralXYT, SpectralXYZT
  - `is_fuse_4d()`: Added GridXYZT and SpectralXYZT as 4D types
  - `fuse_trailing_n()`: Added time dimension extraction for GridXYZT and SpectralXYZT
  - `fused_slots()`: Added slot counting logic for new 3D types with time dimension

## Type Hierarchy
All new dimension types now inherit from the appropriate abstract supertypes:

AbstractVariableDim2D
├── Grid2D
└── Spectral2D

AbstractVariableDim3D
├── Grid3D
├── GridXYZ
├── GridXYT
├── LandXYZ
├── LandXYT
├── OceanXYZ
├── OceanXYT
├── Spectral3D
├── SpectralXYZ
└── SpectralXYT

AbstractVariableDim4D
├── Grid4D
├── GridXYZT
├── Land4D
├── Ocean4D
├── Spectral4D
└── SpectralXYZT



## Testing & Verification
✅ Model initializes successfully with new dimension types
✅ 1-hour simulation completes without errors
✅ All dimension types properly categorized in hierarchy
✅ Fusion helper functions support all new types

## Total Changes
- **10 files modified**
- **73 variable dimension updates**
- **All changes preserve backward compatibility** (generic types Grid3D/Grid4D/Spectral3D/Spectral4D still available)

maximilian-gelbrecht:
Before you spend more time on adapting everything. What about the performance regression?

milankl:
> Before you spend more time on adapting everything. What about the performance regression?

It doesn't look that bad anymore, but I'm double checking with your suggestions now as well.

milankl:
Fable found the culprit:

Regression confirmed, root cause found, and fixed with a one-line change.

Confirmation (T31/T127, 8 layers, min times, reproducible across two rounds):

- T31 spec→grid: 65.8 µs (main) → 71.7 µs (branch), +9%
- T31 grid→spec: 69.1 → 73.5 µs, +6%
- T127 spec→grid: 1946 → 2047 µs, +5%
- Allocations byte-identical, as you observed.

Localization: benchmarking the stages separately showed the Fourier stages identical, but both Legendre stages slower (inverse 27.3→32.0 µs +17%, forward 37.1→41.2 µs +11%). Line-level profiles showed the extra time on the lines touching `legendre_polynomials` / `specs` inside the m-loop.

Root cause: the singleton dims do compile away everywhere — except one spot. The SpectralTransform constructor at spectral_transform.jl:204 built the `LowerTriangularArrayType` type parameter by hand with the old 4-parameter form:

```julia
LowerTriangularArray{NF, 2, array_type(architecture, NF, 2), typeof(spectrum)}
```

With the new 5th Dims parameter this is no longer concrete — it's a UnionAll (`... where Dims<:AbstractArrayDimensions`), which I spotted in the `@code_native` signature dump. So `S.legendre_polynomials` became an abstractly-typed field, and every access in the Legendre hot loop paid runtime type checks (no allocations, just checks — which is why allocation counts didn't catch it).

Fix: replaced that expression with `typeof(legendre_polynomials`). After the fix the branch matches main exactly: Legendre stages 27.2/37.1 µs, full transforms 65.7/69.3/1945.7/2500.9 µs. You updated the type-parameter comment in the struct accordingly.

One thing to be aware of: the same hand-written-type-parameter pattern existed in spectral_grid.jl, but those aliases were already updated with the 5th parameter on this branch, and the `zeros(LowerTriangularArray{...4 params...}, ...)` calls in truncation.jl are harmless (UnionAll dispatch, not a stored field).

milankl:
Note, the Legendre polynomials now have dimension LM+ (LM with unspecified additional dimension(s)) though technically should be LMY (LM x latitude) but I don't think this is important so let's leave it as is

```julia
julia> S.legendre_polynomials
560×24 (LM+) LowerTriangularArray{Float32, 2, Array{...}}
...
```

SpeedyWeather/src/dynamics/spectral_grid.jl
```diff
@@ -64,12 +64,21 @@ struct SpectralGrid{
...
 
-    "[DERIVED] Type of spectral variable in 4D (horizontal only + e.g. vertical and time, flattened into 3D array)"
+    "[DERIVED] Type of spectral variable in 3D (horizontal + vertical, flattened into 2D matrix)"
+    SpectralVariableXYZ::Type{<:AbstractArray}
+
+    "[DERIVED] Type of spectral variable in 3D (horizontal + time, flattened into 2D matrix)"
+    SpectralVariableXYT::Type{<:AbstractArray}
+
+    "[DERIVED] Type of spectral variable in 4D (horizontal + 2 unspecified, flattened into 3D array)"
     SpectralVariable4D::Type{<:AbstractArray}
```
maximilian-gelbrecht, line 77: Why do we need both `SpectralVariable4D` and `SpectralVariableXYZT`?

SpeedyWeather/src/dynamics/spectral_grid.jl
```diff
@@ -222,14 +240,20 @@ function SpectralGrid(
...
 
     # Spectral variable types
-    SpectralVariable2D = LowerTriangularArray{Complex{NF}, 1, array_type(architecture, Complex{NF}, 1), typeof(spectrum)}
-    SpectralVariable3D = LowerTriangularArray{Complex{NF}, 2, array_type(architecture, Complex{NF}, 2), typeof(spectrum)}
-    SpectralVariable4D = LowerTriangularArray{Complex{NF}, 3, array_type(architecture, Complex{NF}, 3), typeof(spectrum)}
+    SpectralVariable2D = LowerTriangularArray{Complex{NF}, 1, array_type(architecture, Complex{NF}, 1), typeof(spectrum), ArrayDimensions.LM}
+    SpectralVariable3D = LowerTriangularArray{Complex{NF}, 2, array_type(architecture, Complex{NF}, 2), typeof(spectrum), ArrayDimensions.LM}
+    SpectralVariableXYZ = LowerTriangularArray{Complex{NF}, 2, array_type(architecture, Complex{NF}, 2), typeof(spectrum), ArrayDimensions.LMZ}
+    SpectralVariableXYT = LowerTriangularArray{Complex{NF}, 2, array_type(architecture, Complex{NF}, 2), typeof(spectrum), ArrayDimensions.LMT}
+    SpectralVariable4D = LowerTriangularArray{Complex{NF}, 3, array_type(architecture, Complex{NF}, 3), typeof(spectrum), ArrayDimensions.LM}
```
maximilian-gelbrecht, line 247: As above. Do we need `4D` and `XYZT`?

SpeedyWeather/src/variables/dimensions.jl
```diff
@@ -127,18 +197,27 @@ allocate(::AbstractVariable{LocatorDim}, model::AbstractModel) = RingGrids.Anvil
...
 fuse_family(::Grid3D) = :grid
+fuse_family(::GridXYZ) = :grid
+fuse_family(::GridXYT) = :grid
 fuse_family(::Grid4D) = :grid
+fuse_family(::GridXYZT) = :grid
 fuse_family(::Spectral2D) = :spectral
 fuse_family(::Spectral3D) = :spectral
+fuse_family(::SpectralXYZ) = :spectral
+fuse_family(::SpectralXYT) = :spectral
 fuse_family(::Spectral4D) = :spectral
```
maximilian-gelbrecht, line 208: I might slightly revise this in a follow up. I think this new dimension systems allows to simplify some things of the fuse logic

SpeedyWeatherInternals/src/ArrayDimensions/ArrayDimensions.jl
```diff
@@ -0,0 +1,104 @@
...
+
+"""Dimension of LowerTriangularArray: 3D horizontal + time"""
+struct LMT <: ThreeDimensions end
+@inline Base.ndims(::LMT) = 2
+@inline Base.getindex(::LMT, ::Colon, s::Union{Integer, CartesianIndex}, args...) = LM()
+
+"""Dimension of LowerTriangularArray: 4D horizontal + vertical + time"""
+struct LMZT <: FourDimensions end
+@inline Base.ndims(::LMZT) = 3
+@inline Base.getindex(::LMZT, ::Colon, k::Union{Integer, CartesianIndex}, args...) = LMT()
```
maximilian-gelbrecht, line 64: Why was this actually previously type unstable and now is stable?

maximilian-gelbrecht (commented):
Again, I really like this. Just a few comments before merging.

SpeedyWeather/src/dynamics/spectral_grid.jl
milankl: Technically we don't but the `SpectralVariable4D` will yield LM+ (so 2 unspecified dimensions) but `SpectralVariableXYZT` will use LMZT. The latter is now encouraged, but when creating variables one can still say

```julia
variables(::MyComponent) = (
     DynamicsVariables(:a, Grid4D, ...) # only XY specified
     DynamicsVariables(:b, SpectralLMZT, ...) # all 4 dimensions specified
```

so in a way it's a backwards compatibility we don't make use of anymore but who knows maybe we do one day, so I'd just leave it in for now?

milankl: But good catch there's still some inconsistences in the definitions of Grid3D, Spectral4D etc

LowerTriangularArrays/src/LowerTriangularArrays.jl
```diff
@@ -31,4 +46,12 @@ include("lower_triangular_array.jl")
 include("rotate_reverse.jl")
 include("truncation.jl")
 
+# Trait-based dispatch on dimensions
+"""Type alias for all LowerTriangularArrays with a time dimension"""
+const LowerTriangularArrayWithTime = LowerTriangularArray{T, N, ArrayType, S, Dims} where {T, N, ArrayType, S, Dims <: DimensionsWithTime}
```
maximilian-gelbrecht, line 51: I like these trait based definitions, but this doesn't seem to be used anywhere at them moment, right?

milankl: no not used yet, but one could e.g. extend the `get_step` methods with it, which would simplify for `get_step(x) = x` in case field/lta doesn't have time

SpeedyWeatherInternals/src/ArrayDimensions/ArrayDimensions.jl
milankl: I think this was Claude making assumptions

SpeedyWeather/src/variables/dimensions.jl
milankl: Yes I agree it seems this can be adapted now, but indeed follow up

maximilian-gelbrecht (approved):

PR #1151, opened 2026-07-06
Reduce dynamic dispatches and revise Enzyme Rules to improve differentiability 

maximilian-gelbrecht:


SpeedyTransforms/src/fourier.jl
```diff
@@ -11,12 +11,12 @@ function _fourier!(f_north, f_south, field::AbstractField, S::SpectralTransform)
     end
 end
 
-function _fourier!(field::AbstractField, f_north, f_south, S::SpectralTransform)
+function _fourier!(field::AbstractField, f_north, f_south, S::SpectralTransform; add::Bool = false)
```
maximilian-gelbrecht, line 14: @milankl This PR reimplements some of the Enzyme rules to make it more stable, 1.12 compatible and also comptable with the chunked transform of the batched PRs. The only real change that also affects things outside of Enzyme is that I've added this `add` kwarg to `_fourier!` and `_legendre!` that allows to accumulate into memory (used by the adjoint rules)

maximilian-gelbrecht:
At the very least we restored compat with the most recent Enzyme version and Julia 1.10 here. Maybe also with 1.12? I hope the job actually finishes successfully and doesn't just time out.

SpeedyTransforms/ext/SpeedyTransformsEnzymeExt.jl
```diff
@@ -10,7 +10,15 @@ using SpeedyTransforms
...
-import SpeedyTransforms: _fourier!
+import SpeedyTransforms: _fourier!, wrapped_view
+
+# The spectral transform `S` is fixed geometry (Legendre polynomials, quadrature weights, FFT
+# plans) and is never a differentiation target. Marking it inactive stops Enzyme from building a
+# shadow of the large `SpectralTransform` aggregate when it is loaded out of a `Duplicated`
+# (mutable) model — the "cannot deduce type of copy" type-analysis failure on Julia >= 1.11 that
+# otherwise requires `Enzyme.API.maxtypeoffset!`. The transform! adjoint rules below read only
+# `S.val`, so treating `S` as a constant everywhere is consistent.
+EnzymeRules.inactive_type(::Type{<:SpectralTransform}) = true
```
milankl, line 21: just a question: Would that counteract the possibility to recompute the Legendre polynomials on the fly? The array entries would change then but they still depend on formulas and geometry only not on other data

SpeedyTransforms/src/fourier.jl
milankl: AH nice, yeah I think this would effectively mean that one can now also force temperature and humidity in spectral space which currently doesn't work because we directly transform into those LTAs and therefore overwrite anything that's in there before. But if I see this right then this only applies to grid -> spectral where this here is spectral -> grid?

SpeedyTransforms/src/fourier.jl
```diff
@@ -61,14 +61,17 @@ function _apply_batched_fft!(
...
     )
     nlayers = size(field, 2)        # number of vertical layers
     brfft_plan = S.brfft_plans[nlayers][j]  # plan pre-built for this K and this ring
     nfreq = nlon ÷ 2 + 1
 
     if not_equator  # skip FFT, redundant when north already did that latitude
-        view(field.data, ilons, :) .= brfft_plan * view_only_on_cpu(g_in, 1:nfreq, 1:nlayers, j)
+        dest = view(field.data, ilons, :)
+        rhs = brfft_plan * view_only_on_cpu(g_in, 1:nfreq, 1:nlayers, j)
+        add ? (dest .+= rhs) : (dest .= rhs)
```
milankl, line 74: the `?` has to be evaluated at runtime, is that a problem? I believe if `add` was somehow translated to a type this can be even compiled away?

milankl: Maybe not worth it ...

milankl: Just checking: The `add` wouldn't be available on GPU?

milankl (approved):
I don't know enough about the Enzyme rules, but otherwise this looks really good to me!!!

maximilian-gelbrecht:
I used the oppurtunity here now also to reduce some dynamic dispatches that were happening (as they can trigger Enzyme errors as well). The `BarotropicModel` is now completely free of them, the `PrimitiveWetModel` still has a few, that I will look into in a follow up.

SpeedyTransforms/ext/SpeedyTransformsEnzymeExt.jl
maximilian-gelbrecht: I don't think so. This just means we can't differentiate wrt parameters of the transform which is completely fine.

SpeedyTransforms/src/fourier.jl
maximilian-gelbrecht: I think things like this are compiled away, but I will run a benchmark here on this branch before merging to make sure everything is still in order

maximilian-gelbrecht: > Just checking: The `add` wouldn't be available on GPU?

Good shoutout, there was actually also still a GPU error here

SpeedyWeather/src/dynamics/scaling.jl
```diff
@@ -32,25 +32,30 @@ function scale_tendencies!(vars::Variables, model::AbstractModel)
     haskey(vars.fused, :grid_tendencies) &&
         scale!(get_tendency_step(parent(vars.fused.grid_tendencies), TS, DummyParameterization()), scale)
 
-    # Scale the standalone (non-fused) tendencies individually. 
-    # spectral
-    for varname in tendency_names(vars)
```
maximilian-gelbrecht, line 37: @milankl So apparently the compiled-time names alone weren't enough to mitigate dynamic dispatches. We have no `@generated` functions for most of these loops instead

RingGrids/src/field.jl
```diff
@@ -584,7 +584,7 @@ function fields_match(
         horizontal_only::Bool = false,
         vertical_only::Bool = false,
     )
-    @assert ~(horizontal_only && vertical_only) "Conflicting options: horizontal_only = $horizontal_only and vertical_only = $vertical_only"
```
maximilian-gelbrecht, line 587: this interpolation here seemingly caused a dynamic dispatch according to `JET`

SpeedyTransforms/src/fourier.jl
```diff
@@ -294,29 +306,32 @@ function plan_FFTs!(
...
-
+    nfreqj(j) = nlons[j] ÷ 2 + 1
+
+    # SERIAL (K=1, 1-D) plans — always built. Comprehensions give concretely-typed `Vector`s.
+    rfft_plan_serial = [AbstractFFTs.plan_rfft(view_only_on_cpu(fake_grid_data.data, rings[j], 1), 1)
+                        for j in 1:nlat_half]
+    brfft_plan_serial = [AbstractFFTs.plan_brfft(view_only_on_cpu(scratch_memory_north, 1:nfreqj(j), 1, j), nlons[j], 1)
+                         for j in 1:nlat_half]
+
+    # BATCHED (K>1, 2-D) plans — a `Dict{Int, Vector{P2}}` keyed by K with CONCRETE plan type P2.
```
maximilian-gelbrecht, line 317: apparently serial and batched FFT plans are not of the same type in Julia... 

So this had caused a minor type instability and dynamic dispatch as well

SpeedyWeather/src/time_stepping/steppers/leapfrog.jl
```diff
@@ -79,28 +79,33 @@ tendency_steps(::AbstractLeapfrog) = 1
...
-        var_old, var_new = get_steps(getfield(prognostic.ocean, varname))
-        var_new .= var_old
+    # but statically compiled into the *_names generators shared with the drivers
+    initialize_steps_unrolled!(vars)
+    return nothing
+end
+
+# copy step 1 -> step 2 for one variable; steps bound individually via get_step
+@inline function copy_step_forward!(var)
+    var_old = get_step(var, 1)
```
maximilian-gelbrecht, line 92: I encountered this problem before: the `get_steps` isn't fully type stable because it's a runtime value `NTuple`

SpeedyWeather/test/dynamics/dispatch.jl
maximilian-gelbrecht, line 1: Now all tested in the CI :)

maximilian-gelbrecht:
The original aim of the PR was to restore Enzyme compatibility with the current version and enable it for 1.12. The former it also did, the latter unfortunately not, even though I also made some progress with that. 

One thing I also notice here is that Enzyme with 1.10 is now compiling the gradient much faster than before. Many of the CI jobs just take 15 or 20 minutes instead of 45 minutes before. 

Still outstanding are some benchmarks. I'll push those later.

SpeedyWeather/src/dynamics/horizontal_diffusion.jl
```diff
@@ -209,9 +211,9 @@ function horizontal_diffusion!(             # explicit version
         tendency::LowerTriangularArray,     # tendency of var
         var::LowerTriangularArray,          # spectral horizontal field to diffuse
         expl::AbstractMatrix,               # explicit spectral damping (lmax x nlayers matrix)
-        impl::Nothing,                      # pass on nothing to dispatch to explicit diffusion                         
+        impl::Nothing,                      # pass on nothing to dispatch to explicit diffusion
     )
-    lmax, mmax = size(tendency, OneBased, as = Matrix)
+    lmax, mmax = size(tendency, OneBased, Matrix)    # positional size, see implicit version above
```
maximilian-gelbrecht, line 216: apparently the compiler just sometimes gives up on things like this deep down the model and `JET` flags it as a dynamic dispatch when using it with `kwargs`

maximilian-gelbrecht:
Benchmarks look good and also show a small speed up thanks to the reduced dynamic dispatches. I'll merge this once all pass (the GPU CI fail isn't related to this PR, and it passed before)

maximilian-gelbrecht:
@nviebig @SieglStefan I merge this PR a bit later. With the updated Enzyme version and these fixes here, the gradient compile time seems to be reduced by a lot. I'm seeing roughly ~half here. So this should be pretty helpful for you.

PR #1152, opened 2026-07-08
Allocation-free copy for Terrarium extension

maximilian-gelbrecht:


SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -129,9 +143,12 @@ function TerrariumLand(
     # as the layer thickness for the SpeedyWeather LandGeometry. It's not actually used, but
     # we set it here for consistency.
     geometry = LandGeometry(1, NF[Δz_arr[end]])
+    # `unmasked_indices` treats `true` as masked-out; `model.grid.mask` is `true` at land
```
milankl, line 146: Oh no, why is for a land model the `model.grid.mask` true over land? 🙈 😉

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
```diff
@@ -129,9 +143,12 @@ function TerrariumLand(
...
+    # points, so invert it to get the indices of the (unmasked) land columns.
+    mask_indices = RingGrids.unmasked_indices(.!model.grid.mask)
```
milankl, line 148: Just double checking because I don't think this is tested in RingGrids, this preservies the array type?

milankl (approved):
Yeah nice!!!

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: On the one hand the word "mask" implies that true is masked out to me as well. But on the other hand it's just so practical to just do a `array[mask]` to get all the points that are not masked out. 

We are definitely not the only ones misusing this / or rather being inconsistent about this.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: It doesn't preserve that it's a `Field` but that's good. There are some problems with indexing with `Field` having a direct array and not a wrapper type is preferable. The array type within `Field` it does preserve.

bgroenks96 (approved):

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
bgroenks96: :upside_down_face: 

I guess we should actually change that... but I agree with Max that the logical definition of "mask" is quite annoying to work with outside of bitmasks.

milankl: You could call it an ocean mask, because as a land model you want to mask the ocean. Or call it land fraction, as we do now in SpeedyWeather, sure you want it boolean and not fractional but that's less inaccurate at least?

milankl: Just an idea for the future, this is not to prevent this PR from being merged of course

maximilian-gelbrecht: Yeah, I'll merge now. But

@bgroenks96 I like Milan's suggestions as it would be non-breaking as well.

PR #1155, opened 2026-07-14
Terrarium output (and more flexible output definitions) 

maximilian-gelbrecht:
This provides output writers for Terrarium. For their usage, see the edited documentation. 

I had to make a few changes to the `define_variables!` to be a bit more flexible, so that it also allows for different vertical dimensions. Aside from that all changes are in the Terrarium extension that is also slightly reorganised.


SpeedyWeather/ext/SpeedyWeatherTerrariumExt/output.jl
```diff
@@ -0,0 +1,240 @@
...
+# simulations. Constructor methods are added to the `SpeedyWeather.TerrariumOutput`
+# function stub; they build `TerrariumOutputVariable`s with name, units, long name
+# and dimensionality derived from Terrarium's variable descriptors.
+
+import SpeedyWeather: TerrariumOutput, AbstractOutput, AbstractSimulation,
+    path, path_or_nothing, output!, write_array!, hastime, is3D,
+    vertical_dimension, get_nlayers, define_dimension!, get_dimension_length, define_coordinate!
+
+# name of the vertical output dimension shared by all 3D Terrarium output variables
+const SOIL_DEPTH_DIM_NAME = "soil_depth"
```
maximilian-gelbrecht, line 11: Currently we only have one vertical dimension in Terrarium but that will change. 

I only implemented it with that dimension in mind, once we know how different vertical dimensions look like it should be easy to adjust (probably just mirroring what we already do in SpeedyWeather for land/atmospheric output, because in the end it will still only be two or three different vertical dimensions in Terrarium)

SpeedyWeather/ext/SpeedyWeatherTerrariumExt/output.jl
```diff
@@ -0,0 +1,240 @@
...
+        variable.scratch_grid = is3D(variable) ?
+            RingGrids.Field(output_NF, grid, variable.nlayers) : RingGrids.Field(output_NF, grid)
+        fill!(variable.scratch_grid, output_NF(variable.missing_value))
+        if is3D(variable)
+            variable.scratch_output = RingGrids.Field(output_NF, output.field2D.grid, variable.nlayers)
+        end
+    end
+
+    # gather the land columns on CPU, drop the singleton y-dimension,
+    # then scatter onto the (land points of the) full ring grid
```
maximilian-gelbrecht, line 181: This here below is actually the main output computation. Currently it's all done on CPU because that's also how we do it in SpeedyWeather. We should probably eventually change that to do the interpolation still on GPU and only then transfer to CPU. But I would keep that like this for now here, and do change that when it's also changed in the main code.

bgroenks96 (approved):
Looks great! No comments.

PR #1156, opened 2026-07-14
Add a neutral wind model

gregrmunday:
The `neutral_wind` is useful for a number of calculations involving the surface fluxes (including the surface roughness, see the [IFS docs, Chapter 3](https://www.ecmwf.int/en/elibrary/81734-ifs-documentation-cy50r1-part-4-physical-processes)) because it doesn't include effects local thermal instability in the boundary layer.

Here I've fit a neural network followed by a symbolic regression model that shortcuts a lot of expensive physics involving calculating stability gradients in the boundary layer and the Obukhov length. Instead, this produces the `neutral_wind` from the `surface_wind_speed` and local thermal conditions. 

neutral_wind


maximilian-gelbrecht (commented):
Really cool! 

I am curious about some details of how you trained it, but I guess you'll also include this in an upcoming manuscript, right? 

In terms of the implementation, I think it would be better to implement this directy as an `AbstractParameterization` according to our interface for that.

SpeedyWeather/src/parameterizations/surface_fluxes/surface_condition.jl
```diff
@@ -13,18 +13,23 @@ atmospheric relationships. Fields are $(TYPEDFIELDS)"""
 
     "[OPTION] Wind speed of sub-grid scale gusts [m/s]"
     gust_speed::NF = 5
+
+    "[OPTION] Calculate the neutral wind speed [m/s]"
+    neutral_wind::Bool = true
```
maximilian-gelbrecht: With the neutral wind as a `AbstractParameterization` this would just be `neutral_wind::NW = NeutralWind(spectral_grid)` and to turn it off, you'd just do `neutral_wind = nothing`

gregrmunday:
> Really cool!
> 
> I am curious about some details of how you trained it, but I guess you'll also include this in an upcoming manuscript, right?
> 
> In terms of the implementation, I think it would be better to implement this directy as an `AbstractParameterization` according to our interface for that.

Thanks Max! Yes indeed, I'll share a draft with you and Milan hopefully by the end of the week :)

I'll have a go doing as you suggest here!

SpeedyWeather/src/parameterizations/surface_fluxes/surface_condition.jl
gregrmunday: I've given this a go - it runs but would be great if you could see that it's been set up right!

milankl: I think there's 3 options

1. do as part of `SurfaceCondition`
2. make it part of `model.boundary_layer`
3. let it have it's own `model.neutral_wind` parameterization

Which one to chose probably should scale with complexity and modularity. If you bake it into `SurfaceCondition` then you can't easily modify it beyond a on/off; putting it into 2 follows a philosophy where we'd try to group parameterizations, 3 would be more flat. And I think that's the question I want to ask: Should we have a flat structure with many `model.parameterization` directly reachable as that or should we attempt to group like `model.group.parameterization`. The latter requires more boiler plate code and you need to set `neutral_wind = nothing` when building a custom boundary layer scheme and then pass that into the model. The former might just end up having many parameterizations as a long list. We currently group

- land (temperature, soil moisture, vegetation, ...)
- long wave radiation (transmissivity, transfer)
- short wave radiation (clouds, transmissivity, transfer)
- boundary layer (roughness, drag)

So what are people's thoughts where this would fit in best?

maximilian-gelbrecht: My preference would be grouping in this case to keep it a bit ordered.

milankl: Also my preference I think

milankl: Then I would suggest to move the neutral wind calculation into the `model.boundary_layer` and extend this to cover

- neutral wind
- surface_roughness
- drag
- surface_condition (currently separate but I don't think it should be)

Hope the order is right, but you can decide which order makes the most sense

gregrmunday: Makes sense to me! I'll have a go

gregrmunday:
Alright it seems to be working correctly... is this what you both had in mind? @milankl @maximilian-gelbrecht 

The only thing is the pretty printing for `model.boundary_layer` doesn't seem to be working correctly and I'm not sure why!

SpeedyWeather/src/parameterizations/surface_fluxes/surface_condition.jl
gregrmunday: And I think the order is correct.

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
```diff
@@ -146,3 +167,38 @@ For vertical stability in the boundary layer."""
...
+    return neutral_wind_speed(ij, vars, scheme, model)
+end
+
+"""Ocean-based neutral wind speed calculation from actual wind speed, 
+derived from ERA5 data via symbolic regression."""
+@propagate_inbounds function neutral_wind_speed(ij, vars, scheme::NeutralWindSpeed{NF}, model) where {NF}
+    (; surface_wind_speed) = vars.parameterizations
+    (; surface_air_temperature) = vars.parameterizations
+
+    c1 = NF(-0.039317116)
```
maximilian-gelbrecht: I would still put the parameters in the `NeutralWindSpeed` struct as in your earlier version. Maybe we could even fine-tune/calibrate them online later. 

This would also resolve the issue that makes the GPU CI fail if I see it correctly. The GPU CI is failing because `adapt_structure` apparently doesn't really know how to deal with empty structs in this case.

gregrmunday: Ah I see ok, I'll move them in then

gregrmunday: Done!

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
```diff
@@ -1,7 +1,11 @@
 abstract type AbstractBoundaryLayer <: AbstractParameterization end
 
 variables(::AbstractBoundaryLayer) = (
+    ParameterizationVariable(:neutral_wind_speed, Grid2D(), desc = "Neutral surface wind speed", units = "m/s"),
```
maximilian-gelbrecht:
```suggestion
```

I would not make this a part of the `AbstractBoundaryLayer` because the parametererization might also just be `nothing`, as below the concrete `variables(BL::BoundaryLayer)` is given, this one is used anyway

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
```diff
@@ -146,3 +178,28 @@ For vertical stability in the boundary layer."""
...
+
+    sst = vars.prognostic.ocean.sea_surface_temperature[ij]
+    t_diff = surface_air_temperature[ij] - sst # TODO: replace SST with ocean skin temperature
+    ws_safe = max(surface_wind_speed[ij], NF(1.0e-6))
+    log_arg = max(nw.c1 * ws_safe * t_diff + exp(t_diff), NF(1.0e-8))
+
+    numerator = 2 * t_diff + nw.c8 * exp(t_diff) - nw.c3 * (nw.c4^surface_air_temperature[ij])
+    denominator = t_diff * (log(log_arg) + nw.c2) + nw.c5 * (nw.c6^ws_safe) + nw.c9 * (ws_safe^nw.c7) + ws_safe
+
+    vars.parameterizations.neutral_wind_speed[ij] = max(surface_wind_speed[ij] - (numerator / denominator), 0)
```
maximilian-gelbrecht, line 239: I am a bit confused now in this version, where is the `neutral_wind_speed` actually used? I don't see it.

maximilian-gelbrecht: If it's not actually used at the moment, I think I wouldn't include the parameterization as a default.

gregrmunday: It's not actually used in this PR but it's going to be used in an upcoming surface roughness PR as the main predictor for ocean surface roughness :) 

I just split this into its own PR to make it simpler

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
gregrmunday: I think following Milan's comment above:

> Then I would suggest to move the neutral wind calculation into the model.boundary_layer and extend this to cover
> 
> neutral wind
> surface_roughness
> drag
> surface_condition (currently separate but I don't think it should be)

It kind of makes sense to have it here, because the rest of the boundary layer parameterisations will all depend on the neutral wind (after the surface roughness PR!) Unless I'm misunderstanding what you're saying

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
maximilian-gelbrecht: Ah yes. Yeah, smaller PRs are always better to review :). Okay, I think to be consistent, I would still prefer to set the default here in the `BoundaryLayer` to `nothing` then though and then you can just turn it on in the surface roughness PR.

gregrmunday: Sounds good, I've done that in the latest and hopefully fixed the tests...

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
maximilian-gelbrecht: We have two different variable declarations here. One for the concrete `BoundaryLayer` type and one for the `AbstractBoundaryLayer`. Currently the `AbstractBoundaryLayer` is nowhere used because with multiple dispatch we will always hit the `variables(::BoundaryLayer)` below that you also modified. 

The definition for  `AbstractBoundaryLayer` is more a convenience thing for any further/additional kinds of boundary layers that other people might implement. I would think of the variables in those kind of definitions therefore as the absolute minimum that every boundary layer parameterisation should have. And I am not sure every single boundary layer parameterisation would need the `neutral_wind` (as we also didn't have it so far)>

gregrmunday: I see ok - I'll remove it from here!

gregrmunday: Done in latest, just waiting on tests now!

maximilian-gelbrecht (approved):
Great! From my side this can be merged once all tests pass

gregrmunday:
@milankl are you happy with this? If so I can merge!

docs/src/surface_fluxes.md
```diff
@@ -22,11 +22,11 @@ subtypes(SpeedyWeather.AbstractSurfaceMomentumFlux)
...
 !!! note "Interdependence of surface flux computations"
-    `model.surface_condition` extrapolates atmospheric variables (wind, temperature, humidity)
+    `model.boundary_layer.surface_condition` extrapolates atmospheric variables (wind, temperature, humidity)
     to the surface. `model.boundary_layer` computes centrally a drag coefficient for surface
-    momentum, heat and humidity fluxes. Setting `model.surface_condition = nothing` or
-    `model.boundar_layer = nothing` will therefore disable all other surface fluxes unless
-    those are have `use_boundar_layer_drag = false` which lets them use independently their
+    momentum, heat and humidity fluxes. Setting `model.boundary_layer.surface_condition = nothing` or
+    `model.boundary_layer = nothing` will therefore disable all other surface fluxes unless
+    those are have `use_boundary_layer_drag = false` which lets them use independently their
```
milankl: typo

docs/src/surface_fluxes.md
```diff
@@ -22,11 +22,11 @@ subtypes(SpeedyWeather.AbstractSurfaceMomentumFlux)
...
 
 !!! note "Interdependence of surface flux computations"
-    `model.surface_condition` extrapolates atmospheric variables (wind, temperature, humidity)
+    `model.boundary_layer.surface_condition` extrapolates atmospheric variables (wind, temperature, humidity)
     to the surface. `model.boundary_layer` computes centrally a drag coefficient for surface
-    momentum, heat and humidity fluxes. Setting `model.surface_condition = nothing` or
-    `model.boundar_layer = nothing` will therefore disable all other surface fluxes unless
-    those are have `use_boundar_layer_drag = false` which lets them use independently their
+    momentum, heat and humidity fluxes. Setting `model.boundary_layer.surface_condition = nothing` or
+    `model.boundary_layer = nothing` will therefore disable all other surface fluxes unless
```
milankl, line 28: do we test that these `= nothing` actually work? Like if the defaults define variables that other components need then it might not ...

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
```diff
@@ -22,39 +25,67 @@ initialize!(::ConstantDrag, ::PrimitiveEquation) = nothing
...
+    # Parameters for neutral wind calculation
+    @param c1::NF = NF(-0.039317116)
+    @param c2::NF = NF(-2.9858496)
+    @param c3::NF = NF(2.0046231e-10)
+    @param c4::NF = NF(1.0768474)
+    @param c5::NF = NF(0.20268184)
+    @param c6::NF = NF(1.2684147)
+    @param c7::NF = NF(-0.94933933)
+    @param c8::NF = NF(0.041551278)
+    @param c9::NF = NF(5.8649142)
```
milankl: the `NF(...)` is unnecessary

```julia
julia> @kwdef struct A{T}
           a::T = 1e-1
       end

julia> A{Float32}()
A{Float32}(0.1f0)
```

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
```diff
@@ -22,39 +25,67 @@ initialize!(::ConstantDrag, ::PrimitiveEquation) = nothing
...
+
 export BoundaryLayer
 """Composite type, containing surface roughness computation
 and drag coefficient computation. Fields are $(TYPEDFIELDS)"""
-@parameterized @kwdef struct BoundaryLayer{SR, D} <: AbstractBoundaryLayer
+@parameterized @kwdef struct BoundaryLayer{SR, D, NW, SC} <: AbstractBoundaryLayer
+    @component neutral_wind_speed::NW
     @component surface_roughness::SR
     @component drag::D
+    @component surface_condition::SC
```
milankl: change order to
1. surface condition
2. neutral wind
3. roughness
4. drag

doesn't matter here but for consistency still

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
```diff
@@ -22,39 +25,67 @@ initialize!(::ConstantDrag, ::PrimitiveEquation) = nothing
     return nothing
 end
 
+export NeutralWindSpeed
+@parameterized @kwdef struct NeutralWindSpeed{NF} <: AbstractBoundaryLayer
```
milankl: add docstring to explain what these mean

milankl (approved):

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
gregrmunday: done in latest

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
gregrmunday: done in latest

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
gregrmunday: done in latest

docs/src/surface_fluxes.md
gregrmunday: fixed

docs/src/surface_fluxes.md
gregrmunday: They both seem to work!

gregrmunday:
Tests pass, I shall merge :0

PR #1157, opened 2026-07-15
Terrarium 0.1.4 compat

maximilian-gelbrecht:
Adjusts the Terrarium extension to the changes that put the timestepper into the model


maximilian-gelbrecht:
I realised this is actually the change needed for the version we haven't released yet. So this will stay open until then.

bgroenks96:
v0.1.4 is now released

PR #1158, opened 2026-07-15
Reduce CI time of vertical coordinates tests 

maximilian-gelbrecht:
cc @milankl


maximilian-gelbrecht:
Ok it helped to massively reduce CI time. It's still 8 min here in 1.12, but that's just 25% of what it was before, so that's arlready great

milankl:
Ah yeah fantastic!!

PR #1159, opened 2026-07-15
Terrarium coupling: improve mask consistency

maximilian-gelbrecht:
This PR 

* Sets proper fallback values over the ocean for land variables in the Terrarium coupling, to mitigate issues at the coastline when interpolating in the output (reported e.g. in https://github.com/NumericalEarth/Terrarium.jl/issues/139) 
* It also explicitly checks whether the land sea masks of SpeedyWeather and Terrarium are consistent when constructing the coupled model


SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
bgroenks96, line 38: I just noticed we're missing the `initializers` here. Without that, there is no way for the user to customize the land initialization.

bgroenks96 (changes requested):
One small issue unrelated to the changes in this PR, but otherwise looks good.

SpeedyWeather/ext/SpeedyWeatherTerrariumExt.jl
maximilian-gelbrecht: This confused me briefly, but no, it's not missing. This is the `initialize` of the `StateVariables` that in this version is still called `initialize` as we haven't released a new version yet.

maximilian-gelbrecht: The `initializers` are used when constructing the `ModelIntegrator` in `SpeedyWeather.initialize!(vars::Variables, land::AbstractTerrariumLandModel, model::PrimitiveWetModel,)`

bgroenks96 (approved):

PR #1160, opened 2026-07-16
Fix ZarrOutput ensembles

maximilian-gelbrecht:
A Claude guided attempt to fix issues with the `ZarrOutput` with ensembles as brought up by @bgroenks96 . What do you think?

### DIFF
--- a/SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
+++ b/SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
@@ -101,8 +101,9 @@ one store; member 1 (the *creator*) builds the store schema, coordinates and the
 `time` axis, then drops a readiness marker; members `2..ensemble_size` (the *writers*) wait
 for that marker, open the existing store and write only their own ensemble slice.
 Writer members never create Zarr arrays or coordinates (concurrent metadata writes from
-parallel processes would corrupt the store) and write their side files (parameters txt,
-progress txt, restart file) under member-specific filenames. When rerunning an ensemble
+parallel processes would corrupt the store). Every member (creator included) writes its
+side files (parameters txt, progress txt, restart file) under member-specific filenames
+(`_member\$index` suffix) so they don't clobber each other. When rerunning an ensemble
 into the same run folder with `overwrite=true`, start the creator before (or together
 with) the writers: a writer launched against a leftover store from a previous run cannot
 distinguish its readiness marker from the current run's."""
@@ -135,9 +136,10 @@ function initialize!(
     if ensemble
         setup_ensemble_run_folder!(output)
         # all members share one run folder but run as independent parallel processes:
-        # give the side-file callbacks (parameters/progress txt, restart file) of writer
-        # members unique filenames so they don't clobber each other's files
-        output.ensemble_index != 1 && set_ensemble_member_filenames!(model, output.ensemble_index)
+        # give the side-file callbacks (parameters/progress txt, restart file) of every
+        # member — including the creator (member 1) — unique filenames so they don't
+        # clobber each other's files and share one consistent naming scheme
+        set_ensemble_member_filenames!(model, output.ensemble_index)
     end
 
     # Total number of output snapshots: IC + one per `output_every_n_steps`.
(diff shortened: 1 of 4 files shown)


maximilian-gelbrecht:
Test script for a local multi threaded test (can't run it in the CI because it's not multi threaded) 

```Julia 
# Run a local ZarrOutput ensemble: N members run concurrently as Julia threads,
# all writing into one shared Zarr store (one ensemble slice each).
#
# Member 1 (the creator) builds the store schema, coordinates and the shared time
# axis, then drops a readiness marker; members 2..N wait for that marker inside
# run! and then only write their own ensemble slice — exactly the same protocol
# as independent parallel jobs on a cluster, just emulated with threads here.
#
# Usage:
#   julia --project=SpeedyWeather -t auto run_zarr_ensemble_threads.jl
#
# (works with a single thread too — waiting members yield while polling — but
# with ≥ 2 threads the members actually run in parallel)

using Zarr
using SpeedyWeather
using SpeedyWeather.Dates: Dates, Day, now

ensemble_size = 4
period = Day(5)
trunc = 31
output_path = mkpath(joinpath(@__DIR__, "ensemble_output"))

# a fresh, deterministic run id per invocation: all members must resolve to the
# same run folder, but reusing a folder from a previous invocation would race
# the previous run's leftover readiness marker
run_id = Dates.format(now(), "yyyymmdd-HHMMSS")

spectral_grid = SpectralGrid(; trunc, nlayers = 1)

# Build and initialize all members serially (model setup, e.g. FFT planning, is
# not guaranteed thread-safe), then run them concurrently below. The Zarr store
# itself is only created inside run! when the output is initialized.
simulations = map(1:ensemble_size) do member
    output = ZarrOutput(
        spectral_grid, ShallowWater;
        path = output_path, id = run_id,
        ensemble_index = member, ensemble_size = ensemble_size,
    )
    # ensemble spread: Galewsky et al. (2004) zonal jet, barotropically unstable,
    # with the initial height perturbation at a different longitude per member
    initial_conditions = ZonalJet(spectral_grid, perturb_lon = 270 + 5member)
    model = ShallowWaterModel(spectral_grid; output, initial_conditions)
    return initialize!(model)
end

# run all members concurrently; wait rethrows if any member fails
tasks = map(simulations) do simulation
    Threads.@spawn run!(simulation, output = true; period)
end
foreach(wait, tasks)

output = simulations[1].model.output
store_path = joinpath(output.run_path, output.filename)
println("Ensemble store written to: $store_path")

# quick sanity check: every ensemble slice holds finite data
g = Zarr.zopen(store_path)
vor = g["vor"]
println("vor dimensions (row-major): ", vor.attrs["_ARRAY_DIMENSIONS"], ", size ", size(vor))
for member in 1:ensemble_size
    slice = vor[:, :, :, :, member]
    println("member $member: all finite = ", all(isfinite, slice))
end

# open in Python with:  xarray.open_zarr("$store_path")
```

bgroenks96:
This will not work with multi-processing because there are issues with trying to serialize Speedy models/simulations via `pmap`. To really test this, you should do the full simulation setup on each worker, not serially.

bgroenks96 (approved):

bgroenks96:
Approved but I would add tests with `Distributed`.

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
```diff
@@ -235,11 +270,73 @@ function write_zarr_coordinates!(g::Zarr.ZGroup, output::ZarrOutput, model::Abst
...
 shared store schema, coordinates and time axis are in place."""
 ensemble_marker_path(output::ZarrOutput) = joinpath(output.run_path, ".zarr_ensemble_ready")
 
+"""$(TYPEDSIGNATURES)
+Give the side-file callbacks added by the output initialization (parameters txt,
+progress txt, restart file) of an ensemble writer `member` unique filenames with a
+`_member\$member` suffix, e.g. `progress_member2.txt`. All ensemble members share one
+run folder but run as independent parallel processes; with the default filenames every
+member would concurrently write the same `parameters.txt`, `progress.txt` and
+`restart.jld2`, garbling or corrupting them. The creator (member 1) keeps the defaults."""
```
bgroenks96: I do think it's a bit weird that the creator keeps the default names. Isn't it an ensemble member also?

maximilian-gelbrecht:
> Approved but I would add tests with `Distributed`.

Would those work in GitHub CI?

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
maximilian-gelbrecht: Agreed, I'll change that

bgroenks96:
I don't see why not?

maximilian-gelbrecht:
I thought it's just a single process and you might not be able to spawn more, but I also never tried.

bgroenks96:
It's probably a single core, so adding more processes will be slower, but I would be surprised if they somehow block `addprocs` from adding more processes.

maximilian-gelbrecht:
Ok, I added a multiprocessing test

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
maximilian-gelbrecht: done

PR #1162, opened 2026-07-21
Add buffering of time chunks in `ZarrOutputWriter`

### LINKED ISSUE #1161: Degraded performance of `ZarrOutputWriter` when `time_chunk` > 1
There are two issues worth documenting here:

1. Defaulting to `time_chunk = 1` is actually not a very good idea. Even for only modest simulation lengths with subdaily output frequency, the number of files created by ZARR quickly climbs to the order of tens or hundreds of thousands. This is especially problematic on networked file systems where large numbers of file pointers result in extremely high I/O overhead. We should probably set the default `time_chunk` automatically based on the user-specified output interval and simulation length.
2. The current implementation of `ZarrOutputWriter` is extremely inefficient with chunking (see benchmark script below) and, as a result, simulation speed degrades roughly linearly as `time_chunk` increases.

The following is a summary of the benchmark findings generated by Claude 🤖.

# ZarrOutput `time_chunk` performance degradation

Date: 2026-07-21

## Summary

Setting `time_chunk > 1` on `ZarrOutput` substantially degrades simulation
runtime, with the slowdown scaling roughly linearly with the chunk size. The
cost is entirely in the output write path — the dynamical core does identical
work regardless of `time_chunk`. The benchmark that produced these numbers is
[`scratch/benchmark_zarr_time_chunk.jl`](benchmark_zarr_time_chunk.jl).

## Root cause

A Zarr chunk is the atomic unit of (de)compression and I/O. The writer emits one
time slice per output step in `write_array!`
([`ext/SpeedyWeatherZarrExt.jl:491-503`](../ext/SpeedyWeatherZarrExt.jl)):

```julia
i = hastime(variable) ? output.output_counter : 1
indices = get_indices(i, variable, output.ensemble_index)
z[indices...] = parent_array(field)
```

When `time_chunk = N > 1`, a single time slice lands inside a chunk that spans
`N` steps. Writing that slice forces Zarr to perform a **read-modify-write** of
the whole chunk on *every* output step:

1. read the chunk from disk,
2. decompress it,
3. patch the one new time slice,
4. recompress it,
5. write it back.

The same chunk is thus rewritten up to `N` times over the run. With
`time_chunk = 1` each step writes its own fresh chunk (write-only, no
decompress/read), so no read-modify-write occurs.

## Benchmark configuration

- Model: `PrimitiveWetModel`, T31, 8 vertical layers
- Period: 30 days (~one month), hourly output => 721 output steps
- Small spectral truncation deliberately keeps runtime dominated by output I/O so
  the write-path cost is isolated
- JIT compilation excluded via an unmeasured warmup run before timing `run!`
- Each run uses a fresh temp output directory that is removed afterwards

## Results

T31, 8 layers, 30 days, hourly output (721 steps):

| time_chunk | run! elapsed (s) | vs tc=1 |
|-----------:|-----------------:|--------:|
| 1 (baseline) | 20.33 | 1.00x |
| 6  | 22.39 | 1.10x |
| 12 | 25.02 | 1.23x |
| 24 | 30.98 | 1.52x |
| 48 | 42.23 | 2.08x |
| 168 | 104.14 | 5.12x |

The slowdown grows monotonically — approximately linearly — with `time_chunk`,
matching the read-modify-write prediction (each chunk rewritten ~`N` times). At
`time_chunk = 168` (weekly chunks) the run is >5x slower doing identical physics.

## Caveats

- These factors are close to the pure I/O penalty because the T31 core is cheap;
  at higher resolution the absolute overhead grows but its ratio to the
  dynamical core shrinks.
- `time_chunk` values that do not divide `n_outputs = 721` (6, 48, 168) leave a
  partial final chunk, which does not affect the trend.

## Suggested fix

Buffer `time_chunk` time steps in memory and flush a full chunk in a single
write (write-only, no read-modify-write), rather than writing one time slice per
step in `write_array!`. This makes `time_chunk > 1` cheap while retaining the
larger on-disk chunking users typically want for read performance.

## Benchmark script

```julia
# MWE: benchmark SpeedyWeather simulation time for ZarrOutput as a function of
# the `time_chunk` size, with hourly output.
#
# Hypothesis: with `time_chunk > 1`, each output step writes a single time slice
# into a chunk that spans `time_chunk` steps. Because a Zarr chunk is the atomic
# unit of (de)compression and I/O, writing one slice forces a read-modify-write
# of the whole chunk on *every* output step (read + decompress -> patch one slice
# -> recompress + write). With `time_chunk = 1` each step writes its own chunk and
# no read-modify-write occurs. We therefore expect runtime to grow substantially
# with `time_chunk` even though the amount of physics is identical.
#
# Run from the SpeedyWeather package directory with an environment that has both
# SpeedyWeather and Zarr available:
#
#   julia --project=<env-with-SpeedyWeather+Zarr> benchmark_zarr_time_chunk.jl

using SpeedyWeather
using Zarr
using Dates
using Printf

# ---------------------------------------------------------------------------
# Configuration: hourly output over a one-month period so many output steps are
# written; small spectral truncation so runtime is dominated by output I/O, not
# by the dynamical core (this isolates the write path we want to measure).
# ---------------------------------------------------------------------------
const TRUNC       = 31                    # spectral resolution (T31)
const NLAYERS     = 8                      # vertical layers
const PERIOD      = Day(30)                # simulation length (~one month)
const INTERVAL    = Hour(1)               # hourly output => 721 output steps
const TIME_CHUNKS = [1, 6, 12, 24, 48, 168]  # sweep; 1 is the write-per-chunk baseline

"""
Run one simulation writing ZarrOutput with the given `time_chunk` and return the
elapsed wall-clock time for `run!` in seconds. A fresh temp output directory is
used per run and removed afterwards so no store is reused or left on disk.
"""
function benchmark_time_chunk(time_chunk::Int; trunc = TRUNC, nlayers = NLAYERS,
                              period = PERIOD, interval = INTERVAL)
    tmp_output_path = mktempdir(; prefix = "zarr_bench_tc$(time_chunk)_")
    try
        spectral_grid = SpectralGrid(; trunc, nlayers)
        output = ZarrOutput(
            spectral_grid, PrimitiveWet;
            path = tmp_output_path,
            interval,
            time_chunk,
            write_restart = false,
            write_parameters_txt = false,
            write_progress_txt = false,
        )

        # @elapsed measures the second run of a given configuration in `main`
        # (a warmup run triggers JIT for run!/output! beforehand), so compilation
        # time is excluded from the reported numbers.
        elapsed = @elapsed run!(simulation, output = true; period)

        n_outputs = Int(period / interval) + 1
        return (; elapsed, n_outputs)
    finally
        # clean up the entire per-run output directory (Zarr store + side files)
        rm(tmp_output_path; recursive = true, force = true)
    end
end

function main()
    @info "Warming up (JIT compilation, not measured)"
    benchmark_time_chunk(1; period = Hour(3))   # tiny warmup run

    println()
    @printf("%-12s %-12s %-14s %-16s %-12s\n",
            "time_chunk", "period", "n_outputs", "run! elapsed (s)", "vs tc=1")
    println("-"^70)

    results = Dict{Int, Float64}()
    base = nothing
    for tc in TIME_CHUNKS
        r = benchmark_time_chunk(tc)
        results[tc] = r.elapsed
        tc == 1 && (base = r.elapsed)
        slowdown = base === nothing ? "-" : @sprintf("%.2fx", r.elapsed / base)
        @printf("%-12d %-12s %-14d %-16.3f %-12s\n",
                tc, string(PERIOD), r.n_outputs, r.elapsed, slowdown)
    end
    println("-"^70)
    return results
end

main()
```

CC @maximilian-gelbrecht

bgroenks96:
Claude-guided attempt at addressing #1161. Adds in-memory buffers for each `time_chunk`.

Resolves #1161.


SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
```diff
@@ -487,18 +487,92 @@ end
...
+
+A Zarr chunk is the atomic unit of (de)compression and I/O, so writing one time
+slice into a chunk that spans `time_chunk > 1` steps would force a read-modify-write
+of the whole chunk on every step (read + decompress → patch one slice → recompress
++ write). To avoid that, time-varying variables are buffered `time_chunk` slices deep
+(see `output.time_buffers`) and flushed a full chunk at a time via
+[`flush_time_chunk!`](@ref); a chunk-aligned full-chunk write is fulfilled by Zarr's
+single-chunk fast path with no read. The trailing partial chunk is flushed on `close`
+(see [`flush_partial_time_chunks!`](@ref)). With `time_chunk == 1` (and for static,
+non-time variables) each write is already its own chunk and is written directly."""
```
maximilian-gelbrecht: For me the paragraph reads very claudy. 

Would you also call this "flush"? I would maybe just call it `write_time_chunk!`?

bgroenks96:
I think the buildkite failure is unrelated to the changes in this PR.

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
```diff
@@ -487,18 +487,92 @@ end
...
+    end
+
+    # Buffered path: copy this slice into the variable's time buffer (lazily allocated
+    # to the chunk shape (spatial..., time_chunk)) at its offset within the current chunk.
+    i = output.output_counter
+    offset = mod1(i, time_chunk)    # 1..time_chunk position within the current chunk
+    buffer = get!(output.time_buffers, variable.name) do
+        Array{eltype(data)}(undef, size(data)..., time_chunk)
+    end
+    selectdim(buffer, ndims(buffer), offset) .= data
```
maximilian-gelbrecht, line 531: So this is always used, also for `time_chunk = 1`?

maximilian-gelbrecht (commented):
Thank you already for this. It's less complex than I thought. 

I am just wondering whether we should include some heuristic to warn the user that the chunking might result in an OOM. For high resolutions it's really not possible to use very large buffers.

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
bgroenks96: Flush is correct because it's a buffer, but yes, I can rewrite the paragraph.

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
bgroenks96: At the moment, yes.

maximilian-gelbrecht: I am wondering about it. On the one hand, it's probably a reasonable assumption for most situations, and reduces code complexity a little bit. On the other hand, it does mean that more memory is needed even for a configuration with `time_chunk = 1`. 

I think I am fine with keeping it like that for the moment. What do you think?

bgroenks96: I don't think it will add too much complexity to just skip the buffer when `time_chunk = 1`.

maximilian-gelbrecht: Then, let's do that.

bgroenks96: I was actually wrong, the code already did this. But I had Claude refactor a bit to make it more clear.

See [6edf0d9](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1162/commits/6edf0d91f7f071ae847172ea48ba7e1fda77a7f2).

SpeedyWeather/ext/SpeedyWeatherZarrExt.jl
bgroenks96: Fixed, I think, in [b361c68](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1162/commits/b361c680df2f0ad2a95079b8fa30055882a7266c)

maximilian-gelbrecht (approved):
Looks good now, thanks!

PR #1167, opened 2026-07-31
Revise extended differentiability tests

maximilian-gelbrecht:
This revises the tests in `SpeedyWeather/test/differentiabilty` aiming that they actually finally all pass and are revised with the new syntax. I want to run them regularly again from now on.


.github/workflows/CI_extended_tests.yml
```diff
@@ -1,25 +1,13 @@
 name: Extended Tests
 on:
-  pull_request:
-    branches:
-      - main
```
maximilian-gelbrecht, line 5: Currently I have this set up now so that it runs on every commit on `main`, and can be additionally manually triggered in the Actions tab. 

Running it on every commit in a PR definitely is too much, so this seems like a decent compromise, but I am open for suggestions. 

@milankl what do you think?

PR #1177, opened 2026-08-05
1-based `truncation` as resolution parameter

milankl:
This one is for @bgroenks96. No more T31, 63, ... 124 eh what???-business anymore! 

- deprecate `trunc` as resolution parameter and call it `truncation`, which is 1-based
- so old typical resolutions 31, 63, 127, ... become neat powers of 2, i.e. 32, 64, 128 etc.
- the "in-between" resolutions 42, 85, etc therefore become 43, 86, 171, ... which is similarly hard to guess
- but other "in-between" resolutions can easily be chosen as 2^n * 3, i.e. 48, 96, ... etc

Downside: Mathematically spherical harmonics use 0-based wavenumbers and every other spectral model uses those too, but I don't think that's bad to be honest. Opinions?

EDIT: __This is a non-breaking change mostly__, as `trunc` is still supported but throws a deprecation warning. However, I have changed the interface to functions in the submodules that take in `trunc` to use consistently `truncation` (1-based). Previously we had a mix of 0 and 1-based assumed arguments which wasn't great, now everything is 1 based I believe.


milankl:
@pierluigividale do you majorly dislike this as someone used to the T???? business from other spectral models?

bgroenks96:
As a computer scientist, I am always happy to see powers of two :D

pierluigividale:
> @pierluigividale do you majorly dislike this as someone used to the T???? business from other spectral models?

Are you saying that all we will see is the power of two, or are you saying that we are going to see labels like T32, T64, T128 and so on? I am quite happy with that, as the other model I use, the UM, has "Node Numbers", so N96, N640 and my current model, N1280. Of course the true resolution is not the same, but the naming would be quite aligned.

milankl:
> Are you saying that all we will see is the power of two, or are you saying that we are going to see labels like T32, T64, T128 and so on? I am quite happy with that, as the other model I use, the UM, has "Node Numbers", so N96, N640 and my current model, N1280. Of course the true resolution is not the same, but the naming would be quite aligned.

Indeed, great then I consider this move to 1-based truncation as resolution parameter as approved!

SpeedyWeather/src/dynamics/horizontal_diffusion.jl
```diff
@@ -35,7 +35,7 @@ $(TYPEDFIELDS)"""
 
     # DIMENSIONS
     "spectral resolution"
```
maximilian-gelbrecht: best to denote (1-based) here as well

SpeedyWeather/src/dynamics/horizontal_diffusion.jl
```diff
@@ -369,13 +370,13 @@ export SpectralFilter
...
```
maximilian-gelbrecht:
```suggestion
    "spectral resolution (1-based)"
```

SpeedyWeather/src/dynamics/horizontal_diffusion.jl
maximilian-gelbrecht:
```suggestion
    "spectral resolution (1-based)"
```

maximilian-gelbrecht (approved):

PR #1178, opened 2026-08-05
Lower wind gust speed

gregrmunday:
Lowering the gust speed constant from free convection to a more realistic value, from 5m/s to 1m/s.

This reduces the sensible heat flux over land, lowers rainfall over the ocean significantly and reduces surface temperatures everywhere.

Screenshot 2026-08-05 at 17 57 52

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Lower free convection gust constant [#1178](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1178)
 - Add ocean neutral surface wind speed model [#1156](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1156)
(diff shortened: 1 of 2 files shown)


milankl (approved):

milankl:
Documentation preview will be here https://speedyweather.github.io/SpeedyWeatherDocumentation/previews/PR1178

milankl:
Could you do a quick check how cold the land gets vs how warm the upper layer soil temperature? I'm just worried temperatures get unrealistic over land with the reduced flux

gregrmunday:
> Could you do a quick check how cold the land gets vs how warm the upper layer soil temperature? I'm just worried temperatures get unrealistic over land with the reduced flux

Sure! I can't get the soil temperature from these simulations, but here's the lowest layer temperature:

Screenshot 2026-08-06 at 10 36 55

PR #1180, opened 2026-08-07
Optimize legendre transform for better gpu performance

maximilian-gelbrecht:


milankl (approved):
Awesome work, communicated comments verbally!

PR #1183, opened 2026-08-07
Time stepping soil temperature and moisture

milankl:
Next one after #1140 

- hooks up soil temperature and moisture into the general time stepping procedure (e.g. Leapfrog) instead of the current Euler step that's hardcoded into the kernel
- leaves snow as Euler forward for now
- actually turns leapfrogging for ocean, sea ice and land on, by moving the tendency to the 2nd, i.e. current step.


maximilian-gelbrecht (dismissed):
Looks good to me!

maximilian-gelbrecht:
I could maybe use this for my FP experiments so I briefly just let Claude weigh in on its opinion about the instability

> ### What causes the instability
> 
> The leapfrog integrates the land over 2Δt, and the land surface energy balance is only marginally stable at Δt.
> 
> The top soil layer is thin (z₁ = 0.2 m, z₁C₁ ≈ 2.3·10⁵ J/(m² K)) but strongly coupled to the atmosphere through εσT₁⁴ and ρ cₚ C_H V₀ (T₁ − T_air). The relaxation rate λ = −∂F/∂T₁ / (z₁C₁) approaches 1/Δt when surface winds are large — explicit stability needs Δt_step · λ < 2, and the old hardcoded Euler step with Δt sat just inside. Moving the update into update_prognostic! doubles the step to 2Δt and crosses the limit.
> 
> Reproducer: dynamics/forcing_drag.jl "Forcing and drag: 3D" (StartFromRest + StochasticStirring + LinearVorticityDrag, T32, 15 days) — NaNs in both PrimitiveDry and PrimitiveWet. Soil temperature runs away in ~20 steps (228 K → 72 K → −1886 K → −1.7·10⁷ → NaN, sensible heat flux reaching 10⁵ W/m²). The default PrimitiveWetModel is unaffected, which is why it isn't obvious.

This does sound plausible to me. But Claude's suggestions is effectively just to not do it at all though, which isn't very helpful. If it were true then integrating implicitly would be an option but certainly not trivial

milankl-claude:
Review of #1183 at 8b7100d, done by Claude on @milankl's request. The first section covers the long-integration instability, which I reproduced and have a fix for. The second lists other bugs and inconsistencies.

## 1. The instability

### Reproduction
- `dynamics/forcing_drag.jl` "Forcing and drag: 3D" (T32, StochasticStirring): NaN at step ~20 for both Dry and Wet.
- `long_integrations/default_primitive_dry.jl`: NaN at step 6301 (24 June 2000). This is the same failure CI shows at 6351.

In both cases the first soil layer at a single land point develops a 2Δt oscillation between the two leapfrog steps, and it grows by about 1.3× per step. In the year-long run the point is Antarctic, in June, with a stable boundary layer and 20–30 m/s surface winds:

```
 step   T₁ [step1 step2]      sensible heat flux
 6249   184.3  229.9           -870 W/m²
 6256   216.0   85.4          10204 W/m²
 6260   205.0 -155.7          25681 W/m²
 6274  -329.1 -20776          5·10⁵ W/m²  → NaN
```
SST and sea ice stay quiet (the gap between steps is < 0.1 K), so the ocean isn't the cause.

### Cause
@maximilian-gelbrecht's comment above already points in this direction: the land physics is explicit, and its relaxation timescale is shorter than the doubled time step.

- The surface fluxes (sensible, longwave up, latent) are parameterizations, so they read soil temperature at step 1. The tendency is now applied by the leapfrog, i.e. `T(t+Δt) = T(t−Δt) + 2Δt·f(T(t−Δt))`. That is forward Euler with **2Δt** on each of the two leapfrog chains, which are coupled only through the RAW filter.
- Linearise around T₁. The damping rate is κ = k/(z₁C₁), where k = ∂F/∂T₁ ≈ 4σT³ + cₚρC_H V₀ ≈ 40–100 W/m²/K in strong winds, and z₁C₁ = 0.2 m · 1.13·10⁶ J/m³/K ≈ 2.3·10⁵ J/m²/K. Explicit Euler is stable for 2Δt·κ < 2. At T31 (Δt = 2400 s) that gives 2Δt·κ ≈ 0.8…2.1.
- `main` did Euler with **Δt** from the current state, so it had twice the margin.

Note that switching only to `which_prognostic_step = 1` ("staggered Euler") does **not** restore `main`'s behaviour. The update is still a 2Δt step from t−Δt. I tested this: with step-1 reads, forcing_drag still NaNs.

### Fix A (recommended): Euler forward with Δt for ocean, sea ice and land, like on `main`
The time stepping of the surface namespaces goes through a separate `update_prognostic_surface!` hook. It defaults to the model's `update_prognostic!`, so NCycleLorenz etc. will use the general path later. For `Leapfrog` it does Euler over Δt and writes the new state into **both** steps:

```julia
# time_integration.jl: generated update_prognostic_namespaces! calls this instead of update_prognostic!
update_prognostic_surface!(var, tendency, clock, time_stepping, implicit, model) =
    update_prognostic!(var, tendency, clock, time_stepping, implicit, model)
surface_time_step(time_stepping, clock) = time_step(time_stepping, clock)

# leapfrog.jl
surface_time_step(L::Leapfrog, clock::Clock) = ifelse(clock.step_counter <= 1, L.Δt / 2, L.Δt)  # matches the clock

function update_prognostic_surface!(var, tendency, clock::Clock, time_stepping::Leapfrog, implicit, model)
    Δt_euler = surface_time_step(time_stepping, clock)
    var_old = get_step(var, 1)
    var_new = get_step(var, 2)
    var_tend = get_tendency_step(tendency, time_stepping, time_stepping)
    var_old .+= Δt_euler .* var_tend
    var_new .= var_old                      # both steps hold the current state
    return nothing
end

@inline which_prognostic_step(var, ::AbstractLeapfrog, ::AbstractDynamic{Ocean,SeaIce,LandComponent}) = 1
```
`ThermodynamicSeaIce` and `LandBucketMoisture` then use `surface_time_step` rather than `time_step` for their "within one time step" terms. `SnowModel` already uses Δt, so it's consistent again.

This keeps what the PR generalised: tendencies are written into `vars.tendencies.land/ocean`, and the kernels no longer do the update themselves. Only the leapfrog path is special-cased.

### Fix B (optional, if the surface should ever be leapfrogged or use longer steps): linearised implicit flux
Divide the top-layer tendency by `1 + Δt_eff·k/(z₁C₁)`, with `k = 4σT₁³ + cₚρ C_H V₀ + Δ`. This is the linearised backward-Euler trick SPEEDY uses for skin temperature (`dhfdt`/`dslr`). It leaves the equilibrium unchanged and stabilises forcing_drag and both year runs even with the 2Δt leapfrog. A clean version would have the surface-flux parameterizations write `∂F/∂T_skin` into a variable. With A this isn't needed now, but it would matter for NCycleLorenz with a longer Δt.

### Results (T31/T32)

| run | PR as is | Fix A | Fix B (+ step-1 reads, snow Δt fix) | `main` |
|---|---|---|---|---|
| forcing_drag 3D Dry, 15 d | NaN at ~step 20 | ok, T₁ ∈ [190.4, 298.3] K | ok, [192.8, 296.3] K | ok, [190.3, 298.3] K |
| forcing_drag 3D Wet, 15 d | NaN | ok, [227.4, 340.4] K | ok, [227.8, 339.5] K | ok, [226.2, 340.7] K |
| default PrimitiveDry, 1 yr | NaN at step 6301 | NaN at step 12151 (3 Dec), from the atmosphere† | ok | ok |
| default PrimitiveWet, 1 yr | – | ok | ok | ok |

† With Fix A the Dry year blows up in the **dynamical core**, not the land. It starts over the open North Pacific (165°E, 39°N, land fraction 0) in layers 3–5, under an 85–100 m/s jet, while soil temperatures are all normal (≤ 303 K). `main` passes the same window, but its trajectory is different by then. I read this as the default T31/Δt = 40 min setup being marginal for strong jets, which the chaotic trajectory happens to hit. It may still fail CI's `long_integrations/default_primitive_dry.jl` on some Julia/OS combinations, so it's worth watching.

Tests (with `--check-bounds=yes`): `parameterizations/{land,ocean_sea_ice,surface_fluxes,longwave_radiation}.jl`, `dynamics/{forcing_drag,set}.jl` and `variables/steps.jl` pass with Fix A (305/305). The same set plus `long_integrations/default_primitive_{dry,wet}.jl` passes with Fix B (307/307).

### Related: snow melt uses the wrong latent heat (pre-existing, also on `main`)
`land_bucket_temperature_kernel!` computes `Q_melt = Lᵢ * M` with `Lᵢ = latent_heat_sublimation(...)` (2.8·10⁶ J/kg). But `SnowModel` computes the melt rate `M` from the available energy with `latent_heat_fusion` (3.3·10⁵ J/kg), so melting cools the soil **8.5× too much**. In the Wet run this produced 40 K jumps between steps at points sitting on the melting threshold (`Q_melt` ≈ 2900 W/m²). It should be `latent_heat_fusion`, with a zero fallback for dry atmospheres. The fix above includes this. (If the surface were leapfrogged, as in Fix B, `SnowModel`'s `E_avail = cₛ δT z₁ / Δt` would additionally need the 2Δt.)

### Why step 2 was problematic anyway
None of the land, ocean and sea-ice components has oscillatory or advective terms, only damping and relaxation terms (soil diffusion, soil moisture diffusion, the `ThermodynamicSeaIce` restore-to-freezing). Evaluating those at the centred step is formally unstable under leapfrog. For example, the sea-ice term `dsst -= min(sst − T_f, 0)/Δt`, read at step 2 with 2Δt, becomes `y⁺ = y⁻ − min(y, 0)`. In a scalar model with RAW filter and constant cooling it settles into a limit cycle ~4 K *above* freezing instead of pinning at T_f.

## 2. Other bugs and inconsistencies

1. **The new abstract types break the type hierarchy.** `AbstractDynamicSeaIce`, `AbstractPrescribedSeaIce`, `AbstractDynamicLandComponent` and `AbstractPrescribedLandComponent` are all `<: AbstractModelComponent`, so e.g. `ThermodynamicSeaIce isa AbstractSeaIce` and `LandBucketTemperature isa AbstractLandTemperature` are now `false`. `AbstractDynamicSnow`/`…SoilMoisture`/`…LandTemperature` can't be subtypes of both `AbstractSnow` and `AbstractDynamicLandComponent` (single inheritance). As a result `subtypes(SpeedyWeather.AbstractSeaIce)` in sea_ice.md and `subtypes(AbstractLandTemperature)`/`subtypes(AbstractSoilMoisture)` in land.md now list no concrete types, and `AbstractSnow`/`AbstractSoilMoisture`/`AbstractLandTemperature` are orphaned. Suggestion: keep the component hierarchy, and express dynamic vs prescribed as a trait, e.g. `isdynamic(::LandBucketTemperature) = true`, used in `variables` to decide on the time dimension. With Fix A, `which_prognostic_step` doesn't need the split at all.
2. **Custom oceans no longer get an SST allocated.** `variables(::AbstractOcean)` became `variables(::AbstractPrescribedOcean)`, so the `CustomOceanModel <: AbstractOcean` in `docs/src/custom_ocean.md` gets no `sea_surface_temperature`. Keep a fallback `variables(::AbstractOcean)` (GridXY).
3. **`LandBucketMoisture`: percolation and river runoff are dead code, and water is not conserved.** `filter!` clamps *all* steps of `soil_moisture` to [0, 1] after every time step, so at the next kernel call `W₁ ≤ 1` and `δW₁ = W₁ − min(W₁, 1) ≡ 0`. The `p·δW₁` transfer to layer 2 and `R += …` never happen, and excess water in layer 1 is deleted by the clamp. On `main` it went to layer 2 and runoff. Do the excess handling in `filter!`, where `δW₁` of the new state is known, or from the predicted state in the kernel. (Minor and pre-existing: `R[ij] += Δt * (1 - p) * δW₁ * f₁` accumulates m·s while the variable is declared m/s.)
4. **`SnowModel` declares `soil_temperature` as `LandXYZT(pg)`**, which conflicts with a prescribed land temperature (`LandXYZ`). It only works because `filter_variables` keeps the first declaration and `land.temperature` comes before `land.snow`. Snow only reads the variable, so it shouldn't declare it. `tg` is unused there.
5. **The `isfinite` guards were removed** in `SurfaceLandHeatFlux` and `land_snow_kernel!`. With `mask = false` in `LandBucketTemperature`/`SeasonalLandTemperature`, NaN soil temperatures at ocean points give `flux_land = NaN`, and `NaN * land_fraction` is NaN even for `land_fraction = 0`. Either keep the guards or remove the `mask` option.
6. `vegetation.jl`: `@boundscheck ndims(soil_moisture) == 2` has no `|| throw(...)`, so it never checks anything. Several new checks use `throw(BoundsError)` (the type) rather than `throw(BoundsError())`.
7. `set.jl` test: `set!(…, soil_moisture = D; namespace = :land)` without `step` is only checked for step 1. `set!` on land and ocean variables should probably write all steps. With Fix A both steps must hold the current state.
8. Minor and pre-existing: land points where the input file has NaN are initialised with `ocean_temperature = 285 K`. This includes Greenland and Antarctica in January (e.g. ij = 244: T₁ = 285 K, sensible heat flux ≈ 1000 W/m² at step 1), which causes a large initial shock.

The Fix A prototype diff is below; happy to push it to this branch as proper commits.

<details><summary>Fix A prototype diff</summary>

```diff
diff --git a/SpeedyWeather/src/dynamics/atmosphere.jl b/SpeedyWeather/src/dynamics/atmosphere.jl
index 6b9efa5b..234a8e1d 100644
--- a/SpeedyWeather/src/dynamics/atmosphere.jl
+++ b/SpeedyWeather/src/dynamics/atmosphere.jl
@@ -126,6 +126,9 @@ latent_heat_condensation(A::AbstractDryAtmosphere) = zero(eltype(A))
 latent_heat_sublimation(A::AbstractWetAtmosphere) = A.latent_heat_sublimation
 latent_heat_sublimation(A::AbstractDryAtmosphere) = zero(eltype(A))
 
+latent_heat_fusion(A::AbstractWetAtmosphere) = A.latent_heat_fusion
+latent_heat_fusion(A::AbstractDryAtmosphere) = zero(eltype(A))
+
 saturation_vapor_pressure(T, A::AbstractDryAtmosphere) = zero(T)
 
 """$(TYPEDSIGNATURES)
diff --git a/SpeedyWeather/src/parameterizations/land/soil_moisture.jl b/SpeedyWeather/src/parameterizations/land/soil_moisture.jl
index d474e24e..b6177376 100644
--- a/SpeedyWeather/src/parameterizations/land/soil_moisture.jl
+++ b/SpeedyWeather/src/parameterizations/land/soil_moisture.jl
@@ -231,7 +231,7 @@ function timestep!(
     soil_moisture = get_prognostic_step(vars.prognostic.land.soil_moisture, model.time_stepping, soil)
     soil_moisture_tendency = get_tendency_step(vars.tendencies.land.soil_moisture, model.time_stepping, soil)
 
-    Δt = time_step(model.time_stepping, vars.prognostic.clock)
+    Δt = surface_time_step(model.time_stepping, vars.prognostic.clock)
     ρ = model.atmosphere.water_density
     (; land_fraction) = model.land_sea_mask
 
diff --git a/SpeedyWeather/src/parameterizations/land/soil_temperature.jl b/SpeedyWeather/src/parameterizations/land/soil_temperature.jl
index f9ca2ad8..99164d4f 100644
--- a/SpeedyWeather/src/parameterizations/land/soil_temperature.jl
+++ b/SpeedyWeather/src/parameterizations/land/soil_temperature.jl
@@ -226,7 +226,7 @@ function timestep!(
         get_prognostic_step(vars.prognostic.land.soil_moisture, model.time_stepping, land) : nothing
 
     Lᵥ = latent_heat_condensation(model.atmosphere)
-    Lᵢ = latent_heat_sublimation(model.atmosphere)
+    Lᵢ = latent_heat_fusion(model.atmosphere)
 
     (; land_fraction) = model.land_sea_mask
     (; thermodynamics, geometry) = model.land
diff --git a/SpeedyWeather/src/parameterizations/sea_ice.jl b/SpeedyWeather/src/parameterizations/sea_ice.jl
index 7df05cc9..d321bbed 100644
--- a/SpeedyWeather/src/parameterizations/sea_ice.jl
+++ b/SpeedyWeather/src/parameterizations/sea_ice.jl
@@ -63,7 +63,7 @@ function timestep!(vars::Variables, sea_ice_model::ThermodynamicSeaIce, model::P
     
     @boundscheck size(dsst) == size(dℵ) == size(sst) || throw(BoundsError)
 
-    Δt = time_step(model.time_stepping, vars.prognostic.clock)
+    Δt = surface_time_step(model.time_stepping, vars.prognostic.clock)
     (; land_fraction) = model.land_sea_mask
 
     m = sea_ice_model.melt_rate             # melt rate [m²/m²/s/K]
diff --git a/SpeedyWeather/src/time_stepping/steppers/leapfrog.jl b/SpeedyWeather/src/time_stepping/steppers/leapfrog.jl
index 7e29bb4e..2b8e7f62 100644
--- a/SpeedyWeather/src/time_stepping/steppers/leapfrog.jl
+++ b/SpeedyWeather/src/time_stepping/steppers/leapfrog.jl
@@ -69,9 +69,9 @@ tendency_steps(::AbstractLeapfrog) = 1
 @inline which_prognostic_step(var, ::AbstractLeapfrog, ::AbstractLandComponent) = 1
 
 # dynamic models are then generally leapfrogged
-@inline which_prognostic_step(var, ::AbstractLeapfrog, ::AbstractDynamicOcean) = 2
-@inline which_prognostic_step(var, ::AbstractLeapfrog, ::AbstractDynamicSeaIce) = 2
-@inline which_prognostic_step(var, ::AbstractLeapfrog, ::AbstractDynamicLandComponent) = 2
+@inline which_prognostic_step(var, ::AbstractLeapfrog, ::AbstractDynamicOcean) = 1
+@inline which_prognostic_step(var, ::AbstractLeapfrog, ::AbstractDynamicSeaIce) = 1
+@inline which_prognostic_step(var, ::AbstractLeapfrog, ::AbstractDynamicLandComponent) = 1
 
 # particle advection using u, v at current not previous time step
 @inline which_prognostic_step(var, ::AbstractLeapfrog, ::AbstractParticleAdvection) = 2
@@ -289,3 +289,22 @@ end
     var_old[lmk] = var_lf[lmk] + w1 * update
     var_new[lmk] = new - w2 * update
 end
+
+"""$(TYPEDSIGNATURES)
+With leapfrog, ocean, sea ice and land variables are stepped with Euler forward over Δt
+(Δt/2 on the first two steps, matching the clock) from the current state, which is written
+into both steps so that step 1 and step 2 always hold the current state."""
+function update_prognostic_surface!(var, tendency, clock::Clock, time_stepping::Leapfrog, implicit, model)
+    (; Δt) = time_stepping
+    Δt_euler = surface_time_step(time_stepping, clock)
+    var_old = get_step(var, 1)
+    var_new = get_step(var, 2)
+    var_tend = get_tendency_step(tendency, time_stepping, time_stepping)
+    @boundscheck size(var_old) == size(var_new) == size(var_tend) || throw(BoundsError())
+    var_old .+= Δt_euler .* var_tend        # Field/LTA broadcasting over the step views
+    var_new .= var_old
+    return nothing
+end
+
+"""Time step [s] over which ocean, sea ice and land tendencies are applied, see `update_prognostic_surface!`."""
+surface_time_step(L::Leapfrog, clock::Clock) = ifelse(clock.step_counter <= 1, L.Δt / 2, L.Δt)
diff --git a/SpeedyWeather/src/time_stepping/time_integration.jl b/SpeedyWeather/src/time_stepping/time_integration.jl
index 05d8dc67..b44a9125 100644
--- a/SpeedyWeather/src/time_stepping/time_integration.jl
+++ b/SpeedyWeather/src/time_stepping/time_integration.jl
@@ -169,7 +169,7 @@ end
     for namespace in (:ocean, :land), name in _namespace_names(T, namespace)
         push!(
             calls, :(
-                update_prognostic!(
+                update_prognostic_surface!(
                     getfield(getfield(vars.prognostic, $(QuoteNode(namespace))), $(QuoteNode(name))),
                     getfield(getfield(vars.tendencies, $(QuoteNode(namespace))), $(QuoteNode(name))),
                     clock, time_stepping, implicit, model,   # no scale: ocean/land use the unscaled time step
@@ -179,3 +179,8 @@ end
     end
     return Expr(:block, calls..., :(return nothing))
 end
+
+# ocean/land/sea ice: by default use the same time stepper as the atmosphere
+update_prognostic_surface!(var, tendency, clock, time_stepping, implicit, model) =
+    update_prognostic!(var, tendency, clock, time_stepping, implicit, model)
+surface_time_step(time_stepping, clock) = time_step(time_stepping, clock)
```
</details>

PR #1184, opened 2026-08-08
Learned surface roughness scheme

gregrmunday:
Superseding #959... 

**Land** and **ocean** momentum, heat and moisture surface roughness represented by analytical expressions distilled from neural networks.

_Boundary_ _layer_ drag with constant land and ocean roughnesses:

bld_m_const

and with the new scheme:

bld_m_learned

---

_Surface_ _heat_ _flux_ with constant land and ocean roughnesses:

shf_h_const

and with the new scheme:

shf_h_learned

---

Benchmarks below 👇


gregrmunday:
One small thing that needs to be sorted is how we name the surface flux - I've called it `humidity flux` as it was but I've also been using `moisture roughness` when dealing with the surface roughness. 

We could maybe change the name of the flux to the `moisture flux`, or use `humidity roughness`, which sounds a bit weird imo. Personally I think we should use `moisture roughness` and apply this to the `moisture flux`. Either way it should be consistent between surface roughnesses, boundary layer drags and surface fluxes I think.

gregrmunday:
**Benchmarks**:

Default model: `1.412 ms (8 allocations: 256 bytes)`
WIth new scheme: `1.639 ms (8 allocations: 256 bytes)`

milankl:
Naming isn't easy I agree. I normally prefer to call every humidity related actually humidity something rather than moisture as we use specific humidity. Yes we use soil *moisture* but that's liquid and not in vapour form. 

I think there's still a bunch of hardcoded parameters, e.g. the charnock calculation could be a functor no? And also it seems that the charnock parameter doesn't need to be stored and can be calculated on the fly as it's only used once afterwards?

SpeedyWeather/src/parameterizations/land/vegetation.jl
```diff
@@ -92,6 +98,12 @@ export VegetationClimatology
 
     "Low vegetation cover [1], interpolated onto Grid"
     low_cover::GridVariable2D
+
+    "High vegetation LAI [1], interpolated onto Grid"
+    high_lai::GridVariable2D
```
milankl: Can we call these high_leaf_area or similar to avoid the acronym? I consider the "index" to be the unit and we normally don't include units in the names so "leaf area" is telling enough what this is about.

SpeedyWeather/src/parameterizations/land/vegetation.jl
```diff
@@ -233,6 +274,8 @@ function variables(::VegetationClimatology)
         PrognosticVariable(:soil_moisture, LandXYZ(), desc = "Soil moisture content (fraction of capacity)", units = "1", namespace = :land),
         ParameterizationVariable(:vegetation_high, Grid2D(), desc = "Vegetation high cover", units = "1", namespace = :land),
         ParameterizationVariable(:vegetation_low, Grid2D(), desc = "Vegetation low cover", units = "1", namespace = :land),
+        ParameterizationVariable(:lai_vegetation_high, Grid2D(), desc = "Leaf Area Index of high vegetation", units = "1", namespace = :land),
+        ParameterizationVariable(:lai_vegetation_low, Grid2D(), desc = "Leaf Area Index of low vegetation", units = "1", namespace = :land),
```
milankl: Particularly here. `high_leaf_area`? Etc. maybe for consistency also high_vegetation_cover and high_vegetation_leaf_area? Or similar?

SpeedyWeather/src/parameterizations/ocean.jl
```diff
@@ -289,6 +289,7 @@ function variables(::SlabOcean, model::AbstractModel)
     pg = nsteps.prognostic_grid
     tg = nsteps.tendency_grid
     return (
+        ParameterizationVariable(:charnock_parameter, Grid2D(), namespace = :ocean, desc = "Charnock parameter", units = ""),
```
milankl: Maybe not needed see below

SpeedyWeather/src/parameterizations/ocean.jl
```diff
@@ -350,17 +351,34 @@ function timestep!(vars::Variables, ocean_model::SlabOcean, model::PrimitiveEqua
...
     launch!(
         architecture(dsst), LinearWorkOrder, size(dsst), slab_ocean_kernel!,
-        dsst, land_fraction, Rsd, Rsu, Rld, Rlu, H, S, params
+        dsst, land_fraction, Rsd, Rsu, Rld, Rlu, H, S, αᵪ, uₙ, params
     )
     return nothing
 end
 
-@kernel inbounds = true function slab_ocean_kernel!(dsst, land_fraction, Rsd, Rsu, Rld, Rlu, H, S, params)
+@inline function calculate_charnock(uₙ::T) where {T <: Real}
```
milankl: As functor to expose the parameters?

SpeedyWeather/src/parameterizations/ocean.jl
```diff
@@ -350,17 +351,34 @@ function timestep!(vars::Variables, ocean_model::SlabOcean, model::PrimitiveEqua
...
+    )
+    return log(charnock)
+end
+
+@kernel inbounds = true function slab_ocean_kernel!(dsst, land_fraction, Rsd, Rsu, Rld, Rlu, H, S, αᵪ, uₙ, params)
     ij = @index(Global, Linear)         # every grid point ij
     if land_fraction[ij] < 1            # at least partially ocean
         (; C₀⁻¹, Lᵥ) = params
         dsst[ij] = C₀⁻¹ * (Rsd[ij] - Rsu[ij] - Rlu[ij] + Rld[ij] - Lᵥ * H[ij] - S[ij])
+        αᵪ[ij] = calculate_charnock(uₙ[ij])
```
milankl: Calculate in roughness not ocean kernel?

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
```diff
@@ -14,7 +14,9 @@ export ConstantDrag
 end
 
 variables(::ConstantDrag) = (
-    ParameterizationVariable(:boundary_layer_drag, Grid2D(), desc = "Boundary layer drag coefficient", units = "1"),
+    ParameterizationVariable(:boundary_layer_drag_momentum, Grid2D(), desc = "Boundary layer drag coefficient for momentum", units = "1"),
+    ParameterizationVariable(:boundary_layer_drag_heat, Grid2D(), desc = "Boundary layer drag coefficient for heat", units = "1"),
+    ParameterizationVariable(:boundary_layer_drag_humidity, Grid2D(), desc = "Boundary layer drag coefficient for humidity", units = "1"),
```
milankl, line 19: I thought these were the same

SpeedyWeather/src/parameterizations/surface_fluxes/surface_roughness.jl
```diff
@@ -1,41 +1,313 @@
...
+    return nothing
+end
+
+@inline function sea_ice_momentum_roughness(siconc::NF) where {NF <: Real}
+    # Following IFS documentation
+    base = NF(0.93e-3) * (1 - siconc) + NF(6.05e-3) * exp(NF(-17) * (siconc - 0.5)^2)
+    return max(base, NF(1.0e-3))
+end
+
+@inline function maximum_momentum_roughness(uₙ::NF) where {NF <: Real}
```
milankl, line 291: Also functor?

SpeedyWeather/src/parameterizations/surface_fluxes/surface_roughness.jl
```diff
@@ -1,41 +1,313 @@
...
+    U_cap = NF(33.0)
+    U_high = NF(55.0)
+
+    m = (z_high - z_cap) / (U_high - U_cap)
+    z_linear = muladd(m, uₙ - U_cap, z_cap)
+
+    return clamp(z_linear, z_high, z_cap)
+end
+
+@inline function ocean_momentum_roughness(uₙ::NF, αᵪ::NF, siconc::NF) where {NF <: Real}
```
milankl: Calculate charnock here on the fly?

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
gregrmunday: They're actually slightly different over the ocean!

SpeedyWeather/src/parameterizations/ocean.jl
gregrmunday: Yep can do, I just put it here since it's technically an ocean-related quantity (and should be provided by a wave model, but I guess this is our wave model!)

SpeedyWeather/src/parameterizations/land/vegetation.jl
gregrmunday: Updated in latest, gone for `high_vegetation_cover` `high_vegetation_leaf_area` combo.

SpeedyWeather/src/parameterizations/land/vegetation.jl
gregrmunday: Agreed this is clearer, done in latest commit

SpeedyWeather/src/parameterizations/surface_fluxes/surface_roughness.jl
gregrmunday: Yep, I've moved into the roughness to calculate on the fly

SpeedyWeather/src/parameterizations/ocean.jl
gregrmunday: Removed in latest

SpeedyWeather/src/parameterizations/ocean.jl
gregrmunday: Done in latest!

gregrmunday:
Something I think it would be good to aim for with the surface fluxes is roughly matching observations in terms of magnitude and spatial pattern. Here's the surface heat flux after this model is in (low res), colormap magnitude is the same.

**Observations**

shf_TRUE

**SpeedyWeather**

shf_latest

milankl:
I agree with you but you probably have to run this for longer to allow the ocean to be in equilibrium with the atmosphere. I believe this is why these fluxes are so clearly cooling the ocean in a pattern that probably still results from the initial conditions of SST?

gregrmunday:
> I agree with you but you probably have to run this for longer to allow the ocean to be in equilibrium with the atmosphere. I believe this is why these fluxes are so clearly cooling the ocean in a pattern that probably still results from the initial conditions of SST?

I did think that yes but I've done a run with 2 yrs of spinnup (at low res) and 3 years of output:

Screenshot 2026-08-08 at 16 42 57

and SSTs (which seem way too high?):

Screenshot 2026-08-08 at 16 52 37

gregrmunday:
We also seem to get a double **anti** ITCZ? Which seems to be anti-correlated with the SST pattern in the tropics... 

Screenshot 2026-08-08 at 17 12 46

milankl:
Oh gosh, yeah not sure why SSTs now are much higher, maybe because they don't lose enough latent heat after you turned the fluxes down with weaker gusts ;)

We absolutely need to investigate the surface flux balances and check what's off what's not. Thanks for starting this here.

gregrmunday:
> Oh gosh, yeah not sure why SSTs now are much higher, maybe because they don't lose enough latent heat after you turned the fluxes down with weaker gusts ;)
> 
> We absolutely need to investigate the surface flux balances and check what's off what's not. Thanks for starting this here.

I think you're right!! Although they do look a little high still with the gusts back up at 5m/s (35C instead of 45C...)

Given how sensitive the surface climate seems to be to the gust speed it could be worth making a small PR to calculate gusts properly on the fly depending on surface conditions. Gemini found a nice approximation the other day that's fast and straightforward, so I might make a PR.

milankl-claude (commented):
Nice work Greg — distilling the roughness closures into analytical expressions is a genuinely good idea, and the before/after plots in the description make the case well. Splitting `boundary_layer_drag` and `surface_roughness` into momentum/heat/humidity is also overdue and I'm glad it's happening.

That said, I don't think this is mergeable yet. Comments inline; the highlights:

### Blockers

1. **`BoundaryLayer` components are silently mis-assigned.** The struct fields were reordered but the positional constructor call wasn't, and all four type parameters are unconstrained — so `surface_roughness` gets the `SurfaceCondition`, `surface_condition` gets the `NeutralWindSpeed`, and so on. No error, just the wrong execution order and wrong `show` output. (`boundary_layer.jl:102`)
2. **`ConstantDrag` is broken.** Its `variables()` was renamed to the three split fields but `boundary_layer_drag!` still writes `boundary_layer_drag`. (`boundary_layer.jl:29`)
3. **`ConstantWindSlowdown` is a `MethodError`** — five-argument method, six-argument call site. (`surface_condition.jl:12`)
4. **Two output variables point at fields that don't exist**: `boundary_layer_drag_moisture` (should be `_humidity`) and, outside the diff, `output/variables/boundary_layer.jl:72` still reads `parameterizations.surface_roughness`. Both are in the default `BoundaryLayerOutput()`.
5. **The `z₀` fallback in `LearnedWindSlowdown` fires almost everywhere.** `land.momentum_roughness` is identically zero at ocean points (and vice versa), so the `||` throws away the correctly computed roughness and substitutes the hardcoded default outside mixed cells. The learned roughness barely reaches the slowdown. (`surface_condition.jl:39`)
6. **Tests weren't updated.** `test/parameterizations/surface_roughness.jl` still uses `roughness_length_ocean` / `parameterizations.surface_roughness` throughout, so the suite won't even load. There are also no tests for any of the five new schemes. Given CI is skipped on this PR, please get a green run before the next review round.

### Device compatibility

The learned coefficients are hardcoded `Float64` literals, not `NF`-typed fields. With `NF = Float32` they promote every expression to `Float64`, so:

- **Metal**: no `Float64` at all — these kernels won't compile, and Metal is Float32-only by design.
- **CUDA**: compiles, but you pay ~30× on `exp`/`log`/`pow`, and this is the boundary layer, called at every grid point every step.
- **Reactant/AD**: `where {NF <: Real}` signatures that pin several arguments to one identical concrete type will `MethodError` on `Dual`s and on any mixed-precision setup.

`sea_ice_momentum_roughness`'s `(siconc - 0.5)^2` has the same problem in a function that otherwise carefully wraps everything in `NF(...)`.

### Coding standards

- **`@parameterized` / `@param`**: none of `LearnedLandRoughness`, `LearnedOceanRoughness`, `Charnock`, `LearnedWindSlowdown` use it, so ~45 fitted coefficients are invisible to the parameter interface and can't be calibrated. `NeutralWindSpeed` in the same directory is the pattern to follow — and it fixes the Float32 problem at the same time. This matters more than usual here: the whole point of a *learned* scheme is that the numbers are retunable.
- **Naming**: `moisture_roughness` vs `boundary_layer_drag_humidity` for the same physical pair. Per your #959 thread with Milan, let's use *humidity* for vapour-phase and keep *moisture* for soil moisture. Also `sr` / `sr_h` for the roughness outputs where the drags got `bld_m` / `bld_h` / `bld_q`.
- **Docstrings**: most new structs have none, and most new fields have no `"[OPTION]"` line, so `$(TYPEDFIELDS)` renders blanks. For the opaque `m_c1…m_c9` please put the functional form the regression produced into the struct docstring.
- **Exports**: `LearnedLandRoughness`, `LearnedOceanRoughness` and `Charnock` aren't exported, so they can't be configured from user code.
- **Magic numbers**: `0.62/0.4` in `ConstantSurfaceRoughness` duplicates `a_q/a_h`; `1e-3` sea-ice roughness appears twice; `maximum_momentum_roughness` and `sea_ice_momentum_roughness` are entirely hardcoded.
- Runic: trailing whitespace at `surface_condition.jl:50,52,57,61`, and a ~190-character line in `vegetation.jl:120`.

### Physics worth a second look

- `C_H` and `C_Q` are computed as `κ²/log(z/z₀X)²`. The standard neutral-limit bulk coefficient is `κ²/(log(z/z₀M)·log(z/z₀H))` — the wind profile always supplies the momentum scale. Since the learned `z₀H` is orders of magnitude below `z₀M` over the ocean, the current form makes the scalar drags substantially too small. Given the too-warm SSTs and missing latent heat flux you found downthread, this seems worth testing before concluding the fluxes are wrong.
- `land_slowdown` is unbounded and goes negative above roughly 22 m/s at `z₀ = 0.1 m` — the surface wind reverses relative to the lowest level. It also diverges as `U₁ → 0` (`U₁^-0.999`). Needs a clamp and a `min_wind` guard.
- `BeljaarsWindGust` defaults `min_gust = 0`, replacing the old `gust_speed = 1`. In neutral/stable conditions the gust is exactly zero, which puts a zero into the denominator of `bulk_richardson_surface`. Given how sensitive you found the surface climate to gust speed, I'd keep a non-zero floor.

### Scope

This PR changes four defaults at once (`surface_roughness`, `neutral_wind_speed`, `wind_slowdown`, `wind_gust`), and `LearnedSurfaceRoughness` hard-requires `VegetationClimatology`, snow and sea ice — so `NoVegetation` and any sea-ice-free setup now fail on default settings. My suggestion: land the machinery with the **old** defaults, with the dependency reads guarded the way `heat.jl`/`humidity.jl` guard theirs, then flip the defaults in a follow-up once the surface flux balance is understood. That keeps the climate regression bisectable and gets the refactor (which is good and independently useful) in much sooner.

Happy to look again once the blockers are fixed — and I'd be interested to see the `C_H` formulation change on its own, that one might move the SSTs.

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
```diff
@@ -83,18 +85,18 @@ export BoundaryLayer
...
 function BoundaryLayer(
         SG::SpectralGrid;
         surface_condition = SurfaceCondition(SG),
-        neutral_wind_speed = nothing,
-        surface_roughness = ConstantSurfaceRoughness(SG),
+        neutral_wind_speed = NeutralWindSpeed(SG),
+        surface_roughness = LearnedSurfaceRoughness(SG),
         drag = BulkRichardsonDrag(SG),
     )
     return BoundaryLayer(surface_condition, neutral_wind_speed, surface_roughness, drag)
```
milankl-claude, line 102: **Blocker:** the struct field order was changed to `surface_roughness, surface_condition, neutral_wind_speed, drag`, but this positional call still passes the old order. All four type parameters are unconstrained, so nothing errors — the components are silently assigned to the wrong fields:

- `BL.surface_roughness` ← `SurfaceCondition`
- `BL.surface_condition` ← `NeutralWindSpeed`
- `BL.neutral_wind_speed` ← `LearnedSurfaceRoughness`

`parameterization!` below still calls `surface_condition, neutral_wind_speed, surface_roughness, drag` by *field name*, so the actual execution order ends up `neutral wind → roughness → surface condition → drag`, and `model.boundary_layer.surface_roughness` shows the wrong thing to users. Use keywords so this can't silently break again:

```suggestion
    return BoundaryLayer(; surface_condition, neutral_wind_speed, surface_roughness, drag)
```

Separately: if the intent of the field reorder *was* to change execution order (roughness before surface condition, because `LearnedWindSlowdown` needs `z₀`), then `parameterization!` is the place to do that, not the struct — please reorder there explicitly and add a comment explaining the dependency chain.

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
milankl-claude, line 19: `variables(::ConstantDrag)` now declares the three split fields, but `boundary_layer_drag!(ij, vars, ::ConstantDrag)` at line 29 still writes the old name:

```julia
vars.parameterizations.boundary_layer_drag[ij] = scheme.drag
```

That field no longer exists, so `ConstantDrag` errors on first use — and even if it didn't, the three fields that `heat.jl`/`humidity.jl`/`momentum.jl` now read would never be set. Needs:

```julia
@propagate_inbounds function boundary_layer_drag!(ij, vars, scheme::ConstantDrag)
    vars.parameterizations.boundary_layer_drag_momentum[ij] = scheme.drag
    vars.parameterizations.boundary_layer_drag_heat[ij] = scheme.drag
    vars.parameterizations.boundary_layer_drag_humidity[ij] = scheme.drag
    return nothing
end
```

SpeedyWeather/src/parameterizations/surface_fluxes/boundary_layer.jl
```diff
@@ -173,12 +177,20 @@ initialize!(::BulkRichardsonDrag, ::PrimitiveEquation) = nothing
...
+        return (κ / log(z_calc / z₀))^2
+    end
 
-    # should be z > z₀, z=z₀ means an infinitely high drag, choose one order higher than roughness length at least
-    # 0 < z < z₀ doesn't make sense so cap here
-    z = max(z, 10z₀)
-    drag_max = (κ / log(z / z₀))^2
+    drag_max_momentum = calc_drag_max(z₀M)
+    drag_max_heat = calc_drag_max(z₀H)
+    drag_max_humidity = calc_drag_max(z₀Q)
```
milankl-claude, line 193: Two things here.

**1. Physics.** Splitting the drag coefficient by roughness length is the right move, but the bulk formulation isn't `(κ/log(z/z₀X))²` for scalars. The wind profile always sets the momentum scale, so the standard Monin–Obukhov neutral-limit bulk coefficients are

```
C_D = κ² / log(z/z₀M)²
C_H = κ² / (log(z/z₀M) · log(z/z₀H))
C_Q = κ² / (log(z/z₀M) · log(z/z₀Q))
```

Since the learned `z₀H` is typically orders of magnitude below `z₀M` over the ocean, `log(z/z₀H)²` makes the heat/moisture drag much too small — quite possibly a chunk of the missing latent heat flux and the +35 °C SSTs you're seeing downthread. Worth testing.

**2. Style/GPU.** A closure capturing `z` and `κ` inside a `@propagate_inbounds` per-gridpoint function is something I'd rather avoid in code that has to go through KernelAbstractions and Reactant. A top-level `@inline` helper is free and clearer:

```julia
@inline function drag_max(z, z₀M, z₀X, κ)
    # should be z > z₀, z = z₀ means infinitely high drag, cap one order above the roughness length
    zM = max(z, 10z₀M)
    zX = max(z, 10z₀X)
    return κ^2 / (log(zM / z₀M) * log(zX / z₀X))
end
```

```suggestion
    drag_max_momentum = drag_max(z, z₀M, z₀M, κ)
    drag_max_heat = drag_max(z, z₀M, z₀H, κ)
    drag_max_humidity = drag_max(z, z₀M, z₀Q, κ)
```

One more, just below: `ΔΦ₀ = gravity * z` on line 197 now uses the *unclamped* `z`, whereas on `main` `z` had already been overwritten by `z = max(z, 10z₀)` before that line. I think the new behaviour is more correct, but it silently changes the bulk Richardson number in every simulation — please call it out in the CHANGELOG (or restore it if unintended).

SpeedyWeather/src/parameterizations/surface_fluxes/surface_condition.jl
```diff
@@ -1,21 +1,167 @@
...
+@parameterized @kwdef struct ConstantWindSlowdown{NF} <: AbstractParameterization
+    "[OPTION] Ratio of near-surface wind to lowest-level wind [1]"
+    @param wind_slowdown::NF = 0.95 (bounds = 0 .. 1,)
+end
+
+Adapt.@adapt_structure ConstantWindSlowdown
+ConstantWindSlowdown(SG::SpectralGrid; kwargs...) = ConstantWindSlowdown{SG.NF}(; kwargs...)
+initialize!(::ConstantWindSlowdown, ::PrimitiveEquation) = nothing
+
+@propagate_inbounds function calc_wind_slowdown(ij, vars, ws::ConstantWindSlowdown, model, U₁)
```
milankl-claude, line 12: **Blocker:** arity mismatch. The call site at line 200 is

```julia
calc_wind_slowdown(ij, vars, wind_slowdown, model, surface_condition, U₁)
```

six arguments, but this method takes five, so `ConstantWindSlowdown` is a `MethodError`. Since `ConstantWindSlowdown` is the fallback everyone will reach for when the learned scheme misbehaves, it needs to work:

```suggestion
@propagate_inbounds function calc_wind_slowdown(ij, vars, ws::ConstantWindSlowdown, model, surface_condition, U₁)
```

SpeedyWeather/src/parameterizations/surface_fluxes/surface_condition.jl
```diff
@@ -1,21 +1,167 @@
...
+    return ws.wind_slowdown
+end
+
+"""Calculates the near-surface wind speed from the lowest model
+level wind speed using surface roughness. Fields are $(TYPEDFIELDS)"""
+@kwdef struct LearnedWindSlowdown{NF} <: AbstractParameterization
```
milankl-claude, line 18: This is the part I'd most like to see reworked before merge, for three reasons.

**1. No parameters.** `LearnedWindSlowdown{NF}` has no fields, and all ~15 fitted coefficients are hardcoded literals in the function body. That means they can't be tuned, can't be calibrated, and aren't visible in `model.boundary_layer`. We already have exactly this situation solved one file over — `NeutralWindSpeed` is `@parameterized` with `@param c1::NF = -0.039317116` etc. Please follow that precedent here (and in `LearnedLandRoughness` / `LearnedOceanRoughness` / `Charnock`). It also makes the empty-struct `Adapt.adapt_structure` workaround on line 22 unnecessary.

**2. Float32 / Metal.** Every literal is a `Float64`. With `NF = Float32`, `0.83675 - (0.049608 * log_z₀_ocean::Float32)` promotes the whole expression to `Float64`, so this function returns `Float64` regardless of `NF`. On CUDA that's a silent ~30× slowdown on the transcendentals; on Metal there is no `Float64` at all and the kernel won't compile. Making them `@param ::NF` fields fixes this at the source.

**3. Naming.** `LearnedWindSlowdown` sits next to `NeutralWindSpeed`, which is also a symbolic-regression fit but named for what it computes, not how it was obtained. Consider `SurfaceLayerWindSlowdown` or similar — "learned" will age badly once we have three of these. Same thought for `LearnedSurfaceRoughness`.

SpeedyWeather/src/parameterizations/surface_fluxes/surface_condition.jl
```diff
@@ -1,21 +1,167 @@
...
+
+    # Handle initialisation since z_₀ is not yet set
+    if !(z₀_raw_ocean > 0) || !(z₀_raw_land > 0)
+        # 0.1m typical for land, 0.0001m typical for ocean
+        z₀_land = convert(typeof(z₀_raw_land), 0.1)
+        z₀_ocean = convert(typeof(z₀_raw_ocean), 0.0001)
+    else
+        z₀_land = z₀_raw_land
+        z₀_ocean = z₀_raw_ocean
+    end
```
milankl-claude, line 47: **Bug.** This fallback fires far more often than intended. At a pure ocean point `land.momentum_roughness[ij]` is unconditionally set to zero by `surface_roughness!(::LearnedLandRoughness, …)`, so `!(z₀_raw_land > 0)` is always true, and the `||` then discards the *correctly computed ocean* roughness and substitutes `1e-4`. Symmetrically over land. So outside mixed land/ocean cells the learned roughness never actually reaches the wind slowdown — which would explain why the slowdown looks insensitive to the new scheme.

Independent fallbacks per surface:

```suggestion
    # roughness is filled by the surface roughness scheme; fall back per surface where it is
    # still zero (first time step, or the other surface type is absent at this grid point)
    z₀_land = z₀_raw_land > 0 ? z₀_raw_land : convert(typeof(z₀_raw_land), 0.1)
    z₀_ocean = z₀_raw_ocean > 0 ? z₀_raw_ocean : convert(typeof(z₀_raw_ocean), 1.0e-4)
```

Better still, make `0.1` and `1e-4` `@param` fields, and initialise `land.momentum_roughness` / `ocean.momentum_roughness` to them in `initialize!` so the branch disappears from the hot loop entirely.

SpeedyWeather/src/parameterizations/surface_fluxes/surface_condition.jl
```diff
@@ -1,21 +1,167 @@
...
+
+    T₁ = get_prognostic_step(vars.grid.temperature, time_stepping, ws)[ij, nlayers]
+    q₁ = hasproperty(vars.grid, :humidity) ? 
+         get_prognostic_step(vars.grid.humidity, time_stepping, ws)[ij, nlayers] : zero(T₁)
```
milankl-claude, line 51: `hasproperty` — the convention for `vars.grid` in this codebase is `haskey`, and `surface_condition!` further down in this same PR uses `haskey(vars.grid, :humidity)`. Also trailing whitespace on this and lines 52, 57 and 61; Runic will want those gone.

```suggestion
    q₁ = haskey(vars.grid, :humidity) ?
        get_prognostic_step(vars.grid.humidity, time_stepping, ws)[ij, nlayers] : zero(T₁)
```

SpeedyWeather/src/parameterizations/surface_fluxes/surface_condition.jl
```diff
@@ -1,21 +1,167 @@
...
+    Tᵥ_skin_ocean = virtual_temperature(T_skin_ocean, q_skin_ocean, atmosphere)
+    Δθᵥ_land = (Tᵥ₁ + (gravity / cₚ) * z₁) - Tᵥ_skin_land
+    Δθᵥ_ocean = (Tᵥ₁ + (gravity / cₚ) * z₁) - Tᵥ_skin_ocean
+
+    log_z₀_ocean = log(z₀_ocean)
+    log_z₀_land = log(z₀_land)
+
+    sqrt_abs(x) = sqrt(abs(x))
+    ocean_slowdown = (((1.4852 - ((0.83675 - ((0.10807 * 0.45903) * log_z₀_ocean)) ^ log_z₀_ocean)) + (-0.13767 * exp((sqrt_abs(log_z₀_ocean) * log_z₀_ocean) + log(2.9248)))) - (-0.13038 * log_z₀_ocean)) * (((z₁ * ((exp(log(U₁ * 0.07611)) ^ -1.1987) + 0.97005)) - Δθᵥ_ocean) * 0.01083)
+    land_slowdown = -0.18860915 - ((sqrt_abs(exp(log_z₀_land * -0.7117916)) * (sqrt_abs(log_z₀_land) - exp(1.2022188))) * (U₁ ^ -0.99909467))
```
milankl-claude, line 72: A few things about these two expressions.

**Unbounded output.** `land_slowdown` is not constrained to `(0, 1]`. Plugging in `z₀_land = 0.1` it evaluates to roughly `4.11/U₁ - 0.189`, which crosses zero around `U₁ ≈ 22 m/s` and goes *negative* above that — the surface wind then points opposite the lowest-level wind, and the surface momentum flux changes sign in exactly the storms where you least want that. It also blows up as `U₁ → 0` (`U₁^-0.999`). Please clamp, e.g. `clamp(slowdown, ws.min_slowdown, ws.max_slowdown)` with sensible `@param` defaults, and guard `U₁` with a `max(U₁, ws.min_wind)`.

**Redundant work.** `exp(log(U₁ * 0.07611))` is just `U₁ * 0.07611`, and `log(2.9248)` is a compile-time constant that's being recomputed per grid point per time step. `exp(log_z₀_land * -0.7117916)` is `z₀_land^-0.7117916`. Between this and the roughness functions we're at ~20 `exp`/`log`/`^` per grid point per time step in the boundary layer alone — worth counting. Also, both branches are evaluated at every point and then blended by `land_fraction`, so ocean points pay for the land fit and vice versa.

**Float64.** As above, every literal here promotes the result to `Float64`.

**Readability.** These are unreadable as one-liners. Even with the coefficients as fields, please break them into named intermediates and add a comment with the functional form the regression produced — otherwise nobody (including you in six months) can check them.

SpeedyWeather/src/parameterizations/surface_fluxes/surface_condition.jl
```diff
@@ -1,21 +1,167 @@
...
+    @param β::NF = 1.2
+
+    "[OPTION] Boundary layer depth for free convection scaling [m]"
+    @param z_i::NF = 1000.0
+
+    "[OPTION] Typical heat transfer coefficient for scaling [1]"
+    @param C_H::NF = 1.0e-3
+
+    "[OPTION] Minimum background gustiness [m/s]"
+    @param min_gust::NF = 0.0
```
milankl-claude, line 103: `min_gust = 0` is a meaningful change from the old `gust_speed = 1` default: in neutral or stable conditions `ΔTᵥ = 0`, so the gust is exactly zero and `surface_wind_speed` collapses to `|u_s|`. Downstream that's a `0` in the denominator of `bulk_richardson_surface` (`Θ₀ * Vₛ^2`) and in `exp(h_c4 / uₙ)` in `ocean_heat_roughness`. Given the sensitivity to gust speed you and Milan found in the thread, I'd keep a small non-zero floor (0.5–1 m/s) as the default rather than 0.

Also `C_H = 1e-3` is a hardcoded stand-in for the heat drag coefficient we now actually compute (`boundary_layer_drag_heat`). Worth a `# TODO` noting that, once the ordering dependency is sorted, this should use the real value.

SpeedyWeather/src/parameterizations/surface_fluxes/surface_condition.jl
```diff
@@ -1,21 +1,167 @@
...
+    return gust
+end
 
 """Surface condition parameterization that calculates near-surface atmospheric
 variables needed for surface flux calculations. Computes surface wind speed
 including sub-grid scale gusts, surface air density, and surface air temperature
 by extrapolating from the lowest model level to the surface using standard
 atmospheric relationships. Fields are $(TYPEDFIELDS)"""
-@kwdef struct SurfaceCondition{NF} <: AbstractParameterization
+@kwdef struct SurfaceCondition{NF, WS, WG} <: AbstractParameterization
```
milankl-claude, line 139: `NF` is no longer used by any field, and that breaks GPU adaptation. `Adapt.@adapt_structure` generates `SurfaceCondition(adapt(to, wind_slowdown), adapt(to, wind_gust))` — the bare-name positional constructor — but Julia does not generate an outer constructor when a type parameter can't be inferred from the field types. Verified:

```julia
julia> Base.@kwdef struct SC{NF, WS, WG}; wind_slowdown::WS; wind_gust::WG; end
julia> SC(1.0, 2.0)
ERROR: MethodError: no method matching SC(::Float64, ::Float64)
```

So `adapt(CuArray, model)` throws. (You hit the same thing with `LearnedWindSlowdown`, which is why it needed a hand-written `adapt_structure`.) Simplest fix is to drop the unused parameter, which also lets the outer constructor collapse to one line:

```suggestion
@kwdef struct SurfaceCondition{WS, WG} <: AbstractParameterization
```

and then

```julia
function SurfaceCondition(
        SG::SpectralGrid;
        wind_slowdown = ConstantWindSlowdown(SG),
        wind_gust = BeljaarsWindGust(SG),
    )
    return SurfaceCondition(wind_slowdown, wind_gust)
end
```

Also: `SurfaceCondition` and `LearnedSurfaceRoughness` are composites of sub-components but are plain `@kwdef struct` — they should be `@parameterized` with `@component` fields like `BoundaryLayer` is, otherwise the nested parameters are invisible to the parameter interface.

SpeedyWeather/src/parameterizations/surface_fluxes/surface_roughness.jl
```diff
@@ -1,41 +1,357 @@
...
+
+    # Heat roughness
+    land.heat_roughness[ij] = ifelse(land_fraction > 0, z₀H_land, zero(z₀H_land))
+    ocean.heat_roughness[ij] = ifelse(land_fraction < 1, z₀H_ocean, zero(z₀H_ocean))
+    vars.parameterizations.heat_roughness[ij] = land_fraction * land.heat_roughness[ij] +
+        (1 - land_fraction) * ocean.heat_roughness[ij]
+
+    # Moisture roughness (just heat times a constant)
+    land.moisture_roughness[ij] = ifelse(land_fraction > 0, z₀H_land * (0.62 / 0.4), zero(z₀H_land))
+    ocean.moisture_roughness[ij] = ifelse(land_fraction < 1, z₀H_ocean * (0.62 / 0.4), zero(z₀H_ocean))
```
milankl-claude, line 73: `0.62 / 0.4` is the same `a_q / a_h` pair that appears as named fields on `LearnedOceanRoughness` (line 147). Hardcoding it here means the two schemes can silently drift apart, and `ConstantSurfaceRoughness` loses the ability to set the moisture roughness directly — which is a small regression given it's the "I want full control" option.

I'd add explicit options instead:

```julia
"[OPTION] constant humidity roughness length over land [m]"
@param humidity_roughness_land::NF = 0.5

"[OPTION] constant humidity roughness length over ocean [m]"
@param humidity_roughness_ocean::NF = 1.55e-4
```

Minor: all three blocks here repeat the same land/ocean/blend pattern — a small local helper taking `(land_field, ocean_field, z_land, z_ocean)` would cut this to a third of the length.

SpeedyWeather/src/parameterizations/surface_fluxes/surface_roughness.jl
```diff
@@ -1,41 +1,357 @@
...
+    vars.parameterizations.moisture_roughness[ij] = land_fraction * land.moisture_roughness[ij] +
+        (1 - land_fraction) * ocean.moisture_roughness[ij]
+
+    return nothing
+end
+
+@kwdef struct LearnedLandRoughness{NF} <: AbstractSurfaceRoughness
```
milankl-claude, line 80: None of the new schemes are `@parameterized`, and none of the ~30 fitted coefficients across `LearnedLandRoughness`, `LearnedOceanRoughness` and `Charnock` are `@param`. For a *learned* scheme that seems backwards — these are precisely the numbers we'd want to recalibrate, and `NeutralWindSpeed` in `boundary_layer.jl` already shows the pattern:

```julia
@parameterized @kwdef struct LearnedLandRoughness{NF} <: AbstractSurfaceRoughness
    "[OPTION] symbolic regression parameter 1 for momentum roughness over land [1]"
    @param m_c1::NF = -7.231684
    ...
```

Also: only `m_c1` and `h_c1` carry a docstring, so `$(TYPEDFIELDS)` renders eight undocumented fields. And there's no docstring on the struct itself (`ConstantSurfaceRoughness` has one). Since `m_c1…m_c9` are inherently opaque, please put the functional form the regression produced into the struct docstring so a reader can map coefficient → term.

Finally, `LearnedLandRoughness`, `LearnedOceanRoughness` and `Charnock` aren't exported, so users can't write `LearnedOceanRoughness(spectral_grid, charnock_param = …)` without reaching into the module.

SpeedyWeather/src/parameterizations/surface_fluxes/surface_roughness.jl
```diff
@@ -1,41 +1,357 @@
...
+    c3 = T(charnock_param.c3)
+    c4 = T(charnock_param.c4)
+    c5 = T(charnock_param.c5)
+
+    val = ifelse(
+        uₙ ≤ Uₜ,
+        muladd(uₙ * uₙ, muladd(c1, uₙ, c2), c3),
+        muladd(c4, uₙ - Uₜ, c5) # for uₙ > Uₜ, linear increase
+    )
+    return log(val)
```
milankl-claude, line 143: `Charnock <: AbstractParameterization` but implements none of the parameterization interface (no `parameterization!`, no `variables`) — it's used purely as a functor. I'd either subtype something narrower or leave it as a plain struct, otherwise it looks like something you can drop into a model.

Also the method returns `log(val)`, not the Charnock parameter — and the call site names the result `αᵪ`, which then gets used as `muladd(scheme.m_c7, αᵪ, -scheme.m_c8)`. Please rename to `log_αᵪ` (both here and at the call site) or return `val` and take the log where it's needed; as written it's a trap for the next person tuning `m_c7`.

SpeedyWeather/src/parameterizations/surface_fluxes/surface_roughness.jl
```diff
@@ -1,41 +1,357 @@
...
+    end
+
+    # Unpack predictors from the land parameterization
+    vₕ = vars.parameterizations.land.high_vegetation_cover[ij]
+    vₗ = vars.parameterizations.land.low_vegetation_cover[ij]
+    laiₕ = vars.parameterizations.land.high_vegetation_leaf_area[ij]
+    laiₗ = vars.parameterizations.land.low_vegetation_leaf_area[ij]
+    sd = vars.prognostic.land.snow_depth[ij]
+    soil_moisture = vars.prognostic.land.soil_moisture[ij, begin]  # currently top layer
+    soil_temperature = vars.prognostic.land.soil_temperature[ij, end]  # currently bottom layer
```
milankl-claude, line 246: **Compatibility:** these six reads are unguarded, and `LearnedSurfaceRoughness` is now the *default* `surface_roughness` for every `BoundaryLayer`. But `high_vegetation_cover` / `low_vegetation_cover` / `high_vegetation_leaf_area` / `low_vegetation_leaf_area` are only registered by `VegetationClimatology` — `NoVegetation` (see `variables(::NoVegetation)`) registers none of them. Likewise `snow_depth` is optional (`heat.jl` guards it with `haskey(vars.prognostic.land, :snow_depth)`) and `sea_ice_concentration` is optional (`humidity.jl` guards it too).

So `PrimitiveWetModel(spectral_grid, land = LandModel(spectral_grid, vegetation = NoVegetation()))` now fails on the default settings, and so does anything without sea ice. Either guard each read the way the flux schemes do, or make `LearnedSurfaceRoughness` require its inputs explicitly and keep `ConstantSurfaceRoughness` as the default until the dependencies are declared.

On the default change generally: flipping `surface_roughness`, `neutral_wind_speed`, `wind_slowdown` and `wind_gust` all to new schemes in one PR changes every simulation's climate. Given the SST/ITCZ oddities in the thread, I'd merge the machinery with the old defaults and switch the defaults in a follow-up once the flux balance is understood — that also makes the regression bisectable.

Minor: `soil_temperature[ij, end]` is the *bottom* soil layer — a deep-soil temperature as a predictor for surface roughness is surprising. The `# currently bottom layer` comment suggests it may be a placeholder; if the fit really was trained on the bottom layer, please say so in the comment, otherwise `begin` looks more likely to be what you want.

SpeedyWeather/src/parameterizations/surface_fluxes/surface_roughness.jl
```diff
@@ -1,41 +1,357 @@
...
-    vars.parameterizations.surface_roughness[ij] = land_fraction * land.surface_roughness[ij] +
-        (1 - land_fraction) * ocean.surface_roughness[ij]
     return nothing
 end
+
+@inline function sea_ice_momentum_roughness(siconc::NF) where {NF <: Real}
+    # Following IFS documentation
+    base = NF(0.93e-3) * (1 - siconc) + NF(6.05e-3) * exp(NF(-17) * (siconc - 0.5)^2)
+    return max(base, NF(1.0e-3))
+end
```
milankl-claude, line 289: **Float32/Metal:** `0.5` here is a `Float64` literal, so with `NF = Float32` the expression `(siconc - 0.5)^2` promotes and this function returns `Float64`. Same class of issue as in `LearnedWindSlowdown`. The `NF(...)` wrappers on the other constants show the intent — this one just got missed:

```suggestion
@inline function sea_ice_momentum_roughness(siconc::NF) where {NF <: Real}
    # Following IFS documentation
    base = NF(0.93e-3) * (1 - siconc) + NF(6.05e-3) * exp(NF(-17) * (siconc - NF(0.5))^2)
    return max(base, NF(1.0e-3))
end
```

More generally: hardcoded magic numbers in `sea_ice_momentum_roughness` (`0.93e-3`, `6.05e-3`, `17`), `maximum_momentum_roughness` (`6.74e-3`, `1.3e-3`, `33`, `55`) and `ocean_heat_roughness` (`1.0e-3` sea-ice roughness) should all be `@param` fields on the owning scheme — same argument as for the regression coefficients, plus it removes the promotion hazard by construction. The `1.0e-3` ice roughness in particular is duplicated between `sea_ice_momentum_roughness` and `ocean_heat_roughness`.

SpeedyWeather/src/parameterizations/surface_fluxes/surface_roughness.jl
```diff
@@ -1,41 +1,357 @@
...
+    term4 = min(L_H, W)
+
+    denominator = term3 + term4
+
+    log_roughness = (numerator / denominator) + scheme.m_c9
+    return exp(log_roughness)
+end
+
+@inline function land_heat_roughness(scheme, C_H::NF, C_L::NF, L_H::NF, L_L::NF, T::NF, W::NF) where {NF <: Real}
+    gamma_term1 = ((scheme.h_c1^scheme.h_c2) * min(C_H, L_H)) / ((L_L + T)^scheme.h_c2) + L_H
```
milankl-claude, line 347: `scheme.h_c1^scheme.h_c2` is `373.86^5.98 ≈ 2.4e15`, recomputed at every grid point at every time step, and then divided by `(L_L + T)^5.98 ≈ 5e14`. Two `pow`s and a huge intermediate to get a ratio of order 5.

Algebraically identical, one `pow`, no large intermediate, and far safer in Float32:

```suggestion
    gamma_term1 = (scheme.h_c1 / (L_L + T))^scheme.h_c2 * min(C_H, L_H) + L_H
```

(If `h_c1` is ever tuned upward this matters more than it looks — `Float32` tops out at `3.4e38`, and `h_c1^h_c2` is already within ~23 orders of it.)

SpeedyWeather/src/parameterizations/surface_fluxes/surface_roughness.jl
```diff
@@ -1,41 +1,357 @@
...
+    # Weight by sea ice concentration
+    return sea_ice_momentum_roughness(siconc) * siconc + (1 - siconc) * z₀M
+end
+
+@inline function ocean_heat_roughness(scheme, uₙ::NF, siconc::NF) where {NF <: Real}
+    log_roughness = (scheme.h_c1 * ((uₙ + scheme.h_c2)^scheme.h_c3)) - exp(scheme.h_c4 / uₙ)
+    return NF(1.0e-3) * siconc + (1 - siconc) * exp(log_roughness) # following IFS documentation, ice roughness is 1e-3 m
+end
+
+@inline function land_momentum_roughness(scheme, C_H::NF, C_L::NF, L_H::NF, L_L::NF, S::NF, T::NF, W::NF) where {NF <: Real}
```
milankl-claude, line 325: These signatures over-constrain: `where {NF <: Real}` forces `uₙ`, `αᵪ` and `siconc` to be the *same* concrete type. `surface_wind_speed` and `sea_ice_concentration` come from different variable groups, and the moment anything is mixed-precision (or a `ForwardDiff.Dual` shows up from the parameter interface) this is a `MethodError` rather than a promotion. `land_momentum_roughness` is worse — seven arguments all pinned to one type, including `soil_temperature` in Kelvin and `snow_depth` in metres.

Prefer `::Real` arguments and let promotion do its job, or `promote(...)` explicitly at the top. Also, `scheme` is untyped in all four — please annotate (`scheme::LearnedOceanRoughness`, `scheme::LearnedLandRoughness`) both for dispatch safety and so the reader knows which coefficient set is expected.

Naming: `C_H`, `C_L`, `L_H`, `L_L`, `S`, `T`, `W` are hard to follow at the call site, where they're passed as `vₕ, vₗ, laiₕ, laiₗ, sd, soil_temperature, soil_moisture`. I'd use the descriptive names in both places.

SpeedyWeather/src/parameterizations/surface_fluxes/surface_roughness.jl
```diff
@@ -1,41 +1,357 @@
...
+end
+
+Adapt.@adapt_structure LearnedOceanRoughness
+function LearnedOceanRoughness(SG::SpectralGrid; charnock_param = Charnock(SG), kwargs...)
+    return LearnedOceanRoughness{SG.NF, typeof(charnock_param)}(; charnock_param, kwargs...)
+end
+initialize!(::LearnedOceanRoughness, ::PrimitiveEquation) = nothing
+
+"""
+Learned ocean surface roughness for momentum, heat and moisture fluxes."""
```
milankl-claude, line 177: The docstring says "Learned **ocean** surface roughness" but this type is the land+ocean composite. Also `LearnedLandRoughness`, `LearnedOceanRoughness` and `Charnock` have no docstrings at all, which the `@parameterized`/`TYPEDFIELDS` docs convention in this repo expects.

Terminology: this PR uses `moisture_roughness` and `boundary_layer_drag_humidity` for the same physical pair, and `a_q` for the coefficient. Per Milan's earlier comment on #959, we call the vapour-phase quantity *humidity* (specific humidity) and reserve *moisture* for liquid soil moisture. Suggest `humidity_roughness` throughout, which also makes it consistent with `boundary_layer_drag_humidity`, `SurfaceHumidityFlux` and `surface_humidity_flux`.

SpeedyWeather/src/output/variables/boundary_layer.jl
```diff
@@ -150,55 +150,55 @@ end
...
     missing_value::Float64 = NaN
     compression_level::Int = 3
     shuffle::Bool = true
     keepbits::Int = 7
 end
 
-path(::OceanSurfaceRoughnessOutput, simulation) =
-    simulation.variables.parameterizations.ocean.surface_roughness
+path(::MoistureBoundaryLayerDragOutput, simulation) =
+    simulation.variables.parameterizations.boundary_layer_drag_moisture
```
milankl-claude, line 197: Typo — the variable is registered as `boundary_layer_drag_humidity` (see `variables(::BulkRichardsonDrag)`), so any run with `BoundaryLayerOutput()` errors here. Since `MoistureBoundaryLayerDragOutput` is in the default `BoundaryLayerOutput()` tuple below, that's every run with boundary layer output.

If you take the `moisture → humidity` renaming suggested in `surface_roughness.jl`, this becomes `HumidityBoundaryLayerDragOutput` / `bld_q` / `boundary_layer_drag_humidity`.

```suggestion
    simulation.variables.parameterizations.boundary_layer_drag_humidity
```

SpeedyWeather/src/output/variables/boundary_layer.jl
```diff
@@ -150,55 +150,55 @@ end
...
 
 """Defines netCDF output for a specific variables, see [`VorticityOutput`](@ref) for details.
 Fields are: $(TYPEDFIELDS)"""
-@kwdef mutable struct SurfaceRoughnessOutput <: AbstractOutputVariable
+@kwdef mutable struct MomentumSurfaceRoughnessOutput <: AbstractOutputVariable
     name::String = "sr"
     unit::String = "m"
     long_name::String = "Surface roughness length"
```
milankl-claude, line 204: The drag outputs became `bld_m` / `bld_h` / `bld_q`, but momentum roughness kept the bare `sr` while heat roughness got `sr_h`. Inconsistent, and it means `sr` silently changes meaning from "the surface roughness" to "the momentum roughness". I'd rename for symmetry:

```suggestion
@kwdef mutable struct MomentumSurfaceRoughnessOutput <: AbstractOutputVariable
    name::String = "sr_m"
    unit::String = "m"
    long_name::String = "Momentum surface roughness length"
```

Two more in this file:
- `LandSurfaceRoughnessOutput` (`lsr`) and `OceanSurfaceRoughnessOutput` (`osr`) were removed rather than renamed. They're public API — that's a breaking change and needs a CHANGELOG note, or better, keep them pointing at `land.momentum_roughness` / `ocean.momentum_roughness`.
- Line 72 (in `SurfaceTemperatureOutput`/10m-wind path, outside the diff) still reads `simulation.variables.parameterizations.surface_roughness`, which no longer exists. That's a third output-time error.

SpeedyWeather/src/parameterizations/land/vegetation.jl
```diff
@@ -83,27 +83,41 @@ export VegetationClimatology
...
 
 # generator function
 function VegetationClimatology(SG::SpectralGrid, geometry::LandGeometryOrNothing = nothing; kwargs...)
     (; NF, GridVariable2D, grid) = SG
     high_cover = zeros(GridVariable2D, grid)
     low_cover = zeros(GridVariable2D, grid)
-    return VegetationClimatology{NF, GridVariable2D}(; high_cover, low_cover, kwargs...)
+    high_lai = zeros(GridVariable2D, grid)
+    low_lai = zeros(GridVariable2D, grid)
+    return VegetationClimatology{NF, GridVariable2D}(; high_vegetation_cover = high_cover, low_vegetation_cover = low_cover, high_vegetation_leaf_area = high_lai, low_vegetation_leaf_area = low_lai, kwargs...)
```
milankl-claude, line 120: Two things.

**Asset version.** `lai_hv` / `lai_lv` are new variables being read out of `vegetation.nc` at `DEFAULT_ASSETS_VERSION`. If the currently pinned asset release doesn't contain them, `initialize!` fails for everyone the moment `VegetationClimatology` is constructed — and it's the default vegetation. Please confirm the asset version has been bumped (and bump `DEFAULT_ASSETS_VERSION` in this PR if not); a CI run would catch it but the checks on this PR are all skipped.

**Formatting.** This line is ~190 characters; Runic won't split it for you but it's well past what the rest of the file does. Also the local variable names (`high_cover`, `high_lai`) no longer match the field names:

```suggestion
    high_vegetation_cover = zeros(GridVariable2D, grid)
    low_vegetation_cover = zeros(GridVariable2D, grid)
    high_vegetation_leaf_area = zeros(GridVariable2D, grid)
    low_vegetation_leaf_area = zeros(GridVariable2D, grid)
    return VegetationClimatology{NF, GridVariable2D}(;
        high_vegetation_cover,
        low_vegetation_cover,
        high_vegetation_leaf_area,
        low_vegetation_leaf_area,
        kwargs...,
    )
```

(the four `zeros` lines above, 116–119, get replaced too)

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Learned land and ocean surface roughness [1184](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1184)
```
milankl-claude, line 5: Missing the `#` — GitHub won't render the PR link without it. Also I'd split this into the three user-visible changes, since two of them are breaking:

```suggestion
- Learned land and ocean surface roughness [#1184](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1184)
- [BREAKING] `boundary_layer_drag` and `surface_roughness` split into momentum/heat/humidity variants, output variables renamed [#1184](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1184)
- [BREAKING] `BoundaryLayer` defaults changed to `LearnedSurfaceRoughness`, `NeutralWindSpeed`, `LearnedWindSlowdown` and `BeljaarsWindGust` [#1184](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1184)
```

PR #1185, opened 2026-08-08
Further GPU optimisations: `Feedback` and FFT plans 

maximilian-gelbrecht:


maximilian-gelbrecht:
Two main culprits were here: 
* the batched FFT plans for K=2 not existing which forced two quite slow 2D transforms in the regular FFT without graphs. These were a major bottleneck 
* The feedback printing forcing a `synchronize!` -> now only done every 50 step per default  

So far, timing for `time_step!` 

| trunc | H100 before | H100 after | speedup | A40 before | A40 after | speedup |
|------:|------------:|-----------:|--------:|-----------:|----------:|--------:|
| 31  | 2.760  | **1.126**  | **2.45×** | 3.590 | **1.277** | **2.81×** |
| 63  | 5.120  | **2.138**  | **2.39×** | 6.256 | **2.386** | **2.62×** |
| 127 | 10.485 | **4.777**  | **2.20×** | — | — | — |
| 255 | 23.833 | **14.280** | **1.67×** | — | — | — |

GPU busy time 66-78% (up from just 25%)

Some other insights from the profiling

| phase | busy ms/step | share | kernels/step |
|---|---:|---:|---:|
| `dynamics` (includes the transforms it calls) | 5.574 | 53 % | 1414 |
| `transform` | 4.552 | 44 % | 1387 |
| `parameterizations` | 0.213 | 2 % | 7 |
| everything else | 0.126 | 1 % | 35 |

The count of kernels/step includes the kernels that are recorded by the graphs (and those are the vast majority)

maximilian-gelbrecht:
I only wait for unit tests and the benchmark on the H100 to finish

SpeedyWeather/benchmark/README.md
```diff
@@ -37,30 +37,30 @@ Simulated years per wallclock day (SYPD) for the `PrimitiveWetModel` resolution
...
+| 127 | 16 | LT+FFT | 21 | 15 | 243 |
+| 127 | 16 | MT | 2.1 | 0.2 | 15 |
+| 127 | 24 | LT+FFT | 15 | 11 | 220 |
+| 127 | 24 | MT | 1.7 | 0.1 | 9.5 |
+| 170 | 8 | LT+FFT | 5.5 | 3.9 | 138 |
+| 170 | 16 | LT+FFT | 7.7 | 5.6 | 122 |
+| 170 | 24 | LT+FFT | 4.3 | 4.5 | 110 |
+| 255 | 8 | LT+FFT | 1.4 | 1.0 | 52 |
+| 255 | 16 | LT+FFT | 1.9 | 1.3 | 45 |
+| 255 | 24 | LT+FFT | 1.5 | 1.0 | 39 |
```
maximilian-gelbrecht, line 63: These seem to be actual true results now. I wonder if one of the other PRs helped?

SpeedyWeather/benchmark/README.md
```diff
@@ -397,94 +397,94 @@ Toolchain:
 
 | Model | T | L | Physics | Δt | SYPD | Memory|
 | --- | --- | --- | --- | --- | --- | --- |
-| BarotropicModel | 31 | 1 | false | 1800 | 2187 | 489.56 KB |
-| ShallowWaterModel | 31 | 1 | false | 2400 | 1787 | 493.06 KB |
-| PrimitiveDryModel | 31 | 8 | true | 2400 | 322 | 649.18 KB |
-| PrimitiveWetModel | 31 | 8 | true | 2400 | 2366 | 654.31 KB |
+| BarotropicModel | 31 | 1 | false | 1800 | 2124 | 489.57 KB |
+| ShallowWaterModel | 31 | 1 | false | 2400 | 1838 | 493.08 KB |
+| PrimitiveDryModel | 31 | 8 | true | 2400 | 363 | 661.18 KB |
```
maximilian-gelbrecht, line 402: The dry model is still very slow. But we can look into that in a separate issue. 

My guess: the CUDA graphs are not working correctly for it.

SpeedyWeather/benchmark/README.md
milankl: You mean gained stability doesn't make these crash anymore and so additional layers are sensibly slowing performance down a bit instead of blowing up?

SpeedyWeather/src/output/feedback.jl
```diff
@@ -16,8 +16,8 @@ $(TYPEDFIELDS)"""
     "[OPTION] Progress description"
     description::String = ""
 
-    "[OPTION] Progress bar length, nothing = full window width"
-    progress_bar_length::Int = 0
+    "[OPTION] Progress bar length in characters, `nothing` fits it to the terminal width, `0` shows no bar"
+    progress_bar_length::Union{Int, Nothing} = nothing
```
milankl: I had deliberately taken out the bar, so length = 0 as the percentage is still shown and in the future I wanted to modularize what's printed in the progress, e.g. other stats

maximilian-gelbrecht: Did not know that. Happy to switch the default back to 0

SpeedyWeather/benchmark/README.md
maximilian-gelbrecht: Yes exactly. We still don't have a proper NaN check in the benchmarks, but the numbers seem to be what you'd expect for a stable rollout.

SpeedyWeather/src/output/feedback.jl
maximilian-gelbrecht: Ok done

SpeedyWeather/benchmark/README.md
milankl: Could that be related to fusing variables, given that humidity does not exist in that model?

SpeedyWeather/src/output/feedback.jl
milankl: awesome!

SpeedyWeather/benchmark/README.md
milankl: great!

milankl (approved):

SpeedyWeather/benchmark/README.md
maximilian-gelbrecht: Yes both combined. So the batching not correctly set so that the transforms without CUDA graphs are then used. But I don't know. This just seems most likely.

PR #1186, opened 2026-08-10
Add `PrescribedAlbedo` for coupling 

maximilian-gelbrecht:
Adds a prescribed albedo for coupling as wished for by @bgroenks96 

The difference to `ManualAlbedo` is that here the user is expected to copy the albedo into the `Variables`.


milankl (approved):
looks good to me!

PR #1188, opened 2026-08-10
Add snow to Terrarium-SpeedyWeather coupling

### LINKED ISSUE #1189: `SpeedyWeather.animate` fails for soil temperature output
There is a longstanding bug in `animate` that prevents it from working for soil temperature (`st`) output. I've known about this for a while but apparently never made an issue (sorry). The problem is that soil layers are declared as a separate coordinate in NetCDF files `soil_layer` and thus the following line in [`SpeedyWeatherGeoMakieExt`](https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/main/SpeedyWeather/ext/SpeedyWeatherGeoMakieExt.jl#L98),

```julia
# Check if the variable is 3D (has a vertical dimension)
is_3d = "layer" in dimnames(ds[variable])
```

fails to pick it up and then subsequently causes an indexing error. The fix is imply to include `"soil_layer"` in the above conditional.

Since this is a minor change and is relevant to Terrarium-Speedy coupling, I will just include the fix for this in #1188.

bgroenks96:
Updates `SpeedyWeatherTerrariumExt` to include snow depth water equivalent (SWE) and albedo. Also fixes broken GPU compatibility.

## TODO
- [x] Add SWE coupling (when `snow` is present in `LandModel`)
- [x] Add albedo coupling using #1186 
- [x] Fix GPU compat
- [x] Bonus finally fix #1189

### DIFF
--- a/SpeedyWeather/ext/SpeedyWeatherTerrariumExt/coupling.jl
+++ b/SpeedyWeather/ext/SpeedyWeatherTerrariumExt/coupling.jl
@@ -435,7 +435,7 @@ function SpeedyWeather.initialize!(
 
     # fill ocean/non-Terrarium points with fallback values first, then seed the land columns
     fill_fallback!(vars, land)
-    RingGrids.copy_unmasked!(vars.prognostic.land.soil_temperature, interior(state.skin_temperature) .+ NF(273.15), indices)
+    RingGrids.copy_unmasked!(vars.prognostic.land.soil_temperature, @view(interior(state.temperature)[:, 1, end]) .+ NF(273.15), indices)
     return nothing
 end
 
--- a/SpeedyWeather/ext/SpeedyWeatherTerrariumExt/coupling.jl
+++ b/SpeedyWeather/ext/SpeedyWeatherTerrariumExt/coupling.jl
@@ -247,7 +247,7 @@ function TerrariumLand(
     )
 end
 
-function SpeedyWeather.variables(::AbstractTerrariumLandModel)
+function SpeedyWeather.variables(land_model::AbstractTerrariumLandModel)
     return (
         # The full Terrarium state, owned by SpeedyWeather's Variables tree.
         SpeedyWeather.PrognosticVariable(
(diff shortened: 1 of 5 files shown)


SpeedyWeather/ext/SpeedyWeatherTerrariumExt/coupling.jl
```diff
@@ -365,11 +373,8 @@ function SpeedyWeather.timestep!(
         RingGrids.copy_unmasked!(vars.prognostic.land.surface_humidity_flux, state.latent_heat_flux, indices)
         vars.prognostic.land.surface_humidity_flux.data ./= consts.thermodynamics.latent_heat_vaporization
     end
-    if haskey(vars.parameterizations, :surface_longwave_up)
-        RingGrids.copy_unmasked!(vars.parameterizations.surface_longwave_up, state.surface_longwave_up, indices)
```
maximilian-gelbrecht, line 369: This is removed because it's not used anyway, correct? As part of the whole discussion about the order how parameterisations and the coupling are executed.

maximilian-gelbrecht (approved):

SpeedyWeather/ext/SpeedyWeatherTerrariumExt/coupling.jl
bgroenks96: Yes it addresses one of the concerns raised in https://github.com/NumericalEarth/Terrarium.jl/issues/148

maximilian-gelbrecht:
We want to release a new (major) version tomorrow, do you still want this in before that? A quick minor release afterwards also shouldn't be a problem though.

bgroenks96:
Ideally yes, I just need to fix the tests...

bgroenks96:
Quick proof that #1189 is fixed:

https://drive.google.com/file/d/12pfa4Pnl98mtB6x_h_z-MNOcvvnUxUgQ/view?usp=sharing

PR #1191, opened 2026-08-13
Update checkpointed sensitivity examples

maximilian-gelbrecht:
We missed this so far. Updates the checkpointed sensitivity examples to the new variables and time stepping syntax.

### DIFF
--- a/SpeedyWeather/test/differentiability/sensitivity_examples/checkpointed_sensitivity.jl
+++ b/SpeedyWeather/test/differentiability/sensitivity_examples/checkpointed_sensitivity.jl
@@ -18,17 +18,14 @@ run!(simulation, period = Day(20))
 # do the scaling and init again because we need it for the timestepping when calling it manually
 initialize!(simulation, steps = N)
 
-(; prognostic_variables, diagnostic_variables, model) = simulation
-(; Δt, Δt_millisec) = model.time_stepping
-dt = 2Δt
+(; variables, model) = simulation
+vars = variables
 
-progn = prognostic_variables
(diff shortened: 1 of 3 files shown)


SpeedyWeather/test/differentiability/sensitivity_examples/checkpointed_sensitivity.jl
```diff
@@ -1,8 +1,17 @@
...
 Pkg.activate("SpeedyWeather/test/differentiability/sensitivity_examples")
 
 using SpeedyWeather, Enzyme, JLD2, Checkpointing
 
+# Enzyme runs the LLVM Attributor pass on Julia < 1.12
+# (`Enzyme.Compiler.RunAttributor = Ref(VERSION < v"1.12")`). Stepping the clock inside the
+# `@ad_checkpoint` loop sends the Attributor's AAPotentialValues analysis into unbounded
+# recursion (`AAPotentialValuesFloating::updateImpl` -> `getAssumedSimplified` -> ... -> itself),
+# which overflows the C++ stack and shows up as a segfault. Disabling the pass avoids it.
+Enzyme.Compiler.RunAttributor[] = false
```
maximilian-gelbrecht, line 13: @vchuravy Through a bit of guided Claude trial and error, I found that the script here just works when I turn this off manually, but I don't actually know what it does. I can't seem to find a lot of docs on it. Without this, this script segfaults. It's a bit hard to condense this to a MWE because the two individual components of the loop that is differentiated do work fine just by themselves. 

What does RunAttributor actually do?

PR #1192, opened 2026-08-13
Add implementation of `cat` for RingGrids Fields

bgroenks96:
Bringing `Field`s together with :cat:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Add implementation of `cat` for `RingGrids` `Field`s [#1192](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1192)
+
 ## v0.22.0
 
 - [BREAKING] 1-based truncation as resolution parameter [#1177](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1177)
--- a/RingGrids/src/field.jl
+++ b/RingGrids/src/field.jl
@@ -571,8 +571,9 @@ function Base.cat(Fs::AbstractField...; dims)
     @assert length(Fs) > 0 "At least one Field must be provided to `cat`"
     @assert all(map(grids_match, Fs[1:(end - 1)], Fs[2:end])) "All concatenated Fields must have matching grids"
     @assert length(dims) < length(first(Fs)) "Too many `dims` specified"
+    # we let the standard implementation of `cat` handle the case where dims is greater than the number of axes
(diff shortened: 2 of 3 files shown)


maximilian-gelbrecht (approved):
Looks good to me!

milankl (approved):
Looks great, the `dims` keyword is always an integer? Just checking that `dims > 1` wouldn't error ...

bgroenks96:
Hmm, some array types like `Rasters`/`DimArray` use non-integer types for `dims`, but that's non-standard for the `Array` interface. I don't think the rest of `RingGrids` would anyway be compatible with this, so I probably wouldn't change it just for that.

Maybe slightly confusing is that `Field` now also has a field `dims` which elsewhere refers to the overall shape (number of axes) rather than the dimension index.

bgroenks96:
Ah but `dims` can be an iterable for multiple dimensions in the standard interface. That we should support.

milankl:
> Maybe slightly confusing is that `Field` now also has a field `dims` which elsewhere refers to the overall shape (number of axes) rather than the dimension index.

~~Eh, what are you referring to? You didn't change the definition of `Field`?~~ Ah yes, we use it as an empty type just to have a flag to know what the dimensions mean. Why would you need to save the number of axes? Isn't that clear from `size(field.data)`?

EDIT: We could rename `field.dims`, it's not supposed to to be public interface for now

bgroenks96:
> Why would you need to save the number of axes? Isn't that clear from size(field.data)?

We don't, I just meant that it might confuse readers a bit to see `dims` used in different ways. But I think it's clear from context.

[2510828](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1192/commits/251082897a8016dc78e5cdf6eb71f19a5590a061) adds support for multi-dimension concat.

milankl:
We forgot to move RingGrids from v0.1.8 to v0.2-DEV. I believe because this adds a new feature this isn't backwards compatible so we need to go to v0.2 with this. If the next pull request going in could do that, that would be fab!

bgroenks96:
Are you referring to the last SpeedyWeather release? Because I don't think there's anything in this PR that would be considered backwards incompatible. Adding a new feature does not automatically make a new version backwards incompatible, only breaking changes do.

The [Julia semantic versioning](https://pkgdocs.julialang.org/v1/toml-files/#The-version-field) standard requires a minor version bump only for breaking changes on 0.x versions, so new features should only get revision bumps.

milankl:
Sorry my bad, got confused with pre-1.0 vs post-1.0. But then `v0.1.8+DEV` so that when we do a new SpeedyWeather release we know RingGrids needs a new release too.

PR #1193, opened 2026-08-14
Work around AMDGPU/LLVM miscompile of triangular loop in implicit kernel

gregordecristoforo:
Implementation of Ludovic's suggested work around for bug reported in https://github.com/JuliaGPU/AMDGPU.jl/issues/1015 for `implicit_primitive_leapfrog_kernel!` I did not find any other kernel that is affected by this pattern.

The bug has already been fixed in https://github.com/llvm/llvm-project/pull/215829 However, it may still take quite some time until it gets integrated it in a new version.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -24,7 +24,7 @@ steps:
     matrix:
       setup:
         version:
-          - "1.12"
+          - "1.11"
     plugins:
       - JuliaCI/julia#v1:
           version: "{{matrix.version}}"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -24,7 +24,7 @@ steps:
     matrix:
       setup:
         version:
-          - "1.11"
+          - "1.12"
     plugins:
       - JuliaCI/julia#v1:
           version: "{{matrix.version}}"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Work around an AMDGPU/LLVM miscompile of a triangular loop with dynamic bounds in the semi-implicit primitive-equation kernel [#1193](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1193)
+- Add implementation of `cat` for `RingGrids` `Field`s [#1192](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1192)
 
 ## v0.22.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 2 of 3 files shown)


gregordecristoforo:
When running the RingGrids test suite locally with Julia 1.12.6 all tests pass, including `Interpolate meridionally-constant field `
```Julia
Test Summary:                           |  Pass  Total  Time
Interpolate meridionally-constant field | 24000  24000  0.4s
```
Also, running `SpeedyWeather/test/GPU/runtests.jl` finally doesn't result in a segfault anymore.

maximilian-gelbrecht (approved):
Ah so it really only was this one loop. 

Yeah, great. Thanks.

maximilian-gelbrecht:
I'll just directly test here if the AMD CI is now also passing in 1.12.

gregordecristoforo:
Thanks! BTW, shall I leave merging to you and Milan or may I do so if the tests pass and you guys approved it?

maximilian-gelbrecht:
Who ever sees all green first, can merge when the PR is approved :) 

With GPU CI, there are generally longer queue times. It's generally okay to ignore the GPU CI if the PR is small and has nothing to do with GPU (in which case we also usually put the tag "skip-gpu-ci" on it), but for everything actually GPU related we wait for it of course.

maximilian-gelbrecht:
The 1.12 AMD CI did fail again, however at a different point. 

I still find it surprising that there seems to be such a difference between 1.12 and 1.11

gregordecristoforo:
~~The `Vertical Integration` test that fails in the buildkite runner passes on LUMI for Julia 1.11.9~~
~~Does anyone know what might be causing this discrepancy, or is there any reason to believe this test could be flaky~~

Didn't realize that the tests failed for Julia v.1.12 since the title was still AMDGPU, Julia v1.11. However, when I run the GPU test suite on LUMI with Julia v 1.12all tests pass.

gregordecristoforo:
I ran the same test suite now with the main branch in #1196 and the buildkite tests fail at the same point. I will create a separate issue about this as this PR doesn't seem to cause the crash.

maximilian-gelbrecht (approved):
Yeah, I've forgetten to rename the title of the CI yaml when testing it. 

Okay let's merge this though after the CI passes again, and then we can tackle other issues in separate PRs.

vchuravy:
Should be fixed when https://github.com/JuliaPackaging/Yggdrasil/pull/14465 lands

gregordecristoforo:
> Okay let's merge this though after the CI passes again, and then we can tackle other issues in separate PRs.

Okay, great. Can I merge or should the skipped checks also be run?

maximilian-gelbrecht:
> > Okay let's merge this though after the CI passes again, and then we can tackle other issues in separate PRs.
> 
> Okay, great. Can I merge or should the skipped checks also be run?

I already did now. In general the skipped checks don't have to run if they have nothing to do with the PR.

maximilian-gelbrecht:
> Should be fixed when [JuliaPackaging/Yggdrasil#14465](https://github.com/JuliaPackaging/Yggdrasil/pull/14465) lands

Oh that's much quicker than I thought. Ludovic was already telling me we might have to wait several months for this. But I guess with a backport that's not the case.

PR #1194, opened 2026-08-14
WhichTransform depending on resolution and architecture

milankl:
Makes the default transform for all models `WhichTransform` that decides as follows

```julia
function WhichTransform(spectral_grid::SpectralGrid; kwargs...)
    if (spectral_grid.truncation <= 64 && spectral_grid.architecture isa GPU) || spectral_grid.architecture isa ReactantDevice
        return MatrixSpectralTransform(spectral_grid; kwargs...)
    else
        return SpectralTransform(spectral_grid; kwargs...)
    end
end
```

The criterion are hardcoded but I think that's fine as it's just a rough guide for convenience, one can always create their own transform and pass that one on, just if one doesn't, these defaults would be picked.


SpeedyWeather/src/dynamics/spectral_grid.jl
```diff
@@ -337,6 +337,17 @@ function (::Type{S})(
     return S(spectrum, grid; NF, nlayers = scratch_nlayers, transform_batch, kwargs...)
 end
 
+"""$(TYPEDSIGNATURES)
+Chooses the transform based on resolution and architecture. For low-resolution
+GPU a MatrixSpectralTransform is returned, otherwise the SpectralTransform."""
+function WhichTransform(spectral_grid::SpectralGrid; kwargs...)
+    if spectral_grid.truncation <= 64 && spectral_grid.architecture isa GPU
```
maximilian-gelbrecht:
```suggestion
    if (spectral_grid.truncation <= 64 && spectral_grid.architecture isa GPU) || spectral_grid.architecture isa ReactantDevice
```

maximilian-gelbrecht: Let's also directly include Reactant in the criteria. Currently Reactant only works for the matrix transform.

milankl:
We got

```julia
GPU Barotropic: Error During Test at /var/lib/buildkite-agent/builds/gpuci-6/julialang/speedyweather-dot-jl/SpeedyWeather/test/GPU/barotropic.jl:3
  ArgumentError: Illegal conversion of a CUDACore.DeviceMemory to a Ptr{Float32}
```

maximilian-gelbrecht:
This has nothing to do with this PR but I am looking into it

maximilian-gelbrecht:
> We got
> 
> ```julia
> GPU Barotropic: Error During Test at /var/lib/buildkite-agent/builds/gpuci-6/julialang/speedyweather-dot-jl/SpeedyWeather/test/GPU/barotropic.jl:3
>   Got exception outside of a @test
>   ArgumentError: Illegal conversion of a CUDACore.DeviceMemory to a Ptr{Float32}
> ```

Solved by #1201

Issue #1187, opened 2026-08-10
Model not working with Julia 1.12 on AMD GPUs


maximilian-gelbrecht:
As already initially discussed in #1171, while the model runs just fine with Nvidia/CUDA GPUs in Julia 1.12, it doesn't for AMD GPUs. In Julia 1.11 this works, but not in 1.12. 

As neither Milan nor I have AMD GPUs at our disposal, this will be hard to figure out for us, but maybe @gregordecristoforo finds a solution.

maximilian-gelbrecht:
@luraess Are you aware of any breaking changes that might cause this or where to look? Our model runs just fine on AMD with Julia 1.11 but with Julia 1.12 we get segmentation faults.

luraess:
> [@luraess](https://github.com/luraess) Are you aware of any breaking changes that might cause this or where to look? Our model runs just fine on AMD with Julia 1.11 but with Julia 1.12 we get segmentation faults.

Not fun. A MWE would be helpful. Ideally, not a whole solver call but if you could nail down where it breaks I could have a look.

maximilian-gelbrecht:
@gregordecristoforo Could you maybe try to set up a MWE for Ludovic? Inspecting the stack trace in the failed CI runs, it seems that `set!` might be the culprit. Most likely the `set!` with functions as an input if I'd take a guess.

gregordecristoforo:
The following kernel call triggers the crash on LUMI (MI250X, ROCm 6.3.4, AMDGPU.jl 2.7.1, Julia 1.12.6):

```Julia
using SpeedyWeather
using AMDGPU
using KernelAbstractions

lmax, mmax = 34, 33   # trunc=32 -> lmax = trunc+2, mmax = trunc+1
l_indices_cpu = [l for m in 1:mmax for l in m:lmax]
nlayers = 8
lm = length(l_indices_cpu)   # 594

l_indices = ROCArray(l_indices_cpu)
temp_tend = ROCArray(rand(ComplexF32, lm, nlayers))
pres_tend = ROCArray(rand(ComplexF32, lm))
div_tend = ROCArray(rand(ComplexF32, lm, nlayers))
G = ROCArray(zeros(ComplexF32, lm, nlayers))
geopotential = ROCArray(zeros(ComplexF32, lm, nlayers))
div_old = ROCArray(rand(ComplexF32, lm, nlayers))
div_new = ROCArray(rand(ComplexF32, lm, nlayers))
[...]

PR #1196, opened 2026-08-16
Run AMDGPU CI with 1.12 (without Reactant) 

gregordecristoforo:
Testing whether the buildkite runner fails for the main branch. Trying to reproduce #1193 failing tests.


maximilian-gelbrecht:
The AMD build kite runs on Ludovic's machines. If we can break this down to a difference between different AMD machines, then we can also easily ask him to just run some targeted scripts when he has some time.

maximilian-gelbrecht:
@gregordecristoforo be aware that I changed the environments for the GPU CI in one of the recent PRs to make handling dependencies and the envs easier. Every vendor gets its own Project.toml now. We could do want the Reactant test with Nvidia, but we could easily remove Reactant and the respective test from AMD now

maximilian-gelbrecht (approved):
Thanks, looks good to me!

gregordecristoforo:
Thanks! I'll merge once all tests pass.

gregordecristoforo:
fix #1187

Issue #1202, opened 2026-08-18
Architectures.synchronize is exported but never bound, breaks on import


gregordecristoforo:
`Architectures.jl` does export synchronize but only ever extends `KernelAbstractions.synchronize` via its fully-qualified name, without first importing it. So anything that imports `SpeedyWeatherInternals.Architectures: synchronize` gets a binding with no value, and fails with `UndefVarError` the moment it's called.

```Julia
julia> using SpeedyWeather
julia> import SpeedyWeatherInternals.Architectures: synchronize, CPU
WARNING: Imported binding Architectures.synchronize was undeclared at import time during import to Main.
julia> synchronize(CPU())
ERROR: UndefVarError: `synchronize` not defined in `Main`
Suggestion: this global was defined as `SpeedyWeatherInternals.Architectures.synchronize` but not assigned a value.
Hint: a global variable of this name also exists in KernelAbstractions.
Stacktrace:
 [1] top-level scope
   @ REPL[3]:1

```

This might look a bit silly for `CPU()` but I ran into it when working on #1147 with GPU hardware.

gregordecristoforo:
As pointed out in #1203 this is not the intended way.

PR #1203, opened 2026-08-18
fix synchronize being exported without ever being bound to a value

gregordecristoforo:


gregordecristoforo:
fix #1202

maximilian-gelbrecht:
I think this isn't the right solution. I am all a bit confused whether the actually works now to really synchronize the devices like this. 

Reexporting `synchronize` should work, but we probably just made an error how exactly we did it. I am just a bit confused myself with it within your package structure with multiple packages and within `SpeedyWeatherInternals` also multiple modules.

gregordecristoforo:
Thanks, I'll abandon this approach

PR #1206, opened 2026-08-19
Restrict Reactant compat

maximilian-gelbrecht:


maximilian-gelbrecht:
The CI ignored my compat bound here for some reason, but the issue is connected to the recent Reactant version and already fixed upstream. The next released version will fix it github.com/EnzymeAD/Reactant.jl/issues/3196

PR #1207, opened 2026-08-19
First implementation for FFTs using Metal.jl

Gabrielgerez:
I have created this draft pull request to show work I have done, and what needs to be done. Open for feedback.

Currently in my test range Metal is still slower than CPU, but it gets closer and closer the higher the resolution. I want to try and find this crossover point, as it seems there should be one at higher resolutions.

Most of the work has been done on the `SpeedyTransformsMetalExt.jl` where most of the machinery related to FFTs has been adapted to MTLArrays and MPS. 

Another point is the addition of two separate methods of structuring the command bufffer, merged hemispheres and unmerged. In the merged strategy both north and south hemispheres get their own buffers and are sent further together, while on the unmerged they each get their own buffer and are sent separately. The reason to keep both is that originally they had different ranges where each was better than the other. 

Further testing seems to point that this difference in domains might not be the case, as i might have had not rigorous enough benchmarking. So far it seems that the merged is the better method overall.


milankl:
Awesome, will look a little closer in a minute but note that we use Runic formatting, many changes were due to a different formatter. We have https://github.com/fredrikekre/Runic.jl installed in VS Code and so shift + option + F just formats everything into Runic before. Doesn't have to be perfect, we're not strict on this and can just reformat the whole repo every few months to avoid diverting from it too much. But just to filter out the lines of code changed here due to format changes only

Then I also suggest to redirect this PR here into a branch that lives in this repo otherwise the docs don't deploy and we can't see a preview therefore. Also means we don't have to approve the workflows all the time

milankl:
Also in general I'm curious about the preformance of the `MatrixSpectralTransform` on Metal. We don't really need Metal for high resolution (this should be done on a big GPU) and so I believe here we should focus on performance in the T16-T128 range

SpeedyTransforms/ext/SpeedyTransformsMetalExt.jl
```diff
@@ -0,0 +1,308 @@
...
+        nfreq = nlons[j] ÷ 2 + 1
+        copyto!(ring_complex_north[j], view(g_north, 1:nfreq, :, j))
+        (has_equator && j == j_equator) && continue   # south equator ring: north already covers it
+        copyto!(ring_complex_south[j], view(g_south, 1:nfreq, :, j))
+    end
+
+    cmdbuf = Metal.MPS.MPSCommandBuffer(queue)
+    @inbounds for j in 1:nlat_half
+        encode_fft!(cmdbuf, brfft_plans[j], ring_complex_north[j], ring_real_north[j])
+        (has_equator && j == j_equator) && continue
```
milankl, line 147: Not sure about this, but looping only over one hemisphere and then over the other was originally implemented for the CPU due to the symmetry of the FFT lengths. However, I'm not sure this is actually beneficial on the GPU as this means that the memory access on the southern hemisphere is reversed so not 1:N but N:-1:1 and hence not contiguous

SpeedyTransforms/ext/SpeedyTransformsMetalExt.jl
```diff
@@ -0,0 +1,308 @@
...
+based on `S.metal_merge_hemispheres`; see the module-level comment for the benchmarked
+crossover between them."""
+function SpeedyTransforms._fourier_batched!(
+        f_north::MtlArray{<:Complex, 3},
+        f_south::MtlArray{<:Complex, 3},
+        field::AbstractField,
+        S::SpectralTransform,
+    )
+    @assert eltype(field) == eltype(S) "Number format of grid $(eltype(field)) and SpectralTransform $(eltype(S)) need to match."
+    if !S.cuda_graphs   # generic toggle for "use accelerated batched Fourier path"; also gates Metal batching
```
milankl, line 268: with Gregor's #1147 this should become `S.gpu_graphs`

SpeedyTransforms/src/spectral_transform.jl
```diff
@@ -101,6 +102,12 @@ struct SpectralTransform{
     # Set to `false` to fall back to the generic (allocating) per-ring GPU
     # Meaningless for non-CUDA architectures.
     cuda_graphs::Bool
+
+    # METAL: MERGE HEMISPHERES
+    # toggle for merging the northern and southern hemisphere's per-ring FFTs into a
+    # single shared Metal command buffer/commit (`true`) instead of one commit per
+    # hemisphere (`false`).
+    metal_merge_hemispheres::B2
```
milankl, line 110: Why does this bool need to parameteric but `cuda_graphs` not? It's generally fine to hardcode types like `Int` or `Bool` but this needs to be checked with Reactant particularly as Reactant replaces runtime scalar variables with things like `TracedRNumber` in order to built up a computational graph, hence you'll see some struct fields being parametric, not sure this applies here though, Max?

milankl: Also generally, I'm wondering whether all CPU/GPU Fourier transforms should forget about the latitude symmetry of the FFT lengths and just treat them as 1:nlat. There might be some batching actually possible as the first and last (second and second last etc) rings have the same FFT applied to them but with the graphs I'm not sure that's useful and it's definitely not contiguous in memory

milankl (commented):
Just some comments from my side, if you have a macbook you can of course check locally but maybe we can setup some simple Metal CI too. Oceananigans seems to do this here

https://github.com/CliMA/Oceananigans.jl/blob/a932918b2d42977c598a764313f4c8823e2152d3/.github/workflows/ci.yml#L362

Before we do this we should check though what the free limits on this are, I'm just cautious that Metal CI might trigger some GitHub billing earlier or later if do a lot of it. @giordano How does Oceananigans do Metal CI? Just basic functionality but then it's otherwise a free service from GitHub?

giordano:
> Before we do this we should check though what the free limits on this are, I'm just cautious that Metal CI might trigger some GitHub billing earlier or later if do a lot of it. @giordano How does Oceananigans do Metal CI? Just basic functionality but then it's otherwise a free service from GitHub?

It's free, the only limitation is that you get 5 concurrent macOS runners across the entire organisation

gregordecristoforo:
> Then I also suggest to redirect this PR here into a branch that lives in this repo otherwise the docs don't deploy and we can't see a preview therefore. Also means we don't have to approve the workflows all the time

Sure. Could @Gabrielgerez  (and @Jullcifer) be added as members to the Github organization so that they can edit a branch in this repo themselves?

giordano:
> Then I also suggest to redirect this PR here into a branch that lives in this repo otherwise the docs don't deploy and we can't see a preview therefore.

I always [upload docs build as artifact](https://github.com/NumericalEarth/Breeze.jl/blob/20b78645cd31a97f9fd9034b382b06aa90818478/.github/workflows/Documentation.yml#L112-L117) for this reason

maximilian-gelbrecht:
> > Then I also suggest to redirect this PR here into a branch that lives in this repo otherwise the docs don't deploy and we can't see a preview therefore. Also means we don't have to approve the workflows all the time
> 
> Sure. Could @Gabrielgerez (and @Jullcifer) be added as members to the Github organization so that they can edit a branch in this repo themselves?

Just did.

Gabrielgerez:
I will close this pull request and link it from a new one on our fork

milankl:
Just open a new branch here and redirect the pull request, and merge directly?

SpeedyTransforms/src/spectral_transform.jl
Gabrielgerez: Just responding to this comment here. I originally tested these two methods, and they were generally good at different Trunc values. Though, now there is just one implementation, and it seems to be better than both these strategies, and thus this bool is now no more. 

Though the MPSGraphs do not seem to be affected by how contiguous the arrays are in memory, in a meaningful way that is, though this needs to be properly profiled.

PR #1209, opened 2026-08-19
3D interpolation and particle advection

Jullcifer:
I tried to implement the steps we discussed for particle advection here. I have not tested the performance yet, so I wanted to make sure it looks right before I start with it.

### Short description
Adds 3D continuous particle tracking (`ParticleAdvection3D`) in $\sigma$-coordinates. Particles now carry their own σ coordinate and are advected by the full 3D velocity field (u, v, w) using the same Heun time integration already used for 2D.

### Key Changes
* 3D Particle Struct & Allocation: `ParticleAdvection3D` struct and constructor — identical to `ParticleAdvection2D` but without the fixed `layer` field; each particle tracks its own $\sigma$
* Vertical Layer Bracketing: ` find_vertical_bracket(σ, σ_levels_full)` — finds the two adjacent full levels bracketing a particle's σ and returns the blending weight α
* `interpolate_3D!` - bilinear (anvil) horizontal interpolation with vertical blending between the two bracketing layers; reuses the existing `AnvilLocator` weights


SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -152,6 +212,50 @@ end
...
+    v0 = vars.particles.v
+    w0 = vars.particles.w
+    interpolate_3D!(u0, u_3d, locator, geometry, particles, σ_levels_full)
+    interpolate_3D!(v0, v_3d, locator, geometry, particles, σ_levels_full)
+    interpolate_3D!(w0, w_3d, locator, geometry, particles, σ_levels_full)
+    return nothing
+end
+
+# Modulo particles and extract horizontal coordinates; σ is each particle's own
+@kernel inbounds = true function _initialize_3D_particles_kernel!(
```
milankl, line 221: Is that any different from the 2D kernel?

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -54,6 +52,34 @@ function ParticleAdvection2D(SG::SpectralGrid; kwargs...)
...
+    } <: AbstractParticleAdvection
+
+    "[OPTION] Number of particles"
+    nparticles::IntType = 10
+
+    "[OPTION] Execute particle advection every n timesteps"
+    every_n_time_steps::IntType = 6
+
+    "[DERIVED] Time step used for particle advection (scaled by radius, converted to degrees) [s*˚/m]"
+    Δt::Base.RefValue{NF} = Ref(zero(NF))
```
milankl: I think this ref value can actually be removed and just computed on the fly, same for 2D

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -82,6 +120,16 @@ function initialize!(
     return particle_advection
 end
 
+function initialize!(
+        particle_advection::ParticleAdvection3D,
+        model::AbstractModel,
+    )
+    (; every_n_time_steps) = particle_advection
+    particle_advection.Δt[] = every_n_time_steps * model.time_stepping.Δt
+    particle_advection.Δt[] *= (360 / 2π)
```
milankl: I think we should just compute this on the fly, it's so cheap

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -227,6 +331,133 @@ function particle_advection!(
...
+    (; geometry) = particle_advection
+    (; time_stepping) = model
+
+    n = particle_advection.every_n_time_steps
+    clock.time_step_counter % n == (n - 1) || return nothing
+
+    Δt = particle_advection.Δt[]
+    Δt_half = Δt / 2
+    Δt_vert = Δt * oftype(Δt, π / 180)  # convert [s*˚/m] → [s] for dσ = w[s⁻¹]*Δt_vert[s]
+    Δt_vert_half = Δt_vert / 2
```
milankl: Compute the time step here on the fly as you already do for `_vert` to remove the ref value?

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -227,6 +331,133 @@ function particle_advection!(
...
+# σ outside [σ_levels_full[1], σ_levels_full[end]] is pinned, not extrapolated
+@inline function find_vertical_bracket(σ::NF, σ_levels_full) where NF
+    n = length(σ_levels_full)
+    k_lo = 1
+    for k in 1:(n - 1)
+        σ_levels_full[k] <= σ && (k_lo = k)
+    end
+    k_hi = k_lo + 1 
+    Δσ = σ_levels_full[k_hi] - σ_levels_full[k_lo]
+    α = clamp((σ - σ_levels_full[k_lo]) / Δσ, zero(NF), one(NF))
```
milankl, line 387: `, 0, 1)` is fine here, the type promotion is automatically done
```julia
julia> clamp(1.5f0, 0, 1)
1.0f0
```

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -227,6 +331,133 @@ function particle_advection!(
...
+    interpolate_3D!(v_new, v_3d, locator, geometry, particles, σ_levels_full)
+    interpolate_3D!(w_new, w_3d, locator, geometry, particles, σ_levels_full)
+    return nothing
+end
+
+# σ outside [σ_levels_full[1], σ_levels_full[end]] is pinned, not extrapolated
+@inline function find_vertical_bracket(σ::NF, σ_levels_full) where NF
+    n = length(σ_levels_full)
+    k_lo = 1
+    for k in 1:(n - 1)
```
milankl, line 382: Also for the GPU it might be fine to have a while loop here, in #1193 @gregordecristoforo also replaced a for loop with while

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -227,6 +331,133 @@ function particle_advection!(
...
+
+# Vertically blended anvil interpolation at each particle's σ coordinate.
+function interpolate_3D!(Aout, A, locator, geometry, positions, σ_levels_full)
+    (; ij_as, ij_bs, ij_cs, ij_ds, Δabs, Δcds, Δys, npoints_output) = locator
+    A_data = A.data
+    nlayers = size(A_data, 2)
+    rings = geometry.grid.rings             # CPU array, used only for pole averages
+    T = eltype(A_data)
+    north_vals_cpu = Vector{T}(undef, nlayers)
+    south_vals_cpu = Vector{T}(undef, nlayers)
```
milankl, line 399: Can we avoid the allocation here? Especially not as hardcoded `Vector`. I see the `on_architecture` below but I'm sure that's also easily to formulate as small kernel and `north_vals` could be just declared as required scratch arrays given how small they are, see https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/variable_system#Declare-variables

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -227,6 +331,133 @@ function particle_advection!(
...
+    b = ij_as[i] == 0 ? north_vals[k_lo] : A_data[ij_bs[i], k_lo]
+    c = ij_cs[i] == -1 ? south_vals[k_lo] : A_data[ij_cs[i], k_lo]
+    d = ij_cs[i] == -1 ? south_vals[k_lo] : A_data[ij_ds[i], k_lo]
+    ab_lo = a + (b - a) * Δabs[i]
+    val_lo = ab_lo + (c + (d - c) * Δcds[i] - ab_lo) * Δys[i]
+
+    a = ij_as[i] == 0 ? north_vals[k_hi] : A_data[ij_as[i], k_hi]
+    b = ij_as[i] == 0 ? north_vals[k_hi] : A_data[ij_bs[i], k_hi]
+    c = ij_cs[i] == -1 ? south_vals[k_hi] : A_data[ij_cs[i], k_hi]
+    d = ij_cs[i] == -1 ? south_vals[k_hi] : A_data[ij_ds[i], k_hi]
```
milankl, line 434: Maybe easier to write with tuples to cut down on the number of `?`s? Like

```julia
(a, b) = cond ? (r, p) : (x, y)
```

but might be just a matter of style, could be that the compiler merges that anyway!

milankl (commented):
Awesome!! Just some comments to start! Note that to test it might be good to check what's currently done in the docs

https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/particles#Tracking-particles (but oh gosh there's clearly something wrong here, I'm wondering whether there's some scaling wrong since https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1139)

or reuse the functionality from github.com/SpeedyWeather/TravellingSailorProblem.jl to make some nice plots to illustrate that the particles indeed fly smoothly in x, y, z

milankl:
Can you create a branch here like `jm/3D-particle-tracking` so that we don't have to approve every commit and also you can look at the deployed docs (which isn't allowed for pull requests from external)

SpeedyWeather/src/dynamics/particle_advection.jl
milankl: So previously our time step in `model.time_stepping.Δt` was scaled, i.e. actually Δt/radius, now we do this on the fly, I believe that caused a bug here such that we now need to change the last line to `360 / (2π*radius)` (but note the implicit type conversion to Float64, that's only avoided here because writing into a ref value with a fixed type

```julia
julia> 2π * 1f0
6.283185307179586
```

instead

```julia
julia> 2*(π * 1f0)
6.2831855f0
```

wouldn't do that, see https://docs.julialang.org/en/v1/manual/conversion-and-promotion/)

this makes currently the time steps for particle advection massive and explains the doc plots being ziggy zaggy lines everywhere

milankl:
Skipping some of the CI here for now, should be switched back on once it's clear it's working on CPU.

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -227,6 +331,133 @@ function particle_advection!(
...
+        σ_levels_full[k] <= σ && (k_lo = k)
+    end
+    k_hi = k_lo + 1 
+    Δσ = σ_levels_full[k_hi] - σ_levels_full[k_lo]
+    α = clamp((σ - σ_levels_full[k_lo]) / Δσ, zero(NF), one(NF))
+    return k_lo, k_hi, α
+end
+
+# Vertically blended anvil interpolation at each particle's σ coordinate.
+function interpolate_3D!(Aout, A, locator, geometry, positions, σ_levels_full)
```
milankl, line 392: Happy to have this here to get started but in the end I believe 3D interpolation should be moved to RingGrids

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -97,6 +145,18 @@ function initialize!(
...
+function initialize!(
+        particles::AbstractVector{P},
+        vars::Variables,
+        particle_advection::ParticleAdvection3D,
+        model::AbstractModel,
+    ) where {P <: Particle}
+    particles .= on_architecture(architecture(particles), rand(P, length(particles)))
+    return particles
+end
+
```
milankl: This is code duplication from above, just turn `::ParticleAdvection2D` into `::AbstractParticleAdvection` to have a method for both 2D and 3D

SpeedyWeather/src/dynamics/particle_advection.jl
milankl: Note that you can also use dispatch in kernels, though often I agree, it's more readable to just give it a slightly different name

SpeedyWeather/src/dynamics/particle_advection.jl
milankl: did that in [e90d360](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1209/commits/e90d36067b2c9cd7c478701b11f91df01f71383d)

SpeedyWeather/src/dynamics/particle_advection.jl
milankl: did that in [e90d360](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1209/commits/e90d36067b2c9cd7c478701b11f91df01f71383d)

SpeedyWeather/src/dynamics/particle_advection.jl
milankl: Did this in [e90d360](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1209/commits/e90d36067b2c9cd7c478701b11f91df01f71383d)

SpeedyWeather/src/dynamics/particle_advection.jl
milankl: addressed this in [e90d360](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1209/commits/e90d36067b2c9cd7c478701b11f91df01f71383d)

milankl:
I think the time step is correctly scaled now, at T32 with particle advection every 6 time steps it's ~0.13 s*˚/m as computed before the kernel on the fly. Now with a velocity of 1m/s the change in latitude of a particle would therefore be 0.13˚ over 6 time steps (40min each) i.e. 4 hours. 1m/s over 4 hours is 14.4km. As 1˚ = 111.19km at the Equator moving 14.4km is equivalent to ~0.13˚, so yay just a /radius was missing here after #1139 my bad!

milankl:
Skipping submodule CI now as currently all changes are in SpeedyWeather itself

SpeedyWeather/src/dynamics/particle_advection.jl
Jullcifer: Very good point! Merged them in 07de068a5a16ea1337aa701b0d56ef03047f9bb9

PR #1210, opened 2026-08-20
Rewrite `sind(x)` as `sin(deg2rad(x))` etc for AMDGPU

gregordecristoforo:
Currently the AMDGPU testsuite produces multiple warnings about detected hostcalls that are triggered by `Base`'s degree-trig functions (`sind/cosd/tand`).[ This buildkite runner ](https://buildkite.com/julialang/speedyweather-dot-jl/builds/5130/list?jid=01a0194b-0bd6-43ac-ab21-bd0208460ecb&tab=output)shows them for example. By using radians plain `sin/cos/tan` avoid AMDGPU hostcalls. Running the GPU testsuite on LUMI now doesn't detect any hostcalls anymore. 

I didn't run any CPU benchmarks so I don't know whether this impacts non AMD related performance significantly. Is this  worth testing?


milankl (approved):
Absolutely don't mind the changes here but why is not solved on the AMDGPU.jl side? @luraess could one not define `cosd(x) = cos(deg2rad(x))` within AMD kernels somehow? Feels like that should happen somewhere here https://github.com/JuliaGPU/AMDGPU.jl/blob/aef0a5bea88a8d6f9e2e4d6e90cb16768c5ac2e4/src/device/gcn/math.jl ?

luraess:
@milankl we have fix in AMDGPU and are actually planning also to fix things in Base.

gregordecristoforo:
Shall I still merge this PR for backwards compatibility once the builkite runner finished successfully? 

I would like to do some benchmarking, equivalently to the CUDA benchmarks in this repo. For that the host calls should probably be avoided.

milankl:
Happy for you to merge, can you add a comment in the hot loops (particle advection and transmissivity only I believe) to move back to `cosd` when upstream pull requests are merged and new AMDGPU version is out?

maximilian-gelbrecht:
Some of these changes are also arready in parallel applied in #985 , because Reactant also currently only has limited support for some of these functions like `sind` . So they might have to be changed anyway.

maximilian-gelbrecht (approved):

gregordecristoforo:
> Happy for you to merge, can you add a comment in the hot loops (particle advection and transmissivity only I believe) to move back to cosd when upstream pull requests are merged and new AMDGPU version is out?

Done in [09bed87](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1210/commits/09bed87b931046c4999357f5d7a03486e5891002). I'll merge this PR once all tests passed.

PR #1211, opened 2026-08-20
Reapply Runic formatting

maximilian-gelbrecht:


milankl (approved):
Now we compare this to blame and check who's the biggest culprit 😝 it looks like Claude isn't quite innocent either!

PR #1212, opened 2026-08-20
Reduce model initialization compile time

milankl:
Claude proposes

1. Type-erase the set! keyword loop — funnel the unchanged public kwarg API into a core taking Vector{Pair{Symbol,Any}}. I prototyped this: 5 signatures cost 5.5s vs 14.4s for 4 signatures currently. The dynamic dispatch it introduces is irrelevant on a setup-only path (all of step 3 runs in 0.024s warm).
2. Collect variable definitions into a Vector — the element type is already erased downstream, so the tuple detour buys nothing.


maximilian-gelbrecht (commented):
Huh, I did already know that the initial conditions are what takes the longest, but I didn't directly connect this to `set!`. So that's definitely valuable insight. 

As long as we don't use `set!` in a hot loop this is fine, right? 

The implementation that it did now seems a bit convoluted to me, but maybe that's also just the way to do it. Still you could maybe ask it to simplify it.

PR #1215, opened 2026-08-21
ParticleAdvection3D via 3D interpolation

Jullcifer:
Copying over from a fork, still working on comments from #1209


SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -54,34 +49,68 @@ function ParticleAdvection2D(SG::SpectralGrid; kwargs...)
...
-        particle_advection::ParticleAdvection2D,
-        model::AbstractModel,
+# same as 2D but also add vertical velocity and pole scratch arrays for 3D interpolation
+function variables(P::ParticleAdvection3D)
+    (; nparticles, nlayers) = P
+    return (
+        variables(ParticleAdvection2D, nparticles)...,
+        ParticleVariable(:w, VectorDim(nparticles), desc = "Vertical velocity dσ/dt at particle location", units = "1/s"),
+        ScratchVariable(:north_pole_vals, VectorDim(nlayers), desc = "North pole ring averages for 3D interpolation"),
+        ScratchVariable(:south_pole_vals, VectorDim(nlayers), desc = "South pole ring averages for 3D interpolation"),
```
milankl: you can pull `nlayers` directly from `model` when providing that as the 2nd argument here. Then you don't need the "derived" `nlayers` field in the struct. 

```julia
function variables(::MyComponent, model::AbstractModel)
    (; nlayers) = model.spectral_grid
    ...
end
```

but in the end you'd want these arrays to live inside RingGrids-owned objects, otherwise you can't move the 3D interpolation from SpeedyWeather where it belongs: in RingGrids

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -139,19 +168,65 @@ function initialize!(
...
+    )
+
+    RingGrids.update_locator!(locator, geometry, lons, lats)
+    u0 = vars.particles.u
+    v0 = vars.particles.v
+    w0 = vars.particles.w
+
+    interpolate_3D!(u0, u_3d, locator, geometry, particles, σ_levels_full, vars.scratch.north_pole_vals, vars.scratch.south_pole_vals)
+    interpolate_3D!(v0, v_3d, locator, geometry, particles, σ_levels_full, vars.scratch.north_pole_vals, vars.scratch.south_pole_vals)
+    interpolate_3D!(w0, w_3d, locator, geometry, particles, σ_levels_full, vars.scratch.north_pole_vals, vars.scratch.south_pole_vals)
```
milankl: given this is how you call it, could one not add the scratch vectors to `locator` or define a 3D locator with them?

milankl (commented):
Yay getting there!

SpeedyWeather/src/dynamics/particle_advection.jl
Jullcifer: As suggested, now instead of passing `north_pole_vals`/`south_pole_vals` as separate arguments to `interpolate_3D!`, they are embedded in the locator itself. 
- `variables(ParticleAdvection3D, model)` now uses `LocatorDim3D` for the locator, which allocates an `AnvilLocator` with north_pole_vals/south_pole_vals of length `nlayers` built in

I also added a `find_vertical_bracket` test.
Sorry everything took so long.

SpeedyWeather/src/dynamics/particle_advection.jl
Jullcifer: `nlayers` is now pulled from `model.spectral_grid` directly in `variables(P::ParticleAdvection3D, model::AbstractModel)`, and the pole arrays have been moved into `AnvilLocator` (using `LocatorDim3D`), so they live in a RingGrids-owned object as suggested.

RingGrids/src/interpolation.jl
```diff
@@ -103,6 +103,10 @@ between two latitude rings."""
     Δys::VectorType = zero(VectorType(undef, npoints_output))    # distance fractions between rings
     Δabs::VectorType = zero(VectorType(undef, npoints_output))    # distance fractions between a, b
     Δcds::VectorType = zero(VectorType(undef, npoints_output))    # distance fractions between c, d
+
+    # optional pole averages per vertical layer for 3D interpolation; nothing for 2D
+    north_pole_vals::Union{Nothing, VectorType} = nothing
+    south_pole_vals::Union{Nothing, VectorType} = nothing
```
milankl: What's the motivation for allowing `nothing` here? That means the compiler can't infer the type at compile time which may lead to slower code. Given `nlayers` is a mandatory positional argument in the constructor method below, I'd just include an `nlayers` field here that's used to allocate the right length of vector. For backwards compatibility it could just be `nlayers = 1` by default

RingGrids/src/interpolation.jl
milankl: you could just add `nlayers = 1` as positional optional argument here then you don't need the `AnvilLocator` constructor below

RingGrids/src/interpolation.jl
```diff
@@ -142,6 +148,23 @@ end
...
+        nlayers::Integer;
+        architecture::AbstractArchitecture = DEFAULT_ARCHITECTURE(),
+    )
+    VectorType = array_type(architecture, NF, 1)
+    VectorIntType = array_type(architecture, Int, 1)
+    north_pole_vals = on_architecture(architecture, zeros(NF, nlayers))
+    south_pole_vals = on_architecture(architecture, zeros(NF, nlayers))
+    return AnvilLocator{VectorType, VectorIntType, typeof(npoints)}(;
+        npoints_output = npoints, north_pole_vals, south_pole_vals)
+end
```
milankl: redundant? See above

RingGrids/src/interpolation_3d.jl
```diff
@@ -0,0 +1,74 @@
...
+# Compute north and south pole ring averages per vertical layer on device.
+@kernel inbounds = true function _compute_pole_averages_kernel!(north_vals, south_vals, A_data, ring_starts, nlons, nlat)
+    k = @index(Global, Linear)
+    n_north = nlons[1]
+    rs_north = ring_starts[1]
+    north_sum = zero(eltype(north_vals))
+    for i in 0:(n_north - 1)
+        north_sum += A_data[rs_north + i, k]
+    end
+    north_vals[k] = north_sum / n_north
```
milankl: given this is an average rename to `north_pole_average`?

RingGrids/src/interpolation_3d.jl
```diff
@@ -0,0 +1,74 @@
+export interpolate_3D!
+
+# σ outside [σ_levels_full[1], σ_levels_full[end]] is pinned, not extrapolated
+@inline function find_vertical_bracket(σ::NF, σ_levels_full) where {NF}
```
milankl: I think the `where NF` can be removed now

RingGrids/src/interpolation.jl
milankl: rename to `north_pole_average` etc see below?

SpeedyWeather/src/variables/dimensions.jl
```diff
@@ -160,6 +160,13 @@ allocate(::AbstractVariable{TransformScratchMemory}, model::AbstractModel) = mod
 end
 allocate(::AbstractVariable{LocatorDim}, model::AbstractModel) = RingGrids.AnvilLocator(model.spectral_grid.NF, model.particle_advection.nparticles; architecture = model.spectral_grid.architecture)
 
+"""Dimension for particle locator with embedded vertical pole scratch for 3D interpolation."""
+@kwdef struct LocatorDim3D <: AbstractVariableDim
+    nparticles::Int = 1
+    nlayers::Int = 1
+end
+allocate(v::AbstractVariable{LocatorDim3D}, model::AbstractModel) = RingGrids.AnvilLocator(model.spectral_grid.NF, v.dims.nparticles, v.dims.nlayers; architecture = model.spectral_grid.architecture)
```
milankl: I think this can be merged with `LocatorDim` above

SpeedyWeather/src/variables/dimensions.jl
milankl: the `n` isn't even used, can you remove? `model.particle_advection` is supposed to determine the number of particles, and so I believe should `spectral_grid` determine the number of layers, hence `LocatorDim3D` below is redundant

milankl (commented):
Some more comments, I feel like we're getting there and just a few things can be merged into one. 

- While you could define an `AnvilLocator2D` and `3D` it seems you only really need to have 2 nlayers-length vectors in addition to the precomputed horizontal arrays. Think 10,000 particles/grid points and 100 vertical layers, there's no problem always storing those in general, no `nothing` fallback needed
- that also simplifies their constructors and dimensions below

Just to summarize the approach, while the horizontal indices for interpolation are precomputed, the vertical ones are computed on the fly. So `update_locator!` is only needed when horizontal locations change (e.g. particles moving horizontally) but not for vertical changes, e.g. outputting on a different grid (that's always the same) with changing pressure levels. Would be good to write this up into the docs so that also all the limitations and flexibility is known.

milankl:
Btw, the time stepping issue has been resolved already here! https://speedyweather.github.io/SpeedyWeatherDocumentation/previews/PR1215/particles#Tracking-particles

compared to main https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/particles#Tracking-particles

milankl-claude:
I tested this branch on GPU (NVIDIA L4, CUDA backend) with three checks: 3D interpolation of random data, 3D particle advection in isolation, and a full model run with 3D particle advection.

**1. 3D interpolation of random data — works.** `RingGrids.interpolate_3D!` compiled and ran correctly on a random `Field` with an `AnvilLocator` built for multiple layers.

**2. 3D particle advection in isolation — no errors.** Calling `particle_advection!` directly (bypassing `run!`) on a GPU-resident `PrimitiveDryModel` + `ParticleAdvection3D` completed without error.

**3. Full model simulation with 3D particle advection — fails to compile on GPU:**

```
ERROR: LoadError: GPU compilation of MethodInstance for SpeedyWeather.gpu_column_parameterizations_kernel!(...) failed
KernelError: passing non-bitstype argument
  .particles is of type ... which is not isbits.
    .locator is of type RingGrids.AnvilLocator{...} which is not isbits.
      .north_pole_vals is of type Union{Nothing, CUDACore.CuDeviceVector{Float32, 1}} which is not isbits.
      .south_pole_vals is of type Union{Nothing, CUDACore.CuDeviceVector{Float32, 1}} which is not isbits.
```
Failure originates in `column_parameterizations!` (`SpeedyWeather/src/parameterizations/tendencies.jl:39`), called from `time_step!`.

**Root cause:** `north_pole_vals`/`south_pole_vals` are added to `RingGrids.AnvilLocator` typed as `Union{Nothing, VectorType}` (`RingGrids/src/interpolation.jl`). A field whose *static* type is a `Union` is never `isbitstype`, regardless of the runtime value — confirmed `isbitstype(typeof(loc))` is `false` even for a plain 2D locator. Since `vars.particles.locator` lives inside the top-level `Variables` struct, and unrelated kernels (e.g. column parameterizations) take the entire `vars` object as an argument, this breaks GPU compilation for the whole time step, not just the interpolation kernel.

**This is a regression affecting `ParticleAdvection2D` too, not just the new 3D feature.** I re-ran the same full-simulation test with the pre-existing `ParticleAdvection2D` and got the identical `not isbits` error at the identical call site. As currently written, this PR breaks GPU support for all particle advection.

**Suggested fix:** avoid the `Union{Nothing, VectorType}` field type — e.g. always allocate `north_pole_vals`/`south_pole_vals` as concrete (possibly zero-length) vectors instead of `nothing` for the 2D case, or split into a separate 3D-only locator type so the 2D `AnvilLocator` isn't touched at all.

RingGrids/src/interpolation.jl
Jullcifer: I just wanted to avoid touching the 2D code parts, hence `nothing`; clearly i didn't think this through. Will fix

RingGrids/src/interpolation_3d.jl
```diff
@@ -0,0 +1,74 @@
+export interpolate_3D!
+
+# σ outside [σ_levels_full[1], σ_levels_full[end]] is pinned, not extrapolated
```
milankl: This is still an extrapolation but one that assumes that the gradient at the boundary is zero (von Neumann boundary condition) so the boundary value just continues as is. Note that the "boundary" here is actually not at 0, 1 but at the first/last full sigma level

RingGrids/src/interpolation_3d.jl
milankl: this should be renamed `σ_levels` probably as it can be both half and full

RingGrids/src/interpolation_3d.jl
```diff
@@ -0,0 +1,74 @@
...
+@inline function find_vertical_bracket(σ::NF, σ_levels_full) where {NF}
+    n = length(σ_levels_full)
+    k_lo = 1
+    while k_lo < n - 1 && σ_levels_full[k_lo + 1] <= σ
+        k_lo += 1
+    end
+    k_hi = k_lo + 1
+    Δσ = σ_levels_full[k_hi] - σ_levels_full[k_lo]
+    α = clamp((σ - σ_levels_full[k_lo]) / Δσ, 0, 1)
```
milankl: We can probably also do this for half-level interpolation, just that if `k_lo = 1` this value isn't explicitly stored for vertical velocity but is zero

RingGrids/src/interpolation_3d.jl
```diff
@@ -0,0 +1,74 @@
...
+    south_vals[k] = south_sum / n_south
+end
+
+@kernel inbounds = true function _interpolate_3D_kernel!(
+        Aout, A_data,
+        ij_as, ij_bs, ij_cs, ij_ds, Δabs, Δcds, Δys,
+        positions, σ_levels_full, north_vals, south_vals,
+    )
+    i = @index(Global, Linear)
+    k_lo, k_hi, α = find_vertical_bracket(positions[i].σ, σ_levels_full)
```
milankl: if `k_lo = 1` in the half-level case then `a, b, c, d` all zero, otherwise `A_data[ij, k_lo-1]`

Jullcifer:
I think I have incorporated all your comments now, @milankl. I tried to fix the code using suggestions from milankl-claude (regarding the types), but the full simulation still fails for me locally, with the same error type (isbitstype). I am looking into it and will try to fix it this afternoon.

milankl (commented):
There's another complication we have to account for that I mentioned before which is that in general the 3D interpolation may be from vertical sigma levels that sit on full levels _or on half levels_. E.g. what you wrote works when a 5-layer field is located vertically on full levels but not if it sits on half levels. For that one technically needs 6 layers as you see below, 

```julia
julia> SigmaCoordinates(SpectralGrid(nlayers=5))
5-layer SigmaCoordinates{Int64, Vector{Float32}}
 ├─ 0.0000  k = 0.5
 │× 0.1000  k = 1
 ├─ 0.2000  k = 1.5
 │× 0.3000  k = 2
 ├─ 0.4000  k = 2.5
 │× 0.5000  k = 3
 ├─ 0.6000  k = 3.5
 │× 0.7000  k = 4
 ├─ 0.8000  k = 4.5
 │× 0.9000  k = 5
 └─ 1.0000  k = 5.5
```

however, we only ever need this for the vertical velocity where we have $$w=0$$ at $$\sigma = 0, 1$$, so in practice we do use a 5-layer field to represent velocity at $$k+1/2$$, so at half levels below, the lowermost layer of that field is therefore explicitly 0 whereas the 0 at $$k=1/2$$ isn't explicitly stored (we use a kinematic boundary condition that no velocity goes through the $$\sigma = 0, 1$$ boundaries which can be hardcoded). I propose we stick to how the vertical velocities are stored unless changes are actually necessary. In that sense, we need some kind of `Center` (full levels), `FaceBelow`, `FaceAbove` (that's half levels) flag, as it's normally called for finite volume staggered grids, probably easiest as a dummy type to dispatch over and then for the faces we'd need to pass on the sigma half vectors instead of full to the bracket function. But then we need a specialised `interpolate_3D_kernel!`, as indicated, which one can probably just dispatch over given that center/face type too

Jullcifer:
https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1215#pullrequestreview-5050182768 About the full and half levels - For now, I only used `Center` and `Face` (where `Face` corresponds to the lower interface $k+1/2$ where $w$ lives). Since $w$ is currently the only staggered field in `SpeedyWeather`, keeping it as `Face` avoids unused code paths. If you prefer to split it explicitly into `FaceBelow` and `FaceAbove`, let me know.

SpeedyWeather/src/variables/dimensions.jl
Jullcifer: merged in ef9a3e5ef1af13fb2d9f7a3188c5753003d25079

RingGrids/src/interpolation.jl
Jullcifer: fixed in 09d724c3?

RingGrids/src/interpolation_3d.jl
```diff
@@ -34,13 +43,14 @@ end
     south_pole_average[k] = south_sum / n_south
 end
 
+# Full-level (Center) interpolation kernel.
 @kernel inbounds = true function _interpolate_3D_kernel!(
         Aout, A_data,
         ij_as, ij_bs, ij_cs, ij_ds, Δabs, Δcds, Δys,
```
milankl: I think you can also pass on the whole `locator` and then unpack inside the kernel

RingGrids/src/interpolation_3d.jl
```diff
@@ -1,15 +1,24 @@
-export interpolate_3D!
+export interpolate_3D!, Center, Face
 
-# σ outside [σ_levels_full[1], σ_levels_full[end]] is pinned, not extrapolated
-@inline function find_vertical_bracket(σ, σ_levels_full)
-    n = length(σ_levels_full)
+"""Marker for fields located on full vertical levels (cell centers): u, v, T, etc."""
+struct Center end
```
milankl: maybe more general like

```julia
abstract type AbstractVerticalStaggering end
struct SigmaCenter{V} <: AbstractVerticalStaggering
    sigma::V
end

struct SigmaFaceAbove{V, T} <: AbstractVerticalStaggering
    sigma::V
    bottom_boundary_condition::T
end

struct SigmaFaceBelow{V, T} <: AbstractVerticalStaggering
    sigma::V
    top_boundary_condition::T
end

boundary(V::AbstractVerticalStaggering) = zero(eltype(V.sigma))
boundary(V::SigmaFaceAbove) = V.bottom_boundary_condition
boundary(V::SigmaFaceBelow) = V.top_boundary_condition

interpolate_3D!(Aout, A, locator, geometry, positions, σ::AbstractVector) = 
interpolate_3D!(Aout, A, locator, geometry, positions, SigmaCenter(σ))

function interpolate_3D!(..., sigma::AbstractVerticalStaggering)
    ...
    # decide based on sigma how to shift indices
    k_data_lo = ...
    # and how to define exceptions e.g.
    val_lo = k_data_lo == 0
        boundary(sigma)
    else
```

then you don't have to hardcode the boundary condition into the kernel and can simply pass it on.

RingGrids/src/interpolation.jl
```diff
@@ -521,7 +521,10 @@ function find_rings!(
         #@assert latd[end] == -90-ϵ "Latitudes latd are expected to contain -90˚, the south pole."
     end
 
-    return find_rings_unsafe!(js, Δys, θs, latd, architecture)
+    # Convert θs to the target architecture (moves CPU Vector to GPU CuArray when on GPU)
+    θs_arch = on_architecture(architecture, θs)
```
milankl: hold on this can be part of a hot loop, you don't want to move data here, this should be addressed on the callsite

Jullcifer: Oh shoot! I forgot to remove it after fixing the gpu error. I will revert this and lambda now as well. Thank you!

RingGrids/src/interpolation_3d.jl
Jullcifer: should be here 15bdf59d63960fa9fd6763d2d4a0c63fe86254ad

gregordecristoforo:
The documentation build error looks similar to #1237

RingGrids/src/interpolation_3d.jl
```diff
@@ -0,0 +1,155 @@
...
+    for i in 0:(n_north - 1)
+        north_sum += A_data[rs_north + i, k]
+    end
+    north_pole_average[k] = north_sum / n_north
+    n_south = nlons[nlat]
+    rs_south = ring_starts[nlat]
+    south_sum = zero(eltype(south_pole_average))
+    for i in 0:(n_south - 1)
+        south_sum += A_data[rs_south + i, k]
+    end
```
milankl, line 86: All our grids are symmetric around the equator so technically both loops could be merged, but given this is probably a really cheap kernel I probably suggest not to just in case, this version makes fewer assumptions which I like

RingGrids/src/interpolation_3d.jl
```diff
@@ -0,0 +1,155 @@
...
+    south_pole_average[k] = south_sum / n_south
+end
+
+# Vertically-blended anvil interpolation kernel, generic over the vertical staggering
+# (full level / face above / face below): staggering determines the data-column shift
+# and the boundary value substituted when a bracket index falls outside stored data.
+@kernel inbounds = true function _interpolate_3D_kernel!(
+        Aout, A_data, locator, positions, staggering,
+    )
+    (; ij_as, ij_bs, ij_cs, ij_ds, Δabs, Δcds, Δys, north_pole_average, south_pole_average) = locator
```
milankl, line 96: ah yeah nice the unpacking here, makes the kernel interface (all positional arguments) less prone to errors, great!

RingGrids/src/interpolation_3d.jl
```diff
@@ -0,0 +1,155 @@
...
+    i = @index(Global, Linear)
+    k_lo, k_hi, α = find_vertical_bracket(positions[i].σ, staggering.sigma)
```
milankl, line 98: one could also pass on `staggering` directly and then it's possible to dispatch in case there's changes needed in this function depending on centre vs face. But that doesn't seem to be the case so all good here and would be an easy change in the future if needed

RingGrids/src/interpolation_3d.jl
```diff
@@ -0,0 +1,155 @@
...
+    val_lo = if k_data_lo < 1 || k_data_lo > nlayers
+        boundary(staggering)
+    else
+        a = ij_as[i] == 0 ? north_pole_average[k_data_lo] : A_data[ij_as[i], k_data_lo]
+        b = ij_as[i] == 0 ? north_pole_average[k_data_lo] : A_data[ij_bs[i], k_data_lo]
+        c = ij_cs[i] == -1 ? south_pole_average[k_data_lo] : A_data[ij_cs[i], k_data_lo]
+        d = ij_cs[i] == -1 ? south_pole_average[k_data_lo] : A_data[ij_ds[i], k_data_lo]
+        ab = a + (b - a) * Δabs[i]
+        ab + (c + (d - c) * Δcds[i] - ab) * Δys[i]
+    end
```
milankl, line 114: just wondering whether there's a GPU performance penalty here with branching. Alternatively, one could do

```julia
k_data_lo = clamp(k_lo+s, 1, nlayers)
k_data_hi = clamp(k_hi+s, 1, nlayers)
outside = ~(1 <= k_data_lo <= nlayers)

a = ...
...

val_lo = ab + (c + (d - c) * Δcds[i] - ab) * Δys[i]
val_lo = outside ? boundary(staggering) : val_lo
```

to have the same data access for all threads but maybe that's also completely irrelevant because the `ij_*` arrays already cause a lot of nearby but not contiguous memory access ... maybe worth checking though what the impact would be

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -26,10 +26,17 @@ particle_advection!(vars, ::Nothing, ::AbstractModel) = nothing
 
 export ParticleAdvection2D
 
-export ParticleAdvection2D
-@kwdef struct ParticleAdvection2D{
+# geometry is `mutable` and untyped by architecture (AbstractGridGeometry, not a parametric
+# GeometryType) so that `initialize!` can rebind it onto model.architecture in place, since
+# geometry is fixed at construction time (typically before model.architecture is known).
```
milankl: don't agree with that, at construction the `spectral_grid` is known and the `architecture` in there is the same as in model or elsewhere and so known at construction
```julia
julia> model.architecture === spectral_grid.architecture === spectral_grid.grid.architecture === spectral_grid.spectrum.architecture
true
```

so mutability isn't needed as long as you construct with the `spectral_grid.architecture ` (or via `.grid`)

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -26,10 +26,17 @@ particle_advection!(vars, ::Nothing, ::AbstractModel) = nothing
...
+"""
+$(TYPEDFIELDS)
+
+Particle advection on a fixed model layer: particles are advected horizontally with the
+velocity of a single, user-specified vertical layer, their σ coordinate stays constant.
+"""
+@kwdef mutable struct ParticleAdvection2D{
```
milankl: I don't think there's a need to make this mutable and the `reconcile!` below

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -42,44 +49,98 @@ export ParticleAdvection2D
     "[OPTION] Advect with velocities from this vertical layer index"
     layer::IntType = 1
 
-    "[DERIVED] Time step used for particle advection (scaled by radius, converted to degrees) [s*˚/m]"
-    Δt::Base.RefValue{NF} = Ref(zero(NF))
-
     "[DERIVED] Interpolation geometry used during advection"
-    geometry::GeometryType
+    geometry::RingGrids.AbstractGridGeometry
```
milankl: I'd revert that

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -42,44 +49,98 @@ export ParticleAdvection2D
...
 end
 
 function ParticleAdvection2D(SG::SpectralGrid; kwargs...)
     geometry = GridGeometry(SG.grid; NF = SG.NF)
-    return ParticleAdvection2D{SG.NF, typeof(geometry), typeof(SG.truncation)}(; geometry, kwargs...)
+    return ParticleAdvection2D{SG.NF, typeof(SG.truncation)}(; geometry, kwargs...)
```
milankl: just make the geometry parametric again

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -42,44 +49,98 @@ export ParticleAdvection2D
...
+
+    "[OPTION] Execute particle advection every n timesteps"
+    every_n_time_steps::IntType = 6
+
+    "[DERIVED] Interpolation geometry used during advection"
+    geometry::RingGrids.AbstractGridGeometry
+end
+
+function ParticleAdvection3D(SG::SpectralGrid; kwargs...)
+    geometry = GridGeometry(SG.grid; NF = SG.NF)
```
milankl, line 92: Same here

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -42,44 +49,98 @@ export ParticleAdvection2D
...
+    } <: AbstractParticleAdvection
+
+    "[OPTION] Number of particles"
+    nparticles::IntType = 10
+
+    "[OPTION] Execute particle advection every n timesteps"
+    every_n_time_steps::IntType = 6
+
+    "[DERIVED] Interpolation geometry used during advection"
+    geometry::RingGrids.AbstractGridGeometry
```
milankl: same here

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -42,44 +49,98 @@ export ParticleAdvection2D
...
+
+# particle_advection.geometry is fixed at ParticleAdvection2D/3D construction time (typically
+# before model.architecture is known), so it may not match model.architecture. Rebind it onto the
+# model's architecture in place if needed (geometry is mutable for this reason).
+function _reconcile_architecture!(particle_advection::AbstractParticleAdvection, model::AbstractModel)
+    (; geometry) = particle_advection
+    ismatching(model.architecture, geometry.grid.architecture) && return particle_advection
+    particle_advection.geometry = on_architecture(model.architecture, geometry)
+    return particle_advection
+end
```
milankl: this is not needed, as long as the particle advection has been created from `spectral_grid` the architectures always match!

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -152,6 +214,59 @@ end
...
+    )
+
+    RingGrids.update_locator!(locator, geometry, lons, lats)
+    u0 = vars.particles.u
+    v0 = vars.particles.v
+    w0 = vars.particles.w
+
+    interpolate_3D!(u0, u_3d, locator, geometry, particles, SigmaCenter(σ_levels_full))
+    interpolate_3D!(v0, v_3d, locator, geometry, particles, SigmaCenter(σ_levels_full))
+    interpolate_3D!(w0, w_3d, locator, geometry, particles, SigmaFaceBelow(σ_levels_half, zero(eltype(σ_levels_half))))
```
milankl, line 234: brief explanation here on the callsite about centre vs face?

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -152,6 +214,59 @@ end
...
+    return nothing
+end
+
+# Modulo particles and extract horizontal coordinates; σ is each particle's own
+@kernel inbounds = true function _initialize_3D_particles_kernel!(
+        particles, lons, lats
+    )
+    i = @index(Global, Linear)
+    particles[i] = mod(particles[i])
```
milankl: you can also pass functions to kernels in case you wanted to avoid writing 2 almost identical `_initialize_particles_kernel!`

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -173,18 +288,23 @@ function particle_advection!(
...
 
     # HEUN: PREDICTOR STEP, use u, v at previous time step and location
-    Δt = particle_advection.Δt[]        # time step [s*˚/m]
+    # calculate time step on the fly
+    Δt = model.time_stepping.Δt                 # time step [s]
+    Δt *= n * convert(NF, 180 / (π * radius))   # scale to [s*°/m] to obtain [˚] when multiplied with velocity in [m/s]
     Δt_half = Δt / 2                    # /2 because Heun is average of Euler+corrected step
 
-    u_old = vars.particles.u            # from previous time step and location
+    u_old = vars.particles.u            # from previous time step and locationq
```
milankl: typo

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -173,18 +288,23 @@ function particle_advection!(
...
     clock.time_step_counter % n == (n - 1) || return nothing
+    NF = eltype(eltype(particles))      # number format used for particle locations
+    (; radius) = model.planet
 
     # HEUN: PREDICTOR STEP, use u, v at previous time step and location
-    Δt = particle_advection.Δt[]        # time step [s*˚/m]
+    # calculate time step on the fly
+    Δt = model.time_stepping.Δt                 # time step [s]
+    Δt *= n * convert(NF, 180 / (π * radius))   # scale to [s*°/m] to obtain [˚] when multiplied with velocity in [m/s]
     Δt_half = Δt / 2                    # /2 because Heun is average of Euler+corrected step
```
milankl, line 285: can we include the backwards particle advection by dt -> -dt from #1136 here?

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -227,22 +347,106 @@ function particle_advection!(
...
         u::NF,                                  # zonal velocity [m/s]
         v::NF,                                  # meridional velocity [m/s]
-        dt::NF,                                 # scaled time step [s*˚/m]
+        Δt::NF,                                 # scaled time step [s*˚/m]
     ) where {NF}
 
-    dlat = v * dt                               # increment in latitude [˚N]
-    
-    # TODO: Replace `cos(deg2rad(particle.lat))` with `cosd(particle.lat)` once 
-    # JuliaGPU/AMDGPU.jl#1041 is merged/released and `cosd` is supported on AMD GPUs.
```
milankl, line 240: retain this comment so we know why it's not simply `cosd` which would pass everywhere but on AMDGPU

SpeedyWeather/src/variables/dimensions.jl
```diff
@@ -235,15 +243,15 @@ fuse_trailing_n(d::SpectralXYZT) = d.n
...
-            d isa Spectral3D ? d.n :
-            d isa SpectralXYT ? d.n :
-            d isa Grid3D ? d.n :
-            d isa GridXYT ? d.n :
-            get_nlayers(model)
+        d isa Spectral3D ? d.n :
+        d isa SpectralXYT ? d.n :
+        d isa Grid3D ? d.n :
+        d isa GridXYT ? d.n :
+        get_nlayers(model)
```
milankl, line 254: I bet this is just undoing runic formatting ...

milankl (commented):
Super cool! I think there's just the point about the struct doesn't have to be mutable. Otherwise just some small comments

SpeedyWeather/src/variables/dimensions.jl
Jullcifer: When I couldn't figure out why the documentation build was failing, I tasked Claude to scan through the changed scripts and find problems in formatting - didn't run it through runic afterward. Will re-run runic and push a clean diff.

milankl: All good, we don't enforce Runic because that seems tedious and silly but every now and then we just reformat the whole repo, so that people and Claude write following a style as they see it.

RingGrids/src/interpolation_3d.jl
milankl: just realised "nearby" is only true for a semi-Lagrangian advection scheme (which I eventually plan to build on top of this) for particle advection the `ij`s are typically all over the place because chaos and there's no "sorting" of the particles like a semi-Lagrangian scheme would do

SpeedyWeather/src/dynamics/particle_advection.jl
Jullcifer: So, the `model.architecture` is always the same object as `spectral_grid.architecture`, and the  ParticleAdvection2D/3D is built from the same `spectral_grid` that later gets passed into the model (if I understand correctly now?) So there is no need for geometry to be mutable..

Jullcifer: Reverted geometry back to immutable and removed _reconcile_architecture!

milankl: > So, the `model.architecture` is always the same object as `spectral_grid.architecture`, and the ParticleAdvection2D/3D is built from the same `spectral_grid` that later gets passed into the model (if I understand correctly now?) So there is no need for geometry to be mutable..

Yes, when you create a model this happens

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/e7c569df54f64cdba89916303901b259307f8887/SpeedyWeather/src/models/primitive_wet.jl#L59-L60

so because you write `model = PrimitiveWetModel(spectral_grid, ...)` then the "two" architectures are the same and in fact I don't know what would happen if you deliberate chose them to be different, it's not something we ever intended (or easily makes sense). But to be precise, because architectures are immutable objects like integers, their egality doesn't mean it's actually the same object in memory like `1` isn't distinguishable from another `1` as it doesn't have a single address in memory like mutable objects have, see

```julia
help?> ===
search: === == !==

  ===(x,y) -> Bool
  ≡(x,y) -> Bool

  Determine whether x and y are identical, in the sense that no program could
  distinguish them. First the types of x and y are compared. If those are
  identical, mutable objects are compared by address in memory and immutable
  objects (such as numbers) are compared by contents at the bit level. This
  function is sometimes called "egal".
```

RingGrids/src/interpolation_3d.jl
Jullcifer: I ran some tests locally, just on my laptop, and branch doesn't look like the bottleneck, but I don't want to say anything for sure yet. I will rerun it on the cluster first, and report back.

Jullcifer: We ran benchmark comparissons with @gregordecristoforo today, between the current (branching) and branchless (clamp+select) across 4mil particles, and the branchless variant is definitely faster on the clusters' GPUs. However, on my laptop it is actually slower: 

| Hardware | Benchmark Case | branching | branchless (clamp+select) | Speedup |
| :--- | :--- | :--- | :--- | :--- |
| NVIDIA GH200 96GB | large grid (271k points): random σ order, nlayers=24 | 0.201 ms | 0.191 ms| 1.05x (+5.5%)|
| NVIDIA GH200 96GB| random σ order, nlayers=8 (~1/8 divergent per warp) | 0.669 ms | 0.174 ms| 3.83x (+283%) |
| AMD MI250X | large grid (271k points): random σ order, nlayers=24 | 1.485 ms | 1.449 ms | 1.02x (+2.5%) |
| NVIDIA MX550 (Laptop GPU)| large grid (271k points): random σ order, nlayers=24 | 47.54 ms | 53.72 ms | 0.88x (-11.5%) |
| AMD EPYC (32 CPU Threads) | large grid (271k points)| 9.82 ms | 9.95 ms | 0.99x (-1.3%) |
| AMD EPYC (32 CPU Threads) | sorted σ order, nlayers=8| 2.04 ms | 3.68 ms | 0.55x (-44.6%) |

It makes sence to me to go for the branchless solution, I'll work on that. But we can also probably use multiple dispatch on `architecture` to launch `_interpolate_3D_kernel_branchless!` for `SpeedyWeather.GPU()` while preserving `_interpolate_3D_kernel!` for `SpeedyWeather.CPU()` (to prevent performance regressions for CPU users at least). What do you think is best, @milankl?

milankl: How did you choose the horizontal locations here, are they also random or in some way sorted? Overall a ~2x performance penalty in CPU is considered acceptable if it leads to higher GPU performance. But you're right here we can just do both with dispatching. It's a bit of duplicated code but given it's quite clear what the difference is I'd be happy to pay that price.

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -227,26 +219,30 @@ function initialize!(
...
-    )
+# Modulo particles and extract horizontal coordinates. `f` is applied to each particle
+# before moduloing: `identity` keeps its own σ (3D advection), `p -> set(p; σ)` pins σ to a
+# fixed layer (2D advection) — passing the function in avoids two near-identical kernels.
+@kernel inbounds = true function _initialize_particles_kernel!(
+        particles, lons, lats, f::F,
+    ) where {F}
     i = @index(Global, Linear)
-    particles[i] = mod(particles[i])
+    particles[i] = mod(f(particles[i]))
```
milankl, line 251: Yes, absolutely! Every time I see code like this I'm again amazed about the power in simplicity here. Pass on a function as an argument and because a function is basically it's own type the compiler will inline this function into the kernel, compile an independent version for every function and so you only have to write it once but you get infinitely many efficient GPU kernels for free, so cool.

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -190,6 +190,10 @@ function initialize!(
     return nothing
 end
 
+# `vars.dynamics.w` is radius*σ̇ as vor, div stay radius-scaled throughout `run!`. Particles move
+# in physical coordinates, the vertical counterpart of the 180/(π*radius) factor for u, v below.
+@inline unscale_vertical_velocity!(w, radius) = (w ./= radius)   # radius*σ̇ -> σ̇ [1/s]
```
milankl: could be done as `*` with the inverse, the floating-point result isn't the same because of double rounding but that's OK

RingGrids/src/interpolation_3d.jl
Jullcifer: So, horizontal locations were random (uniform lon/lat), we didn't sort them. But since you mentioned it, I reran it, including the horizontally sorted cases, on the server (only NVIDIA GH200 and CPU): 100 trials × 50 calls per case. (and good thing I did!) 
The 3.8x speedup on GH200 didn't reproduce. 9 of the 10 timings from my earlier run agree within ~7%, but branching in the second case went from 0.669 to 0.160 ms. 

**GH200, 4M particles:** 0.93–1.05x across all 8 cases, including the horizontally sorted ones, so no measurable GPU difference.
**EPYC, 32 threads:** branchless is 8–21% slower in 7/8 cases (only 1.18x faster in the nlayers=2 worst case).
(The MI250X number was a run on a different system and I haven't rerun it.)

So with no GPU gain, maybe we keep the current branching kernel? Happy to revisit once the semi-Lagrangian scheme exists.

_NB: the 8 cases_ = random/sorted σ with nlayers=8 and random σ with nlayers=2 on FullGaussianGrid(64), plus random/sorted σ, horizontally sorted with random/sorted σ, and a uniform-lat sampling variant, all with nlayers=24 on OctahedralGaussianGrid(256).

milankl: Happy for you to make this call, I just wanted to flag here that branching may have an impact on performance, but it doesn't necessarily need to!

Also feel free to "resolve conversation", I'm loosing a little bit the overview whether there's any remaining issues

SpeedyWeather/src/dynamics/particle_advection.jl
Jullcifer: But it is an easy fix, so let's just fix it :) changed

milankl:
Seems like all points here were resolved? I'll ask @milankl-claude to do another review in case my last missed something, what about testing this and adding some simple 3D to the documentation, e.g. could just be a super simple plot with random initial conditions that just shows lines in x=time, y=sigma added to the horizontal visualisation we already have? Numerically, particle advection is pretty stable so it would be good to have a visual way to verify that the trajectories are actually smooth and reasonable. Including this into the docs we can track when something goes wrong!

milankl-claude (commented):
Hi @Jullcifer, @milankl asked me for another review pass. First of all, congratulations on your first big SpeedyWeather feature, this is a really solid piece of work! 🎉

I checked the current state of the branch against the earlier discussion. Almost everything has been resolved well:
- ✅ `nothing`-typed pole vectors are gone, and the locator is isbits-friendly again (GPU full simulation test 🎉)
- ✅ `nlayers` lives in the RingGrids-owned `AnvilLocator` with `nlayers = 1` default, so no separate 3D constructor
- ✅ `north_pole_average` naming, `σ_levels` naming, von Neumann comment
- ✅ `AbstractVerticalStaggering` with `SigmaCenter`/`SigmaFaceAbove`/`SigmaFaceBelow`, which is very clean
- ✅ geometry immutable again, no `reconcile!`
- ✅ `backwards` from #1136 included for both 2D and 3D (and the old `Δt` RefValue gone from `ParticleAdvection2D`, nice cleanup)
- ✅ one `_initialize_particles_kernel!` with a function argument
- ✅ branching vs branchless: benchmarked carefully, well-reasoned decision to keep branching

**Highlights** 🌟 the staggering design (`shift` + `boundary` dispatch), passing `f` into the init kernel, finding the radius scaling of `w` and writing a test that pins it down, and the physically meaningful tests (2D/3D consistency, forward/backward symmetry). Also the new docs section on horizontal precompute vs vertical on-the-fly is very clear.

**Still open from earlier discussion**
- Runic reverts in `dimensions.jl` are still in the diff
- σ-vs-time plot in the docs requested by @milankl (sketch inline)

**Should fix before merging**
- `interpolate_3D!` has no size checks. With the *default* `AnvilLocator(NF, npoints)` (`nlayers = 1`) and a 3D field, the `inbounds` pole-average kernel writes out of bounds silently
- No test covers `SigmaFaceBelow`/`SigmaFaceAbove`, although that's the path `w` takes (test suggested inline)

**Simplifications** (optional but would make the code smaller and more consistent)
- `LocatorDim`: always allocate with `nlayers`, no `isa` branch
- symmetric `variables(P::ParticleAdvection2D/3D)` methods
- reuse `anvil_average` in the 3D kernel via a small helper, removing the duplicated lo/hi blocks
- fold `1/radius` into `Δt_vert` instead of `unscale_vertical_velocity!` (saves broadcasts, mirrors the horizontal `Δt`)
- default-zero `SigmaFaceBelow(σ)` constructor and build the staggerings once per call
- `advect_2D` = `advect_3D` with `w = 0`
- informative error for `ParticleAdvection3D` with non-primitive models

Everything else is small (exports, stale comments, test placement, and the `cosin-distribution` typo in the `initialize!` docstring at L139, which is outside the diff so I can't comment inline). I didn't run anything since tests pass and nothing looked broken beyond the out-of-bounds footgun above. Great job! 🚀

RingGrids/src/interpolation_3d.jl
```diff
@@ -0,0 +1,155 @@
+export interpolate_3D!, SigmaCenter, SigmaFaceAbove, SigmaFaceBelow
```
milankl-claude, line 1: Exports in RingGrids are collected in `RingGrids/src/RingGrids.jl`, and `interpolate_3D!` is now exported in both places. I'd remove this line and add the three staggering types to the `export` block over there instead.

RingGrids/src/interpolation_3d.jl
```diff
@@ -0,0 +1,155 @@
...
+
+# data column index = bracket index k + shift(staggering); out-of-[1, nlayers] means
+# the value isn't stored and boundary(staggering) is substituted instead.
+@inline shift(::SigmaCenter) = 0
+@inline shift(::SigmaFaceAbove) = 0
+@inline shift(::SigmaFaceBelow) = -1
+
+@inline boundary(s::AbstractVerticalStaggering) = zero(eltype(s.sigma))
+@inline boundary(s::SigmaFaceAbove) = s.bottom_boundary_condition
+@inline boundary(s::SigmaFaceBelow) = s.top_boundary_condition
```
milankl-claude, line 53: This is really nice ✨ `shift` + `boundary` reduces the whole centre/face distinction to two one-line methods, and the kernel doesn't need to know about any of it. It's exactly the kind of small dispatch-based design we want more of.

Small addition: since the kinematic boundary condition (zero) is by far the most common case, a default would make the call sites shorter:

```julia
SigmaFaceAbove(σ::AbstractVector) = SigmaFaceAbove(σ, zero(eltype(σ)))
SigmaFaceBelow(σ::AbstractVector) = SigmaFaceBelow(σ, zero(eltype(σ)))
```

Then it's just `SigmaFaceBelow(σ_levels_half)` in particle_advection.jl.

RingGrids/src/interpolation_3d.jl
```diff
@@ -0,0 +1,155 @@
...
+@inline function find_vertical_bracket(σ, σ_levels)
+    n = length(σ_levels)
+    k_lo = 1
+    for k in 1:(n - 1)
+        σ_levels[k] <= σ && (k_lo = k)
+    end
+    k_hi = k_lo + 1
+    Δσ = σ_levels[k_hi] - σ_levels[k_lo]
+    α = clamp((σ - σ_levels[k_lo]) / Δσ, zero(σ), one(σ))
+    return k_lo, k_hi, α
```
milankl-claude, line 66: Edge case: with a single level (`SigmaCenter` and `nlayers = 1`) the loop is empty, `k_hi = 2`, and `σ_levels[2]` is read out of bounds. The kernel runs with `inbounds = true`, so this doesn't throw, it just reads garbage. I'm not sure anyone runs a 1-layer primitive equation model with 3D particles, but a check in `interpolate_3D!` (see comment below) would catch it for free.

RingGrids/src/interpolation_3d.jl
```diff
@@ -0,0 +1,155 @@
...
+    val_hi = if k_data_hi < 1 || k_data_hi > nlayers
+        boundary(staggering)
+    else
+        a = ij_as[i] == 0 ? north_pole_average[k_data_hi] : A_data[ij_as[i], k_data_hi]
+        b = ij_as[i] == 0 ? north_pole_average[k_data_hi] : A_data[ij_bs[i], k_data_hi]
+        c = ij_cs[i] == -1 ? south_pole_average[k_data_hi] : A_data[ij_cs[i], k_data_hi]
+        d = ij_cs[i] == -1 ? south_pole_average[k_data_hi] : A_data[ij_ds[i], k_data_hi]
+        ab = a + (b - a) * Δabs[i]
+        ab + (c + (d - c) * Δcds[i] - ab) * Δys[i]
+    end
```
milankl-claude, line 125: The two branches are identical apart from `k_data_lo`/`k_data_hi`, and the last two lines of each re-implement `anvil_average` from interpolation.jl. A small helper would remove the duplication and reuse the 2D code path. It also mirrors the tuple form `_interpolate_kernel!` uses for the poles:

```julia
# anvil interpolation of layer k of A_data onto point i, pole values from the locator
Base.@propagate_inbounds function anvil_average(A_data, k, i, locator)
    (; ij_as, ij_bs, ij_cs, ij_ds, Δabs, Δcds, Δys, north_pole_average, south_pole_average) = locator
    a, b = ij_as[i] == 0 ? (north_pole_average[k], north_pole_average[k]) : (A_data[ij_as[i], k], A_data[ij_bs[i], k])
    c, d = ij_cs[i] == -1 ? (south_pole_average[k], south_pole_average[k]) : (A_data[ij_cs[i], k], A_data[ij_ds[i], k])
    return anvil_average(a, b, c, d, Δabs[i], Δcds[i], Δys[i])
end
```

(`@propagate_inbounds` rather than `@inline` so the kernel's `inbounds = true` still applies inside the helper.) The kernel body then becomes

```suggestion
    val_lo = 1 <= k_data_lo <= nlayers ? anvil_average(A_data, k_data_lo, i, locator) : boundary(staggering)
    val_hi = 1 <= k_data_hi <= nlayers ? anvil_average(A_data, k_data_hi, i, locator) : boundary(staggering)
```

The branching structure is the same, so your benchmark conclusion still holds (and thanks for doing those so carefully, including re-running when a number looked off!). Line 96 can then be removed too.

RingGrids/src/interpolation_3d.jl
```diff
@@ -0,0 +1,155 @@
...
+Pole ring averages are stored in `locator.north_pole_average` and
+`locator.south_pole_average` (pre-allocated via the 3D `AnvilLocator` constructor) and
+reused across calls without CPU allocations."""
+function interpolate_3D!(Aout, A, locator, geometry, positions, staggering::AbstractVerticalStaggering)
+    (; npoints_output, north_pole_average, south_pole_average) = locator
+    A_data = A.data
+    arch = architecture(Aout)
+    (; ring_starts, nlons, nlat) = geometry
+    launch!(arch, LinearWorkOrder, (size(A_data, 2),), _compute_pole_averages_kernel!,
+            north_pole_average, south_pole_average, A_data, ring_starts, nlons, nlat)
```
milankl-claude, line 142: Two things here:

1. **Size checks.** The default `AnvilLocator(NF, npoints)` has `nlayers = 1`, so pole-average vectors of length 1. If someone passes that locator with a 3D field, this kernel writes `north_pole_average[k]` for `k > 1`, and because of `inbounds = true` that's silent memory corruption, not an error. Given that's the *default* constructor, I think this needs a guard, similar to the checks in the 2D `interpolate!`:

```julia
nlayers = size(A.data, 2)
length(north_pole_average) >= nlayers || throw(DimensionMismatch(
    "Locator has pole buffers for $(length(north_pole_average)) layers but field has $nlayers layers, " *
    "create it with AnvilLocator(NF, npoints, nlayers)."))
length(Aout) == npoints_output || throw(DimensionMismatch("Aout has length $(length(Aout)) but locator $npoints_output."))
```

(optionally also check that `length(staggering.sigma)` is `nlayers` for `SigmaCenter` and `nlayers + 1` for the faces, which would also cover the single-layer edge case above).

2. For consistency with the interpolation kernel below, which gets the whole `locator` and unpacks it inside, you could pass `geometry` here and unpack `ring_starts, nlons, nlat` in the kernel.

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -54,34 +59,80 @@ function ParticleAdvection2D(SG::SpectralGrid; kwargs...)
     return ParticleAdvection2D{SG.NF, typeof(geometry), typeof(SG.truncation)}(; geometry, kwargs...)
 end
 
-function variables(P::ParticleAdvection2D)
-    (; nparticles) = P
+export ParticleAdvection3D
+
+# σ_levels_full is read from model.geometry at runtime.
```
milankl-claude, line 64: Stray comment, and it now sits between the `export` and the docstring. I'd remove it.

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -54,34 +59,80 @@ function ParticleAdvection2D(SG::SpectralGrid; kwargs...)
...
+# model.particle_advection isa ParticleAdvection3D to embed north/south_pole_average
+# arrays in the locator (type-stable, avoids per-call allocation)
+function variables(P::ParticleAdvection3D, ::AbstractModel)
+    (; nparticles) = P
+    return (
+        _common_particle_variables(nparticles)...,
+        ParticleVariable(:locator, LocatorDim(), desc = "Particle locator with embedded pole averages for 3D interpolation", units = "1"),
+        ParticleVariable(:w, VectorDim(nparticles), desc = "Vertical velocity dσ/dt at particle location", units = "1/s"),
+    )
+end
```
milankl-claude, line 125: The `variables` interface ended up asymmetric: 2D goes `variables(P)` → `variables(typeof(P), nparticles)`, while 3D defines `variables(P, ::AbstractModel)` with an unused model argument. The comment on L115-117 is also stale now that `LocatorDim` has no fields. Since `LocatorDim` can be the same for both (see my comment in dimensions.jl), this could simply be:

```julia
function particle_variables(nparticles)
    return (
        PrognosticVariable(:particles, ParticleVectorDim(nparticles), desc = "Particle locations", units = "˚/1"),
        ParticleVariable(:locations, ParticleVectorDim(nparticles), desc = "Predicted particle locations", units = "˚/1"),
        ParticleVariable(:u, VectorDim(nparticles), desc = "Zonal velocity at particle location", units = "m/s"),
        ParticleVariable(:v, VectorDim(nparticles), desc = "Meridional velocity at particle location", units = "m/s"),
        ParticleVariable(:locator, LocatorDim(), desc = "Particle locator for horizontal interpolation", units = "1"),
    )
end

variables(P::ParticleAdvection2D) = particle_variables(P.nparticles)
variables(P::ParticleAdvection3D) = (
    particle_variables(P.nparticles)...,
    ParticleVariable(:w, VectorDim(nparticles), desc = "Vertical velocity at particle location", units = "1/s"),
)
```

(I also tweaked the duplicated "Particle locations" desc of `:locations`. That one was already there before this PR, but it's a good moment to fix it.)

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -54,34 +59,80 @@ function ParticleAdvection2D(SG::SpectralGrid; kwargs...)
...
-    (; every_n_time_steps) = particle_advection
-    # Δt [˚*s/m] is scaled by radius to convert more easily from velocity [m/s]
-    # to [˚/s] for particle locations in degree
-    particle_advection.Δt[] = every_n_time_steps * model.time_stepping.Δt
-    particle_advection.Δt[] *= (360 / 2π)
-    return particle_advection
+    return nothing
 end
 
+initialize!(::ParticleAdvection3D, ::AbstractModel) = nothing
```
milankl-claude, line 134: The docs say 3D advection only works for the primitive equation models. With a `BarotropicModel` this currently fails later with an obscure error (no `vars.dynamics.w`, or `SigmaCoordinates` of 1 layer). An informative error at initialization would help users:

```suggestion
initialize!(::ParticleAdvection3D, ::PrimitiveEquation) = nothing
initialize!(::ParticleAdvection3D, model::AbstractModel) =
    throw(ArgumentError("ParticleAdvection3D requires a PrimitiveDryModel or PrimitiveWetModel, use ParticleAdvection2D for $(nameof(typeof(model)))."))
```

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -139,15 +190,65 @@ function initialize!(
     return nothing
 end
 
-# Kernel to modulo particles and extract their coordinates
-@kernel inbounds = true function _initialize_particles_kernel!(
-        particles, lons, lats, σ
+# `vars.dynamics.w` is radius*σ̇ as vor, div stay radius-scaled throughout `run!`. Particles move
+# in physical coordinates, the vertical counterpart of the 180/(π*radius) factor for u, v below.
+@inline unscale_vertical_velocity!(w, radius) = (w .*= inv(radius))   # radius*σ̇ -> σ̇ [1/s]
```
milankl-claude, line 195: Nice that you tracked down the radius scaling of `w`, and thanks for the regression test in time_stepping.jl that pins it down!

One simplification: `unscale_vertical_velocity!` adds a broadcast over all particles after every `w` interpolation (twice per advection step, plus once at init). You could instead fold `1/radius` into the vertical time step, which is exactly how the horizontal part already handles it (`Δt` in [s*˚/m]):

```julia
# vars.dynamics.w is radius*σ̇ [m/s] (vor, div stay radius-scaled throughout run!), so like the
# horizontal Δt [s*˚/m] also scale the vertical time step by 1/radius to get dσ = w*Δt_vert
Δt_vert = model.time_stepping.Δt * n / radius   # [s/m]
```

Then `vars.particles.w` is in m/s like `u, v` (update the `units` in `variables`), and `unscale_vertical_velocity!` plus the three calls to it can go.

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -139,15 +190,65 @@ function initialize!(
...
+    RingGrids.update_locator!(locator, geometry, lons, lats)
+    u0 = vars.particles.u
+    v0 = vars.particles.v
+    w0 = vars.particles.w
+
+    # u, v sit on full (center) levels; w sits on half levels, storing the lower face
+    # (k+½) with an implicit zero at the top boundary (σ=0, kinematic condition)
+    interpolate_3D!(u0, u_3d, locator, geometry, particles, SigmaCenter(σ_levels_full))
+    interpolate_3D!(v0, v_3d, locator, geometry, particles, SigmaCenter(σ_levels_full))
+    interpolate_3D!(w0, w_3d, locator, geometry, particles, SigmaFaceBelow(σ_levels_half, zero(eltype(σ_levels_half))))
```
milankl-claude, line 239: Minor: the same `SigmaFaceBelow(σ_levels_half, zero(eltype(σ_levels_half)))` (and `SigmaCenter(σ_levels_full)`) is constructed 3× here and 6× in `particle_advection!`. With the default-zero constructor suggested in interpolation_3d.jl you could define them once at the top:

```julia
# u, v on full levels; w on half levels storing the lower face (k+½), zero at σ=0 (kinematic BC)
uv_levels = SigmaCenter(model.geometry.σ_levels_full)
w_levels = SigmaFaceBelow(model.geometry.σ_levels_half)
```

and every `interpolate_3D!` call fits on one short line.

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -139,15 +190,65 @@ function initialize!(
...
+end
+
+# Modulo particles and extract horizontal coordinates. `f` is applied to each particle
+# before moduloing: `identity` keeps its own σ (3D advection), `p -> set(p; σ)` pins σ to a
+# fixed layer (2D advection) — passing the function in avoids two near-identical kernels.
+@kernel inbounds = true function _initialize_particles_kernel!(
+        particles, lons, lats, f::F,
+    ) where {F}
+    i = @index(Global, Linear)
+    particles[i] = mod(f(particles[i]))
```
milankl-claude, line 251: Love this ❤️ One kernel, specialised by the compiler for `identity` and the `set` closure, and the `f::F ... where {F}` makes sure it really specialises.

SpeedyWeather/src/dynamics/particle_advection.jl
```diff
@@ -227,22 +334,119 @@ function particle_advection!(
...
+        Δt::NF,         # horizontal: [s*˚/m]
+        Δt_vert::NF,    # vertical: [s]; dσ = w[s⁻¹] * dt_vert[s]
+    ) where {NF}
+
+    dlat = v * Δt
+    coslat = max(cos(deg2rad(particle.lat)), eps(NF))
+    dlon = u * Δt / coslat
+    dσ = w * Δt_vert
+    return mod(move(particle, dlon, dlat, dσ))
+end
```
milankl-claude, line 448: `advect_2D` is exactly `advect_3D` with `w = 0`, so one could be written in terms of the other:

```julia
# 2D: no vertical movement, σ is unchanged (and only clamped by mod)
@inline advect_2D(particle::Particle{NF}, u::NF, v::NF, Δt::NF) where {NF} =
    advect_3D(particle, u, v, zero(NF), Δt, zero(NF))
```

Then the AMDGPU `cosd` TODO comment only needs to live in `advect_3D`, where it's currently missing. The same idea would work for the predictor/corrector kernels and even the two `particle_advection!` methods, which are nearly identical apart from which velocities get interpolated. But that's a bigger refactor I'm happy to leave for a follow-up so this PR doesn't grow further.

SpeedyWeather/src/variables/dimensions.jl
```diff
@@ -154,11 +154,19 @@ allocate(v::AbstractVariable{MatrixDim}, model::AbstractModel) = fill!(model.spe
...
+function allocate(::AbstractVariable{LocatorDim}, model::AbstractModel)
+    (; NF, architecture, nlayers) = model.spectral_grid
+    (; nparticles) = model.particle_advection
+
+    if model.particle_advection isa ParticleAdvection3D
+        return RingGrids.AnvilLocator(NF, nparticles, nlayers; architecture)
+    else
+        return RingGrids.AnvilLocator(NF, nparticles; architecture)
+    end
 end
```
milankl-claude, line 169: Following up on @milankl's earlier comment that it's fine to always store the two `nlayers`-length vectors: the `isa ParticleAdvection3D` branch (and the reference to a dynamics type from the variables code) isn't needed:

```suggestion
function allocate(::AbstractVariable{LocatorDim}, model::AbstractModel)
    (; NF, architecture, nlayers) = model.spectral_grid
    return RingGrids.AnvilLocator(NF, model.particle_advection.nparticles, nlayers; architecture)
end
```

SpeedyWeather/src/variables/dimensions.jl
```diff
@@ -209,8 +217,8 @@ fuse_family(::Spectral4D) = :spectral
 fuse_family(::SpectralXYZT) = :spectral
 fuse_family(d::AbstractVariableDim) = error(
     "Fusion is not supported for dim type $(typeof(d)). " *
-        "Supported dim types: Grid2D, Grid3D, GridXYZ, GridXYT, Grid4D, GridXYZT, " *
-        "Spectral2D, Spectral3D, SpectralXYZ, SpectralXYT, Spectral4D, SpectralXYZT."
+    "Supported dim types: Grid2D, Grid3D, GridXYZ, GridXYT, Grid4D, GridXYZT, " *
+    "Spectral2D, Spectral3D, SpectralXYZ, SpectralXYT, Spectral4D, SpectralXYZT."
 )
```
milankl-claude, line 222: These hunks (and the ones further down in `fused_slots` / `_fuse_rank_and_n`) still undo the Runic formatting mentioned earlier. `git checkout origin/main -- ` for those lines, or running Runic on just this file, would drop them from the diff.

SpeedyWeather/test/GPU/particle_advection_3d.jl
```diff
@@ -0,0 +1,57 @@
...
+    spectral_grid = SpectralGrid(truncation = 31, nlayers = 4, architecture = arch)
+    particle_advection = ParticleAdvection3D(spectral_grid, nparticles = 20)
+
+    model = PrimitiveDryModel(spectral_grid; particle_advection)
+    model.feedback.verbose = false
+    simulation = initialize!(model)
+    vars = simulation.variables
+
+    # call particle_advection! directly, bypassing the time_step!/column_parameterizations!
+    # pipeline that currently fails to compile on GPU, see testset below and PR #1215
```
milankl-claude, line 36: Stale comment: the full simulation compiles on GPU now (testset below 🎉).

Also, @milankl-claude found earlier that the `Union{Nothing, ...}` locator broke **2D** particle advection on GPU as well. Since there's no GPU test running a full simulation with `ParticleAdvection2D`, would you add one next to the 3D one below? That way the regression can't come back unnoticed.

SpeedyWeather/test/GPU/CUDA/Project.toml
```diff
@@ -4,9 +4,11 @@ CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
...
 KernelAbstractions = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
 Reactant = "3c362404-f566-11ee-1572-e11a4b42c853"
+RingGrids = "d1845624-ad4f-453b-8ff4-a8db365bf3a7"
 SpeedyWeather = "9e226e20-d153-4fed-8a5b-493def4f21a9"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Zarr = "0a941bbe-ad1d-11e8-39d9-ab76183a1d99"
 
 [sources]
 SpeedyWeather = {path = "../../.."}
+RingGrids = {path = "../../../../RingGrids"}
```
milankl-claude, line 14: If the RingGrids source is needed here to pick up the local RingGrids changes, the AMDGPU and MetalGPU test environments need it too, since they run the same `particle_advection_3d.jl`.

SpeedyWeather/test/dynamics/particle_advection.jl
```diff
@@ -30,3 +30,182 @@
         end
     end
 end
+
+@testset "find_vertical_bracket" begin
+    σf = Float32[0.1, 0.3, 0.5, 0.7, 0.9]
```
milankl-claude, line 35: `find_vertical_bracket` is a RingGrids function, so I'd move this testset to `RingGrids/test/interpolation_3d.jl`. It's also good to see the boundary cases covered explicitly.

SpeedyWeather/test/dynamics/particle_advection.jl
```diff
@@ -30,3 +30,182 @@
...
+        spectral_grid = SpectralGrid(truncation = 32, nlayers = 4)
+        pa = ParticleAdvection3D(spectral_grid, nparticles = 20)
+        @test pa.nparticles == 20
+        @test pa.every_n_time_steps == 6
+
+        model = Model(spectral_grid; particle_advection = pa)
+        model.feedback.verbose = false
+        simulation = initialize!(model)
+
+        σ_initial = copy([p.σ for p in simulation.variables.prognostic.particles])
```
milankl-claude, line 67:
```suggestion
        σ_initial = [p.σ for p in simulation.variables.prognostic.particles]
```
the comprehension already allocates a new array

SpeedyWeather/test/dynamics/particle_advection.jl
```diff
@@ -30,3 +30,182 @@
...
+    run!(simulation_3d, period = Hour(6), output = false)
+    run!(simulation_2d, period = Hour(6), output = false)
+
+    particles_3d = simulation_3d.variables.prognostic.particles
+    particles_2d = simulation_2d.variables.prognostic.particles
+    dlon = maximum(abs(wrapped_dlon(particles_3d[i].lon, particles_2d[i].lon)) for i in eachindex(lons))
+    dlat = maximum(abs(particles_3d[i].lat - particles_2d[i].lat) for i in eachindex(lons))
+
+    # tracks agree to ~0.03˚ while the 3D particles stay near layer k; they diverged by ~4˚ when
+    # the radius scaling of w flung them onto σ=0/1, into the top/bottom layer winds.
```
milankl-claude, line 191: These tests are really good: the 2D/3D consistency check and the forward/backward symmetry test are clever physical sanity checks, not just "does it run" tests 👏

Two small things:
- Comments like "they diverged by ~4˚ when the radius scaling of w flung them…" (and "was 0.0 … before the fix" below) describe the history of this PR. That's useful in a PR comment, but in the long run it's confusing in the code. I'd just describe what's being tested.
- Test time: "no boundary pile-up" runs another T32 `PrimitiveWetModel` for a day and checks essentially what `any(0 .< σ_final .< 1)` in the "ParticleAdvection3D" testset above already checks. You could merge them by using `@test count(s -> 0 < s < 1, σ_final) / length(σ_final) > 0.8` up there and dropping this testset.

RingGrids/test/interpolation_3d.jl
```diff
@@ -0,0 +1,98 @@
...
+    RingGrids.update_locator!(locator, geometry, NF[0], NF[89.99])   # just south of the north pole
+
+    # exactly on the 2nd σ level so no vertical blending occurs either
+    positions = [TestParticle3D(σ_levels_full[2])]
+    Aout = zeros(NF, 1)
+    RingGrids.interpolate_3D!(Aout, A, locator, geometry, positions, σ_levels_full)
+
+    # north pole value is the average of the first ring, i.e. its (constant) latitude
+    @test Aout[1] ≈ geometry.latd[2]
+end
```
milankl-claude, line 98: Nice test coverage for the centre case, especially the pole averaging test! What's still missing is the face staggering, i.e. the path `w` actually takes. Something like this would test both `SigmaFaceBelow` and `SigmaFaceAbove` including the unstored boundary. It uses a profile that vanishes at σ = 0 and 1, so the zero boundary value is consistent:

```julia
@testset "3D interpolation: face staggering" begin
    NF = Float32
    grid = FullGaussianGrid(8)
    nlayers = 4
    σ_half = NF[0, 0.2, 0.5, 0.7, 1]
    f(σ) = σ * (1 - σ)                      # zero at σ = 0, 1 like the kinematic BC

    # expected: piecewise linear interpolation of f between half levels
    function expected(σ)
        k = findlast(≤(σ), σ_half[1:(end - 1)])
        α = (σ - σ_half[k]) / (σ_half[k + 1] - σ_half[k])
        return f(σ_half[k]) + (f(σ_half[k + 1]) - f(σ_half[k])) * α
    end

    geometry = RingGrids.GridGeometry(zeros(NF, grid, nlayers))
    locator = RingGrids.AnvilLocator(NF, 6, nlayers)
    RingGrids.update_locator!(locator, geometry, NF[0, 60, 120, 180, 240, 300], NF[80, 45, 10, -10, -45, -80])
    σs = NF[0, 0.1, 0.2, 0.35, 0.95, 1]
    positions = [TestParticle3D(σ) for σ in σs]

    # FaceBelow: layer k stores σ_half[k+1], top (σ=0) not stored
    A = zeros(NF, grid, nlayers)
    for k in 1:nlayers
        A[:, k] .= f(σ_half[k + 1])
    end
    Aout = zeros(NF, 6)
    RingGrids.interpolate_3D!(Aout, A, locator, geometry, positions, SigmaFaceBelow(σ_half, zero(NF)))
    @test Aout ≈ expected.(σs)

    # FaceAbove: layer k stores σ_half[k], bottom (σ=1) not stored
    for k in 1:nlayers
        A[:, k] .= f(σ_half[k])
    end
    RingGrids.interpolate_3D!(Aout, A, locator, geometry, positions, SigmaFaceAbove(σ_half, zero(NF)))
    @test Aout ≈ expected.(σs)
end
```

docs/src/particles.md
```diff
@@ -102,6 +105,82 @@ However, for typical time steps of 1 hour and velocities not exceeding 100 m/s
...
+This is a deliberate trade-off that assumes there are many fewer vertical levels than horizontal
+grid points, typically true (e.g. 8 vertical levels against many thousands of horizontal grid
+points). Some consequences:
+
+- Increasing the vertical resolution (`nlayers`) adds essentially no cost or storage to particle
+  advection beyond the ``\sigma`` level vector the model already needs, plus two small
+  per-level pole-average buffers (see [`RingGrids.AnvilLocator`](@ref)).
+- Because the same horizontal indices and weights are reused for both bracketing ``\sigma``
+  levels, horizontal and vertical motion are effectively decoupled: a particle moving only
+  vertically never triggers a new horizontal search.
```
milankl-claude, line 160: Not quite what the code does: `particle_advection!` always calls `update_locator!` after the predictor and the corrector step, whether or not the particle moved horizontally. What's true is that the locator doesn't *depend* on σ, so the same horizontal indices/weights serve every level and every staggering. Maybe:

```suggestion
  levels, horizontal and vertical location are decoupled: the (horizontal) locator does not depend
  on ``\sigma`` and is reused for all levels and for fields on full and half levels alike.
```

(note the adjusted first line, so check the wrapping)

docs/src/particles.md
```diff
@@ -102,6 +105,82 @@ However, for typical time steps of 1 hour and velocities not exceeding 100 m/s
...
+particle_advection = ParticleAdvection3D(spectral_grid, nparticles = 3)
+model = PrimitiveDryModel(spectral_grid; particle_advection)
+simulation = initialize!(model)
+run!(simulation, period = Day(2))
+simulation.variables.prognostic.particles
+```
+
+Every particle's `σ` field is now updated over time, whereas with `ParticleAdvection2D` it would
+have remained fixed at the `layer`-th value of `model.geometry.σ_levels_full`.
+
```
milankl-claude, line 183: As @milankl suggested, a σ-over-time plot here would be great. It's a visual check that the trajectories are smooth, and since the docs are rebuilt on every PR we'd notice if something breaks. `ParticleTracker` already writes `sigma`, so it can reuse the pattern from "Tracking particles" below, e.g.

````markdown
```@example particle3d
particle_tracker = ParticleTracker(spectral_grid, schedule = Schedule(every = Hour(1)))
particle_advection = ParticleAdvection3D(spectral_grid, nparticles = 20)
model = PrimitiveWetModel(spectral_grid; particle_advection)
add!(model.callbacks, particle_tracker)
run!(simulation, period = Day(10))

using NCDatasets, CairoMakie
ds = NCDataset(joinpath(model.output.run_folder, particle_tracker.filename))
σ = ds["sigma"][:, :]
hours = 0:(size(σ, 2) - 1)

fig = Figure()
ax = Axis(fig[1, 1], xlabel = "time [hours]", ylabel = "σ", yreversed = true)
[lines!(ax, hours, σ[i, :]) for i in axes(σ, 1)]
save("particles_sigma.png", fig) # hide
nothing # hide
```
![Particle trajectories in σ](particles_sigma.png)
````

`yreversed = true` so that the surface (σ = 1) is at the bottom.

File: SpeedyWeather/src/parameterizations/albedo.jl [hunk 55]
```diff
@@ -167,6 +167,7 @@ function initialize!(albedo::AlbedoClimatology, model::PrimitiveEquation)
         version = albedo.version
     )
 
+    a = on_architecture(model.architecture, a)
```
milankl-claude, line 170: Good fix, and consistent with how `ocean.jl`, `soil_temperature.jl`, `vegetation.jl` etc. move data at the call site (as discussed above) 👍 Since it's a bugfix independent of particle advection, it may deserve its own line in the CHANGELOG.

RingGrids/src/interpolation_3d.jl
milankl: yeah, would be good to retain the possibility to run the 1-layer primitive models!

PR #1217, opened 2026-08-21
Metal support for SpeedyTransforms 

gregordecristoforo:
* First implementation for FFTs using metal.jl. Create a SpeedyTransformsMetalExt.jl file to adapt existing functions and methods to MtlArrays. Add two strategies for FFTs, unmerged hemispheres and merged hemispheres.

* Minor edits to very verbose comments

* Add first implementation for FFTs using metal.jl

* Apply runic formatting

---------


milankl:
Could we distinguish here between `SpectralTransform` (FFT + Legendre transform) vs `MatrixSpectralTransform`? Are you only considering to do the former? While generally I'd be curious about the former also on Metal, it's much easier code for the latter. And given we won't likely do any high resolution stuff on a macbook anytime soon the `MatrixSpectralTransform` might be just fine and do the job.

Gabrielgerez:
> Could we distinguish here between `SpectralTransform` (FFT + Legendre transform) vs `MatrixSpectralTransform`? Are you only considering to do the former? While generally I'd be curious about the former also on Metal, it's much easier code for the latter. And given we won't likely do any high resolution stuff on a macbook anytime soon the `MatrixSpectralTransform` might be just fine and do the job.

I have only been considering the former, but have been recently testing the `MatrixSpectralTransform` as well, which at first sight does seem promising.

Though what I have now should be better than cpu T80-T128 range.

Ill do a bit of clean up and push it up for now.

maximilian-gelbrecht:
Before going into the code review into more detail. 

This is using Metal's equivalent of CUDA/HIP Graphs, right? It does seem to me like that. @gregordecristoforo just set up a common interface for the Nvidia and AMD ones, could the Metal code also use that or are there some fundamental differences between them?

Gabrielgerez:
> > Could we distinguish here between `SpectralTransform` (FFT + Legendre transform) vs `MatrixSpectralTransform`? Are you only considering to do the former? While generally I'd be curious about the former also on Metal, it's much easier code for the latter. And given we won't likely do any high resolution stuff on a macbook anytime soon the `MatrixSpectralTransform` might be just fine and do the job.
> 
> I have only been considering the former, but have been recently testing the `MatrixSpectralTransform` as well, which at first sight does seem promising.
> 
> Though what I have now should be better than cpu T80-T128 range.
> 
> Ill do a bit of clean up and push it up for now.

following up on this, I had a quick benchmarking session for this which gave me these numbers
| trunc | CPU (FFT) | Metal (FFT) | Metal (Matrix) | Matrix vs CPU | Matrix vs Metal-FFT |
|---|---|---|---|---|---|
| 8   | 0.065 ms  | 1.862 ms  | 0.234 ms   | 0.28x | 7.96x |
| 16  | 0.213 ms  | 2.771 ms  | 0.322 ms   | 0.66x | 8.62x |
| 24  | 0.717 ms  | 5.034 ms  | 0.717 ms   | 1.00x | 7.02x |
| 32  | 0.957 ms  | 5.703 ms  | 1.753 ms   | 0.55x | 3.25x |
| 42  | 4.144 ms  | 6.787 ms  | 5.141 ms   | 0.81x | 1.32x |
| 48  | 2.417 ms  | 8.267 ms  | 8.112 ms   | 0.30x | 1.02x |
| 63  | 4.873 ms  | 9.358 ms  | 23.158 ms  | 0.21x | 0.40x |
| 64  | 4.837 ms  | 9.789 ms  | 22.908 ms  | 0.21x | 0.43x |
| 80  | 9.032 ms  | 11.412 ms | 57.166 ms  | 0.16x | 0.20x |
| 96  | 14.230 ms | 15.251 ms | 117.493 ms | 0.12x | 0.13x |
| 112 | 24.548 ms | 20.346 ms | 250.232 ms | 0.10x | 0.08x |
| 128 | 30.654 ms | 24.066 ms | 365.446 ms | 0.08x | 0.07x |

where `MatrixSpectralTransforms`, works out the box and does better on the earlier T-values, and also, should correct where the cpu crossover happens wrt Metal GPU.

Gabrielgerez:
> Before going into the code review into more detail.
> 
> This is using Metal's equivalent of CUDA/HIP Graphs, right? It does seem to me like that. @gregordecristoforo just set up a common interface for the Nvidia and AMD ones, could the Metal code also use that or are there some fundamental differences between them?

I think it might be a bit harder to do this the same way. In this PR I used `MPSGraphs` which do slightly different things than CUDA/HIP Graphs.
 From what I gather (correct me if im wrong) CUDA/HIP Graphs do a first run of the calculation with CPU to build the graphs, then skip any cpu instantiation after that by repeating the same operations. MPSGraphs always have to go through the cpu instantiation every time, so I think the way to make them faster is to try to package as much as possible in a single call, rather than do the replay that CUDA/HIP do.

maximilian-gelbrecht:
> > Before going into the code review into more detail.
> > This is using Metal's equivalent of CUDA/HIP Graphs, right? It does seem to me like that. @gregordecristoforo just set up a common interface for the Nvidia and AMD ones, could the Metal code also use that or are there some fundamental differences between them?
> 
> I think it might be a bit harder to do this the same way. In this PR I used `MPSGraphs` which do slightly different things than CUDA/HIP Graphs. From what I gather (correct me if im wrong) CUDA/HIP Graphs do a first run of the calculation with CPU to build the graphs, then skip any cpu instantiation after that by repeating the same operations. MPSGraphs always have to go through the cpu instantiation every time, so I think the way to make them faster is to try to package as much as possible in a single call, rather than do the replay that CUDA/HIP do.

Okay, good! I just wanted to make sure there's a reason for it. If we can't reuse the common graph interface easily, then let's do it differently here.

maximilian-gelbrecht:
> following up on this, I had a quick benchmarking session for this which gave me these numbers

Thanks! On which Apple processor is this? 

It's nice to see that the FFT is actually faster on Metal, this isn't necessarily a given. 

With respect to the matrix FT, I would have expected it to be a bit faster in high resolutions, but this also isn't really an issue. The matrix FT primarily exists for low resolution transforms/systems. For high resolutions it's not suitable anymore anyway as it takes to much memory.

Gabrielgerez:
In this PR i am using `Metal.jl`'s `MPSGraphs` to compute the forward/backward FFT+legndretransform.

Metal has a native FFT now, though it wasnt straightforward to get that to work.

What is done now is to pack equal-length rings on the north and south hemisphere into one MetalFFT operation, and creating a MPSGraph call for the whole sphere for each pair of rings at the same time, which reduces the CPU communication overhead, which is the biggest cost of this whole thing. 

I tested this on my mac with Apple M1 Pro, please do run this on your own machines as well and see if it works!

Gabrielgerez:
What might be able to be done in the future, which would be analogous to CUDA/HIP graphs is to drop MPSGraphs and work directly with the Metal.MTL layer, which is very far down the abstraction line with automatically generated C-object wrappers for Metal.

maximilian-gelbrecht:
We can't do CI with `Metal`, but we should still keep tests handy as well. We already have `Metal` in the GPU unit tests `SpeedyWeather/test/GPU`, can you give those a go and see if they pass? There shouldn't be a need for Metal specific tests, but they should pass the general GPU transform tests.  

I would hope that with this at least the transform tests pass, but I think for the full model there's still some things failing. We can address those in follow ups later. We can also easily move those non-transform tests that fail with Metal to be architecture specific for both AMD and Nvidia, so that a smaller set of GPU tests passes for Metal.

pierluigividale:
> > > Could we distinguish here between `SpectralTransform` (FFT + Legendre transform) vs `MatrixSpectralTransform`? Are you only considering to do the former? While generally I'd be curious about the former also on Metal, it's much easier code for the latter. And given we won't likely do any high resolution stuff on a macbook anytime soon the `MatrixSpectralTransform` might be just fine and do the job.
> > 
> > 
> > I have only been considering the former, but have been recently testing the `MatrixSpectralTransform` as well, which at first sight does seem promising.
> > Though what I have now should be better than cpu T80-T128 range.
> > Ill do a bit of clean up and push it up for now.
> 
> following up on this, I had a quick benchmarking session for this which gave me these numbers
> 
> trunc	CPU (FFT)	Metal (FFT)	Metal (Matrix)	Matrix vs CPU	Matrix vs Metal-FFT
> 8	0.065 ms	1.862 ms	0.234 ms	0.28x	7.96x
> 16	0.213 ms	2.771 ms	0.322 ms	0.66x	8.62x
> 24	0.717 ms	5.034 ms	0.717 ms	1.00x	7.02x
> 32	0.957 ms	5.703 ms	1.753 ms	0.55x	3.25x
> 42	4.144 ms	6.787 ms	5.141 ms	0.81x	1.32x
> 48	2.417 ms	8.267 ms	8.112 ms	0.30x	1.02x
> 63	4.873 ms	9.358 ms	23.158 ms	0.21x	0.40x
> 64	4.837 ms	9.789 ms	22.908 ms	0.21x	0.43x
> 80	9.032 ms	11.412 ms	57.166 ms	0.16x	0.20x
> 96	14.230 ms	15.251 ms	117.493 ms	0.12x	0.13x
> 112	24.548 ms	20.346 ms	250.232 ms	0.10x	0.08x
> 128	30.654 ms	24.066 ms	365.446 ms	0.08x	0.07x
> where `MatrixSpectralTransforms`, works out the box and does better on the earlier T-values, and also, should correct where the cpu crossover happens wrt Metal GPU.

This is really interesting. Such a strong gradient of performance across a relatively small range of truncations. I am also curious to know what Apple chip this is, and what T256 would look like, because, from what Milan had explained, GPUs should not be that good at FFTs.

milankl:
So I took your table and marked the fastest in bold

| trunc | CPU (FFT) | Metal (FFT) | Metal (Matrix) | Matrix vs CPU | Matrix vs Metal-FFT |
|---|---|---|---|---|---|
| 16  | **0.213 ms**  | 2.771 ms  | 0.322 ms   | 0.66x | 8.62x |
| 32  | **0.957 ms**  | 5.703 ms  | 1.753 ms   | 0.55x | 3.25x |
| 64  | **4.837 ms**  | 9.789 ms  | 22.908 ms  | 0.21x | 0.43x |
| 96  | **14.230 ms** | 15.251 ms | 117.493 ms | 0.12x | 0.13x |
| 128 | 30.654 ms | **24.066 ms** | 365.446 ms | 0.08x | 0.07x |

So I understand that correctly, at the moment CPU is always faster on mac, unless you reach resolutions like T128 where MetalFFTs are faster. Matrix transforms are never the fastest. Is this Float32? Can we rerun this on an M3, M4 or M5? Just curious what changed since the M1 ... at least the CPU got much faster since. Are there any performance constraints for Metal FFT regarding the size? I looked into AppleAccelerate the other day and we can't do it because they can only do FFT lengths of 2^n or 2^n*3, whereas we need more even lengths 4, 8, 12, 16, 20, ..., 4m. Just checking that there aren't specific sizes that work better/worse. FFTW has this multi-algorithm design that checks what's fastest which is then returned as plan.

milankl:
> We can't do CI with `Metal`

I think we can. Oceananigans does it https://github.com/CliMA/Oceananigans.jl/blob/70f2c63dba4f69d3e04a729a85980f68f15dcd28/.github/workflows/ci.yml#L362-L401

maximilian-gelbrecht:
> > We can't do CI with `Metal`
> 
> I think we can. Oceananigans does it https://github.com/CliMA/Oceananigans.jl/blob/70f2c63dba4f69d3e04a729a85980f68f15dcd28/.github/workflows/ci.yml#L362-L401

Huh, true, then it would be nice to set up a new workflow for metal gpu tests here

pierluigividale:
So, while I am thrilled about this, because of my personal work and because of teaching, are Apple chips all that promising compared to the mainstream? I mean, is there any cluster out there based on Apple chips?
For the workflow tests, we do have a MacBook Pro that I think sports an M5 chips. If you point me to the benchmarking suite, I can ask Narender to run it and report back.

maximilian-gelbrecht:
> So, while I am thrilled about this, because of my personal work and because of teaching, are Apple chips all that promising compared to the mainstream? I mean, is there any cluster out there based on Apple chips? For the workflow tests, we do have a MacBook Pro that I think sports an M5 chips. If you point me to the benchmarking suite, I can ask Narender to run it and report back.

Realistically the Metal version is on the one hand for small experiments you do on your own laptop and for teaching and on the other hand for ease of development of the GPU versions in general. Having a working Metal version already helps a lot as most things work identical to Nvidia/AMD and you can test a few things during development on your laptop instead of connecting to a HPC for a Nvidia card. 

Apples hardware is impressive. With the latest Macs (Studio and Mini) they also specifically target local LLM inference (so they probably massively upgraded the GPU), but Apple seems to have very little interest in data centre and HPCs. However, seemingly at least GitHub has an Apple-based cluster for CI. CI is a special case though I guess, as you'll definitely have the demand to at least offer some kind of CI infrastructure for your chips and OS.

milankl:
EDIT: Max you beat me to it! 😃 

> So, while I am thrilled about this, because of my personal work and because of teaching, are Apple chips all that promising compared to the mainstream? I mean, is there any cluster out there based on Apple chips?

Surely there is but I don't want to pay for it or anticipate using it. No this is mainly about GPU acceleration on your laptop, if you want to run a quick <10min simulation either for research, teaching or a live demo, if that's N times faster, great! Also means you can do a lot of GPU testing just on your laptop, like "does this kernel run, error, is fast, could be made faster" and the latter might already be indicative of how to optimize for Nvidia GPUs

pierluigividale:
OK, if the knowledge is transferable to other/larger systems, that is very compelling.

maximilian-gelbrecht:
I added a CI workflow now, and this shows that the Fourier based transform with Metal seems to be incorrect, at least the test fail there. I haven't looked into it in detail though.

pierluigividale:
BTW, I am still unable to run with Metal on my Mac M2. I ended up freeing the DEV versions of Speedy and its friends, and Terrarium is holding back Oceananingans, which then holds back CUDA, Metal etc., so I think that I am stuck until Terrarium is updated. 
This is my status:
Status `~/.julia/environments/v1.12/Project.toml`
  [79e6a3ab] Adapt v4.7.0
  [6e4b80f9] BenchmarkTools v1.8.0
  [336ed68f] CSV v0.10.17
⌃ [052768ef] CUDA v6.2.2
  [13f3f980] CairoMakie v0.15.13
  [35d6a980] ColorSchemes v3.31.0
  [8f4d0f93] Conda v1.10.3
  [96eb917e] ContinuousWavelets v1.2.2
  [a93c6f00] DataFrames v1.8.2
  [7a1cc6ca] FFTW v1.10.0
  [e9467ef8] GLMakie v0.13.13
  [db073c08] GeoMakie v0.7.17
  [7073ff75] IJulia v1.34.4
  [f7bf1975] Impute v0.7.0
  [a98d9a8b] Interpolations v0.16.3
  [0979c8fe] JACC v1.3.1
  [63c18a36] KernelAbstractions v0.9.42
  [ed20b33f] LowerTriangularArrays v0.1.6 `https://github.com/SpeedyWeather/SpeedyWeather.jl:LowerTriangularArrays#main`
⌃ [dde4c033] Metal v1.10.0
  [85f8d34a] NCDatasets v0.14.15
  [904d977b] NumericalEarth v0.6.2
⌅ [9e8cae18] Oceananigans v0.110.20
  [91a5bcdd] Plots v1.41.7
  [92933f4c] ProgressMeter v1.11.0
  [d330b81b] PyPlot v2.11.6
  [a3a2b9e3] Rasters v0.15.0
  [3c362404] Reactant v0.2.284
  [d1845624] RingGrids v0.1.9
  [ed023a3f] SpeedyTransforms v0.2.1
  [9e226e20] SpeedyWeather v0.22.1
  [34489162] SpeedyWeatherInternals v0.2.3 `https://github.com/SpeedyWeather/SpeedyWeather.jl:SpeedyWeatherInternals#main`
  [80418d68] Terrarium v0.1.6
  [29a6e085] Wavelets v0.10.2
  [c0090381] Qt6Base_jll v6.10.2+2
  [ade2ca70] Dates v1.11.0

maximilian-gelbrecht:
Thanks for pointing this out, I just bumped up the compat in Terrarium to allow the new Oceananigans release from a few days ago. This should be merged later today. 

That being said, this PR here isn't yet delivering full model compatibility with Metal. Once this is done here this should be quite simple (and possibly doable easily with an LLM) but it will be in a follow up.

pierluigividale:
Cheers!

gregordecristoforo:
@Gabrielgerez is out of office this week so I looked into the failing GPU test suite. I don't have a Mac myself, so I am relying solely on the CI pipeline which is quite slow. I see a reoccurring pattern here:

Metal has no double-precision support, so any Float64 reaching a GPU kernel fails to compile. This shows in at least these ways:

- scalar args passed to launch! without converting to the field's NF (e.g. time_scale, eigenvalue_norm, land_sea_mask.quantization::Float64)
- Int / Int inside a kernel → Float64
- x^y with a real exponent — Base's Float32^Float32 still routes through Float64 pow_body; needs @fastmath
- bare Float64 literals in kernel/broadcast bodies
- a Float64 field on a params/constants struct passed into a kernel whole

The first category is the dominant pattern (~82 launch! sites). Rather than fixing them one CI cycle at a time, one could make launch! (or a wrapper used in NF-typed contexts) convert/assert its scalar args to NF. 

Since this would affect probably 20+ files, I want to ask first whether this is heading in the right direction. Happy to revert my recent commits that fix such issues one by one if there is a better way forward.

milankl:
>     * scalar args passed to launch! without converting to the field's NF (e.g. time_scale, eigenvalue_norm, land_sea_mask.quantization::Float64)
> 
>     * Int / Int inside a kernel → Float64
> 
>     * x^y with a real exponent — Base's Float32^Float32 still routes through Float64 pow_body; needs @fastmath
> 
>     * bare Float64 literals in kernel/broadcast bodies
> 
>     * a Float64 field on a params/constants struct passed into a kernel whole

Oh yeah this is great, we definitely want to avoid this and so instead of adapting `launch!` I'd just in the lines before the kernel launch make sure that scalars are all Float32 with slight rewrites to bypass a promotion to Float64. E.g. `convert(NF, ::Int) / ::Int` etc. None of these should be precision-critical so I'm happy for a lot of fixes that make sure all scalars are NF too! 

I also don't like adapting `launch!` because I see this as a good test-bed that we are indeed number format stable, clearly we aren't perfect here. The idea is definitely to have, by default, when `NF=Float32` everything in Float32 on the GPU, if we want to do mixed precision, e.g. `SpectralGrid(NF=Float32); planet = Earth(Float64)` etc but only then

gregordecristoforo:
The metal GPU tests are finally passing.  Maybe now it is ready for a review @milankl?

milankl:
Metal CI finishes in 12min that's pretty good. But with this PR we should say in the Readme + docs that we support Metal. You mind updating that before merging?

milankl:
I can run the Barotropic, ShallowWater and PrimitiveDry models no problem with this branch but for the PrimitiveWet model I get a segfault on `Simulation(model)` ... Given this PR is technically about Metal support for the transforms, that's okay and can still be merged but given there's been some big fixes for Float32 only in kernels the scope is already wider anyway. Maybe worth figuring out what causes the segfault?

gregordecristoforo:
> ... but for the PrimitiveWet model I get a segfault on Simulation(model) ... 

I added a test to reproduce this error in the Metal GPU CI pipeline (I don't have a mac to test things locally), the the tests pass: https://github.com/SpeedyWeather/SpeedyWeather.jl/actions/runs/35218274029/job/105191866450?pr=1217

@milankl do you have more details on the segfault you observed?

milankl:
Maybe it's a v1.13 issue with Metal, will investigate, if the Metal GPU CI here passes on v1.12 to create and run all our default models then that's fine, and v1.13 issues shouldn't hold this back in my opinion. Julia v1.13 just came out and it might be there's still some issues upstream in Metal.jl even Base/Core that haven't been ironed out yet. Julia's new minor releases are by experience way more reliable than Python's but jumping on the new version within the first weeks may still be problematic

gregordecristoforo:
> Metal CI finishes in 12min that's pretty good. But with this PR we should say in the Readme + docs that we support Metal. You mind updating that before merging?

Thanks for reminding! Metal support is now mentioned in the docs and the Readme.

This PR was never formally approved, is there anything missing? 

> [!NOTE]  
> GPU graph support is still disabled as discussed by @Gabrielgerez and @maximilian-gelbrecht . I tried to look into it but having to rely on the Github CI pipeline for debugging is really cumbersome. I would like to hand over this task (could be a separate PR, the docs state that graphs are disabled for metal) to someone with access to a mac if this is possible.

SpeedyTransforms/ext/SpeedyTransformsMetalExt.jl
```diff
@@ -0,0 +1,277 @@
...
+using SpeedyTransforms.RingGrids
+using SpeedyTransforms.LowerTriangularArrays
+
+import SpeedyTransforms: SpectralTransform, _fourier_batched!
+import SpeedyTransforms.RingGrids: AbstractField
+import SpeedyWeatherInternals.Architectures: architecture
+
+# Dissable gpu_graphs for debugging with Metal GPU CI pipeline
+import SpeedyTransforms: SpectralTransform, _fourier_batched!, default_gpu_graphs
+default_gpu_graphs(::Metal.MetalBackend) = false
```
milankl, line 18: I see this is where the graphs are still disabled, happy to leave this for later. It works now, it's just slightly slower on GPU than on CPU but that's fine.

milankl (approved):
Happy to approve this, great work, looks good to me! Before merging can you ask Claude to write a summary like the other plans in `docs/dev/` so we can pick this up more easily in future PRs?

gregordecristoforo:
>Happy to approve this, great work, looks good to me! Before merging can you ask Claude to write a summary like the other plans in docs/dev/ so we can pick this up more easily in future PRs?

Done in [9bcfb95](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1217/commits/9bcfb95ec3d626432612629381e9cbdc96458aa8).

gregordecristoforo:
All the tests are passing but the docs build is disabled. Shall we run them before merging?

milankl:
Yes please! Given this changes the docs better test 😉

gregordecristoforo:
The docs build: top entry at https://github.com/SpeedyWeather/SpeedyWeather.jl/actions
(I couldn't figure out how to run them in this PR). Can I merge?

milankl:
Anytime!

pierluigividale:
Right, I was looking into this, to see whether I could also test the PrimitiveDry model, but hit the snag of Julia updating to 1.13. Feared this would break everything, but I can run simple experiments on the CPU, no luck with Metal. But maybe I need to wait for a final merge?

milankl:
So running the whole model works for me no problem, I just had a segfault with v1.13 see #1259 which I'm not sure is our fault if it's not a problem in v1.12. The Metal models aren't necessarily faster at the moment but there's currently potential we're leaving on the road so that'll be follow-up I guess!

milankl:
> The docs build: top entry at https://github.com/SpeedyWeather/SpeedyWeather.jl/actions (I couldn't figure out how to run them in this PR). Can I merge?

I just realised that you removed the `skip-docs` label in which case you need to retrigger CI to get the docs building. I think it's only the enforce changelog that kicks in immediately if you remove the label. But not a problem, it then did the docs building on main: https://github.com/SpeedyWeather/SpeedyWeather.jl/actions/runs/35352538196/job/105623870738 which succeeded and were deployed

PR #1218, opened 2026-08-21
Implemention of  HIP graphs

gregordecristoforo:
Second attempt of implementing HIP graphs. Commit 8289870cbc9b4174daa85b26f00031c3053eae75 seems to have fixed the segfault errors that prevented the first attempt from working properly.


gregordecristoforo:
The GPU test suite passed on LUMI, but let's wait for the buildkite runner to pass before reviewing this PR

gregordecristoforo:
I benchmarked this feature on LUMI's MI250x GPUs with ROCm 6.3 with the following code:

```Julia
using SpeedyWeather, AMDGPU, BenchmarkTools

spectral_grid = SpectralGrid(truncation=42, nlayers=8, architecture=GPU())

transform!(field, L, S)
transform!(L, field, S)
@benchmark transform!(L, field, S) 
```

Without the HIP graph implementation I get the following results on the `main` branch:
```Julia
julia> @benchmark transform!(L, field, S)
BenchmarkTools.Trial: 1150 samples with 1 evaluation per sample.
 Range (min … max):  2.904 ms …   1.022 s  ┊ GC (min … max): 0.00% … 7.94%
 Time  (median):     3.387 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   4.346 ms ± 30.040 ms  ┊ GC (mean ± σ):  1.62% ± 0.23%

                      ▃▄▆█▇▅▂▃                                
  ▂▃▂▃▃▁▂▁▁▁▁▁▁▁▂▁▁▂▃▅████████▇▇▇▇▅▄▄▄▅▅▃▅▄▅▃▃▃▃▃▄▆▅▅▅▄▄▃▄▄▃ ▄
  2.9 ms         Histogram: frequency by time        3.93 ms <

 Memory estimate: 530.00 KiB, allocs estimate: 12999.
```

With HIP graphs in `gd/hip-graphs` I get:
```Julia
julia> @benchmark transform!(L, field, S)
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):   76.529 μs …  1.340 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     438.239 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   431.642 μs ± 53.546 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▂                                                       ▂▇█▄ ▂
  ██▆▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁████ █
  76.5 μs       Histogram: log(frequency) by time       450 μs <

 Memory estimate: 5.80 KiB, allocs estimate: 93.

```

gregordecristoforo:
The tests pass locally on LUMI. I think this PR is worth reviewing now even though the buildkite runner hasn't finished. It looks like the AMDGPU queue is very full in the last days (I am definitely partly to blame for that).

SpeedyTransforms/ext/SpeedyTransformsAMDGPUExt.jl
```diff
@@ -70,37 +63,77 @@ const GPU_GRAPHS_WARNED = Ref(false)
...
     if !GPU_GRAPHS_WARNED[]
-        @warn "GPU-graphs acceleration of the Fourier transform (`gpu_graphs = true`) is not implemented for AMDGPU/HIP; falling back to the unaccelerated path. Set `gpu_graphs = false` to silence this warning."
+        @warn "GPU-graphs acceleration of the Fourier transform (`gpu_graphs = true`) requires a newer AMDGPU.jl exposing the high-level HIP graph API (`AMDGPU.HIP.capture`/`instantiate`/`HIPGraphExec`); falling back to the unaccelerated path. Set `gpu_graphs = false` to silence this warning."
         GPU_GRAPHS_WARNED[] = true
     end
     return nothing
 end
 
+# =====================================================================================
+# Method overrides: dispatch on ROCArray scratch (more specific than the generic
```
maximilian-gelbrecht: Just a detail but it's not method overrides, it extends methods.

maximilian-gelbrecht (approved):
This looks good to me! Thanks a lot! 

So what's now the difference to before? Fixing the other AMD issues also enabled the HIP graphs? 

I hope it passes in the CI as well. If it doesn't but works on LUMI, we should still keep it. Ideally though we should find the reason why.

SpeedyTransforms/ext/SpeedyTransformsAMDGPUExt.jl
```diff
@@ -70,37 +63,77 @@ const GPU_GRAPHS_WARNED = Ref(false)
...
+            g_south::AbstractArray{<:Complex, 3}, S::SpectralTransform; add,
+        )
+    end
+    if !_HIP_GRAPHS_AVAILABLE
+        warn_gpu_graphs_unavailable()
+        return Base.@invoke _fourier_batched!(
+            field::AbstractField, g_north::AbstractArray{<:Complex, 3},
+            g_south::AbstractArray{<:Complex, 3}, S::SpectralTransform; add,
+        )
+    end
```
milankl: isn't that a code redundancy that could be just written as

```julia
if !S.gpu_graphs || !_HIP_GRAPHS_AVAILABLE
    !_HIP_GRAPHS_AVAILABLE && warn_gpu_graphs_unavailable()
    return ...
end
```

SpeedyTransforms/ext/SpeedyTransformsAMDGPUExt.jl
```diff
@@ -70,37 +63,77 @@ const GPU_GRAPHS_WARNED = Ref(false)
...
+            field::AbstractField, S::SpectralTransform,
+        )
+    end
+    if !_HIP_GRAPHS_AVAILABLE
+        warn_gpu_graphs_unavailable()
+        return Base.@invoke _fourier_batched!(
+            f_north::AbstractArray{<:Complex, 3}, f_south::AbstractArray{<:Complex, 3},
+            field::AbstractField, S::SpectralTransform,
+        )
+    end
```
milankl: also code redundancy, see below?

SpeedyWeather/test/GPU/spectral_transform.jl
```diff
@@ -365,48 +365,52 @@ end
...
 
-            ext.clear_fourier_graph_cache!()
+                ext.clear_fourier_graph_cache!()
+            end
         end
     end
 end
 
+test_fourier_batched_gpu_graphs_equivalence(Base.get_extension(SpeedyWeather.SpeedyTransforms, :SpeedyTransformsCUDAExt), "CUDA")
+test_fourier_batched_gpu_graphs_equivalence(Base.get_extension(SpeedyWeather.SpeedyTransforms, :SpeedyTransformsAMDGPUExt), "HIP")
```
milankl: because of the `ext !== nothing` I'm wondering about letting the `test_fourier_` return `true` when tests executed, and `false` when skipped, then

```julia
cuda_executed = test_...
amd_executed = test_...
@test cuda_executed + amd_executed > 0
```

otherwise these tests might be both skipped for some reason, not throwing an error at all

milankl: And I guess technically it would be `cuda_executed + amd_executed == 1` as only one should ever be executed on a given machine

milankl (commented):
Looks great to me!! Going from 3ms down to 0.08ms, so ~40x faster is awesome!!! Can you try this quickly at different resolutions just to get an idea?

- T32L8 (`truncation=32, nlayers=8`) low resolution
- T256L24 (`truncation=256, nlayers=24`) high-ish resolution

just so we know it's faster at low and higher resolution too

maximilian-gelbrecht:
We could also add (in a separate PR) an AMD option to the benchmarks. Milan and I won't be able to run it regularly, but having the possibility would be nice. Should be very simple to add.

milankl:
Could you then also run the GPU benchmarks https://github.com/SpeedyWeather/SpeedyWeather.jl/tree/main/SpeedyWeather/benchmark#architecture-gpu-nvidia ? The information then gets moved into the documentation here https://speedyweather.github.io/SpeedyWeatherDocumentation/dev/benchmarks so we can track benchmark progression over time

EDIT: 
> We could also add

You were faster! 🤣

gregordecristoforo:
Thanks for the feedback! Yes, I was planning to look at the benchmark suite next week (assuming the buildkite runs don't require debugging for multiple days again). I'll incorporate your comments over the weekend or on Monday

gregordecristoforo:
> So what's now the difference to before? Fixing the other AMD issues also enabled the HIP graphs?

At least for LUMI's environment yes!

maximilian-gelbrecht:
In the `SpeedyWeather/benchmark` folder just ignore the `CUDA` folder, it's outdated. we would be only interested in the main benchmarks. Those would just need a very minor modification in the beginning of `manual_benchmarking.jl` to allow for different GPU backends.

SpeedyTransforms/ext/SpeedyTransformsAMDGPUExt.jl
gregordecristoforo: Of course!

SpeedyWeather/test/GPU/spectral_transform.jl
gregordecristoforo: Addressed in [800d199](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1218/commits/800d199c207d3450afe5469cdbd961487cfee48e)

SpeedyTransforms/ext/SpeedyTransformsAMDGPUExt.jl
gregordecristoforo: Addressed in [800d199](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1218/commits/800d199c207d3450afe5469cdbd961487cfee48e)

SpeedyTransforms/ext/SpeedyTransformsAMDGPUExt.jl
gregordecristoforo: Addressed in [800d199](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1218/commits/800d199c207d3450afe5469cdbd961487cfee48e)

SpeedyTransforms/ext/SpeedyTransformsAMDGPUExt.jl
gregordecristoforo: Addressed in [800d199](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1218/commits/800d199c207d3450afe5469cdbd961487cfee48e)

gregordecristoforo:
> Looks great to me!! Going from 3ms down to 0.08ms, so ~40x faster is awesome!!! Can you try this quickly at different resolutions just to get an idea?
>    T32L8 (truncation=32, nlayers=8) low resolution
>    T256L24 (truncation=256, nlayers=24) high-ish resolution
> just so we know it's faster at low and higher resolution too

To be fair, the 0.08ms case looks like an outlier, so the real speedup is closer to 8x. Here are the other resolutions:

## T32L8

### HIP graphs
```Julia
julia> @benchmark transform!(L, field, S) 
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):   62.562 μs … 535.954 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     351.134 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   328.624 μs ±  54.070 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▃▂                                       ▂▃▆▅▃▂▂▄▅▄▂  ▇█▅▄▂▁▁ ▂
  ██▆▄▃▁▁▃▁▁▁▁▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▁▁▁▁▁▁▇███████████▇▇███████ █
  62.6 μs       Histogram: log(frequency) by time        383 μs <

 Memory estimate: 5.80 KiB, allocs estimate: 93.

```
### NO HIP graphs 
```Julia
julia> @benchmark transform!(L, field, S) 
BenchmarkTools.Trial: 1473 samples with 1 evaluation per sample.
 Range (min … max):  2.073 ms … 730.686 ms  ┊ GC (min … max): 0.00% … 9.12%
 Time  (median):     2.424 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.385 ms ±  23.966 ms  ┊ GC (mean ± σ):  2.31% ± 0.33%

                     █▅▅▁                                      
  ▃▃▃▂▂▂▂▂▂▁▂▂▂▁▁▂▃▄██████▅▅▅▄▅▄▄▅▃▃▃▃▃▃▃▃▄▃▃▃▃▄▄▅▆▅▄▄▄▃▃▃▃▄▃ ▃
  2.07 ms         Histogram: frequency by time        2.92 ms <

 Memory estimate: 397.44 KiB, allocs estimate: 9715.


```
## T256L24

### HIP graphs
```Julia
julia> @benchmark transform!(L, field, S)
BenchmarkTools.Trial: 45 samples with 1 evaluation per sample.
 Range (min … max):   92.211 ms … 652.274 ms  ┊ GC (min … max): 0.00% … 11.49%
 Time  (median):      98.623 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   113.280 ms ±  82.504 ms  ┊ GC (mean ± σ):  1.47% ±  1.71%

  █                                                              
  ██▅▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃ ▁
  92.2 ms          Histogram: frequency by time          652 ms <

 Memory estimate: 11.19 MiB, allocs estimate: 330906.
```
### NO HIP graphs
 ```Julia
julia> @benchmark transform!(L, field, S) 

BenchmarkTools.Trial: 44 samples with 1 evaluation per sample.
 Range (min … max):   95.162 ms … 803.223 ms  ┊ GC (min … max): 0.00% … 11.35%
 Time  (median):      95.722 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   114.565 ms ± 106.355 ms  ┊ GC (mean ± σ):  1.81% ±  1.71%

  █                                                              
  █▆▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂ ▁
  95.2 ms          Histogram: frequency by time          803 ms <

 Memory estimate: 11.19 MiB, allocs estimate: 330906.
```

I will look at the benchmarks in a separate PR.

**Comment by gregordecristoforo** [comment]
The tests in the buildkite runner still fail. To me it looks like the different versions in the ROCm related dependencies are a good starting point for further testing. I'll have a closer look at this as a next step.

**Comment by maximilian-gelbrecht** [comment]
A small investigation into this seems good to understand what exactly is the difference, but in general I am not opposed to just merging this when it works on LUMI (and deactivate the tests in CI). 

Just, as long as we don't know exactly what causes these errors I would keep the graphs off by default on AMD (but activated on Nvidia).

**Comment by gregordecristoforo** [comment]
Thanks @maximilian-gelbrecht. I tested the buildkite runner with different package and julia versions which didn't seem to make a difference. However, I found some examples online reporting that HIP graphs are notoriously less reliable on AMD's consumer RDNA GPUs (gfx1100, gfx1101) whereas they work well on Instinct GPUs (gfx90a) which are used for HPC (including LUMI). Here is one example: https://github.com/nerfstudio-project/gsplat/pull/957

So it is probably a good idea to dissable HIP graphs by default. How would you suggest to implement a boolean variable to switch them on (environmental variable, or inside `spectral_grid = SpectralGrid(, architecture=GPU())`)?

Otherwise, I addressed all of your comments. I will do the benchmarks in a separate PR (with HIP graphs enabled).

**Comment by maximilian-gelbrecht** [comment]
I think I would just introduce 

```Julia 
default_gpu_graphs(arch::AbstractArchitecture) 
```
and use this in the call of `SpectralTransfrom`. While we currently don't have our own different SpeedyWeather types to dispatch to distinguish between AMD and Nvidia, I would assume that we could dispatch on the `device` inside from KernelAbstractions. So something like 

```Julia 
@inline default_gpu_graphs(arch::AbstractArchitecture) = default_gpu_graphs(arch.device)
@inline default_gpu_graphs(backend::KernelAbstractions.ROCBackend) = false 
@inline default_gpu_graphs(backend::KernelAbstractions.CPUBackend) = false # might as well do that as well, when we are already on it 
@inline default_gpu_graphs(backend) = true # fallback + Nvidia 
```

gregordecristoforo:
Thanks @maximilian-gelbrecht, by default HIP graphs are now disabled. This is also the case for the testsuite for the buildkite runner. Does the implementation match what you envisioned?

maximilian-gelbrecht:
Yes, this looks good to me. The docs are failing, but #1224 should fix that.

maximilian-gelbrecht (approved):
Once everything passes now, this could be merged from my side

gregordecristoforo:
Thanks! All tests pass, I will merge

PR #1219, opened 2026-08-22
Fix GPU scalar indexing when Field broadcasts against a bare GPU array

milankl-claude:
## Summary

Running the Held-Suarez example from the docs on GPU crashes inside `LinearDrag`:

```julia
@. Fu -= drag.drag_coefs' .* u
```

`drag.drag_coefs'` is an `Adjoint` of a plain `CuVector`; `u`/`Fu` are RingGrids `Field`s. Combining a `Field`'s `FieldGPUStyle` with a bare GPU array's `AbstractGPUArrayStyle` has no explicit `BroadcastStyle` combine rule, so it falls back to `Broadcast.Unknown()` (Base's generic `AbstractArrayStyle` rule requires matching type names, and `CuArrayStyle != FieldGPUStyle`). That silently triggers scalar indexing on the GPU array instead of erroring clearly at the broadcast call site.

- Add `BroadcastStyle` combine rules for `FieldGPUStyle` vs any `AbstractGPUArrayStyle` (mirroring Base's `DefaultArrayStyle` special case for CPU arrays)
- Add a disambiguating rule for combining two `FieldGPUStyle`s of different `N` (needed since the generic rule above and Base's own same-typename rule become ambiguous otherwise)
- Bump `RingGrids` to `0.1.9+DEV`

## Test plan

- [x] Added `SpeedyWeather/test/GPU/broadcasting.jl` covering `Field .* Adjoint(vector)`, `Field .* reshaped vector`, and `Field .* Field` (different `N`, same grid) on GPU
- [x] Verified the Held-Suarez docs example (`PrimitiveDryModel` + `HeldSuarez` forcing + `LinearDrag`) now runs to completion on GPU (`Day(20)`, truncation=32, nlayers=8)
- [x] Ran `RingGrids/test/field_types.jl` (CPU) — no regressions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fix GPU scalar indexing when a `Field` broadcasts against a bare GPU array (e.g. `LinearDrag` in the Held-Suarez example) [#1219](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1219)
 - Reformulate trigonometric functions to avoid hostcalls on AMDGPU - [#1210](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1210)
 
 ## v0.22.1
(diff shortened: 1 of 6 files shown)


maximilian-gelbrecht (approved):
@milankl-claude Long time no see! 

Looks good to me, in cases like this I also really think the committed markdown file is very handy. With that, it was very easy to follow what was wrong and what was fixed.

PR #976, opened 2026-02-03
Add entrainment profiles to Betts-Miller convection

nviebig:
## Summary
- add `AbstractEntrainment` with `LinearEntrainment` and `ConstantEntrainment` profiles
- wire entrainment into `BettsMillerConvection` and pass it to `pseudo_adiabat!`
- apply entrainment mixing in the moist ascent calculation

PR #1221, opened 2026-08-23
Overhaul Betts-Miller convection: fewer GPU branches, linear entrainment (supersedes #976), convective snow

milankl-claude:
## Summary

Overhaul of the Betts-Miller convection scheme:

- **Docs audit**: fixed 5 mismatches between `docs/src/convection.md` and the code (all doc-side — the code was already correct against Frierson 2007). Relaxation time default (2h→4h), `q_ref` formula, integration-limit sign convention, undocumented no-reevaporation clamp, and the load-bearing role of the deep-convection indicator.
- **Branchless GPU restructuring**: `pseudo_adiabat!`/`dry_adiabat!` previously marked levels above the level of zero buoyancy (LZB) with `NaN`, and `convection!` re-looped over the dynamic, per-column range `level_zero_buoyancy:nlayers` five times (wet) / three times (dry). Dynamic-bound loops per GPU thread are a real hazard here — see #1193 for a previous AMDGPU/LLVM miscompile of exactly this pattern. Now the reference profiles are prefilled with the *environmental* temperature/humidity above the LZB instead, so `temp - temp_ref == 0` there exactly, and downstream loops run unconditionally over the full column with `ifelse`-masked corrections applied inline (never written back). Verified **exactly bit-identical** to `main` via a 200-step, bitwise (`==`) comparison of all prognostic fields for both `PrimitiveWetModel` and `PrimitiveDryModel`.
- **Entrainment profiles**, ported from #976 (by @nviebig — thank you for the original implementation!) onto the restructured code. `NoEntrainment`/`LinearEntrainment`/`ConstantEntrainment`, added as an `entrainment` sub-component of `BettsMillerConvection` and `BettsMillerDryConvection`. Deviates from #976 in one respect: defaults to `NoEntrainment` rather than `LinearEntrainment(0.5)`, so this stays a no-op unless a user opts in (verified: still exactly bit-identical to `main` with the default). An `entraining` trait compiles the mixing step away entirely for the default case.
- **Convective snow**: if the lowermost layer's temperature is below a new `freezing_threshold` option (default 273.15K), convective rain now falls as snow instead (`snow::Bool`, default `true`). Mass-conserving (verified exact), new `snow_convection`/`snow_rate_convection` diagnostics and matching netCDF output types. Also fixes a latent bug found while touching this code: `reset_variables!` didn't reset `rain_rate_convection`, so a column that stopped convecting kept a stale rate from an earlier step.

Implementation plan (with full rationale, bisection notes, and revision log) at [`docs/dev/2026-08/betts-miller-convection-overhaul.md`](https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/mc/convection/docs/dev/2026-08/betts-miller-convection-overhaul.md).

## Benchmarks

Local CPU benchmarking (isolation and combined-with-other-parameterizations, no GPU available in this environment), before/after the restructuring, T31/L8, T31/L16, T63/L8 (T31/L32 and T63/L16 excluded — both numerically unstable at the default timestep even on unmodified `main`, unrelated to this PR):

| Config | Arm | main | this PR |
|---|---|---|---|
| T31 L8  | isolation | 1.087 ms | 1.018 ms |
| T31 L8  | combined  | 5.969 ms | 5.872 ms |
| T31 L16 | isolation | 1.455 ms | 1.441 ms |
| T31 L16 | combined  | 10.325 ms | 10.302 ms |
| T63 L8  | isolation | 3.612 ms | 3.402 ms |
| T63 L8  | combined  | 20.435 ms | 20.170 ms |

No regression in any arm; a small (~2-10%) improvement throughout from fewer loop passes. Zero allocations before and after.

## Known issue (flagging, not resolved)

Full `Pkg.test("SpeedyWeather")`: 116,737/116,739 pass. The 2 failures are JET "free of runtime dispatch" checks in `dynamics/dispatch.jl` for `BarotropicModel` and `PrimitiveWetModel` — both pass cleanly on `main`. I bisected this with `git worktree`s: none of the 3 changed files in the convective-snow commit (`convection.jl`, `precipitation.jl`, `tendencies.jl`) reproduces the failure individually when layered onto the entrainment commit — only the full combination does. The dispatch JET actually flags is 100% inside `KernelAbstractions`/dynamics-core internals (`__thread_run`, spectral transform, horizontal diffusion) that `BarotropicModel` exercises — a model with no parameterizations that never touches convection/entrainment/snow code at all. This looks like a global compilation-heuristic sensitivity to the package's total type/method count (similar in kind to #1193) rather than a logic defect in this PR. Happy to dig further if a maintainer has context on this class of JET fragility.

## Test plan

- [x] `SpeedyWeather/test/parameterizations/convection.jl`: 1132/1132 pass with `--check-bounds=yes` (extended with entrainment functor tests, entrainment-in-a-model tests, a deterministic entrainment-effect test, reference-profile invariants, and convective-snow mass-conservation tests)
- [x] `SpeedyWeather/test/parameterizations/all_parametrizations.jl`: 44/44 pass
- [x] `SpeedyWeather/test/output/netcdf_output.jl`: pass (exercises the new `ConvectiveSnowOutput`/`ConvectiveSnowRateOutput`)
- [x] `SpeedyWeather/test/parameters.jl`, `SpeedyWeather/test/type_stability_test.jl`: pass (run manually, excluded from `Pkg.test`)
- [x] Full `Pkg.test("SpeedyWeather")` with `--check-bounds=yes`: 116,737/116,739 (see Known issue above)
- [x] Bit-identity gate (200 steps, exact `==` on all prognostic fields vs. pre-restructure `main`): pass after restructuring, pass after entrainment (default off), intentional divergence after convective snow (verified confined to the wet model via land coupling; dry model stays exact)


docs/dev/2026-08/betts-miller-convection-overhaul.md
```diff
@@ -0,0 +1,443 @@
...
+> Switch to main, start a new branch mc/convection in that I want you to overhaul the Betts
+> Miller convection scheme we have. Check that the code matches the equations in the docs, and
+> make changes towards GPU performance, which you can benchmark locally but only benchmark
+> convection in isolation and together with other parameterizations don't bother doing a full
+> model benchmark. At the moment the convection scheme writes NaNs above the level of zero
+> buoyancy but maybe it's more efficient to just write the environmental profile here and avoid
+> the branching of doing the relaxation only up to a specific layer. But have a look at what else
+> can be improved. Then have a look at PR 976 and include the entrainment profile also into this
+> PR to do both together. And make a change to convection such that if the lowermost layer
+> temperature is below freezing then turn the convective rain into snow. Plan with Opus, execute
```
maximilian-gelbrecht, line 28: Unrelated to the PR itself, but does this work? Can it switch on its own?

docs/src/convection.md
```diff
@@ -107,13 +149,13 @@ the resulting tendencies from this scheme. Those will be calculated in [Correcte
...
-Vertically integration from surface ``p_0`` to level of zero buoyancy in
-pressure coordinates ``p_{LZB}`` yields
+Vertically integration from the level of zero buoyancy ``p_{LZB}`` to the surface ``p_0`` (so that
+``dp > 0``, as ``p_{LZB} < p_0``) yields
 
 ```math
 \begin{aligned}
-P_q &= - \int_{p_0}^{p_{LZB}} \delta q \frac{dp}{g} \\
-P_T &= \int_{p_0}^{p_{LZB}} \frac{c_p}{L_v} \delta T \frac{dp}{g}
+P_q &= - \int_{p_{LZB}}^{p_0} \delta q \frac{dp}{g} \\
```
maximilian-gelbrecht, line 161: wouldn't the sign need to be flipped if the integration direction is changed or was this a mistake?

SpeedyWeather/src/parameterizations/convection.jl
```diff
@@ -146,13 +149,26 @@ and relaxes current vertical profiles to the adjusted references."""
     rain_convection *= pₛΔt_gρ                  # convert to [m] of rain during Δt
     rain_convection = max(rain_convection, 0)   # ensure non-negative precipitation, rounding errors
 
+    # CONVECTIVE SNOW: below freezing, all convective rain falls as snow instead. Convective
+    # precipitation is deposited immediately (not fluxed through layers like large-scale
+    # precipitation), so a single check on the lowermost layer's temperature is sufficient,
+    # unlike the falling-flux melt cascade in large-scale condensation
+    snow_convection::NF = 0
+    let_it_snow = convection.snow
+    rain_convection, snow_convection = let_it_snow && temp[ij, nlayers] < convection.freezing_threshold ?
```
maximilian-gelbrecht: Turn this into an `ifelse` as it's runtime dependent (for Reactant)

SpeedyWeather/src/parameterizations/convection.jl
```diff
@@ -192,18 +217,23 @@ set to NaN instead and should be skipped in the relaxation."""
     Lᵥ = atmosphere.latent_heat_condensation    # latent heat of vaporization
     cₚ = atmosphere.heat_capacity               # heat capacity
 
-    temp_parcel::NF = temp_environment[ij, nlayers]     # start at surface with environment temperature [K]
-    humid_parcel::NF = humid_environment[ij, nlayers]   # and humidity [kg/kg]
-
+    # prefill both reference profiles with the environment (branchless stand-in for the
+    # previous NaN reset); levels above the LZB are never touched again below
```
maximilian-gelbrecht: delete claudey comments that reference previous implementations.

SpeedyWeather/src/parameterizations/convection.jl
```diff
@@ -192,18 +217,23 @@ set to NaN instead and should be skipped in the relaxation."""
...
-    local saturated::Bool = false           # did the parcel reach saturation yet?
-    local buoyant::Bool = true              # is the parcel still buoyant?
-    local k::Int = nlayers                  # layer index top to surface
-    local temp_virt_parcel::NF = virtual_temperature(temp_parcel, humid_parcel, atmosphere)
+    saturated::Bool = false           # did the parcel reach saturation yet?
+    buoyant::Bool = true              # is the parcel still buoyant?
+    k::Int = nlayers                  # layer index top to surface
+    temp_virt_parcel::NF = virtual_temperature(temp_parcel, humid_parcel, atmosphere)
 
     while buoyant && k > 1                  # calculate moist adiabat while buoyant till top
```
maximilian-gelbrecht, line 240: This whole loop is definitely far from branchless. But maybe that's not avoidable?

SpeedyWeather/src/parameterizations/tendencies.jl
```diff
@@ -109,6 +109,10 @@ function reset_variables!(vars::Variables)
     reset_variable!(vars.parameterizations, :cloud_top, nlayers + 1)   # reset to below top layer
     reset_variable!(vars.parameterizations, :rain_rate, 0)
     reset_variable!(vars.parameterizations, :snow_rate, 0)
+    # convection! only assigns these on the non-early-return (convecting) path, so without a
+    # reset a column that stops convecting would otherwise keep a stale rate from an earlier step
```
maximilian-gelbrecht: remove claudey comments

docs/dev/2026-08/betts-miller-convection-overhaul.md
milankl: Yes, Claude has access to its own fork on our servers and controls it via git but also installed `gh` the github command line interface which makes it much easier to read comments, notifications etc from GitHub that aren't part of the `git` functionality

docs/src/convection.md
milankl-claude: Not a mistake — both the bounds order *and* the sign had to change together, and just flipping the bounds without flipping the sign would indeed be wrong. Traced it through the code to double check: `Pq += (humid[k] - humid_ref[k]) * Δσ[k]` summed for `k = 1:nlayers` (nonzero only for `k >= level_zero_buoyancy`, since above the LZB `humid == humid_ref` exactly). With Frierson's eq. (1), `δq = -(humid[k]-humid_ref[k])/τ`, so `humid[k]-humid_ref[k] = -τ·δq`, and `Δσ[k] > 0` running from LZB (top) to surface (bottom) — i.e. increasing pressure. So `Pq ∝ -∫_{p_LZB}^{p_0} δq dp/g`, which is exactly the new doc formula. The old formula, `-∫_{p_0}^{p_LZB}`, is the same integral with bounds swapped and no compensating sign flip, so it evaluates to the *opposite* sign of what the code actually computes — that was the real bug. Same check holds for `P_T`. So the new formula is the one that matches the code.

SpeedyWeather/src/parameterizations/convection.jl
milankl-claude: Good catch, done in cb1df8d — now an explicit `ifelse` over `(freezing, (snow_convection, rain_convection), (rain_convection, snow_convection))` instead of the ternary.

SpeedyWeather/src/parameterizations/convection.jl
milankl-claude: Removed (and the matching one in `dry_adiabat!` further down) in cb1df8d.

SpeedyWeather/src/parameterizations/convection.jl
milankl-claude: Right, not avoidable without a different algorithm, and intentionally left alone here. This loop is *finding* the LZB — each level's parcel state depends on the level below it, so it's a genuinely sequential search, not something that can be vectorized/masked the way the rest of this PR does. It's called out explicitly as a known, accepted limitation (see the plan doc's Known limitations section, and the PR description). Important distinction from the `#1193` hazard this PR actually targets: that was about the *downstream* loops re-deriving `level_zero_buoyancy:nlayers` as a per-column dynamic **loop bound**, used 5x (wet) / 3x (dry) per column — that's now gone, replaced by full-range loops with `ifelse`-masked corrections. This ascent loop's dynamic trip count is a different kind of divergence (search termination, not a re-derived bound feeding array indexing across multiple separate loops), and flagged as future work if GPU profiling ever shows it's a hotspot.

SpeedyWeather/src/parameterizations/tendencies.jl
milankl-claude: Removed in cb1df8d.

docs/src/convection.md
maximilian-gelbrecht: @milankl So we had a major bug in the convection here?

SpeedyWeather/Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "SpeedyWeather"
 uuid = "9e226e20-d153-4fed-8a5b-493def4f21a9"
 authors = ["SpeedyWeather contributors"]
-version = "0.22.1"
+version = "0.23.0-DEV"
```
maximilian-gelbrecht: I wouldn't really consider this a breaking change

maximilian-gelbrecht (commented):
Would be good to actually profile this on GPU. 

Maybe with a GPU at hand @milankl-claude can also see if it actually can still improve the buoyancy/lzb loop.

SpeedyWeather/src/parameterizations/convection.jl
maximilian-gelbrecht: See other comment. When actually GPU profiling, maybe you can still find a way to optimise this.

SpeedyWeather/Project.toml
milankl: Fixed in [5311061](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1221/commits/53110614c77b3fe7a09b95c7c21231ade23c6eef)

docs/src/convection.md
milankl: I don't think so, Frierson 2007 write this

image

which is what's currently in the docs. In simple terms, if humidity ``q`` is above its reference ``q_{ref}`` then it should rain. The humidity tendency ``\delta q`` would then be negative and the resulting precipitation ``P_q`` should be positive. So the integral should not change the sign of ``\delta q`` but if you integrate from the bottom up with pressure decreasing the ``dp`` would indeed be negative so I agree with Claude here that in the docs the integral bounds should be swapped to integrate in the positive direction of the vertical axis (down). This is what we do in the code, which would be completely off in terms of sign between humidity tendency and resulting precipitation if there was a sign error. So I agree that Frierson's original notation is misleading/wrong here and that's what Claude stumbled upon!

SpeedyWeather/src/parameterizations/convection.jl
milankl-claude: Did the GPU profiling (5f6fb33b, new "GPU profiling (nsys)" section in the plan doc). Short version: `ncu` — the tool that would directly show whether *this* loop specifically is a divergence hotspot, via branch/warp-efficiency counters — is blocked on this machine by the driver's `RmProfilingAdminOnly` restriction; Milan's asked the sysadmins for access, still pending. `nsys` works without those counters, so I used it to compare the *whole* isolated-convection kernel's GPU execution time (this loop + the downstream full-range loops, fused together — `nsys` can't separate them) between `main` and this branch, at the same three configs as the CPU benchmark table:

| Config | main (ns) | this branch (ns) | Δ |
|---|---|---|---|
| T31 L8 | 13,438.8 | 14,104.7 | +4.96% |
| T31 L16 | 24,902.5 | 24,958.7 | +0.23% |
| T63 L8 | 13,868.6 | 14,531.5 | +4.78% |

So, unexpectedly: this branch is measurably *slower* on GPU at `nlayers=8` (~5%, both truncations) — opposite sign from the CPU table above. The gap is nearly identical in absolute terms at T31L8 vs T63L8 despite T63 having ~4× the grid points, which points to a fixed per-thread cost rather than a launch-overhead artifact — consistent with the "full-range loops do more work for shallow-LZB columns" risk the plan already flagged: `main`'s dynamic re-loop skips iterations above the LZB, this branch's full-range loops never do. At `nlayers=16` the gap vanishes (within noise).

So I can't yet tell you whether *this specific loop* is the culprit vs. the downstream loops — that needs `ncu`. But it does confirm the branchless restructuring has a real GPU-side cost at coarse vertical resolution that the CPU numbers didn't catch. Whether that's an acceptable trade for the `#1193` GPU-miscompile fix, or worth chasing further once `ncu` access lands, seems like your call.

SpeedyWeather/src/parameterizations/convection.jl
```diff
@@ -5,18 +5,28 @@ export BettsMillerConvection
...
-@parameterized @kwdef struct BettsMillerConvection{NF} <: AbstractConvection
+@parameterized @kwdef struct BettsMillerConvection{NF, Entrainment <: AbstractEntrainment} <: AbstractConvection
     "[OPTION] Relaxation time for profile adjustment"
     time_scale::Second = Hour(4)
 
     "[OPTION] Relative humidity for reference profile [1]"
     @param relative_humidity::NF = 0.7
+
+    "[OPTION] Entrainment profile mixing environmental air into the rising parcel"
+    @component entrainment::Entrainment = LinearEntrainment(DEFAULT_NF)
```
nviebig, line 16: Here, shouldnt this be entrainment::Entrainment = NoEntrainment() (to fit with docs?)

SpeedyWeather/src/parameterizations/convection.jl
```diff
@@ -5,18 +5,28 @@ export BettsMillerConvection
...
+
+    "[OPTION] Temperature of the lowermost layer below which convective rain becomes snow [K]"
+    @param freezing_threshold::NF = 273.15 (bounds = Positive,)
 end
 
 Adapt.@adapt_structure BettsMillerConvection
 
 # generator function
-BettsMillerConvection(SG::SpectralGrid; kwargs...) = BettsMillerConvection{SG.NF}(; kwargs...)
+BettsMillerConvection(SG::SpectralGrid; entrainment = LinearEntrainment(SG), kwargs...) =
```
nviebig, line 28: Here again entrainment::Entrainment = NoEntrainment()?

SpeedyWeather/src/parameterizations/entrainment.jl
```diff
@@ -0,0 +1,50 @@
...
+    "[OPTION] Entrainment rate at the surface [1]"
+    @param surface_entrainment::NF = 0.5 (bounds = 0 .. 1,)
+end
+
+Adapt.@adapt_structure LinearEntrainment
+LinearEntrainment(SG::SpectralGrid; kwargs...) = LinearEntrainment{SG.NF}(; kwargs...)
+LinearEntrainment(::Type{NF}; kwargs...) where {NF} = LinearEntrainment{NF}(; kwargs...)
+
+"""$(TYPEDSIGNATURES) Entrainment rate at sigma level σ for the linear profile."""
+@inline (E::LinearEntrainment)(σ) = E.surface_entrainment * max(zero(σ), σ - E.σ_entrainment) / (1 - E.σ_entrainment)
```
nviebig, line 38: Here, as σ_entrainment::NF = 0.5 (bounds = 0 .. 1,), nothing stops me from declaring σ_entrainment = 1, so dividing by 0 in the functor below (/ (1 - E.σ_entrainment))?

SpeedyWeather/src/parameterizations/convection.jl
milankl: Ah yeah, I just changed this, can adapt the docs too!

SpeedyWeather/src/parameterizations/convection.jl
milankl: see above

SpeedyWeather/src/parameterizations/entrainment.jl
milankl: good call, we could just set the bounds to `0 .. 0.99` I think that practically solves the problem and is good enough?

nviebig: I think so too :)

SpeedyWeather/src/parameterizations/convection.jl
milankl-claude: Docs now updated to match (both here and `entrainment.jl`'s docstrings) in [8622ab4](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1221/commits/8622ab42): `BettsMillerConvection` (wet) defaults to `LinearEntrainment`, `BettsMillerDryConvection` stays at `NoEntrainment`.

SpeedyWeather/src/parameterizations/entrainment.jl
milankl-claude: Bounds narrowed to `0 .. 0.99` in [8622ab4](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1221/commits/8622ab42). One caveat worth flagging: `bounds` here is `ModelParameters.jl`-facing metadata (docs/perturbation tooling), not a construction-time assertion, so this documents the constraint rather than actually stopping `LinearEntrainment(; σ_entrainment = 1)` from being constructed and dividing by zero at call time. Given we'd agreed 0.99 was "good enough" I kept the fix scoped to that — happy to add an actual runtime check if you'd rather have one.

SpeedyWeather/src/parameterizations/convection.jl
milankl-claude: We have `ncu` access now (2× L40), so I profiled the isolated convection kernel properly. Short version: the ascent loop wasn't the problem, but there was something to optimise next to it. Fixed in 21dcc11f.

**What ncu shows**
- **Memory latency, not divergence.** One thread per column gives only ~2 of 12 possible warps per scheduler, even at T127. So the kernel is limited by memory latency: 45% of stall samples are waiting on global loads, and T127 takes about as long as T31.
- **The ascent loop is fine.** Divergence was already 3.7× lower than on `main` thanks to the branchless restructuring.
- **The prefill loop was the real cost.** The environment prefill loop I'd added before the ascent took 21% of all stall samples on its own. It's why this branch did 24% more loads than `main` and ended up ~5% slower in the earlier nsys numbers.

**Fix:** prefill + `while` loop → one fixed-trip pass `for k in nlayers-1:-1:1`. It computes the parcel only while buoyant and writes each level exactly once, either the parcel or the environment via `ifelse`. Same change in `dry_adiabat!`. This also removes the dynamic trip count you flagged. Results are **bitwise identical** to the previous commit (100 steps, wet + dry, with and without entrainment), and convection tests pass with `--check-bounds=yes`.

GPU device time per launch, 256 threads/block:

| | `main` | before | after |
|---|---|---|---|
| wet T31 L8 | 11.68 µs | 12.64 µs | **10.73 µs** (−15%) |
| wet T127 L8 | 13.11 µs | 13.35 µs | **11.21 µs** (−16%) |
| wet T31 L16 | | 22.65 µs | **18.12 µs** (−20%) |
| wet T31 L8, LinearEntrainment | | 14.07 µs | **11.92 µs** (−15%) |
| dry T31 L8 | | 5.72 µs | **5.25 µs** (−8%) |

So it's now faster than `main` as well.

**Bug found on the way:** `BettsMillerDryConvection` couldn't be adapted to GPU. `@adapt_structure` can't infer the `NF` parameter because no field uses it, and `main` had a hand-written method. So `PrimitiveDryModel` failed on GPU on this branch. Fixed, with a new "Adapt" test.

**Other things ncu pointed at, not done here:**
- **Precompute `σ^κ` per level:** the same for every column, currently a `powf` per level and column. Gives another 2–4%, but changes rounding.
- **Smaller workgroups for `LinearWorkOrder`:** up to ~4%, but affects every kernel.
- **Constant-cache misses (~15% of stalls):** from the kernel receiving the whole `vars`/`model`. That's structural, shared by all fused column-parameterization kernels.
- **More threads per column:** the only thing that addresses the real limit, but it's a redesign.

Details are in the "GPU profiling (ncu)" section of the plan doc.

PR #1222, opened 2026-08-24
Move time-stepping mathematics from Usage to Numerics

milankl-claude:
## Summary

- Split `docs/src/time_integration.md`: the Usage page now covers only how to create a time stepper (`Leapfrog`, `NCycleLorenz`), its options, and passing it to the model constructor.
- The mathematical/numerical content (leapfrog derivation, Robert-Asselin/Williams filter derivation, Lorenz N-cycle update equations, and the step-dispatch internals `get_step`/`which_*_step`/`*_steps`) moves to a new `docs/src/time_stepping.md` under the Numerics section.
- While rewriting the Lorenz N-cycle section, cross-checked the docs against `SpeedyWeather/src/time_stepping/steppers/{leapfrog,ncycle_lorenz}.jl` and added several things that were implemented but undocumented:
  - the per-variant weight formulas for `NCycleLorenzA`/`B`/`AB`/`ABBA` (previously only named, not defined)
  - the `tendency_grid_steps`/`tendency_spectral_steps` split for `NCycleLorenz` (only `tendency_spectral_steps` was mentioned before)
  - the diffusion/implicit-correction ordering difference between `Leapfrog` (implicit then diffusion) and `NCycleLorenz` (diffusion then implicit), matching `diffusion_and_implicit!`
  - docstring references for the four `NCycleLorenz*` variant types
- All existing `@ref leapfrog`/`@ref ncycle`/`@ref steps` cross-references from other doc pages (barotropic, shallowwater, primitiveequation, variables, analysis, surface_fluxes, large_scale_condensation, how_to_run_speedy) keep resolving since the anchor ids moved along with their content rather than being renamed.

## Test plan

- [ ] `julia --project=docs docs/make.jl` builds without broken-link warnings (could not run locally due to a stale `docs/Manifest.toml`/`Zarr` resolution issue unrelated to this change)
- [x] Manually verified all `@id`/`@ref` targets touched by this change still resolve to a unique heading
- [x] Diffed moved sections against the original file to confirm no content was dropped or altered beyond the additions listed above


docs/src/lowertriangularmatrices.md
```diff
@@ -352,7 +352,7 @@ L = rand(Float32, spectrum)
 L = rand(ComplexF32, spectrum, 5)
 ```
 
-In the SpeedyWeather.jl model, the `Spectrum` is stored just once in the `SpectralGrid` type,
+In the SpeedyWeather.jl model, the `Spectrum` is stored just once in the [SpectralGrid](@ref) type,
```
milankl: Apostrophes missing?

docs/src/how_to_run_speedy.md
```diff
@@ -209,8 +209,8 @@ example, change the output interval like so
 ```@example howto
 set!(model.output, model, interval=Hour(1))
 ```
-Now, if there's output, it will be every hour. Furthermore the initial
-conditions can be set with the `initial_conditions` model component
+Now, if there's output, it will be every hour. Furthermore the [initial
+conditions](initial_conditions.md) can be set with the `initial_conditions` model component
```
milankl: With @ref?

docs/src/radiation.md
```diff
@@ -140,7 +142,7 @@ planet = Earth(spectral_grid, length_of_day=Hour(24), length_of_year=Day(365)+Ho
 the sun (controlling the seasonal cycle). Both are `Dates` periods, so
 `Hour(24)`, `Day(1)` or `Second(86400)` are equivalent, and both are completely
 independent of the model's time step and of the simulation time you pass to
-`run!`. A planet with a 10-day-long day and an Earth-length year is just
+[`run!`](@ref run). A planet with a 10-day-long day and an Earth-length year is just
```
milankl: Does a section with `run` as id exist?

docs/src/time_stepping.md
```diff
@@ -0,0 +1,272 @@
...
+
+|                    | time at step ``i-1`` | time at step ``i`` | time step at ``i+1`` |
+| ------------------ | -------------------- | ------------------ | -------------------- |
+| Initial conditions | ``t = 0``            |                    |                      |
+| 1: Euler           | (T) ``\quad t = 0``  |  ``t=\Delta t/2``  |                      |
+| 2: Leapfrog with ``\Delta t``|``t = 0``|(T) ``\quad t = \Delta t/2``| ``t = \Delta t``|
+| 3 to ``n``: Leapfrog with ``2\Delta t``|``t-\Delta t``|(T) ``\qquad \quad \quad t``| ``t+\Delta t`` |
+
+The time step that is used to evaluate the tendencies is denoted with (T).
+It is always the time step furthest in time that is available.
```
milankl, line 66: Add a note here that for practical reasons we copy the initial conditions from step 1 to step 2 before the time integration starts so that the step index for the right-hand side remains unchanged.

docs/src/time_stepping.md
```diff
@@ -0,0 +1,272 @@
...
+
+Four weight variants are available, following the naming in Hotta et al. (2016), selected via the
+`variant` option:
+
+| Variant | Weight ``w_k`` for substep ``k = 1, \dots, N-1`` (``w_0 = 1``) | Notes |
+| ------- | --------------------------------------------------------------- | ----- |
+| `NCycleLorenzA` (default) | ``w_k = N / (N - k)`` | |
+| `NCycleLorenzB` | ``w_k = N / k`` | |
+| `NCycleLorenzAB` | alternates a full cycle of `A`, then a full cycle of `B` | 2 cycles per period |
+| `NCycleLorenzABBA` | sequence `A`, `B`, `B`, `A` | only for ``N=4``, 4th-order accurate |
```
milankl: I believe we also allow ABBA with N=3

docs/src/time_stepping.md
```diff
@@ -0,0 +1,272 @@
...
+\end{aligned}
+```
+
+for every spectral coefficient (and vertical layer), with the implicit correction applied to
+``G`` beforehand by `diffusion_and_implicit!`.
+
+Four weight variants are available, following the naming in Hotta et al. (2016), selected via the
+`variant` option:
+
+| Variant | Weight ``w_k`` for substep ``k = 1, \dots, N-1`` (``w_0 = 1``) | Notes |
```
milankl: Can a column with "steps" be added, so total number steps in a period after which everything repeats? Should be N for A, but 4N for ABBA. And ideally with a sentence below explaining that the steps here can be interpreted as substeps like for Runge-Kutta methods (where the step size dt would be only completed after the N steps) or as a multi-step method with changing weights and restarts where every step is of size dt. We use the latter such that the computational cost is independent of N but inversely proportional to dt.

docs/src/variables.md
```diff
@@ -92,7 +93,7 @@ simulation.variables
 
 ## Setting variables
 
-The prognostic variables can be mutated (e.g. to set new initial conditions) with the [`SpeedyWeather.set!`](@ref) function.
+The prognostic variables can be mutated (e.g. to set new [initial conditions](initial_conditions.md)) with the [`SpeedyWeather.set!`](@ref) function.
```
milankl: Use @ref

docs/src/lowertriangularmatrices.md
milankl-claude: Yeah, inconsistent with the "SpeedyWeather.jl's default..." phrasing used elsewhere on this page. Fixed to "In SpeedyWeather.jl's model, ...".

docs/src/how_to_run_speedy.md
milankl-claude: Good catch, switched to `[initial conditions](@ref "Initial conditions")` so it resolves against the actual heading.

docs/src/radiation.md
milankl-claude: No, `run!`'s auto-generated anchor is `run!` not `run` (confirmed no other page uses an explicit id for it either). Fixed to `[`run!`](@ref)`.

docs/src/variables.md
milankl-claude: Same fix as the how_to_run_speedy.md comment: `[initial conditions](@ref "Initial conditions")`.

docs/src/time_stepping.md
milankl-claude: Added — checked `initialize!(vars::Variables, ::AbstractLeapfrog, ::AbstractModel)` in leapfrog.jl, which does exactly this (copies step 1 to step 2 via `copy_step_forward!`) so `which_prognostic_step` can always return step 2.

docs/src/time_stepping.md
milankl-claude: Confirmed — `weight_coefficient` for ABBA has no restriction to N=4, and the test suite exercises it with N=3 too (`w3` in test/dynamics/time_stepping.jl). Softened the note to "4th-order accurate for N=4; other N run but without that accuracy guarantee".

docs/src/time_stepping.md
milankl-claude: Added a "Steps per period" column (N, N, 2N, 4N) plus a paragraph below the table on the substep/multi-step distinction, using your explanation.

docs/src/lowertriangularmatrices.md
milankl-claude: Correction: you were right, it wasn't about the possessive wording. `SpectralGrid` is a type (has a docstring), and Documenter's `@ref` only resolves against docstrings when the link text is backtick-wrapped — plain `[SpectralGrid](@ref)` was instead resolving against the unrelated "## SpectralGrid" heading in how_to_run_speedy.md. Fixed to `[`SpectralGrid`](@ref)`. Kept the "SpeedyWeather.jl's" wording tweak too since it matches the phrasing used elsewhere on this page.

PR #1224, opened 2026-08-24
Update Terrarium output writer to handle new variable containers

bgroenks96:
Terrarium v0.1.6 broke the SpeedyWeather extension output writer because the  `Variables` container now uses `OrderedDict`s instead of `NamedTuple`s. This PR updates the output writer to work with the new format. It should also be backwards compatible with older Terrarium versions, but this is not tested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


maximilian-gelbrecht:
Happy to merge it and issue a release but should we put in a Terrarium 0.1.6 compat or not?

bgroenks96:
I don't think you need to change the compat.

maximilian-gelbrecht (approved):

PR #1228, opened 2026-08-26
Add GPU-safe `clamp!` for `Field`, make `fill!` return a `Field`, only clamp land-sea mask when out of range

milankl-claude:
## Problem
```julia
julia> set!(model, land_sea_mask = (λ, ϕ) -> if 0 <= λ < 180 1 else 0 end)
ERROR: Scalar indexing is disallowed.
```

## Summary

- `set!(model, land_sea_mask = ...)` threw a GPU scalar-indexing error because `AbstractField` had no specialized `clamp!` and fell back to the generic `AbstractArray` implementation, which iterates element-by-element. Add `Base.clamp!(F::AbstractField, lo, hi)` that operates on `F.data` directly (mirroring the existing `fill!` pattern), avoiding scalar indexing on GPU.
- Make `fill!` for `AbstractField` return a `Field` (rather than the bare underlying array), for consistency with the new `clamp!`.
- In `set!` for `AbstractLandSeaMask`, only call `clamp!` inside the `if` branch where the mask is actually out of `[0, 1]`, instead of unconditionally clamping every call.

Thanks to @hannahw0od for surfacing this GPU issue.

## Test plan
- [x] Added GPU-array (`JLArray`) tests for `clamp!` and `fill!` in `RingGrids/test/grids.jl` (`AbstractField: GPU (JLArrays)` testset)
- [x] Added a `set!` clamping test in `SpeedyWeather/test/parameterizations/land_sea_mask.jl`
- [x] Ran `RingGrids/test/grids.jl` and `SpeedyWeather/test/parameterizations/land_sea_mask.jl` locally, all passing
- [x] `runic --check` clean

### DIFF
--- a/RingGrids/src/field.jl
+++ b/RingGrids/src/field.jl
@@ -123,10 +123,10 @@ matrix_size(field::AbstractField) = (matrix_size(field.grid)..., size(field)[2:e
 # simply propagate all indices forward
 Base.@propagate_inbounds Base.getindex(field::AbstractField, ijk...) = getindex(field.data, ijk...)
 Base.@propagate_inbounds Base.setindex!(field::AbstractField, x, ijk...) = setindex!(field.data, x, ijk...)
-Base.fill!(field::AbstractField, x) = Field(fill!(field.data, x), field.grid, field.dims)
+Base.fill!(field::AbstractField, x) = nonparametric_type(typeof(field))(fill!(field.data, x), field.grid, field.dims)
 
 # defined explicitly as the generic AbstractArray fallback uses scalar indexing, which errors on GPU
-Base.clamp!(F::AbstractField, lo, hi) = Field(clamp!(F.data, lo, hi), F.grid, F.dims)
+Base.clamp!(field::AbstractField, lo, hi) = nonparametric_type(typeof(field))(clamp!(field.data, lo, hi), field.grid, field.dims)
 
 # make [:, k...] not escape the Field
 @inline Base.getindex(field::AbstractField, col::Colon, k...) = Field(field.data[col, k...], field.grid, field.dims[col, k...])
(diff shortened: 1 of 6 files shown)


RingGrids/src/field.jl
```diff
@@ -123,7 +123,10 @@ matrix_size(field::AbstractField) = (matrix_size(field.grid)..., size(field)[2:e
 # simply propagate all indices forward
 Base.@propagate_inbounds Base.getindex(field::AbstractField, ijk...) = getindex(field.data, ijk...)
 Base.@propagate_inbounds Base.setindex!(field::AbstractField, x, ijk...) = setindex!(field.data, x, ijk...)
-Base.fill!(field::AbstractField, x) = fill!(field.data, x)
+Base.fill!(field::AbstractField, x) = nonparametric_type(typeof(field))(fill!(field.data, x), field.grid, field.dims)
```
maximilian-gelbrecht:
```suggestion
function Base.fill!(field::AbstractField, x) 
                  fill!(field.data, x)
                  return field
end 
```

maximilian-gelbrecht: the definition of claude feels wrong to me, I think this is better

RingGrids/src/field.jl
```diff
@@ -123,7 +123,10 @@ matrix_size(field::AbstractField) = (matrix_size(field.grid)..., size(field)[2:e
...
+
+# defined explicitly as the generic AbstractArray fallback uses scalar indexing, which errors on GPU
+Base.clamp!(field::AbstractField, lo, hi) = nonparametric_type(typeof(field))(clamp!(field.data, lo, hi), field.grid, field.dims)
```
maximilian-gelbrecht: here as well

milankl: Agree, otherwise also a new field has to be constructed, which means it has to run throw checking data and grid dimensions etc again

PR #1229, opened 2026-08-26
New logo in README and docs

milankl:
@lazarusA is there a problem to use a .gif instead of .png as vite press logo?

output


lazarusA:
a .gif works, but you will need to deal with two versions (if wanted), one for the light mode and another for the dark mode. Otherwise, it should be fine. Let's see how the preview looks like.

milankl:
Yep works!!! Gif is transparent and so looks good in dark and light mode

lazarusA:
oh... the link to deploy docs does not work, consider adding https://luxdl.github.io/DocumenterVitepress.jl/stable/manual/get_started#6.-Add-a-PR-Preview-Comment which works in this case.

lazarusA (approved):

milankl:
> oh... the link to deploy docs does not work, consider adding https://luxdl.github.io/DocumenterVitepress.jl/stable/manual/get_started#6.-Add-a-PR-Preview-Comment which works in this case.

So https://github.com/LuxDL/DocumenterVitepress.jl/pull/355 was actually just a work-around to fix https://github.com/LuxDL/DocumenterVitepress.jl/issues/346 because the actual problem there is actually upstream? Happy to use this for now though looking forward to get this fixed upstream, what's blocking that?

lazarusA:
>  what's blocking that?

the know how 😞 , Documenter's internals (I don't know enough). Probably an LLM could find the solution nowadays.

milankl:
> the know how 😞 , Documenter's internals (I don't know enough). Probably an LLM could find the solution nowadays.

So just bother Morten? 😉

PR #1230, opened 2026-08-27
Fix `StochasticallyPerturbedParameterizationTendencies` on GPU

maximilian-gelbrecht:


SpeedyWeather/src/parameterizations/stochastic_physics.jl
```diff
@@ -29,12 +29,33 @@ variables(::AbstractStochasticPhysics) = ()
...
 function initialize!(sppt::StochasticallyPerturbedParameterizationTendencies, model::PrimitiveEquation)
+    # SPPT perturbs with vars.grid.random_pattern but doesn't define that variable itself, it has
+    # to come from another component, e.g. a random process. Check here, otherwise sppt! errors
+    # only later inside the column parameterizations (on GPU as an unsupported jl_f_getfield call)
+    has_random_pattern = any(all_variables(model)) do var
+        var isa GridVariable && var.name == :random_pattern && var.namespace == Symbol()
+    end
+    @assert has_random_pattern "StochasticallyPerturbedParameterizationTendencies requires a "*
+        "`random_pattern` grid variable, define a random process for the model, e.g. "*
+        "`random_process = SpectralAR1Process(spectral_grid)`."
```
milankl, line 40: I would not have found it wrong to simply require this variable to exist, i.e.

```julia
variables(::StochasticallyPerturbedParameterizationTendencies) = (
    GridVariables(:random_pattern, ...
)
```

if you don't select a random process then it's always 0 sure but it would work.

maximilian-gelbrecht: I prefer the `@assert` here because with your suggestions you would effectively just get a silent failure. Because without a random process `StochasticallyPerturbedParameterizationTendencies` doesn't do what the user expects it would do.

milankl: Yep fair enough I see your point, btw,

```julia
var isa GridVariable && var.name == :random_pattern && var.namespace == Symbol()
```

this looks like equivalent to `haskey(vars.grid, :random_pattern)` given you constrain `grid` via `GridVariable` and namespace to empty and how we filter our variables

```julia
function filter_variables(vars, VariableType)
    vars = filter(v -> v isa VariableType, vars)    # filter by variable type
    vars = unique(v -> identifier(v), vars)         # remove duplicates by identifier (group+namespace+name)
```

with 

```julia
identifier(v::AbstractVariable) = Symbol(v.namespace, :_, v.name)
```

PR #1231, opened 2026-08-27
Docs and tests for 3D particle advection (targets #1215)

milankl-claude:
Adds documentation and tests on top of #1215, leaving the implementation itself to @Jullcifer.

## Documentation
A new "3D particle advection" section in `docs/src/particles.md`:
- explains the vertical bracket + blend interpolation (and the pole-average substitute) conceptually, without introducing new equations
- summarises the horizontal-precompute-via-`update_locator!` vs. vertical-on-the-fly design tradeoff, and what it implies for cost/flexibility
- a minimal `ParticleAdvection3D` interface example

## Tests
- **RingGrids (CPU, new coverage):** `RingGrids/test/interpolation_3d.jl` unit-tests `interpolate_3D!` directly — vertical profile reconstruction, pinning outside the σ range, and pole averaging. This function previously had no dedicated test at the `RingGrids` level.
- **GPU:** `SpeedyWeather/test/GPU/particle_advection_3d.jl` adds the three checks I ran manually while testing this branch on an actual GPU (see [my comment on #1215](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1215#issuecomment-5442657920)):
  1. 3D interpolation of random data — passes
  2. 3D particle advection called in isolation (bypassing `run!`) — passes
  3. A full model run with `ParticleAdvection3D` — currently fails to compile on GPU, so this is included as `@test_broken` rather than a plain `@test`, so CI stays green until it's fixed but will flag loudly (via the `@test_broken` promotion error) once it is.

Root cause of (3), for reference: `AnvilLocator`'s new `north_pole_vals`/`south_pole_vals` fields are typed `Union{Nothing, VectorType}`, which is never `isbits` regardless of the runtime value. Since `vars.particles.locator` lives inside `Variables`, and unrelated kernels (e.g. `gpu_column_parameterizations_kernel!`) take the whole `vars` object as an argument, this breaks GPU compilation of the whole time step for *any* particle advection, not just `ParticleAdvection3D`.

I did not touch `SpeedyWeather/src/dynamics/particle_advection.jl`, `RingGrids/src/interpolation.jl`/`interpolation_3d.jl`, or the existing tests in `SpeedyWeather/test/dynamics/particle_advection.jl` — leaving the fix and any changes to those for the PR author.

## Test plan
- [x] `julia --project=RingGrids --check-bounds=yes -e 'using Pkg; Pkg.test("RingGrids")'` — new `interpolation_3d.jl` tests pass
- [x] Ran `SpeedyWeather/test/GPU/particle_advection_3d.jl` against a CUDA backend (NVIDIA L4): 2 passed, 1 correctly reported broken


gregordecristoforo:
The branch `milankl-claude:mc/docs-tests-3d-particle-advection` was manually merged into `jm/3D-particle-tracking` in 825ff3d57fb6c799992e20d4360e52d96feaa3f5.

PR #1232, opened 2026-08-28
move gpu_graphs to /src

gregordecristoforo:
While going through https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1147 again I noticed that both Milan (https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1147#discussion_r3803411741) and Maximilian (https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1147#issuecomment-5327716605) suggested to put gpu_graphs_common,jl into SpeedyTransforms directly.

I forgot to do that in https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1147 so here is the change as suggested. This is purely cosmetic, so feel free to ignore. I will only ask for reviews if the tests pass first.


milankl:
I think the idea was more to always have it being included as part of the main package if the code doesn't depend on the extension triggering package

gregordecristoforo:
I included now the backend agnostic GPU graph code in the main package as suggested and also tried to reproduce the Enzyme error that I ran into when working on it the first time. The [Enzyme differentiability](https://github.com/SpeedyWeather/SpeedyWeather.jl/actions/runs/33158022758/job/98805388133?pr=1232) tests for this PR don't fail but I found that the following MRE still doesn't run locally:

```Julia
using SpeedyWeather, Enzyme

spectral_grid = SpectralGrid(truncation = 11, nlayers = 1)
S = SpectralTransform(spectral_grid, one_more_degree = true)
dS = deepcopy(S)

vor = rand(ComplexF32, spectral_grid.spectrum, spectral_grid.nlayers)
dvor = zero(vor)
res = zero(vor)
dres = zero(res)
fill!(dres, 1 + im)

autodiff(
    Reverse, SpeedyWeather.SpeedyTransforms.∇²!, Const,
    Duplicated(res, dres), Duplicated(vor, dvor), Duplicated(S, dS),
)

println("PASS: no error — this did not reproduce in this environment")
```

This results in the error
```
cannot handle unknown binary operator:
  %.sroa.0.sroa.0.0.insert.insert = or disjoint i64 %.sroa.0.sroa.6.0.insert.shift, %.sroa.0.sroa.0.0.insert.ext
```
This happens while Enzyme reverse-differentiates `spectral_gradients.jl:420`:
```Julia
∇²alms[I] = add ? ∇²alms[I] + a : a
```
i.e. the `setindex!` into the `LowerTriangularArray`, where `I::CartesianIndex{2}`.

@maximilian-gelbrecht, you asked about an MRE for this error. Should we file a bug for `Enzyme.jl` or do I misinterpret the results?

Is this PR worth merging as this MRE also fails for `main` or should I just close it as it doesn't add any new functionality?

milankl (approved):
Yeah definitely like this, I think it makes it much clearer what the general interface is and what extensions extend given CUDA/AMDGPU specifics. An extension of course shouldn't be as brief as possible for the sake of loading a lot of code that isn't otherwise used, but if there's two extensions using code I think that's very valid to have this code centrally given it adds little to no precompile time? I haven't seen many other packages with a repeated `include` in several extensions so happy as is now!

I feel like the Enzyme questions is unrelated to this PR?

gregordecristoforo:
Cool! Yes I stumbled upon the the Enzyme issue by accident while working on this. I'll create a separate issue for it and merge this PR once the tests pass.

PR #1234, opened 2026-08-28
Optimize the other models by lazy batching of FFT plans 

### LINKED ISSUE #1208: GPU performance of the other models
We spend a lot of time optimising the GPU performance of the `PrimitiveWetModel` - with success as the benchmarks show. But the benchmarks also show that this isn't true for the other models. On GPU the  `PrimitiveWetModel` is actually the fastest of all our models currently. This shouldn't be the case.  

Once we find the time, we should look into fixing performance issues of the other models as well. The most likely culprit is how exactly the variable fusion / FFT batching is done.

maximilian-gelbrecht:
So far just Claude generated solution to close #1208 

# Plan missing FFT batch sizes on first use (GPU)

## Problem

On GPU, a `transform!` whose batch dim `K` has no pre-planned FFT falls back to
`_fourier_serial!`, which applies `nlat*K` individual per-ring FFTs and is not captured into a
GPU graph (only `_fourier_batched!` is). Two model classes were permanently stuck on that path:

- **`PrimitiveDryModel`**: its spectral→grid prognostic batch is `3L+1` (vorticity, divergence,
  temperature, surface pressure), but `default_transform_batch` only lists `4L+1` — the *wet*
  model's batch, which includes humidity. `3L+1` was never planned.
- **All single-layer models** (`BarotropicModel`, `ShallowWaterModel`): `_fourier!` required
  `K > 1` for the batched path and `plan_FFTs` skipped `K = 1` for the batched plans, so every
  single-layer transform went serial regardless of `transform_batch`.

The root cause is structural: `transform_batch` is fixed when the `SpectralGrid` is built, but
which `K` a model emits follows from the model *type*, not from `nlayers` alone — it is not known
at that point.

## Fix

`SpeedyTransforms.ensure_batched_plans!(S, K)` decides the path:

- **CPU** — unchanged, a pure lookup (`K > 1 && haskey(...)`). Unplanned `K` is still split by
  `_transform_chunked!`, and `K = 1` stays serial to avoid FFTW's x86 alignment mismatch.
- **GPU** — a missing `K` (including `K = 1`) is planned on first use via the new
  `plan_batched_FFTs!(S, K)` and cached in the plan `Dict`s, so `transform_batch` becomes a hint
  that avoids planning during the first time step rather than a requirement. Falls back to serial
  only if `K` exceeds the scratch capacity `S.nlayers`.

## Results

Measured on an A40, `Float32`, `OctahedralGaussianGrid`, FFT+Legendre `SpectralTransform` forced
(at T ≤ 64 `WhichTransform` picks `MatrixSpectralTransform`, which is unaffected):

| Model | ms/step before | ms/step after | speedup |
| --- | --- | --- | --- |
| PrimitiveDry T32 L8 | 24.18 | 1.23 | 19.6x |
| PrimitiveDry T128 L8 | 100.42 | 5.66 | 17.7x |
| Barotropic T32 L1 | 2.80 | 0.83 | 3.4x |
| Barotropic T128 L1 | 11.59 | 4.32 | 2.7x |
| ShallowWater T32 L1 | 4.67 | 1.30 | 3.6x |
| ShallowWater T128 L1 | 21.59 | 6.75 | 3.2x |
| PrimitiveWet (unchanged, reference) T128 L8 | 5.99 | 5.99 | — |

Model cost now orders as expected (Barotropic < ShallowWater ≲ PrimitiveDry < PrimitiveWet); before
this, the dry model was ~17x slower than the wet one.

Note the GPU section of `SpeedyWeather/benchmark/README.md` predates #1194 and shows the old
behaviour at T32 as well; it needs a re-run.

## Testing

- `SpeedyTransforms/test/spectral_transform.jl` — asserts the CPU path stays a pure lookup and
  creates no plans on the fly.
- `SpeedyWeather/test/GPU/spectral_transform.jl` — new testset: an unplanned `K` (and `K = 1`) is
  planned on first use, the plan is cached, results are **bit-identical** to the serial path, and
  `K > S.nlayers` still falls back to serial.
- Existing GPU suites pass unchanged (`fourier_batched`/`fourier_serial`/`legendre`, CUDA-graph
  equivalence, Barotropic/ShallowWater/PrimitiveWet model runs), plus a dry-model run on GPU that
  confirms `3L+1` gets planned.

## Known limitations

- Plan creation mutates `S` on first use, so it is not thread-safe if two threads first hit the
  same new `K` concurrently; SpeedyWeather drives the transform single-threaded from the host.
- GPU only. Reactant (`MatrixSpectralTransform`) and CPU keep their previous behaviour.


SpeedyWeather/benchmark/README.md
```diff
@@ -397,94 +397,94 @@ Toolchain:
 
 | Model | truncation | L | Physics | Δt | SYPD | Memory|
 | --- | --- | --- | --- | --- | --- | --- |
-| BarotropicModel | 32 | 1 | false | 1800 | 2113 | 490.02 KB |
-| ShallowWaterModel | 32 | 1 | false | 2400 | 1671 | 493.52 KB |
-| PrimitiveDryModel | 32 | 8 | true | 2400 | 355 | 661.63 KB |
-| PrimitiveWetModel | 32 | 8 | true | 2400 | 5708 | 666.76 KB |
+| BarotropicModel | 32 | 1 | false | 1800 | 7932 | 429.08 KB |
```
maximilian-gelbrecht, line 400: Now this looks good again

SpeedyWeather/benchmark/README.md
```diff
@@ -397,94 +397,94 @@ Toolchain:
...
+| ShallowWaterModel | 32 | 1 | false | 2400 | 8926 | 432.58 KB |
```
maximilian-gelbrecht, line 401: Still a bit odd that ShallowWater is faster than Barotropic but I'll take it for now.

milankl: That might be because we use `NCycleLorenz` here with a shorter time step here by default I believe!

maximilian-gelbrecht: yeah this also checks out with the timings per step reported above in the post

SpeedyTransforms/src/fourier.jl
```diff
@@ -1,25 +1,44 @@
...
+"""$(TYPEDSIGNATURES)
+Can the batched Fourier path be used for a call with batch dim `K`? On GPU any `K` is worth a 
+batched plan, `K = 1` included. Which `K` a model emits only follows from the model type, 
+so it is not generally known when the `SpectralGrid` (and with it `transform_batch`) is constructed. 
+Hence plan a missing `K` on first use and cache it for all subsequent calls. Falls back to the serial 
+path only if `K` exceeds the scratch memory capacity `S.nlayers`."""
+@inline function ensure_batched_plans!(S::SpectralTransform{NF, <:GPU}, K::Integer) where {NF}
+    haskey(S.rfft_plans_batched, K) && return true
+    K <= S.nlayers || return false      # scratch memory too small for a batched call, go serial
+    plan_batched_FFTs!(S, K)
```
maximilian-gelbrecht, line 38: So this is the crucial point of the PR. On GPU only and only if the `K` isn't already in the plans we replan the batches. The default is set so that this doesn't happen for the `PrimitiveWetModel`

SpeedyTransforms/src/fourier.jl
```diff
@@ -339,15 +358,52 @@ function plan_FFTs(
...
+            for j in 1:nlat_half
+    ]
+    return nothing
+end
+
+"""$(TYPEDSIGNATURES)
+Plan the batched FFTs for batch dim `K` on an already constructed `SpectralTransform` and cache
+them in it, so that later calls with that `K` take the batched path, see
+[`ensure_batched_plans!`](@ref)."""
+function plan_batched_FFTs!(S::SpectralTransform{NF}, K::Integer) where {NF}
```
maximilian-gelbrecht, line 402: And here's the replanning itself. Because the Dict s are mutable that's possible.

SpeedyWeather/benchmark/README.md
milankl: At the moment the higher order Lorenz time stepping is slower but also more accurate, maybe we can make this trade off less in the future but right now Lorenz time-stepped simulations are slower

SpeedyTransforms/src/fourier.jl
milankl: This looks good to me, so at runtime the key `K` of the dictionary is checked and if it doesn't exist we plan. Sounds like the only additional work that's done here in the hot loop is that the keys of the dictionary are checked over and over again? Fine with me if that doesn't hurt performance. `haskey` is compile time for named tuples but runtime for dictionaries, just curious whether this causes synchronization or something because it's a central mutable object that could have changed anytime. But happy to let the performance decide!

SpeedyTransforms/src/fourier.jl
```diff
@@ -339,15 +358,52 @@ function plan_FFTs(
...
+        rfft_plans_batched::Dict{Int},
+        brfft_plans_batched::Dict{Int},
+        K::Integer,
+        fake_grid_data::AbstractField,
+        scratch_memory_north::AbstractArray{<:Complex},
+        rings,
+        nlons::Vector{<:Int},
+    )
+    nlat_half = length(nlons)
+    nfreqj(j) = nlons[j] ÷ 2 + 1
```
milankl, line 385: I always find these functions that access objects from the outer scope weird, but I guess it's within a function and so its safe?

milankl (approved):
Yeah really good Max! I've just been wondering whether the `haskey(::Dict, ::Int)` isn't hurting GPU performance, see below but it seems it doesn't!

SpeedyTransforms/src/fourier.jl
maximilian-gelbrecht: Yeah this should be a compile time thing, the benchmarks also show that it's fine

SpeedyTransforms/src/fourier.jl
maximilian-gelbrecht: I would assume that in this case here this actually just gets automatically inlined

PR #1238, opened 2026-08-31
HEALPix Output 

### LINKED ISSUE #1080: HEALPix Output
In #1076 we already started discussing a possible HEALPix output. 

We have to review current standards to encode it in Zarr format and read it across e.g. Nvidia's implementation and other available ones. The actual implementation in our code should be fairly simple. 

cc @pierluigividale

maximilian-gelbrecht:
Closes #1080 

This now uses the RING convention of cuHPX and healpy to save a flat array of the data in a Zarr store. It additionally saves coordinates and ring indices in flat arrays. 

The RING convention seems ideal to me because it also allows to completely skip interpolation and reordering. It just directly writes the arrays to disk as they are. Seems like a great option for a really fast output writer.


SpeedyWeather/ext/SpeedyWeatherZarrExt/healpix_output.jl
```diff
@@ -0,0 +1,294 @@
...
+    n = Int(something(nlat_half, model_nlat_half))
+    n > 0 || throw(ArgumentError("HEALPixOutput: nlat_half=$n must be positive."))
+    iseven(n) && return n
+
+    @info "HEALPixOutput: HEALPixGrid is only defined for even nlat_half, rounding $n up to $(n + 1) (nside=$((n + 1) ÷ 2))."
+    return n + 1
+end
+
+"""$(TYPEDSIGNATURES)
+Constructor for [`HEALPixOutput`](@ref) (extension version, available once `Zarr.jl` is
```
bgroenks96: It doesn't make much sense to put this note about loading `Zarr` in the Zarr extension itself. That note should be on the stub in main source code.

Also, is there a reason why HEALPixOutput needs to be specific to ZARR? Couldn't it also work with JLD2/HDF5?

SpeedyWeather/ext/SpeedyWeatherZarrExt/healpix_output.jl
```diff
@@ -0,0 +1,294 @@
...
+
+The output resolution is given either as `nside`, as `nlat_half = 2nside`, or as a ready-made
+`output_grid::HEALPixGrid`, and defaults to the model grid's `nlat_half` (rounded up to the
+nearest even number, which `HEALPixGrid` requires). An interpolator onto that grid is only
+built when it is actually needed: if the model already runs on the very same HEALPix grid,
+`output.interpolator` stays `nothing` and the data is copied straight out, skipping
+interpolation entirely."""
+function HEALPixOutput(
+        SG::SpectralGrid,
+        Model::Type{<:AbstractModel} = Barotropic;
```
bgroenks96, line 140: Not sure if this is already a standard pattern or not, but I don't think there should be a default argument for the model type. That should not be assumed.

SpeedyWeather/ext/SpeedyWeatherZarrExt/healpix_output.jl
```diff
@@ -0,0 +1,294 @@
...
+        output_grid::Union{HEALPixGrid, Nothing} = nothing,
+        output_NF::DataType = DEFAULT_OUTPUT_NF,
+        interval::Period = Second(DEFAULT_OUTPUT_INTERVAL),
+        compressor = nothing,
+        kwargs...
+    )
+
+    # OUTPUT GRID: from `output_grid`, or built from `nside`/`nlat_half`/the model's own
+    # resolution. Always on CPU, we write from the host. The `::HEALPixGrid` annotation
+    # rejects other (Octa)HEALPix-like grids, which healpy/cuHPX cannot read.
```
bgroenks96: rejects? what is being rejected? the choice of `nlat_half`?

SpeedyWeather/ext/SpeedyWeatherZarrExt/healpix_output.jl
```diff
@@ -0,0 +1,294 @@
...
+    return nothing
+end
+
+"""$(TYPEDSIGNATURES)
+Write the coordinate arrays of the flat HEALPix store into the Zarr group `g`:
+
+- `cell`: the flat index `1:npix` of every grid point, the store's horizontal dimension.
+  HEALPix RING order, so cell `ij` is RING pixel `ij-1` in healpy/cuHPX's 0-based indexing.
+- `lat`, `lon`: latitude (˚N) and longitude (˚E) of every cell, one flat vector each.
+- `ring`: the 1-based latitude ring index of every cell, north to south.
```
bgroenks96, line 283: Did you check what the latitude ordering convention is for cuHPX?

SpeedyWeather/ext/SpeedyWeatherZarrExt/healpix_output.jl
maximilian-gelbrecht: I'm going to edit the comment. 

Wrt JLD2/HDF, yes in principle that would be possible but I really want to avoid the code complexity to make it independent of the backend. You'd need all those definitions of dimensions etc for all formats. I'd rather have it just working with Zarr for now and if this becomes the standard we can still think about extending it

SpeedyWeather/src/output/variables/output_variables.jl
```diff
@@ -38,6 +38,21 @@ get_indices(i, x::Val{true}, y::Val{true}, z::Val{true}, t::Val{false}) = (:, :,
...
+Indices to write the output slice of `variable` at output step `i` into a store that keeps
+the horizontal dimension *unravelled* into a single flat dimension, as a `Field`'s `data`
+already is. The `x` and `y` flags of `variable.dims_xyzt` collapse into that one flat
+dimension, so a 3D+time variable is written as `(:, :, i)` where the equivalent
+[`get_indices`](@ref) for a rectangular `lon`/`lat` store gives `(:, :, :, i)`.
+Used by the flat [`HEALPixOutput`](@ref) backend."""
+get_flat_indices(i, variable::AbstractOutputVariable) =
+    get_flat_indices(i, Val(is3D(variable)), Val(hastime(variable)))
+
+get_flat_indices(i, z::Val{true}, t::Val{true}) = (:, :, i)     # 3D + time
```
bgroenks96, line 51: Is there a reason why the `Val` dispatch pattern is needed here? If not, I would maybe simplify it because it makes the code harder to understand when it's pretty trivial what it's doing.

SpeedyWeather/ext/SpeedyWeatherZarrExt/healpix_output.jl
maximilian-gelbrecht: It rejects OctaHealpix as it says. 

Even though I think we should probably lift that restriction. CuHPX/healpy will not be able to read it, but we could nonetheless output the data with the same convention

SpeedyWeather/ext/SpeedyWeatherZarrExt/healpix_output.jl
maximilian-gelbrecht: That's a thing we have for all output writers to get reasonable defaults for each model.

maximilian-gelbrecht: The default, yeah that's not great but hard to get around it. It exists for the other output writers as well

bgroenks96: but for the other models you must anyway pass the model type as an argument, why have the default for `Barotropic`?

SpeedyWeather/ext/SpeedyWeatherZarrExt/healpix_output.jl
bgroenks96: where does it reject `OctaHEALPix`? I only see it reading in `healpix_nlat_half(SG.grid.nlat_half; nside, nlat_half)`. Does `OctaHEALPix` have incompatible `nside`? Where does the error actually get thrown?

SpeedyWeather/ext/SpeedyWeatherZarrExt/healpix_output.jl
maximilian-gelbrecht: Because it's guaranteed to work and doesn't break the output. The `Model` is used to add the default outputs (and for nothing else). All other models have all variables that the barotropic one has, so its defaults always work, also for the other models. The defaults of the primitive wet model would break the output for all models except for the primitive wet model.

SpeedyWeather/ext/SpeedyWeatherZarrExt/healpix_output.jl
maximilian-gelbrecht: The input argument is typed as `output_grid::Union{HEALPixGrid, Nothing}`, that's why all other grids are rejected.

SpeedyWeather/ext/SpeedyWeatherZarrExt/healpix_output.jl
maximilian-gelbrecht: As far as I can see it cuHPX doesn't specify this anywhere. cuHPX is GPU only, I didn't test yet. 

I checked against `healpy` instead, there we follow the convention with `lonlat=True`, and I tested that it works.

SpeedyWeather/src/output/variables/output_variables.jl
maximilian-gelbrecht: The other implementations do it in the same way. It's compile time, so maybe a bit faster. 
Claude (correctly, I think) just copied the implementation of the other output logic.

SpeedyWeather/ext/SpeedyWeatherZarrExt/healpix_output.jl
maximilian-gelbrecht: I did loosen that restriction now though. But note in the documentation that `OctaHEALPix` isn't supported by cuHPX/healpy.

SpeedyWeather/ext/SpeedyWeatherZarrExt/healpix_output.jl
maximilian-gelbrecht: I just checked with cuHPX and it seems to work to load them and do transforms on GPU with it as well.

SpeedyWeather/ext/SpeedyWeatherZarrExt/healpix_output.jl
bgroenks96: Ok

SpeedyWeather/src/output/variables/output_variables.jl
bgroenks96: It's only compile time if `is3D` and `hastime` are compile-time constants, but I guess they are?

SpeedyWeather/ext/SpeedyWeatherZarrExt/healpix_output.jl
bgroenks96: Hmm... still seems odd but anyway out of scope.

SpeedyWeather/src/output/variables/output_variables.jl
maximilian-gelbrecht: They could be in theory, but I don't think they currently are as it's coded. 

This is an issue that's related to all output writers though, I would prefer to do the optimisation of the output writers in a separate PR/issue. There's also a few other things on my to-do list there like using batched interpolation on GPU. So this will be addressed eventually.

SpeedyWeather/ext/SpeedyWeatherZarrExt/healpix_output.jl
```diff
@@ -0,0 +1,388 @@
...
+
+    # SKIP INTERPOLATION if the model already runs on this very grid: `grids_match`
+    # compares the (nonparametric) grid type and nlat_half, which is exactly the condition
+    # under which `RingGrids.interpolate!` would degenerate to a copy anyway. Not building
+    # the interpolator also saves precomputing its stencil indices and weights.
+    interpolator = if grids_match(output_grid, input_grid)
+        nothing
+    else
+        RingGrids.interpolator(output_grid, input_grid, NF = DEFAULT_OUTPUT_NF)
+    end
```
milankl, line 166: Definitely the way to go for now, but in a follow-up we should do HEALPix to HEALPix and OctaHEALPix to OctaHEALPix "interpolation" via rearranging to nested. Probably the easiest would be to reorder ring to nest into a scratch array (from variables, they'll be of the right size anyway) and then change resolution by always averaging 4 consecutive values, but also I feel like it while we always allowed to output at different resolution, it's not a super popular feature.

SpeedyWeather/ext/SpeedyWeatherZarrExt/healpix_output.jl
```diff
@@ -0,0 +1,388 @@
...
+
+    # pick out the active dims: cell is always on (asserted above), the vertical and time
+    # ones follow the variable's z/t flags
+    active = (true, is3D(var), hastime(var))
+    shape = Tuple(d for (d, on) in zip(full_shape, active) if on)
+    chunks = Tuple(c for (c, on) in zip(full_chunks, active) if on)
+    dims = String[string(d) for (d, on) in zip(all_dims, active) if on]
+
+    # `_ARRAY_DIMENSIONS` is read row-major by xarray, our shape/chunks are column-major
+    reverse!(dims)
```
milankl, line 387: Is that just to guarantee xarray compatibility? So if were to read this data again, we suddenly need to follow row-major?

SpeedyWeather/ext/SpeedyWeatherZarrExt/shared.jl
```diff
@@ -0,0 +1,274 @@
...
+Remove all variables from `output.variables` that don't exist in `simulation`, warning
+about them unless `warn=false`. Returns `output`."""
+function prune_nonexisting_variables!(output::AbstractZarrOutput, simulation::AbstractSimulation; warn::Bool = true)
+    nonexisting_vars = [key for (key, var) in output.variables if isnothing(path_or_nothing(var, simulation))]
+    if !isempty(nonexisting_vars) && warn
+        @warn "Some output.variables do not exist in simulation. Deleting: $(join(nonexisting_vars, ", "))"
+    end
+    delete!(output, nonexisting_vars...)
+    return output
+end
```
milankl: Don't we already have this function? Feels like it could just be generally defined for every output writer that has `output.variables::Dict` ?

SpeedyWeather/test/GPU/barotropic.jl
```diff
@@ -48,3 +49,37 @@ end
...
 end
+
+@testset "GPU Barotropic with HEALPixOutput, interpolation skipped" begin
+    # model already on a HEALPixGrid *on GPU*, output on the same grid on CPU. `grids_match`
+    # compares grid type and nlat_half only, so no interpolator is built and the output path
+    # takes the plain `copyto!` branch — after the GPU -> CPU transfer of the model field.
+    tmp_output_path = mktempdir(pwd(), prefix = "tmp_gpu_healpix_skip_")
+
+    spectral_grid = SpectralGrid(
+        truncation = 33, nlayers = 1, Grid = HEALPixGrid,
```
milankl: That doesn't yield a square nside for the HEALPix. 

- truncation=33 -> 71-ring HEALPix -> nlat_half = 36 -> nside=18 (not square)
- truncation=32 -> 63-ring HEALPix -> nlat_half = 32 -> nside=16 (square!)

milankl: Claude now keeps getting confused what truncation to use, I believe it has seen some old SpeedyWeather code in its training data which doesn't agree anymore ... maybe worth a CLAUDE.md note

milankl (commented):
Looks great to me!

SpeedyWeather/test/GPU/barotropic.jl
maximilian-gelbrecht: Oh I think that was just a much more old school copy+paste error, because all the other tests here use `truncation=33` which might have been wrongly adjusted before in one of the last PRs? 

Anyway, good you saw it.

SpeedyWeather/ext/SpeedyWeatherZarrExt/healpix_output.jl
maximilian-gelbrecht: Yes, but that's handled automatically behind the scenes by Zarr.jl

SpeedyWeather/ext/SpeedyWeatherZarrExt/shared.jl
maximilian-gelbrecht: I revised it now and unified it across different output writer more along what as already there

milankl: don't see this being pushed?

maximilian-gelbrecht: oops, yeah, now it's pushed

milankl (approved):
Great!! Looks good to me

PR #1223, opened 2026-08-24
Add mutating variants for `heatmap` and `globe`

bgroenks96:
This is a long overdue improvement to make it easier to build multi-panel plots. There's probably a better way to do this with plotting recipes, but I can't be bothered to figure that out right now, so hopefully this does the trick for the time being.

Note: This should be merged *after* #1224

Quick example:

```julia
import CairoMakie: heatmap!, resize_to_layout!

grid = RingGrids.OctahedralGaussianGrid(24)
let fig = Figure(size = (600, 500))
    field1 = randn(grid)
    field2 = randn(grid)
    heatmap!(fig[1, 1], field1)
    heatmap!(fig[2, 1], field2)
    resize_to_layout!(fig)
    fig
end
```

image

PR #1247, opened 2026-09-03
Add `rng` argument to `rand` and `randn` for fields and LowerTriangularArrays

milankl-claude:
Follow-up to the discussion in #1223: `randn(rng, grid)` did not exist, so the docs there had to fall back on `Random.seed!`.

`rand` and `randn` for `Field`s (RingGrids) and `LowerTriangularArray`s now accept an optional `Random.AbstractRNG` as first argument, matching the standard `Base`/`Random` interface. All existing signatures gain an `rng` variant, so

```julia
using RingGrids, Random

grid = OctahedralGaussianGrid(24)
randn(Xoshiro(123), grid)               # reproducible
randn(Xoshiro(123), Float32, grid, 8)   # NF and nlayers as before
rand(Xoshiro(123), OctahedralGaussianField{Float32}, 24)

using LowerTriangularArrays
spectrum = Spectrum(10, 10)
randn(Xoshiro(123), spectrum)
rand(Xoshiro(123), LowerTriangularArray{Float32}, spectrum, 8)
```

Implementation-wise the existing `for f in (:zeros, :ones, :rand, :randn)` metaprogramming loops just gained an inner loop over an empty/`rng` argument prefix, so the `rng` methods stay in sync with the non-`rng` ones by construction. `zeros`/`ones` are unchanged.

- `Random` added as a dependency of `RingGrids` and `LowerTriangularArrays`, versions bumped with `+DEV`
- new test sets in `RingGrids/test/grids.jl` and `LowerTriangularArrays/test/lower_triangular_matrix.jl` checking reproducibility for the same seed and that the `rng` variants return the same type/size as the non-`rng` ones
- both package test suites pass with `--check-bounds=yes`, plus a `PrimitiveWetModel` smoke run

@milankl @maximilian-gelbrecht


maximilian-gelbrecht (approved):
Looks good to me.

bgroenks96 (approved):
LGTM as well.

PR #1249, opened 2026-09-07
Enzyme differentiability tested with Julia 1.11 as well and fix issue with Reactant 0.2.285

maximilian-gelbrecht:
After the recent changes differentiating with Enzyme does actually work with Julia 1.11.


SpeedyWeather/src/time_stepping/steppers/leapfrog.jl
```diff
@@ -204,23 +204,21 @@ end
...
-        # rotation and orbit time are dilated, so rewind them by their dilated Δt/2
-        half_Δt = Δt ÷ 2
-        clock.time -= half_Δt
-        clock.rotation_time -= dilate(half_Δt, clock.rotation_dilation)
-        clock.orbit_time -= dilate(half_Δt, clock.orbit_dilation)
-        time_step!(clock, Δt)
-    else                            # later steps: Leapfrog at 2Δt but increase clock by Δt
-        time_step!(clock, Δt)
-    end
+    first_step = i == 0             # first Euler step at Δt/2
```
maximilian-gelbrecht: @milankl this also changes the main `time_step!` so please have a quick look. 

For the Reactant version this is even more clean now then before because with just `ifelse` we don't need any explicit tracing anymore. 

Otherwise this should reproduce exactly the same behaviour as before. Obviously below the `-= 0` are slightly wasteful but it those really shouldn't matter in the grand scheme of things.

PR #1253, opened 2026-09-12
Add convenience methods for Laplacian in grid space

maximilian-gelbrecht:
cc @henrikauestad


docs/src/gradients.md
```diff
@@ -5,11 +5,18 @@ spherical harmonics. These are in particular ``\nabla, \nabla \cdot, \nabla \tim
...
-(i.e. `LowerTriangularMatrix`) or gridded fields (i.e. `<:AbstractGrid`) and
+(i.e. `LowerTriangularArray`) or gridded fields (i.e. `<:AbstractField`) and
 also allow as an additional argument a spectral transform object
 (see [SpectralTransform](@ref SpectralTransform)) which avoids recalculating it
 under the hood.
 
+The gradient operators act in spectral space, so the methods taking a field transform
+to spectral space internally. For `∇²`, `∇⁻²` the result is transformed back, meaning
+they map a field to a field, while `divergence` and `curl` return the spectral
+coefficients (transform them back yourself if you want them on the grid).
```
milankl, line 16: Is there a consistent logic that if we take gradient operators on `Field`s that they are transformed back or not? Maybe they always should be transformed back?

milankl (approved):
Happy with this! Just to check whether we need a consistent logic here e.g. that field in field out and LTA in LTA out always holds with gradient operators?

docs/src/gradients.md
maximilian-gelbrecht: All routines that I add here follow that logic, I thought it would make sense, as they are pure convince routines anyway. 

Otherwise, we don't really have a logic for that. `divergence` and `curl` always return spectral with both grid and spectral input. 

With all the routines that take in spectral, some return grid some spectral but we also need those for the dycore. 

`divergence` and `curl` from grid are also just purely convinience functions. Shall we change those to also transform back to grid? Then all routines that take in gridded data would also return gridded data.

milankl: I think I prefer them not to be transformed back. You need to transform into spectral to take the operation but you don't have to transform back. So better not do that implicitly? So happy as things are sounds sound!

maximilian-gelbrecht: So should the laplacian functions also not transform back?

milankl: Yeah I think it's consistent to not transform back, one can always put a `transform()` call around it? @henrikauestad you think `B = transform(∇²(A, S), S)` is cumbersome? We could bake the `transform` into the `∇²` but then you'd need another round trip transform to apply another gradient operator ... on the other hand if one writes `B = transform(∇²(A))` then actually 2 transforms have to be created ... tricky!

henrikauestad: If unaware of whats going on under the hood I think it is counterintuitive that the function returns spectral when you feed it gridded data. How about a boolean keyword argument that defaults to returning gridded if the input is gridded (and spectral if the input is spectral)? Is that against the JulianSpirit? 

If adding a kwarg requires extra effort cause these are called in the time stepping, then I'd just stay consistent with the other routines.

Issue #268, opened 2023-03-23
Interpolation on pressure levels


milankl:
We currently output on $\sigma = p/p_s$ levels, however at some point we may want to interpolate onto pressure levels, the following function outlines how this can be done

```julia
function interpolate_pressure_levels(A::AbstractArray{T,3},pₛ::AbstractMatrix,σ::AbstractVector,p::Real) where T
    nx,ny,nlev = size(A)
    B = zeros(T,nx,ny)
    
    for i in 1:nx
        for j in 1:ny
            σij = p/pₛ[i,j]            
            if σij < σ[1]
                B[i,j] = A[i,j,1]
            elseif σij >= σ[end]
                B[i,j] = A[i,j,end]
            else
                k = 1
                while (σ[k] < σij) k += 1 end
                Δσ = σ[k+1] - σ[k]       # level thickness
                w1 = (σ[k+1] - σij)/Δσ   # weight for k level
                w2 = (σij - σ[k])/Δσ     # weight for k+1 level
                B[i,j] = w1*A[i,j,k] + w2*A[i,j,k+1]
            end
        end
    end
    
    return B
end 
```

weathermanbarnes:
Hi SpeedyWeather developers. I have just stumbled across this repo and it looks really great! I am new to Julia, but will definitely be keen to play around with this more! Interpolation and output to pressure levels would be really great. I would also be interested in output to isentropic levels and an option for a potential vorticity variable.

milankl:
I think there are important considerations for online vs offline interpolation on pressure levels that have pros and cons

- Interpolate to pressure levels on the fly: For 3D variable the `NetCDFOutput` has a full 3D variable on model levels on a full grid, say `FullGaussianGrid` which data from the internal grid, say `OctahedralGaussianGrid` is first interpolated onto before writing to disk. We could then interpolate in the vertical onto another 3D variable with pressure levels before writing to disk. This would add some compute but would reduce the size of the netCDF file if the number of model levels is higher than the number of pressure levels to output onto.

[...]

PR #1256, opened 2026-09-14
Sigma to pressure interpolation for output

henrikauestad:
ref #268 draft of column-wise interpolation functions from sigma levels to pressure levels. Both normal and logarithmic interpolation. Could be parallelised similarly to the column parameterisations. 

cc @milankl 

! Current function will fail if the number of pressure levels specified is larger than the number of vertical levels in the model...


SpeedyWeather/src/output/vertical_interpolation.jl
```diff
@@ -0,0 +1,148 @@
...
+    vars::Variables,
+    model::AbstractModel,
+)
+    @assert issorted(p)
+
+    n_plev = length(p)
+    σ      = model.geometry.σ_levels_full
+    pₛ     = vars.parameterizations.surface_pressure
+
+    out_field = vars.scratch.grid.a[:,1:n_plev]
```
henrikauestad, line 17: only works if n_plev <= nlev

SpeedyWeather/src/output/vertical_interpolation.jl
```diff
@@ -0,0 +1,148 @@
...
+    model::AbstractModel,
+)
+
+    @assert issorted(p)
+    
+    n_plev = length(p)
+    σ      = model.geometry.σ_levels_full
+    pₛ     = vars.parameterizations.surface_pressure
+
+    out_field = vars.scratch.grid.a[:,1:n_plev]
```
henrikauestad, line 92: only works if n_plev <= nlev

SpeedyWeather/src/output/vertical_interpolation.jl
```diff
@@ -0,0 +1,148 @@
...
+
+    for ij in eachindex(in_field[:,1])
+        column_interpolate_pressure_levels!(ij,out_field,in_field,pₛ,σ,p,n_plev)
+    end
+    
+    return out_field
+end
```
milankl-claude: **Interface**: this should not take `Variables` (or `model`) — pass the externally allocated output field, the surface pressure field and the model's vertical coordinates instead. Three reasons:

- `out_field = vars.scratch.grid.a[:, 1:n_plev]` is where the `n_plev <= nlayers` restriction you mention in the PR description comes from. It's also a *copy*, not a view: `getindex(::AbstractField, ::Colon, k...)` allocates a new `Field` (`RingGrids/src/field.jl:138`), so every call allocates a new array and the result is not the scratch array at all. And the scratch is 2 arrays shared by the whole dynamical core, so writing output into it would be fragile even at matching sizes.
- `eachindex(in_field[:, 1])` allocates a full copy of the first layer just to get `1:npoints` — `eachgridpoint(in_field)` is the iterator you want (but see below, the kernel indexes for you).
- The serial `for ij` loop is CPU-only. The operation is parallel over columns *and* over target levels, so a single kernel over `(ij, k)` covers both architectures — `launch!` with `RingGridWorkOrder` and a 2D worksize does exactly that, see `dynamics/vertical_advection.jl:78` for the same pattern.

```suggestion
function interpolate_pressure_levels!(
        out_field::AbstractField,                   # OUTPUT: (lonlat, npressure), externally allocated
        in_field::AbstractField,                    # INPUT: (lonlat, nlayers) on model levels
        surface_pressure::AbstractField,            # INPUT: (lonlat) surface pressure [Pa]
        p::AbstractVector,                          # pressure levels [Pa], ascending, on the same architecture
        coordinates::AbstractVerticalCoordinates,   # σ or hybrid σ-p coordinates of the model
        interpolation::AbstractVerticalInterpolation = LogInterpolation(),
        extrapolation::AbstractVerticalExtrapolation = ConstantExtrapolation(),
    )
    grids_match(out_field, in_field, surface_pressure) || throw(DimensionMismatch(out_field, in_field))
    size(out_field, 2) == length(p) || throw(DimensionMismatch(
        "$(size(out_field, 2)) pressure levels allocated, but $(length(p)) requested"))

    nlayers = size(in_field, 2)
    launch!(
        architecture(in_field), RingGridWorkOrder, size(out_field),
        interpolate_pressure_levels_kernel!,
        out_field, in_field, surface_pressure, p, coordinates, nlayers,
        interpolation, extrapolation,
    )
    return out_field
end

@kernel inbounds = true function interpolate_pressure_levels_kernel!(
        out_field, in_field, surface_pressure, p, coordinates, nlayers,
        interpolation, extrapolation,
    )
    ij, k = @index(Global, NTuple)
    out_field[ij, k] = interpolate_pressure_level(
        ij, p[k], in_field, surface_pressure[ij], coordinates, nlayers,
        interpolation, extrapolation,
    )
end
```

(`@assert issorted(p)` is worth keeping, but do it once when the output variable is constructed rather than on every output step — and note `@assert` may be compiled away.)

SpeedyWeather/src/output/vertical_interpolation.jl
```diff
@@ -0,0 +1,148 @@
...
+    in_field::SpeedyWeather.AbstractField{T,2}, # input with dims (lonlat, σ-levels)
+    pₛ::SpeedyWeather.AbstractField{T,1},
+    σ::AbstractVector{T},
+    p::AbstractVector,
+    n_plev::Int,
+) where T
+
+    for pk in 1:n_plev
+
+        σij = p[pk] / pₛ[ij]
```
milankl-claude: **This is the line that ties the implementation to σ coordinates.** `σij = p/pₛ` is the model's vertical coordinate only for `SigmaCoordinates`. With `SigmaPressureCoordinates` (already in `main`, and the dynamical core is coming in #1236) full levels sit at `p_k = A[k]*p_ref + B[k]*pₛ`, so comparing `p/pₛ` against `σ` would search the wrong levels — over high orography quite badly so.

You get hybrid support for free by doing the search in pressure instead, `pressure(k, pₛ, coordinates)` is already defined for both coordinate types (`dynamics/vertical_coordinates.jl:107` and `:258`):

```julia
p_k = pressure(k, surface_pressure[ij], coordinates)   # σ[k]*pₛ or A[k]*p_ref + B[k]*pₛ
```

There's precedent for this in output code already, see `output/variables/boundary_layer.jl:69` and the comment above it. Same argument applies to the log variant: `log(σ/σ₁)/log(σ₂/σ₁) == log(p/p₁)/log(p₂/p₁)` holds only because `pₛ` cancels for σ coordinates — with hybrid you really do need `log p`.

milankl-claude (commented):
> [!NOTE]
> Revised after a design discussion with @milankl: the output integration below now attaches the pressure levels to the **output writer** rather than wrapping individual output variables. The numerics part (and all inline suggestions) is unchanged.

Thanks @henrikauestad for kicking this off! The column maths is nearly there and splitting it into a driver + a column function is exactly the right shape. Below is what we'd like this to converge on so that it (a) runs on CPU **and** GPU, (b) works with all output writers, and (c) survives the switch to hybrid sigma-pressure coordinates. Inline comments have concrete suggestions, here's the overall picture.

## TL;DR

1. 🐛 **The interpolation weights are swapped** in both the linear and the log variant — results are mirrored inside the layer. See inline at L53.
2. 🐛 A target level that lands exactly on the topmost full level falls into the `else` branch with `k = 1` and indexes `σ[k-1] == σ[0]` → `BoundsError` (or silent garbage once we add `@inbounds`).
3. The interface should take an **externally allocated output field** and no `Variables`/`model` — that also removes the `n_plev <= nlayers` restriction you flagged in the PR description (it comes from borrowing `vars.scratch.grid.a`).
4. Interpolate **in pressure, not in sigma**, using `pressure(k, pₛ, coordinates)` — then hybrid sigma-pressure coordinates work with *zero* additional changes.
5. **Launch a kernel over (ij, k)**: the operation is embarrassingly parallel across columns *and* target levels, so one kernel covers CPU and GPU.
6. Make the vertical output levels an **option of the output writer**, so all writers (NetCDF/Zarr/HEALPix) get pressure-level output from one hook in the generic `output!`.

## 1. The numerics (no `Variables`, no `model`)

Interpolation and extrapolation as traits, so linear/log is one method each instead of a duplicated file half:

```julia
abstract type AbstractVerticalInterpolation <: AbstractModelComponent end

"""Linear interpolation in pressure."""
struct LinearInterpolation <: AbstractVerticalInterpolation end

"""Linear interpolation in log(pressure), the default for most variables."""
struct LogInterpolation <: AbstractVerticalInterpolation end

@inline interpolation_weight(p, p₁, p₂, ::LinearInterpolation) = (p - p₁) / (p₂ - p₁)
@inline interpolation_weight(p, p₁, p₂, ::LogInterpolation) = log(p / p₁) / log(p₂ / p₁)
```

...and the column function that the kernel calls, in pressure space:

```julia
"""$(TYPEDSIGNATURES)
Value of `in_field` in column `ij` interpolated onto pressure `p` [Pa], given the
`surface_pressure` pₛ [Pa] in that column and the vertical `coordinates` of the model."""
@propagate_inbounds function interpolate_pressure_level(
        ij, p, in_field, pₛ, coordinates, nlayers, interpolation, extrapolation,
    )
    # pressure of the top-most and bottom-most full level in this column.
    # `pressure` dispatches over the vertical coordinates, i.e. σ[k]*pₛ for SigmaCoordinates
    # and A[k]*p_ref + B[k]*pₛ for SigmaPressureCoordinates -> hybrid works out of the box
    p_top = pressure(1, pₛ, coordinates)
    p_bottom = pressure(nlayers, pₛ, coordinates)

    p <= p_top && return extrapolate_above(in_field[ij, 1], p, p_top, extrapolation)
    p >= p_bottom && return extrapolate_below(in_field[ij, nlayers], p, p_bottom, extrapolation)

    # find k with p_full(k-1) < p <= p_full(k), starting at k=2 as p > p_top is guaranteed
    k = 2
    p₂ = pressure(k, pₛ, coordinates)
    while p₂ < p
        k += 1
        p₂ = pressure(k, pₛ, coordinates)
    end
    p₁ = pressure(k - 1, pₛ, coordinates)

    w = interpolation_weight(p, p₁, p₂, interpolation)
    return (1 - w) * in_field[ij, k - 1] + w * in_field[ij, k]
end
```

Note that this is where the σ → p switch buys you hybrid coordinates: `σij = p/pₛ` is only the model's vertical coordinate for `SigmaCoordinates`. For `SigmaPressureCoordinates` the level pressures are `A[k]*p_ref + B[k]*pₛ` and the search above σ would be plain wrong. `pressure(k, pₛ, coordinates)` is already defined for both (`SpeedyWeather/src/dynamics/vertical_coordinates.jl:107` and `:258`), and there's precedent for exactly this in output code, see `output/variables/boundary_layer.jl:69`. Same for log interpolation: `log(σ/σ₁)/log(σ₂/σ₁) == log(p/p₁)/log(p₂/p₁)` holds for σ coordinates only.

## 2. Boundary conditions: hardcode now, but leave the seam

Constant extrapolation above the model top and below the lowest level is fine for a first version, but please put it behind its own dispatch so we don't have to restructure later:

```julia
abstract type AbstractVerticalExtrapolation <: AbstractModelComponent end

"""Constant (nearest-level) extrapolation outside the model levels."""
struct ConstantExtrapolation <: AbstractVerticalExtrapolation end

"""Dry-adiabatic extrapolation below the lowest model level, T(p) = T*(p/p_low)^κ,
for temperature output below the surface, e.g. down to 1000 hPa over orography."""
@kwdef struct DryAdiabaticExtrapolation{NF} <: AbstractVerticalExtrapolation
    "[OPTION] adiabatic exponent κ = R_dry/cₚ, synced with model.atmosphere.κ at initialize!"
    κ::NF = 2 / 7
end

# TODO above the model top we currently hold the top level constant; alternatives are
# NaN/missing (CF-conform "above model domain") or an isothermal continuation
@inline extrapolate_above(ξ_top, p, p_top, ::AbstractVerticalExtrapolation) = ξ_top

# below the lowest model level: constant for u, v, q, ... and dry-adiabatic for temperature
@inline extrapolate_below(ξ_bottom, p, p_bottom, ::ConstantExtrapolation) = ξ_bottom
@inline extrapolate_below(ξ_bottom, p, p_bottom, E::DryAdiabaticExtrapolation) =
    ξ_bottom * (p / p_bottom)^E.κ
```

The dry adiabat is the same one the mean sea-level pressure output already uses to bring the lowest model level down to the surface (`output/variables/dynamics.jl:169-171`, there as `σ^(-κ)`), so the two stay consistent. Geopotential height would eventually want the hypsometric equation with that extrapolated temperature — another `extrapolate_below` method when we get there.

Which extrapolation to use is a per-variable choice (temperature wants the adiabat, humidity and wind don't), so it becomes an **optional field on the output variable**, following the same `hasproperty` pattern the generic `output!` already uses for `transform`, `unscale` and `ocean_fraction`:

```julia
@kwdef mutable struct TemperatureOutput{F, E} <: AbstractOutputVariable
    name::String = "temp"
    # ... as before ...
    "[OPTION] how to extrapolate below the lowest model level when written on pressure levels"
    extrapolation::E = DryAdiabaticExtrapolation()
end

# variables without the field fall back to constant extrapolation
output_extrapolation(variable::AbstractOutputVariable) =
    hasproperty(variable, :extrapolation) ? variable.extrapolation : ConstantExtrapolation()
```

so a user can override it per variable with `TemperatureOutput(extrapolation = ConstantExtrapolation())`, and variables that don't care never gain a field.

## 3. GPU: one kernel over (ij, k)

See the suggestion on the driver function inline. `launch!(arch, RingGridWorkOrder, size(out_field), ...)` gives you `(ij, k)` per thread, `heuristic_workgroup` handles the 2D work layout, and the CPU backend of KernelAbstractions runs the same kernel — no separate CPU code path needed (compare `dynamics/vertical_advection.jl:78` for the same 2D pattern; `dynamics/geopotential.jl:85` only branches CPU/GPU because there the whole column has to be integrated sequentially in one thread).

Two things to watch: the `p` vector must live on the same architecture as the fields (`on_architecture(arch, p)` once, at output initialization, not per output step), and `coordinates` is fine to pass into the kernel as-is — both `SigmaCoordinates` and `SigmaPressureCoordinates` are `Adapt.@adapt_structure`d.

## 4. Output integration: pressure levels as an option of the writer

The vertical output grid is a property of the file being written, not of each variable — just like the horizontal `output_grid` and `interpolator` already are. So rather than wrapping variables, give the writer its vertical levels:

```julia
abstract type AbstractOutputLevels <: AbstractModelComponent end

"""Write 3D variables on the model's vertical levels (σ or hybrid σ-p). Default."""
struct ModelLevels <: AbstractOutputLevels end

"""Write 3D variables interpolated onto `pressure` levels [Pa]. Fields are $(TYPEDFIELDS)"""
@kwdef mutable struct PressureLevels{V, I} <: AbstractOutputLevels
    "[OPTION] pressure levels [Pa], ascending"
    pressure::V = [100, 200, 300, 500, 700, 850, 925, 1000] .* 100

    "[OPTION] interpolate linearly in p or in log(p)?"
    interpolation::I = LogInterpolation()

    "[DERIVED] pressure levels moved onto the model's architecture, set in initialize!"
    pressure_arch::Union{V, Nothing} = nothing

    "[DERIVED] scratch field (npoints_model, npressure) on the model grid, set in initialize!"
    scratch::Union{Field3D, Nothing} = nothing
end

get_nlayers(::ModelLevels, SG::SpectralGrid) = SG.nlayers
get_nlayers(levels::PressureLevels, ::SpectralGrid) = length(levels.pressure)
```

used as

```julia
output = NetCDFOutput(spectral_grid, PrimitiveWet, levels = PressureLevels(pressure = [850, 500, 200] .* 100))
```

Then, in decreasing order of how much they touch:

**(a) The generic `output!`** (`output/writers/general.jl:165`) gains one hook between reading the field and moving it to CPU, so every writer and every ordinary 3D variable is covered at once:

```julia
    ori = has_step ? get_prognostic_step(ori, ts, output) : ori
    ori = interpolate_levels!(output.levels, ori, variable, simulation)   # <-- new, no-op on model levels
    raw = on_architecture(CPU(), ori)
    interpolate_output!(output, var, raw)
```

```julia
interpolate_levels!(::ModelLevels, field, variable, simulation) = field    # no-op, model levels

function interpolate_levels!(levels::PressureLevels, field, variable, simulation)
    is3D(variable) && !is_land(variable) || return field     # only atmospheric 3D variables
    (; model) = simulation
    pₛ = simulation.variables.parameterizations.surface_pressure
    interpolate_pressure_levels!(
        levels.scratch, field, pₛ, levels.pressure_arch,
        model.geometry.vertical_coordinates, levels.interpolation,
        output_extrapolation(variable),
    )
    return levels.scratch
end
```

Note this happens on the model grid and the model's architecture, *before* the copy to CPU and the horizontal interpolation — see the follow-up comment below for why that order, and what a fused 3D interpolator would change.

**(b) Allocation** in `initialize!(core::OutputWriterCore, output, model)` (`output_writer_core.jl:56`), which every writer already calls, so one line covers all of them:

```julia
    hasproperty(output, :levels) && initialize!(output.levels, model)
```

with `initialize!(::ModelLevels, model) = nothing` and the `PressureLevels` method allocating `scratch = Field(NF, model.spectral_grid.grid, length(pressure))` on the model architecture, `pressure_arch = on_architecture(architecture(model.spectral_grid), pressure)`, and syncing `κ` of any `DryAdiabaticExtrapolation` from `model.atmosphere`. Two buffers in total for the whole file — `levels.scratch` plus the writer's existing `field3D` — independent of how many variables are written.

**(c) The writers' constructors**, where `field3D = Field(output_NF, output_grid, nlayers)` becomes `Field(output_NF, output_grid, get_nlayers(levels, SG))` (3 lines, one per writer).

**(d) The dimension name and its coordinate values.** `define_variable!` asks `vertical_dimension(var)` for the dimension name, which should now follow the writer for atmospheric variables while custom variables keep their own ("soil_layer", Terrarium's "soil_depth"):

```julia
vertical_dimension(::ModelLevels) = "layer"
vertical_dimension(::PressureLevels) = "pressure"

# writers call this 2-arg version; the 1-arg method stays the hook for custom dimensions
vertical_dimension(output::AbstractOutput, variable::AbstractOutputVariable) =
    vertical_dimension(variable) == "layer" ? vertical_dimension(output.levels) : vertical_dimension(variable)
```

(the string comparison is a bit inelegant, a `on_model_layers(variable)` trait would be cleaner — either way it's backwards compatible). And in each writer's `initialize!`, the hardcoded `defVar(dataset, "layer", σ, ...)` becomes a `define_vertical_coordinate!(dest, levels)` that writes σ or pressure in hPa with the right `units`/`long_name`/`positive = "down"` attributes — backend-agnostic via the existing `define_coordinate!`.

Variables that override `output!` entirely (mslp, the 10 m wind diagnostics, Terrarium) are untouched by all of this, as they never go through the generic path.

## 5. Housekeeping

- The file isn't `include`d anywhere, so none of it is compiled yet — add it to `SpeedyWeather/src/SpeedyWeather.jl` next to the other `output/` includes, and `export` the user-facing bits.
- No tests yet. A good minimal set: (i) interpolating a field that is linear in p onto p levels reproduces it exactly, (ii) target levels that coincide with model levels return the model level values exactly (that's the edge case that currently throws), (iii) CPU vs GPU agreement, (iv) a `SigmaPressureCoordinates` column once the maths is in pressure space.
- Every PR gets a line in `CHANGELOG.md` under `## Unreleased`.
- Please run [Runic.jl](https://github.com/fredrikekre/Runic.jl) over the file (CI checks this): spaces after commas, no trailing whitespace, aligned `end`s, and a newline at EOF.
- Docstrings: use `"""$(TYPEDSIGNATURES) ..."""` so signatures render in the docs, and document units ([Pa] everywhere, presumably).

A reasonable way to split this: section 1-3 (the numerics + kernel, which is what this PR already is) first, section 4 as a follow-up PR once the interpolation itself is tested. Happy to go through any of it in more detail — and don't be put off by the length, the core is really just "swap the two weights, interpolate in p, take the output field as an argument".

SpeedyWeather/src/output/vertical_interpolation.jl
```diff
@@ -0,0 +1,148 @@
...
+
+        if σij < σ[1]
+            out_field[ij,pk] = in_field[ij,1]
+
+        elseif σij >= σ[end]
+            out_field[ij,pk] = in_field[ij,end]
```
milankl-claude: Two things here.

**Bug:** the `<` is strict, so a target level that coincides *exactly* with the top full level (`σij == σ[1]`, which happens for any `p` that lands on a model level, and exactly is what a user hits when they ask for a level that matches) falls through to the `else` branch, where the `while` loop doesn't advance, `k` stays `1`, and `σ[k-1] == σ[0]` throws a `BoundsError` — or reads out of bounds silently once we add `@inbounds` for the GPU kernel. `<=` fixes it and costs nothing.

**Boundary conditions:** hardcoding constant extrapolation is fine for now, but please mark both ends so we know where to extend. Below the lowest model level, temperature should eventually descend dry-adiabatically (`T(p) = T[end] * (p/p_bottom)^κ` with `κ = model.atmosphere.κ`) so that 1000 hPa over orography is meaningful — that's the same adiabat the mslp output uses in `output/variables/dynamics.jl:169-171`. In the review summary I sketched putting this behind an `extrapolate_above`/`extrapolate_below` dispatch, selected per output variable through an optional `extrapolation` field (`DryAdiabaticExtrapolation()` on `TemperatureOutput`, `ConstantExtrapolation()` for anything without the field, the same `hasproperty` pattern `output!` already uses for `transform` and `unscale`). That way the kernel doesn't need to change when we add it, and a user can override it per variable.

```suggestion
        # TODO above the model top: constant for now, could be NaN/missing or isothermal
        if σij <= σ[1]
            out_field[ij, pk] = in_field[ij, 1]

        # TODO below the lowest model level: constant for now, temperature should descend
        # dry-adiabatically here, T = in_field[ij, end] * (σij/σ[end])^κ, see mslp output
        elseif σij >= σ[end]
            out_field[ij, pk] = in_field[ij, end]
```

SpeedyWeather/src/output/vertical_interpolation.jl
```diff
@@ -0,0 +1,148 @@
...
+
+        else
+            k = 1
+            while (σ[k] < σij) k += 1 end
```
milankl-claude: Start the search at `k = 2`: after the `<=` fix above, `σij > σ[1]` is guaranteed here, so `k-1 >= 1` always holds and the `σ[0]` access is impossible by construction.

Also worth noting for later: the search restarts from the top for every target level, i.e. `O(n_plev * nlayers)` per column. Since `p` is sorted ascending you can walk `k` monotonically across the target levels — though that only pays off in the serialised, one-thread-per-column variant. With one thread per `(ij, k)` (my suggestion above) each thread does its own short search and that's fine; `nlayers` is small.

```suggestion
            k = 2
            while σ[k] < σij
                k += 1
            end
```

SpeedyWeather/src/output/vertical_interpolation.jl
```diff
@@ -0,0 +1,148 @@
...
+                    
+            Δσ, w1, w2 = interpolation_weights(σij,σ[k-1],σ[k])
+            out_field[ij,pk] = ( w1*in_field[ij,k-1] + w2*in_field[ij,k] ) / Δσ
```
milankl-claude: 🐛 **The weights are swapped.** `interpolation_weights` returns `(Δσ, σ-σ₁, σ₂-σ)`, so `w1` is the distance to the *lower* level `σ₁` and therefore has to weight the value at `σ₂` (i.e. `in_field[ij, k]`), and vice versa. As written, the interpolation is mirrored inside the layer: it returns `σ₁`'s value at `σ₂` and vice versa.

Quick check with `σ = [0.2, 0.6]`, values `[10, 20]` and a target at `σij = 0.3`: the answer should be `12.5`, the current code gives `(0.1*10 + 0.3*20)/0.4 = 17.5`, which is the value at `σ = 0.5`.

Same for the log variant at L128-129.

```suggestion
            Δσ, w1, w2 = interpolation_weights(σij, σ[k - 1], σ[k])
            out_field[ij, pk] = (w2 * in_field[ij, k - 1] + w1 * in_field[ij, k]) / Δσ
```

(That's the minimal fix. If you adopt the single normalised weight from the review summary it becomes `w = interpolation_weight(p, p₁, p₂, interpolation); (1 - w) * in_field[ij, k - 1] + w * in_field[ij, k]`, which is one multiplication cheaper and harder to get backwards.)

SpeedyWeather/src/output/vertical_interpolation.jl
```diff
@@ -0,0 +1,148 @@
...
+
+returns:
+
+Δσ  = σ₂ - σ₁
+Δσ₁ = σ  - σ₁  
+Δσ₂ = σ₂ - σ
+"""
+@inline function interpolation_weights(σ,σ₁,σ₂)
+    return σ₂ - σ₁, σ - σ₁, σ₂ - σ
+end
```
milankl-claude: Consider returning a single normalised weight instead of three numbers the caller has to recombine correctly — that's exactly the recombination that went wrong above:

```julia
@inline interpolation_weight(p, p₁, p₂, ::LinearInterpolation) = (p - p₁) / (p₂ - p₁)
@inline interpolation_weight(p, p₁, p₂, ::LogInterpolation) = log(p / p₁) / log(p₂ / p₁)
```

with the caller `(1 - w) * ξ₁ + w * ξ₂`. The `::LinearInterpolation`/`::LogInterpolation` argument is what lets the linear and log variants share one kernel, see below.

SpeedyWeather/src/output/vertical_interpolation.jl
```diff
@@ -0,0 +1,148 @@
...
+            while (σ[k] < σij) k += 1 end
+                    
+            Δσ, w1, w2 = interpolation_log_weights(σij,σ[k-1],σ[k])
+            out_field[ij,pk] = ( w1*in_field[ij,k-1] + w2*in_field[ij,k] ) / Δσ
+
+            end
+    end
+    
+    return out_field
+end
```
milankl-claude: This is a near-verbatim copy of the linear version, and the two will drift apart (the weight bug above is already duplicated here). Everything except the weight formula is identical, so a trait argument collapses the two into one implementation:

```julia
interpolate_pressure_levels!(out, in, pₛ, p, coordinates, LogInterpolation())
interpolate_pressure_levels!(out, in, pₛ, p, coordinates, LinearInterpolation())
```

That also keeps the choice configurable per output variable (log in p is the usual choice for most fields, linear for geopotential height/temperature is defensible) without a second set of exported function names.

While we're here: for hybrid coordinates the log variant must interpolate in `log(p)`, not in `log(σ)` — see my comment at L41.

SpeedyWeather/src/output/vertical_interpolation.jl
```diff
@@ -0,0 +1,148 @@
...
+
+returns:
+
+Δσ  = log(σ₂/σ₁) = log(σ₂) - log(σ₁)
+Δσ₁ = log(σ/σ₁)  = log(σ)  - log(σ₁)
+Δσ₂ = log(σ₂/σ)  = log(σ₂) - log(σ)
+"""
+@inline function interpolation_log_weights(σ,σ₁,σ₂)
+    return log(σ₂/σ₁), log(σ/σ₁), log(σ₂/σ)
+end
```
milankl-claude: Housekeeping for this file:

- It isn't `include`d in `SpeedyWeather/src/SpeedyWeather.jl`, so none of it is compiled or tested yet. Add it next to the other `output/` includes and export the user-facing names.
- Missing newline at end of file, and please run [Runic.jl](https://github.com/fredrikekre/Runic.jl) (the repo's code style, see the badge in the README — there's no CI check for it, so it's on us to remember): spaces after commas, 4-space indent, no stray indentation on `end`.
- Docstrings: `"""$(TYPEDSIGNATURES) ..."""` so signatures render in the docs, and state units — everything here is [Pa], which is worth saying explicitly since users will reach for hPa.
- `SpeedyWeather.` prefixes aren't needed inside the package itself.
- Tests + a `CHANGELOG.md` entry under `## Unreleased` before this comes out of draft.

milankl-claude:
## Follow-up: separate vertical + horizontal interpolation, or one 3D interpolator?

@milankl and I went back and forth on whether the vertical (σ/hybrid → p) and horizontal (model grid → output grid) interpolation should stay two passes or be fused into a single 3D interpolator shared by all output writers. Writing up the conclusion here so it's on the record for this PR and the follow-ups. **Short version: two separate passes for now, composed behind one object, with fusion as a later optimisation behind the same interface.**

### Why vertical first, then horizontal

The two orderings are not equivalent:

- **Vertical first, on the model grid**: each column is converted to pressure levels using *its own* surface pressure, then the p-level field is interpolated horizontally. 
- **Horizontal first**: you would need `pₛ` interpolated onto the output grid, which smooths the orography, and the below-ground extrapolation then no longer matches the model's own surface. 

So vertical-first is the better-defined order (and the conventional one — it's what ECMWF does when producing pressure-level data from model levels). It is also usually cheaper: if you output fewer pressure levels than the model has layers, the horizontal interpolation and the GPU→CPU copy both shrink accordingly. Worth noting the reverse case though — asking for e.g. 37 ERA5-style levels out of 8 model layers makes vertical-first do *more* horizontal work, not less.

### What a fused 3D interpolator would and wouldn't change

- **It wouldn't change the numbers.** A fused kernel that, for each output point, vertically interpolates each of the 4 surrounding source columns (each with its own `pₛ`) and then anvil-averages them is *mathematically identical* to vertical-then-horizontal: both operations are linear and the horizontal weights are the same at every level. Fusion is purely an implementation choice, not a different answer.
- **It can't be a single precomputed weight matrix.** The horizontal weights are computed once for a fixed grid pair and reused (`AnvilLocator`). The vertical weights depend on `pₛ`, which changes every output step, so they must be computed in the kernel. A "3D interpolator" is therefore a *composite* — precomputed horizontal locator + target levels + vertical coordinates + `pₛ` per call — not a fused stencil.
- **What it would buy**: no intermediate `(npoints_model, nlevels)` buffer; a single kernel launch instead of one per level (the 3D `interpolate!` loops over layers, `RingGrids/src/interpolation.jl:412`); and the whole chain could stay on the GPU.
- **What it would cost**: each source column would be vertically interpolated ~4× (once per output point referencing it) instead of once. The vertical work is a short search plus two multiplies, so this may well be cheaper than the extra memory traffic — but that's a guess, and not one worth acting on without a benchmark.

One thing worth correcting from my earlier comment: the horizontal interpolation is **not** inherently CPU-bound. `_interpolate!` already launches a KernelAbstractions kernel (`RingGrids/src/interpolation.jl:318`) and `AnvilLocator` is `Adapt`-ed with an `on_architecture` method. Output runs on CPU only because the writers allocate their scratch fields on a CPU output grid and `output!` does `on_architecture(CPU(), ori)` first (`output/writers/general.jl:180`). Moving output interpolation to the GPU is a separate, independent piece of work — and a bigger win than the vertical kernel alone.

### Conclusion

Build the vertical step as its own GPU-capable operation (this PR), and attach it to the output writer as a composite alongside the existing horizontal interpolator (follow-up). That is a strictly smaller change, leaves `AnvilLocator` untouched, keeps the "no horizontal interpolation needed" case working (`HEALPixOutput` sets `interpolator = nothing` when the model already runs on the output grid, `healpix_output.jl:120`), and shares one set of buffers across all output variables regardless of how many are written. Fusing the two into a single kernel later is then an optimisation behind an unchanged interface.

@henrikauestad — none of this affects the code in this PR: the column maths and the kernel are identical under all of the above. It only decides where the result gets plugged in.

henrikauestad:
All in all, it looks good to me. One thought: personally, I would like to set subsurface grid points to NaN instead of extrapolating. Those NaNs would probably propagate horizontally in the horizontal interpolator if the vertical interpolation is done before the horizontal interpolation. Would it be worth thinking about a route for masking subsurface points already now?

milankl-claude:
@henrikauestad Good instinct, and yes — NaN below ground is just another `extrapolation`, no interface change needed. But your concern about it spreading through the horizontal interpolation is well founded, and it's actually a bit worse than you describe, so I'd suggest doing both things.

### NaN as an extrapolation

Two thresholds matter here and they're not the same one:

- `p > p_bottom` — below the *lowest full model level*. 
- `p > pₛ` — below the *surface*.

Everything between them (e.g. 1000 hPa over a low-lying grid point whose lowest full level sits at ~985 hPa) is above ground and should be extrapolated, not masked. So the mask has to key off `pₛ`, not off the lowest level — which the kernel has anyway. Adding `pₛ` to the `extrapolate_below` signature makes this a three-line addition:

```julia
"""Mask values below the surface with `missing_value`, extrapolate between the lowest
model level and the surface with `above_surface`."""
@kwdef struct SubsurfaceMask{E, NF} <: AbstractVerticalExtrapolation
    above_surface::E = ConstantExtrapolation()
    missing_value::NF = NaN
end

@inline extrapolate_below(ξ, p, p_bottom, pₛ, E::SubsurfaceMask) =
    p > pₛ ? convert(typeof(ξ), E.missing_value) :
        extrapolate_below(ξ, p, p_bottom, pₛ, E.above_surface)
```

and then, per variable:

```julia
add!(output, TemperatureOutput(extrapolation = SubsurfaceMask()))

# or: dry adiabat between the lowest model level and the surface, NaN below it
add!(output, TemperatureOutput(extrapolation = SubsurfaceMask(above_surface = DryAdiabaticExtrapolation())))
```

Use `variable.missing_value` (which every output variable already has, defaulting to `NaN`) rather than a hardcoded `NaN`, because it's already written to the file as `_FillValue` — the mask is then self-describing to whoever reads the netCDF/Zarr.

### The propagation problem

You're right that the NaNs spread, and note that weights don't save you: `anvil_average` is a bilinear blend, `a + (b - a) * Δab`, and `NaN * 0 == NaN` in IEEE. So a single subsurface point anywhere in the 4-point stencil poisons the output point *even when its weight is zero* — i.e. even when the output point coincides exactly with an above-ground source point. The mask dilates by the full stencil, not by "a bit near the edges".

Two ways out:

**1. Mask on the output grid, after the horizontal interpolation.** Interpolate finite (extrapolated) values, then blank out `p > pₛ` using `pₛ` interpolated onto the output grid — one extra `Field2D` and one `interpolate!` per output step, reusing the existing interpolator:

```julia
# once per output step, in the writer
interpolate_output!(output, output.surface_pressure, on_architecture(CPU(), pₛ))

# after interpolate_output! for each 3D variable
function mask_subsurface!(field, pₛ, pressure, missing_value)
    for k in eachindex(pressure), ij in eachgridpoint(field)
        field[ij, k] = pressure[k] > pₛ[ij] ? missing_value : field[ij, k]
    end
end
```

This keeps NaNs out of the interpolator entirely, gives a crisp mask, and has the nice property that the mask edge agrees with the `pres` field written into the same file rather than with the model grid the reader never sees.

**2. Accept the dilation** and use `SubsurfaceMask` as above. Simpler, no writer changes, but the masked region grows by one stencil and the boundary depends on the output grid resolution in a way that's awkward to explain to a user.

A third option — making `anvil_average` NaN-aware by renormalising over the finite stencil points — I'd avoid: it slows down RingGrids' hot path for everyone, and the values it produces are horizontal extrapolations from above-ground neighbours, which is arguably worse than a vertical adiabat.

So my suggestion: build `SubsurfaceMask` now (it's three lines and it makes the point that the `extrapolation` field is genuinely open-ended), and put the output-grid masking in as a `mask_subsurface::Bool` option on `PressureLevels` when the output-writer integration lands, defaulting to whichever @milankl prefers. The two are composable — you'd just not combine masking on the output grid with an extrapolation that already NaNs.

milankl:
Looks like we converged to something here, would you like to code this up or do you want Claude to do it? We could either update the pull request here (may need another external branch as only I would have write access to your branch, Claude doesn't) or let Claude open a new one implementing everything.

henrikauestad:
Yes. Happy to let Claude pull the weight and the request. Thanks!

milankl-claude:
@henrikauestad I've put the numerics part of the review into a PR **against your branch**: henrikauestad/SpeedyWeather.jl#1. Merging it updates this PR with the commit, so #1256 stays yours — I can't push to your branch directly, hence the detour. Take it wholesale, cherry-pick, or ignore it and use it as a reference, whatever suits you.

It covers sections 1-3 of the review: both bug fixes (with tests that fail if the fixes are reverted — I checked), interpolation in pressure so hybrid coordinates come for free, the externally allocated output field (which lifts the `n_plev <= nlayers` restriction), one implementation shared by the linear and log variants, the `(ij, k)` kernel for CPU/GPU, and `SubsurfaceMask` for the NaN masking you asked about. Also `include`s the file in `SpeedyWeather.jl`, which hadn't happened yet. 26 tests, passing locally.

Two notes:

- I renamed `LinearInterpolation`/`LogInterpolation` from my review comment to `LinearInPressure`/`LinearInLogPressure`, because `LinearInterpolation` would clash with `Interpolations.jl` for anyone loading both packages. Not attached to the names.
- Correction to my housekeeping comment above: I said CI enforces Runic formatting. It doesn't — there's no format check in `.github/workflows/`, it's the documented code style (README badge, CLAUDE.md) but unenforced. I've run Runic over the new files anyway.

The output-writer integration (section 4) is still open and probably better as its own PR once this lands.

milankl:
Switching CI on but submodules off (none affected) as well as Enzyme/Reactant (no changes to hot code for now)

milankl:
Looks like vertical interpolation CI adds 8s, easy!

```julia
                                                   │   Test   │   Init   │ Compile │ ──────────────── CPU ──────────────── │
  Test                                    (Worker) │ time (s) │ time (s) │   (%)   │ GC (s) │ GC % │ Alloc (MB) │ RSS (MB) │
  output/vertical_interpolation                (1) │         started at 2026-09-16T13:22:03.844
...
  output/vertical_interpolation                (1) │     7.99 │     7.82 │   95.07 │   0.58 │  7.3 │     679.12 │   822.50 │
...

```

PR #1257, opened 2026-09-16
Output on pressure layers

### LINKED ISSUE #268: Interpolation on pressure levels

milankl-claude:
Follow-up to #1256, which added the vertical interpolation itself. This adds the documentation it didn't have and wires pressure-level output into the output writers.

## User interface

Model levels stay the default; pressure-level output is opt-in with a `levels` keyword argument on the output writers:

```julia
output = NetCDFOutput(spectral_grid, PrimitiveWet, levels = PressureLevels([850, 500, 200] .* 100))
```

`PressureLevels()` without arguments writes 50, 100, 200, 300, 500, 700, 850, 925 and 1000 hPa. The vertical dimension in the file is called `pressure` and holds the levels in hPa. Every 3D atmospheric variable is written on it; 2D variables and soil variables (on `soil_layer`) are unaffected. Works with `NetCDFOutput`, `ZarrOutput` and `HEALPixOutput`.

## What's in here

**The levels types** (`output/writers/output_levels.jl`)

- `AbstractOutputLevels` with `ModelLevels()` (default, a no-op) and `PressureLevels(pressure; interpolation)`.
- `PressureLevels` holds the levels and, allocated at `initialize!`, the levels on the model's architecture plus one scratch field on the model grid. Two buffers for the whole file, shared by every variable, regardless of how many are written.
- `get_nlayers(levels, spectral_grid)` sizes the writers' 3D scratch field on the output grid.

**Writer integration**

- `levels` keyword on all three writers, defaulting to `ModelLevels()`.
- One hook in the generic `output!` (`writers/general.jl`), between selecting the time step and the copy to CPU, so all writers and all ordinary 3D variables are covered at once. The vertical interpolation therefore runs on the model grid and the model's architecture — on the GPU for a GPU simulation — before the horizontal interpolation, per the [ordering discussion](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1256#issuecomment-5679019257) in #1256.
- Allocation in `initialize!(::OutputWriterCore, output, model)`, which every writer already calls, so one place covers all of them. Writers without a `levels` field keep working via `output_levels(output)`.
- `vertical_dimension(output, variable)` picks the writer's dimension for variables on the model's atmospheric layers, while custom dimensions (`soil_layer`, Terrarium's `soil_depth`) are untouched. The hardcoded sigma coordinate in each writer's `initialize!` is replaced by `define_vertical_coordinate!(dest, levels, model)`, backend-agnostic via the existing `define_coordinate!`.

**Extrapolation below the lowest model level**

Per variable, through an optional `extrapolation` field, the same `hasproperty` pattern `output!` already uses for `transform` and `unscale`:

- `TemperatureOutput` defaults to `DryAdiabaticExtrapolation()`, everything else to constant, as requested.
- κ = R_dry/cₚ is taken from the model's atmosphere at `initialize!`, so the adiabat matches the model. It is marked `[DERIVED]` on the struct rather than being a user option.
- `SubsurfaceMask(above_surface = ...)` is available for masking below the surface, but is not a default anywhere.

**Docs**

- New page `docs/src/vertical_interpolation.md` (Advanced): the interface from #1256, the interpolation methods with their weight formulas, the extrapolation methods, and a runnable example interpolating temperature onto 850/500/200 hPa and down to 1000 hPa with subsurface masking.
- New [Output levels](https://speedyweather.github.io/SpeedyWeather.jl/dev/output/) section in `docs/src/output.md`, next to Output grid, covering the `levels` keyword and the extrapolation choices.

**Not exported.** `interpolate_pressure_levels!` and the interpolation/extrapolation types stay internal for now (this PR also removes the exports #1256 added); only `ModelLevels` and `PressureLevels` are exported, as they are the handle for the `levels` keyword. The docs use `SpeedyWeather.`-qualified names accordingly.

## Tests

`test/output/pressure_level_output.jl` (38 tests, passing locally):

- `PressureLevels` construction, `get_nlayers`, the scratch field sizing of `field3D` vs the untouched `field3Dland`, and the monotonic/positive assertions on the levels.
- Extrapolation defaults per variable and the κ sync from the atmosphere, including through a `SubsurfaceMask`.
- **NetCDF round trip**: a `PrimitiveWetModel` on a full grid so the horizontal interpolation is an exact copy, then reading the file back with `NCDataset` and checking that the `pressure` dimension exists with the right values and units, that `layer` is gone, that temp/u/v/vor/humid are on `pressure` while `st` stays on `soil_layer` and `mslp` stays 2D, and that the values match a direct `interpolate_pressure_levels!` on the same state (`rtol = 1e-5`), are finite and decrease with height.
- Default unchanged: a writer without `levels` still writes `layer` with the sigma values.

Zarr and HEALPix pressure-level tests are appended to their existing test files (5 and 4 tests, passing), rather than living with the NetCDF ones — see below.

## A note on local test runs

`netcdf_output.jl` and `zarr_output.jl` cannot be `include`d in the same Julia process on my machine (Julia 1.13.0): the second one segfaults in the GC. This is **not** related to this PR — the same happens with both files unmodified on `main`, and `netcdf_output.jl` alone aborted in 1 of 3 runs with an unmodified tree. CI never hits it because `ParallelTestRunner` gives each test file its own worker. It's the reason the Zarr/HEALPix tests here live in their own files. Might be worth a look independently if anyone else sees it.

## Follow-ups not in here

- A `mask_subsurface` option that masks on the *output* grid after the horizontal interpolation, for a crisp mask without the stencil dilation that `SubsurfaceMask` has (see the [discussion](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1256#issuecomment-5682979199) in #1256).
- Above the model top everything is still held constant, marked with a TODO.
- A GPU test for the output path.


milankl-claude:
One thing worth flagging separately, since #1258 just moved CI to Julia 1.13: while testing this PR I hit repeated **GC aborts** (`GC error (probable corruption)` / segfaults in `gc_mark_outrefs`) running the output tests on Julia 1.13.0 locally.

It is not caused by this PR — I checked with the changes stashed:

- `test/output/netcdf_output.jl` on a **pristine** tree aborted in 3 of 3 runs (exit 134/134/139). With this PR's changes in the tree it passed 1 of 2 — i.e. it's nondeterministic either way, not correlated with the change.
- `test/output/zarr_output.jl` on a pristine tree aborts as well.
- `include`ing `netcdf_output.jl` and `zarr_output.jl` in the *same* process aborts reliably, on a pristine tree too. `ParallelTestRunner` gives each file its own worker, so CI doesn't exercise that combination — it's the reason I put the pressure-level Zarr/HEALPix tests into their existing files rather than alongside the NetCDF ones.

The crashes are in the GC at ordinary allocation sites (e.g. `get_londlatds` during orography initialization), so my guess is a Julia 1.13.0 GC issue or a stale precompile cache in my depot rather than anything in SpeedyWeather. It may well be local to my machine. Mentioning it in case CI on 1.13 starts showing unexplained aborts in the output tests — that would not be from this PR.

milankl:
@henrikauestad I'm a bit undecided about terminology here, I know that ECMWF/ERA slang is to say "level" when you mean any type of discrete vertical steps. However, "level" is quite ambiguous and can mean any discrete steps really, not just the vertical e.g. level 2 in a video game, compression level or high-level overview. The word layer instead clearly has a vertical connotation like layers of sediment, layers of coat, etc. Hence we chose a while ago "nlayers" instead of the ECMWF slang of "nlev" but also because it avoided an unnecessary abbreviation. So what do you think about changing function/type names here to

- `PressureLayers`, `interpolate_pressure_layers!` etc
- `ModelLayer` instead of `ModelLevel` (ECMWF term that is super ambiguous in my opinion, has no clear connection to the vertical, could be different levels of complexity, or whatever else)

Would you be happy with that?

SpeedyWeather/src/output/writers/output_levels.jl
```diff
@@ -0,0 +1,169 @@
...
+"""Write 3D variables on the model's vertical levels, sigma or hybrid sigma-pressure
+depending on `model.geometry.vertical_coordinates`. Default for all output writers."""
+struct ModelLevels <: AbstractOutputLevels end
+
+"""Write 3D atmospheric variables interpolated onto `pressure` levels [Pa] instead of the
+model's vertical levels, see [Vertical interpolation onto pressure levels](@ref
+vertical_interpolation). Pass to an output writer as
+`NetCDFOutput(spectral_grid, levels = PressureLevels())`. Land (soil) variables and 2D
+variables are unaffected. Fields are $(TYPEDFIELDS)"""
+@kwdef mutable struct PressureLevels{V, I} <: AbstractOutputLevels
```
milankl: Can this not be immutable, with pressure directly on `arch` allocated with a `::SpectralGrid` constructor like done for most other model components?

SpeedyWeather/src/output/writers/output_levels.jl
```diff
@@ -0,0 +1,169 @@
...
+    allocated at initialize!"
+    pressure_on_arch::Union{AbstractVector, Nothing} = nothing
+
+    "[DERIVED] scratch field (npoints, npressure) on the model grid to interpolate into
+    before the horizontal interpolation onto the output grid, allocated at initialize!"
+    scratch::Union{AbstractField, Nothing} = nothing
+end
+
+# convenience: PressureLevels([850, 500, 200] .* 100)
+PressureLevels(pressure::AbstractVector; kwargs...) = PressureLevels(; pressure, kwargs...)
```
milankl: add a `::SpectralGrid` constructor which allocates `scratch` on the right architecture

SpeedyWeather/src/output/writers/output_levels.jl
```diff
@@ -0,0 +1,169 @@
...
+
+    (; NF, grid) = model.spectral_grid
+    arch = architecture(grid)
+
+    # in the model's number format and on its architecture so that the interpolation
+    # kernel can run on the GPU without scalar indexing into a host vector
+    levels.pressure_on_arch = on_architecture(arch, NF.(levels.pressure))
+    levels.scratch = Field(NF, grid, length(levels.pressure))
+    return nothing
+end
```
milankl: this would not be needed with the constructor mentioned above

SpeedyWeather/src/output/writers/output_levels.jl
```diff
@@ -0,0 +1,169 @@
...
+"""$(TYPEDSIGNATURES)
+Number of vertical levels that 3D atmospheric variables are written on, used by the output
+writers to allocate their 3D scratch field on the output grid."""
+get_nlayers(::ModelLevels, SG::SpectralGrid) = SG.nlayers
+get_nlayers(levels::PressureLevels, ::SpectralGrid) = length(levels.pressure)
+
+"""$(TYPEDSIGNATURES)
+Name of the vertical dimension in the output file or store for 3D atmospheric variables."""
+vertical_dimension(::ModelLevels) = "layer"
+vertical_dimension(::PressureLevels) = "pressure"
```
milankl: Can this be called `vertical_dimension_name` given it returns a string representing the name used for the dimension?

SpeedyWeather/src/output/writers/netcdf_output.jl
```diff
@@ -219,16 +222,18 @@ Base.close(output::NetCDFOutput) = NCDatasets.close(output.netcdf_file)
 function define_variable!(
         dataset::NCDataset,
         var::AbstractOutputVariable,
-        output_NF::Type{<:AbstractFloat} = DEFAULT_OUTPUT_NF,
+        output_NF::Type{<:AbstractFloat} = DEFAULT_OUTPUT_NF;
+        vertical_dim::String = vertical_dimension(var),
```
milankl: also here call it "*_name" ?

SpeedyWeather/src/output/variables/dynamics.jl
```diff
@@ -214,6 +214,11 @@ Fields are: $(TYPEDFIELDS)"""
     shuffle::Bool = true
     keepbits::Int = 10
     transform::F = (x) -> x - 273.15     # K to ˚C
+
+    "[OPTION] how to extrapolate below the lowest model level when written on pressure
+    levels, dry-adiabatic descent so that e.g. 1000 hPa below the lowest model level is
+    sensible. Only used with `PressureLevels`, see [`AbstractVerticalExtrapolation`](@ref)"
+    extrapolation::AbstractVerticalExtrapolation = DryAdiabaticExtrapolation()
```
milankl: make this parametric?

milankl (commented):
Looks good to me, before we do more extensive testing, essentially two things (1) layers instead of levels as preferred terminology (2) the construction of PressureLevels (...Layers) for an immutable struct with no abstract field types.

SpeedyWeather/src/output/writers/output_levels.jl
milankl-claude: Yes — `PressureLayers` is now an immutable struct with concrete field types, constructed from a `SpectralGrid`:

```julia
struct PressureLayers{V, I, F} <: AbstractOutputLayers
    pressure::V         # [Pa], in the model's number format, on its architecture
    interpolation::I
    scratch::F          # Field(NF, grid, npressure)
end
```

`pressure` is the vector on `arch` directly, so the separate `pressure_on_arch` field is gone.

SpeedyWeather/src/output/writers/output_levels.jl
milankl-claude: Done:

```julia
PressureLayers(spectral_grid; pressure = DEFAULT_PRESSURE_LAYERS, interpolation = LinearInLogPressure())
PressureLayers(spectral_grid, [850, 500, 200] .* 100)    # positional convenience
```

allocating both `pressure` on `SG.architecture` in `SG.NF` and `scratch = Field(NF, SG.grid, length(pressure))`. Usage is now `NetCDFOutput(spectral_grid, PrimitiveWet, layers = PressureLayers(spectral_grid))`.

SpeedyWeather/src/output/writers/output_levels.jl
milankl-claude: Removed — the monotonic/positive assertions moved into the constructor and there is nothing left to allocate. `initialize!(::OutputWriterCore, output, model)` now only calls `sync_extrapolations!` to take κ from the model's atmosphere.

SpeedyWeather/src/output/writers/output_levels.jl
milankl-claude: Renamed to `vertical_dimension_name`, for the 1-argument method on `main` and the Terrarium extension's method too.

SpeedyWeather/src/output/writers/netcdf_output.jl
milankl-claude: Yes, it is `vertical_dim_name` now.

SpeedyWeather/src/output/variables/dynamics.jl
milankl-claude: Done, `TemperatureOutput{F, E}` with `extrapolation::E`. To keep that field type-stable, `sync_extrapolation` now preserves the number format of the extrapolation (`DryAdiabaticExtrapolation{NF}` in, `DryAdiabaticExtrapolation{NF}` out) and `extrapolate_below` converts κ to the number format of the field it works on — which also keeps an NF=Float32 kernel from doing a Float64 `^`.

One consequence: `output.variables[:temp].extrapolation = SubsurfaceMask(...)` no longer type-checks, a different extrapolation has to go through the constructor, `add!(output, TemperatureOutput(extrapolation = ...))`. That is what the docs already showed. I also made the `layers` field of the three writers parametric rather than `::AbstractOutputLayers`.

milankl-claude:
All six inline comments and the terminology point are addressed in 9d3ca55, thanks.

## 1. Layers, not levels

I agree with the reasoning — "layer" has the vertical connotation, "level" does not, and `nlayers` is already the house style. Renamed throughout:

| before | after |
| --- | --- |
| `PressureLevels` | `PressureLayers` |
| `ModelLevels` | `ModelLayers` |
| `AbstractOutputLevels` | `AbstractOutputLayers` |
| `interpolate_pressure_levels!` | `interpolate_pressure_layers!` |
| `interpolate_levels!`, `output_levels` | `interpolate_layers!`, `output_layers` |
| `levels = ...` keyword | `layers = ...` |

plus the prose in the two doc pages and the file names (`output_layers.jl`, `test/output/pressure_layer_output.jl`). `σ_levels_full`, `compression_level` and "mean sea-level pressure" are untouched. @henrikauestad if you would rather keep ECMWF's wording, say so — it is a mechanical rename either way.

## 2. Construction of `PressureLayers`

Immutable, concrete field types, built from a `SpectralGrid` like the other model components:

```julia
struct PressureLayers{V, I, F} <: AbstractOutputLayers
    pressure::V         # [Pa], in the model's number format, on its architecture
    interpolation::I
    scratch::F          # Field(NF, grid, npressure), on the model grid
end

output = NetCDFOutput(spectral_grid, PrimitiveWet, layers = PressureLayers(spectral_grid, [850, 500, 200] .* 100))
```

The constructor allocates both buffers and does the monotonic/positive assertions, so `initialize!(::PressureLayers, ::AbstractModel)` and the `pressure_on_arch` field are gone. What is left at `initialize!` is only `sync_extrapolations!`, which takes κ from the model's atmosphere.

## 3. Naming and parametric fields

- `vertical_dimension` → `vertical_dimension_name` (also the 1-argument method on `main` and the Terrarium extension), and the netCDF keyword is `vertical_dim_name`.
- `TemperatureOutput{F, E}` with `extrapolation::E`, and `layers::Layers` on `NetCDFOutput`, `ZarrOutput` and `HEALPixOutput`. `sync_extrapolation` keeps the extrapolation's number format so the field type stays put, and `extrapolate_below` converts κ to the number format of the field — an NF=Float32 kernel no longer does a Float64 `^`.

Note this makes `output.variables[:temp].extrapolation = SubsurfaceMask(...)` a type error; a non-default extrapolation goes through the constructor now, `add!(output, TemperatureOutput(extrapolation = ...))`, which is what the docs already showed.

## Tests

All green locally, each file in its own process:

| file | result |
| --- | --- |
| `test/output/pressure_layer_output.jl` | 39/39 |
| `test/output/vertical_interpolation.jl` | 26/26 |
| `test/output/zarr_output.jl` | all pass, incl. 5 pressure-layer |
| `test/output/healpix_output.jl` | all pass, incl. 4 pressure-layer |

Runic is clean on every file touched. The dev plan in `docs/dev/2026-09/` has a revision-log entry for this round.

milankl:
@henrikauestad with this branch could you check that all you need here works and is easy to use as you need it? I'd keep the output on model layers for now, but if everything works, in the future I'd be happy to move to some standard e.g. 5-level pressure layer output 100, 200, 500, 825, 1000 hPa for example. What do you think?

henrikauestad:
Quick skim, looks good and I agree with naming and standard output. I will review properly and test, but if you cannot wait to merge then shoot at will. Thanks so much for pushing this! @milankl

milankl:
Top output on $\sigma = 0.5625$, bottom output on $p = 500$ hPa, continents are visible in the top as expected as the mountains push model layers upward (lower pressure, hence lower temperature) which is compensated in the bottom to yield a pretty constant temperature across orography. Neat! 

image

milankl:
@hannahw0od maybe relevant for you ☝🏼 Don't know whether you're currently post-processing onto pressure layers ... with this PR you can simply do

```julia
output = NetCDFOutput(spectral_grid, layers=PressureLayers(spectral_grid, pressure = [200, 500, 850] * 100))
```

to have output on the 200, 500 and 850 hPa layers instead of on layers of constant sigma ...

hannahw0od:
> @hannahw0od maybe relevant for you ☝🏼 Don't know whether you're currently post-processing onto pressure layers ... with this PR you can simply do
> 
> ```julia
> output = NetCDFOutput(spectral_grid, layers=PressureLayers(spectral_grid, pressure = [200, 500, 850] * 100))
> ```
> 
> to have output on the 200, 500 and 850 hPa layers instead of on layers of constant sigma ...

yess this is so useful for me I was going to ask about how to get model levels!

henrikauestad:
testing the subsurface masking and that seems to work fine

test_fig

henrikauestad (commented):
This looks good. Both interpolators and subsurface masking works for me.

milankl:
Great, let's wait for CI after main was merged, then we can merge into main!

PR #1258, opened 2026-09-16
CI with Julia v1.13

milankl:
Move submodules and SpeedyWeather CI to v1.13, docs stays at v1.12 as deliberately done in #1255


milankl:
It seems that here we also reduce CI time by almost 2x

| SpeedyWeather CI | time |
| - | - |
| v1.10 | 25min (here), 37min (main) |
| v1.12 | 62min (main) |
| v1.13 | 37min (here) |

So testing with v1.12 instead of v1.13 also makes CI performance-wise very much sense. Likely related to compiler improvements from 1.12 to 1.13 as e.g. reported also in the State of Julia talk at JuliaCon this year.

milankl:
CUDA passes, AMDGPU still queued but both remain on v1.12 with this PR and so merging as is.

@gregordecristoforo feel free to test with v1.13 too in some future PR, I'm leaving GPU tests untouched here to avoid possible conflicts with other GPU-related changes you've been working on!

Issue #1259, opened 2026-09-18
Segfaults with v1.13


milankl:
I've had this twice now, once on linux + GPU, once on macOS+CPU but within the `simulation = Simulation(model)` call that I get a segfault, looks roughly like this with plenty of numbers printed of some array then boom

```julia
...
0, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000, 1.0000000000000000]
==========
~~~~~~~~~~ ptr queue bottom ~~~~~~~~~~

[41259] signal 6: Abort trap: 6
in expression starting at REPL[19]:1
__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
Allocations: 177292298 (Pool: 177279245; Big: 13053); GC: 260
zsh: abort      julia
```

just flagging this, still not sure what causes it

milankl:
In #1260 AMDGPU CI on 1.13 a similar segfault was just hit, somewhere in the primitive wet model starting like this before endless lines of array entries are printed

```julia
...
_bk;t=1789742949147

_bk;t=1789742949147!!! ERROR in jl_ -- ABORTING !!!

_bk;t=1789742949147==========

_bk;t=1789742949147

_bk;t=1789742949147!!! ERROR in jl_ -- ABORTING !!!

_bk;t=1789742949147==========

[...]

PR #1260, opened 2026-09-18
test gpu and metal tests with julia v1.13

gregordecristoforo:


milankl:
I think testing on 1.13 really improved CI performance, Metal GPU now down to 8min (was 10min on [main](https://github.com/SpeedyWeather/SpeedyWeather.jl/actions/runs/35352537907/job/105623869893#step:7:1)). Makes sense because the compiler is supposedly faster again after a regression in 1.12

milankl:
AMDGPU fails with a segfault similar to #1259, skipping submodules and docs as unaffected

milankl:
Metal tests pass on v1.13, so that should definitely stay, not sure why CUDA didn't run, if AMDGPU has problems with v1.13 we could leave it on v1.12

gregordecristoforo:
This error doesn't appear to be deterministic. For all three extensions (AMDGPU, CUDA and Metal) you will find occasions that pass and others that fail in the commits of this PR.

I suspect these packages haven't caught up yet with Julia v1.13. Shall we just wait and try again once never package versions come out?

milankl:
The release candidate for v1.13.1 is already out: https://github.com/JuliaLang/julia/compare/v1.13.0...release-1.13 so might be a matter of only a week or two. But not sure this fixes anything. But it's good to flag that there might be issues and we can revisit this later!

PR #1261, opened 2026-09-19
Parameterization function stub for parameterizations defined in extensions

milankl-claude:
Adds a `Parameterization` function stub, following @maximilian-gelbrecht's suggestion in #1089 (review thread on `docs/src/co2_forcing.md`).

Julia doesn't let package extensions export new names. Right now, a SpeedyWeather parameterization that lives in another package's extension can only be reached through `Base.get_extension`, e.g. for AnalyticBandRadiation/NumericalRadiation:

```julia
const SpeedyExt = Base.get_extension(AnalyticBandRadiation, :AnalyticBandRadiationSpeedyWeatherExt)
longwave = SpeedyExt.SpeedyAnalyticBandLongwave(spectral_grid)
```

With this PR the extension can add a method to the exported stub instead:

```julia
# ExternalPackage (no SpeedyWeather dependency)
struct AnalyticBandLongwave ... end

# ExternalPackageSpeedyWeatherExt
SpeedyWeather.Parameterization(SG::SpectralGrid, scheme::AnalyticBandLongwave; kwargs...) =
    SpeedyAnalyticBandLongwave(SG, scheme; kwargs...)
```

and users can write

```julia
using SpeedyWeather, NumericalRadiation
longwave_radiation = Parameterization(spectral_grid, AnalyticBandLongwave())
model = PrimitiveWetModel(spectral_grid; longwave_radiation)
```

This isn't type piracy, because the extension owns the type it dispatches on. We already do the same thing with `TerrariumOutput` and `animate`.

Changes
- `function Parameterization end` in `parameterizations/general.jl`, exported, with a docstring
- Fallback `Parameterization(::SpectralGrid, p::AbstractParameterization) = p`, so generic code can call it on SpeedyWeather parameterizations too. Anything else throws a `MethodError`.
- Docs: new section "Parameterizations from other packages" in `parameterizations.md`
- Tests: mock external scheme → `UniformCooling`, which also runs a few time steps

Possible follow-up (not in this PR): model constructors could call `Parameterization(spectral_grid, x)` on every parameterization keyword. Then `PrimitiveWetModel(spectral_grid; longwave_radiation = AnalyticBandLongwave())` would work directly.


maximilian-gelbrecht:
I don't think this will be necessary for NumericalRadiation. We'll rewrite it so that we'll have an extension for NumericalRadiation here (and in Breeze). 

Do you still want to have it for other potential packages?

milankl:
I think it makes sense to have it, I don't see `Parameterization` (singular) to be used otherwise and so I'd just say lets merge it in and if someone has the "can't export new type from extension" problem then this is the solution?

maximilian-gelbrecht (approved):
Yeah fine with me, looks good, can be merged

PR #1263, opened 2026-09-19
Batched interpolation for multi-dimensional fields

bgroenks96:
## Summary

Interpolating a 3D or higher field looped over its layers, launching one kernel per layer and recomputing the pole averages and the four index bounds checks on every one. This collapses the trailing dimensions and interpolates all layers in a single batched launch, sharing the locator's indices and weights across them.

The interpolation itself is unchanged: `anvil_average` and the locator are reused verbatim, and only the indexing differs.

## What changed

- `_interpolate!(::AbstractMatrix, ::AbstractMatrix, ...)`, launching over `(npoints_output, nlayers)` with `ArrayWorkOrder`. Matrix-to-matrix previously hit the vector method and threw on its `length(A) == npoints` check, so nothing that worked before behaves differently.
- `_interpolate_batched_kernel!`, identical to `_interpolate_kernel!` except for `@index(Global, NTuple)` and the layer index on the data reads.
- `average_on_poles(::AbstractMatrix, rings)` for `AbstractFloat` and `Integer`, returning one average per layer. These index rather than view, matching the existing methods, since a view triggers scalar indexing on GPU.
- `interpolate!(::Field, ::Field, ...)` now reshapes both data arrays to `(npoints, :)` and makes one batched call. Since a field's data has the horizontal grid flattened into the first axis, this covers `Field3D` (matrix data) and `Field4D` (3D data) alike, and a matrix reshapes to itself.

The reshape is O(1) only for dense data, so the method falls back to the original per-layer loop when either array is not a `DenseArray`, for example a field wrapping a non-contiguous view. `Array`, `CuArray`, and `JLArray` are all `DenseArray`, so the batched path is the normal one.

## Performance

A40, `FullGaussianGrid(24)` to `FullGaussianGrid(96)`, Float32:

| layers | per-layer loop | batched | speedup |
|---|---|---|---|
| 1 | 214.9 µs | 188.3 µs | 1.1x |
| 7 | 1492.8 µs | 189.1 µs | 7.9x |
| 32 | 6729.1 µs | 194.8 µs | 34.5x |

Per call to `_interpolate!`, the breakdown is roughly:

- the interpolation kernel itself: ~8 µs
- `average_on_poles`: 32 µs batched, 60 µs per layer
- each of the four `extrema_in` bounds checks: ~36 µs, so ~144 µs combined

The bounds checks are GPU reductions whose results feed host-side `throw` decisions, so each forces a device-to-host synchronization. The per-layer loop paid that ~180 µs fixed cost once per layer; batching pays it once in total. That accounts for essentially all of the speedup.

## One behavioral caveat

The batched pole average reduces over all layers at once (`mean(A[rings[1], :], dims=1)`), which accumulates in a different order than the single-layer reduction (`mean(A[rings[1]])`). On some grids the two differ in the last bit.

Measured across `FullGaussianGrid`, `OctahedralGaussianGrid`, `OctaHEALPixGrid`, and `HEALPixGrid` in Float32 and Float64: the results are bit-identical everywhere except the HEALPix-family grids in Float64, where they differ by 1 ULP (1-2e-16) at 4-5 output points, all of which lie on the first or last ring, i.e. exactly the points that take a pole value. Forcing bit-identity would mean constraining how the reduction is written, which did not seem worth it, but it does mean interpolating a field layer by layer and interpolating it as a whole can differ in the last bit near the poles. The added tests use `≈` for this reason, with the cause documented inline.

## Testing

The RingGrids suite passes. Four tests are added:

- batched output matches per-layer output for 3D fields, across six grid types and both precisions
- the same for 4D fields, confirming the trailing dimensions collapse correctly
- non-dense data takes the fallback and agrees with the dense path
- constant fields survive the batched path, which exercises the per-layer pole averages specifically

GPU was checked separately on an A40 for both 3D and 4D fields, agreeing with the per-layer result to 1 ULP.

## Possible follow-up, not in this PR

The four `extrema_in` calls validate the locator's index arrays, which are fixed once `update_locator!` has run and cannot change between calls, yet they are re-validated on every interpolation. At ~144 µs per call on the hardware above, hoisting them into `update_locator!` would speed up every `interpolate!`, 2D included.


RingGrids/src/interpolation.jl
```diff
@@ -358,6 +358,77 @@ end
     Aout[k] = anvil_average(a, b, c, d, Δabs[k], Δcds[k], Δys[k])
 end
 
+# batched version: all layers in a single launch, sharing the locator's indices and weights
+function _interpolate!(
+        Aout::AbstractMatrix,               # Out: interpolated values, (npoints_output, nlayers)
+        A::AbstractMatrix,                  # gridded values to interpolate from, (npoints, nlayers)
```
milankl, line 364: not sure about the dispatch precedence here, you're extending the `_interpolate!` that is `::Any` on both, I guess I agree that there this should be `::AbstractMatrix` to constrain it to 3D (ij, k). Just checking that we skip the `AbstractField` here for `A` (the input) because we want to be able to do `.data`?

RingGrids/src/interpolation.jl
```diff
@@ -409,8 +480,21 @@ function interpolate!(
     fields_match(Aout, A) && return copyto!(Aout.data, A.data)
     @assert ismatching(architecture(A), Aout) "Interpolation is only supported between fields on the same architecture, got $(architecture(A)) and $(architecture(Aout))"
 
-    for k in eachlayer(Aout, A, vertical_only = true)
-        _interpolate!(view(Aout.data, :, k), view(A.data, :, k), locator, geometry, architecture(A))
+    # Interpolate every layer in a single launch by collapsing the trailing dimensions into
+    # one. That reshape is O(1) for dense data only, so anything else (e.g. a field wrapping
+    # a non-contiguous view) would end up in a slower wrapper type and keeps the layer loop.
+    if Aout.data isa DenseArray && A.data isa DenseArray
```
milankl, line 486: I like that!

RingGrids/src/interpolation.jl
```diff
@@ -715,6 +799,27 @@ to return the same number format `NF`."""
...
+"""
+$(TYPEDSIGNATURES)
+Method for a layered `A`, returning a vector of pole averages, one per layer, for the batched
+interpolation kernel. Indexing rather than viewing matches the single-layer methods above: it
+allocates, but a view would trigger scalar indexing on GPU."""
+@inline function average_on_poles(A::AbstractMatrix{NF}, rings) where {NF <: AbstractFloat}
+    A_northpole = vec(mean(A[rings[1], :], dims = 1))    # per layer, around the north pole
+    A_southpole = vec(mean(A[rings[end], :], dims = 1))  # same for south pole
+    return A_northpole, A_southpole
+end
```
milankl, line 811: This is allocating 2 (small) vectors, likely absolutely fine, but in #1215 these vectors went as "scratch" vectors into the AnvilLocator similar as the `ij` vectors are update by `update_locator!` the `interpolate!` call would therefore compute the mean and write into these vectors for every new interpolation. This requires the `AnvilLocator` to support a (maximum) `nlayers` which for 3D interpolation given a specific vertical dimension is fine. Not sure this defeats your points here by using an interpolator for any (?) number of vertical layers that launches a single kernel

milankl (commented):
Generally very supportive of this. Would just be good to coordinate with #1215 to not cause conflicts or end up with similar machinery implemented twice. 3D interpolation here is for the purpose of performance, launching a single kernel instead of one per layer, but does not actually interpolate in the 3rd/vertical dimension, whereas 3D interpolation in #1215 actually computes an interpolation weight for that 3rd/vertical dimension. But there's some overlaps, e.g. because the north/south_pole_averages are computed once upfront in both cases

RingGrids/src/interpolation.jl
maximilian-gelbrecht: A contiguous view is still a `DenseArray` right? 

Because by now contiguous are basically the default for variables in Speedy.

maximilian-gelbrecht: It's an elegant solution to the problem though in general.

maximilian-gelbrecht:
> Generally very supportive of this. Would just be good to coordinate with #1215 to not cause conflicts or end up with similar machinery implemented twice. 3D interpolation here is for the purpose of performance, launching a single kernel instead of one per layer, but does not actually interpolate in the 3rd/vertical dimension, whereas 3D interpolation in #1215 actually computes an interpolation weight for that 3rd/vertical dimension. But there's some overlaps, e.g. because the north/south_pole_averages are computed once upfront in both cases

Yeah, we should just make a design decision together, how to best differentiate the both cases. We can't really rely on the `dims` of the `Field` here because there are legitimate situation where one would want a batched 2D interpolation and also a 3D interpolation for XYZ fields. 

I think in a case like this the very simple approach to just use different function names might be okay, so e.g. `interpolate` and `interpolate_3D`. A bit ugly, but okay. I guess a more Julia-esque way could be `interpolate!(:3D, ...)` `interpolate!(:2D, ...)` or something like this.

maximilian-gelbrecht:
Also just mentioning, that once this is merged, there should be a follow up that adjusts the output writer to use it and also to use GPU in general.

bgroenks96:
I think that, in principle, the `Field` should somehow carry the information about whether or not its third axis should be interpolated? And then this could be a trait as suggested by @maximilian-gelbrecht.

maximilian-gelbrecht:
I am not sure that's possible easily. 

Because the same field may be 3D interpolated in the dycore during particle advection, and then 2D batch interpolated during output. 

So you'd have to do that at call side.

milankl:
I agree with Max that

2D fields
- have 2D interpolation

3D fields
- have batched 2D interpolation (still 2D...)
- 3D interpolation

So the latter isn't distinguishable and both may happen to the same field. Essentially the locator needs to know whether interpolation is on the same vertical coordinates (batched 2D) or different ones (3D).

I think it's easiest to have

- `interpolate_2D!` for 2D
- `interpolate_3D!` for 3D
- `interpolate!` that could decide given the interpolator whether to do 2D/3D or whichever way the vertical locations are communicated

milankl:
Also note that with #1256 and #1257 we're getting output on pressure layers which is a vertical interpolation too. Here we go from 3D with shape IJ x N to 3D with shape IJ x M (so horizontal dimension stays the same, vertical dimension can be different in size). So it's essentially a vertical interpolation only, technically a batched 1D interpolation. Because that depends on the model's vertical coordinate it's not easy to move that into RingGrids (which doesn't know about `AbstractVerticalCoordinates` ... should it?)

```julia
function interpolate_pressure_layers!(
        out_field::AbstractField,                       # OUTPUT: (lonlat, npressure)
        in_field::AbstractField,                        # INPUT: (lonlat, nlayers)
        surface_pressure::AbstractField,                # INPUT: (lonlat), [Pa]
        p::AbstractVector,                              # pressure layers of the output [Pa]
        coordinates::AbstractVerticalCoordinates,       # of the input field
```

so you see here that this interface actually requires the `surface_pressure` to be passed on to convert from sigma/hybrid to pressure coordinates, it feels ugly to put this into RingGrids

maximilian-gelbrecht:
Which one will we merge first? 

The first PR to be merged should just implement the `interpolate_2D!` and `interpolate_3D!` interface and then the other can follow.

PR #1264, opened 2026-09-19
Euler forward for ocean, sea ice and land with leapfrog (into #1183)

milankl-claude:
Follow-up to #1183, targeting `mk/landstepping`. It implements Fix A from the [review comment](https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1183#issuecomment-5743520635) plus the smaller fixes from that review.

Implementation plan: [`docs/dev/2026-09/surface-euler-time-stepping.md`](https://github.com/milankl-claude/SpeedyWeather.jl/blob/mc/landstepping-euler/docs/dev/2026-09/surface-euler-time-stepping.md)

## What changes

- **Euler forward for ocean, sea ice and land with leapfrog.** The new `update_prognostic_surface!` hook defaults to `update_prognostic!`, so NCycleLorenz etc. stay on the general path. For `Leapfrog` it does Euler over Δt (Δt/2 on the first two steps, matching the clock) from step 1 and writes the result into both steps. All surface components read step 1. Only setting `which_prognostic_step = 1` is not enough: it is still a 2Δt step from t−Δt, which halves the stability limit, and forcing_drag still blows up with it.
- `surface_time_step` is used for the "within one time step" terms in `ThermodynamicSeaIce`.
- **`LandBucketMoisture`**: the excess-water handling (infiltration, river runoff, capping) moved into `filter!`, where the excess of the new state is known. Previously the clamp zeroed the excess before the kernel could see it, so the excess water was deleted. River runoff is now accumulated in m, matching the output units.
- **Snow melt** cools the soil with `latent_heat_fusion`, not sublimation, which was 8.5× too much. This bug is also on `main`.
- **Type hierarchy restored**: `AbstractDynamicSnow <: AbstractSnow`, `AbstractDynamicSeaIce <: AbstractSeaIce`, etc. The unused `AbstractDynamic/PrescribedLandComponent` types are removed.
- `variables(::AbstractOcean)` and `variables(::AbstractSeaIce)` fallbacks are back, so custom oceans get an SST allocated.
- `SnowModel` no longer declares `soil_temperature`, which clashed with prescribed land temperatures.
- `@boundscheck` fixes.

Not included: restoring the `isfinite` guards (relevant for `mask = false`) and the 285 K initialisation at ice-sheet points. Those are left for you to decide.

## Testing

- New `test/parameterizations/surface_time_stepping.jl`.
- forcing_drag 3D Dry/Wet and the Wet default year pass. In earlier runs of the same scheme, the Dry default year blew up in the atmosphere on 3 December over the North Pacific jet while soil temperatures were normal (see the review). `long_integrations/default_primitive_dry.jl` is being rerun here to confirm on this branch.


milankl-claude:
Local test results on c14d88f3 (Julia 1.13, `--check-bounds=yes`):

| tests | result |
|---|---|
| `parameterizations/{surface_time_stepping,land,ocean_sea_ice,surface_fluxes,longwave_radiation,all_parametrizations}.jl`, `dynamics/{forcing_drag,set}.jl`, `variables/steps.jl` | 378/378 pass |
| `long_integrations/default_primitive_{dry,wet}.jl` | 2/2 pass, no NaNs |

The PrimitiveDry year passes here too, unlike the earlier atmospheric blow-up I saw with the prototype. That blow-up was chaotic and not caused by the land, so CI on other Julia versions could still differ.

milankl:
Okay this looks really good, I'll merge and then we can figure out rest in the other PR before going into main

Issue #1265, opened 2026-09-20
Semi-Lagrangian advection for the barotropic model: feasibility


milankl-claude:
## Summary

Feasibility assessment for a **semi-Lagrangian (SL) advection scheme in the `BarotropicModel`** as a first step. Short answer: **this fits the current architecture well and needs no restructuring of the dynamical core.** The time stepper abstraction is already the right seam, `RingGrids` already has everything needed for the departure-point interpolation, and `ParticleAdvection2D` is essentially a working trajectory solver that can be reused. The main open questions are accuracy of the available interpolator and how to combine SL transport with the spectral vorticity inversion.

This is a scoping issue, not a plan — no code written yet.

## What the barotropic model looks like today

One time step (`SpeedyWeather/src/time_stepping/time_integration.jl`) is:

```julia
reset_tendencies!(vars, time_stepping)
dynamics_tendencies!(vars, model)     # forcing! drag! -> grid_tendencies! -> transform! -> spectral_tendencies!
diffusion_and_implicit!(vars, model)
update_prognostic!(vars, model)       # dispatches on model.time_stepping
transform!(vars, model)               # spectral vorticity -> grid vor, u, v via UV_from_vor!
particle_advection!(vars, model)
```

Transport is Eulerian and spectral: `vorticity_flux_grid_tendencies!` (`dynamics/tendencies.jl:19`) builds `u_tend = Fᵤ + v(ζ+f)`, `v_tend = Fᵥ - u(ζ+f)` on the grid, these are transformed, and `curl!` gives `∂ζ/∂t` (`dynamics/tendencies.jl:~1046`). The prognostic variable is spectral `vorticity`; `vars.grid.vorticity`, `vars.grid.u`, `vars.grid.v` are diagnosed from it every step in `time_stepping/transform.jl:11`.

## Why this is feasible without many changes

[...]

PR #1266, opened 2026-09-20
Semi-Lagrangian time stepping for the barotropic model

milankl-claude:
Draft, opened to show how the changes sit in the code rather than as something ready to merge. Plan: [`docs/dev/2026-09/semi-lagrangian-barotropic.md`](https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/mc/semi-lagrangian-barotropic/docs/dev/2026-09/semi-lagrangian-barotropic.md). Follows on from the discussion in #1265.

## What this adds

A two-time-level `SemiLagrangian` time stepper for `BarotropicModel`. Absolute vorticity `ζ+f` is shifted to the departure points of backward trajectories instead of being transported through the vorticity flux, which removes the advective CFL restriction. The trajectories use `RingGrids`' existing `AnvilInterpolator`.

```julia
spectral_grid = SpectralGrid(truncation=31, nlayers=1)
model = BarotropicModel(spectral_grid, time_stepping=SemiLagrangian(spectral_grid))
run!(simulation, period=Day(1))
```

## How it fits the existing structure

Spectral and grid step counts are already decoupled and dispatched on the model, which `Leapfrog` uses to get 2 spectral / 1 grid step in 2D. A two-time-level SL scheme wants the mirror image and needs no new mechanism:

```julia
prognostic_spectral_steps(::AbstractSemiLagrangian) = 1   # one spectral state
prognostic_grid_steps(::AbstractSemiLagrangian, ::Barotropic) = 2   # u, v at n and n-1
```

Those two grid levels supply the `3/2 uⁿ - 1/2 uⁿ⁻¹` extrapolation for the trajectory winds. Confirmed working: `size(vars.prognostic.vorticity, 3) == 1`, `size(vars.grid.u, 3) == 2`.

The vorticity flux is switched off through a new `advection_factor` dispatch (1 for Eulerian steppers, 0 for SL), which is a one-line change inside `_vorticity_flux_kernel!`. `dynamics_tendencies!` is then reused **unchanged** and produces exactly the source tendency `S = ∇×(Fᵤ,Fᵥ) - cζ` that the SL update needs.

## Exponential hyperdiffusion (ETD1)

One SL step is a Lie–Trotter split with both operators handled exactly:

```
ζⁿ⁺¹ = exp(z) ζ* + Δt φ₁(z) S,    z = Δt∇²ⁿ,  φ₁(z) = (exp(z)-1)/z
```

`HyperDiffusion` is diagonal in spectral space with a genuine per-degree timescale, and its implicit factor `1/(1-z)` is the [0/1] Padé approximant of `exp(z)`. Substituting the exact form into the existing kernel shape `(tendency + expl*var)*impl` followed by `var += Δt*tendency` shows **`expl` is unchanged and only `impl` differs**:

```julia
impl = φ₁(z)    instead of    1/(1 - z)
```

Same kernel, same arrays, same shapes. Added as an opt-in `HyperDiffusion(..., exponential=true)`; the default stays backward Euler so nothing changes for existing steppers unless asked. `SemiLagrangian` opts in automatically.

Verified numerically — `1 + z φ₁(z) == exp(z)` to machine precision, and `φ₁(0) = 1` with no `0/0`:

| z | `1 + z φ₁(z)` | `exp(z)` | backward Euler |
|---|---|---|---|
| -0.125 | 0.8824969 | 0.8824969 | 0.8888889 |
| -1.0 | 0.36787945 | 0.36787945 | 0.5 |
| -5.0 | 0.0067379475 | 0.006737947 | 0.16666667 |

The -0.125 row is roughly the truncation limit at T32 with the 4-hour default: ~0.7% per step at the smallest resolved scale. This buys **accuracy, not stability** — both forms are unconditionally stable and monotone. It matters more under SL where Δt is larger.

## Why not `NCycleLorenz`

Recorded in the plan. In short: with weights 1, 3/2, 3 the N=3 cycle has amplification `P(z) = 1 + 3z + 4.5z² + 4.5z³`. Feeding it an SL effective tendency `Δt·F = μζ`, `μ = exp(-iθ)-1`, gives agreement with the exact `(1+μ)³ = exp(-3iθ)` only to first order in μ, and `|μ|` grows to 2:

| C (grid-scale Courant) | \|P(μ)\| | exact |
|---|---|---|
| 0.25 | 0.53 | 1 |
| 0.5 | 7.6 | 1 |
| 1.0 | 23 | 1 |

Unstable below C = 0.5 — a worse limit than the Eulerian core it would replace. Any tendency-based multistep method approximates `exp(z)` by a polynomial, and polynomials have bounded stability regions; SL's value is an amplification on the unit circle for arbitrarily large θ. The group-property rescue that works for leapfrog (re-trace over 2Δt) has no analogue since `prognostic_steps(::NCycleLorenz) = 1`.

## Testing

`SpeedyWeather/test/dynamics/semi_lagrangian.jl`, 92 tests, all passing:

```
φ₁ exponential integrator function |   13     13  0.4s
HyperDiffusion exponential option  |   61     61  13.8s
SemiLagrangian: displacement       |    4      4  0.1s
SemiLagrangian: step structure     |    8      8  2.4s
SemiLagrangian: barotropic runs    |    5      5  17.7s
SemiLagrangian: only barotropic    |    1      1  1.4s
```

**The wider test suite has not been run.** The `advection_factor` change touches the shared vorticity flux kernel used by all three model types, so that needs checking before this is anything but a draft.

## Known limitations

- **Interpolation order.** `AnvilInterpolator` is bilinear-class, and SL with bilinear interpolation is strongly damping — poor for a barotropic problem where enstrophy conservation is the point. Expected to dominate the error budget; a cubic/quasi-monotone interpolator on reduced grids is the obvious follow-up.
- **No accuracy verification yet.** No solid-body rotation test, no Courant sweep, no comparison against the Eulerian core, no conservation diagnostics. The scheme runs and stays finite; that is all that has been shown.
- **Tracers stay Eulerian**, so they keep their CFL limit — inconsistent with the vorticity transport.
- **Barotropic only**; `SemiLagrangian` throws on other models.
- **Unmeasured cost**: a locator update per trajectory iteration plus one extra grid→spectral transform.
- **No GPU verification.** Kernels are `KernelAbstractions`-based but untested on device.

## Aside

`ParticleAdvection2D.Δt` looks like it may be missing a division by the planetary radius — `Δt[] = every_n_time_steps * Δt * (360/2π)` with `u` in m/s, while the docstring says `[s*˚/m]`. Not touched here; this PR derives its own conversion (`trajectory_time_step`). Worth a separate look.


SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
```diff
@@ -0,0 +1,218 @@
...
+    Δt_trajectory::Base.RefValue{NF}
+
+    "[DERIVED] Interpolator (grid geometry + locator) used to evaluate fields at departure points"
+    interpolator::IP
+
+    "[DERIVED] Longitudes [˚E] of the arrival points, i.e. of the grid points"
+    arrival_lond::VT
+
+    "[DERIVED] Latitudes [˚N] of the arrival points, i.e. of the grid points"
+    arrival_latd::VT
```
milankl: What about using these as they are in `model.geometry`?

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
```diff
@@ -0,0 +1,218 @@
...
+    ns = :semi_lagrangian
+    return (
+        ScratchVariable(:departure_lond, VectorDim(npoints), namespace = ns, desc = "Departure point longitude", units = "˚E"),
+        ScratchVariable(:departure_latd, VectorDim(npoints), namespace = ns, desc = "Departure point latitude", units = "˚N"),
+        ScratchVariable(:departure_u, VectorDim(npoints), namespace = ns, desc = "Zonal wind at departure point", units = "m/s"),
+        ScratchVariable(:departure_v, VectorDim(npoints), namespace = ns, desc = "Meridional wind at departure point", units = "m/s"),
+        ScratchVariable(:u_star, Grid2D(), namespace = ns, desc = "Time-extrapolated zonal wind for trajectories", units = "m/s"),
+        ScratchVariable(:v_star, Grid2D(), namespace = ns, desc = "Time-extrapolated meridional wind for trajectories", units = "m/s"),
+        ScratchVariable(:absolute_vorticity, Grid2D(), namespace = ns, desc = "Absolute vorticity ζ+f", units = "1/s"),
+        ScratchVariable(:vorticity_departure, Grid2D(), namespace = ns, desc = "Vorticity shifted to departure points", units = "1/s"),
```
milankl: Shouldn't these be `DynamicsVariable`s?

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
```diff
@@ -0,0 +1,218 @@
...
+
+"""$(TYPEDSIGNATURES) Work arrays for the departure point search and the trajectory shift."""
+function variables(L::SemiLagrangian, model::AbstractModel)
+    npoints = RingGrids.get_npoints(model.spectral_grid.grid)
+    ns = :semi_lagrangian
+    return (
+        ScratchVariable(:departure_lond, VectorDim(npoints), namespace = ns, desc = "Departure point longitude", units = "˚E"),
+        ScratchVariable(:departure_latd, VectorDim(npoints), namespace = ns, desc = "Departure point latitude", units = "˚N"),
+        ScratchVariable(:departure_u, VectorDim(npoints), namespace = ns, desc = "Zonal wind at departure point", units = "m/s"),
+        ScratchVariable(:departure_v, VectorDim(npoints), namespace = ns, desc = "Meridional wind at departure point", units = "m/s"),
```
milankl: What about making these also `Grid2D` (or `GridXY`) given they have exactly as many points as a 2D field? For every grid cell this would then correspond to the lon/lat/u/v at the departure point that is associated with this grid cell as arrival point.

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
```diff
@@ -0,0 +1,218 @@
...
+
+    # TRANSPORT. Backward trajectories, then shift ζ+f onto their departure points.
+    departure_points!(vars, time_stepping, model)
+    semi_lagrangian_transport!(vars, time_stepping, model)
+
+    # retain the current grid u, v as the previous ones before `transform!` overwrites them
+    move_prognostic_grid_variables_back!(vars, time_stepping, model)
+
+    # ζ* back to spectral, overwriting the prognostic state
+    vorticity = get_step(vars.prognostic.vorticity, 1)
```
milankl: dispatch over time stepper here, technically this function is only called with a semi lagrangian time stepper, yes, but let's be consistent. And given this is a no-cost abstraction we'd only pay a little compile time

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
```diff
@@ -0,0 +1,218 @@
...
+    transform!(
+        vorticity, vars.scratch.semi_lagrangian.vorticity_departure,
+        vars.scratch.transform_memory, model.spectral_transform
+    )
+
+    # DIFFUSION + SOURCES, as an ETD1 step on the transported state
+    horizontal_diffusion!(vars, model)
```
milankl, line 193: Do we actually need diffusion with an already diffusive interpolation in the semi Lagrangian scheme?

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
```diff
@@ -0,0 +1,218 @@
...
+
+"""$(TYPEDSIGNATURES)
+Generator function for `SemiLagrangian` using `spectral_grid` for resolution."""
+function SemiLagrangian(
+        spectral_grid::SpectralGrid;
+        Δt_at_T32 = Minute(60),         # larger than Leapfrog's default: no advective CFL limit
+        adjust_with_output = true,
+        n_iterations = 2,
+        extrapolate_winds = true,
+        exponential_diffusion = true,
```
milankl: Why not move those into the `@kwdef` declared type definition?

SpeedyWeather/src/dynamics/horizontal_diffusion.jl
```diff
@@ -170,12 +175,35 @@ end
         ∇²ⁿ[l_plus_1, k] = @fastmath -eigenvalue_norm^power / time_scale
         ∇²ⁿ_div[l_plus_1, k] = @fastmath -eigenvalue_norm^p / time_scale_div
 
-        # and implicit part of the diffusion (= 1/(1-Δtν∇²ⁿ))
-        ∇²ⁿ_implicit[l_plus_1, k] = 1 / (1 - Δt * ∇²ⁿ[l_plus_1, k])
-        ∇²ⁿ_div_implicit[l_plus_1, k] = 1 / (1 - Δt * ∇²ⁿ_div[l_plus_1, k])
+        # and implicit part of the diffusion, either backward Euler 1/(1-Δtν∇²ⁿ) or, for
+        # `exponential`, the exact φ₁(Δtν∇²ⁿ) making the whole step an exponential integrator
+        ∇²ⁿ_implicit[l_plus_1, k] = implicit_factor(Δt * ∇²ⁿ[l_plus_1, k], exponential)
+        ∇²ⁿ_div_implicit[l_plus_1, k] = implicit_factor(Δt * ∇²ⁿ_div[l_plus_1, k], exponential)
```
milankl, line 184: Is the `exponential` treatment here not always a good idea? I'm wondering whether this generally should be implemented as an option for implicit HyperDiffusion, regardless of the time stepper?

SpeedyWeather/src/dynamics/semi_lagrangian.jl
```diff
@@ -0,0 +1,178 @@
...
+pole-crossing logic in `move` (crossing a pole flips the longitude by 180˚) and the `mod` wrapping
+back into [0, 360˚E), [-90, 90˚N]."""
+@inline function displace(lond::NF, latd::NF, u, v, Δt_deg) where {NF}
+    dlat = v * Δt_deg
+
+    # TODO: `cos(deg2rad(...))` rather than `cosd` to match `advect_2D`, see JuliaGPU/AMDGPU.jl#1041
+    coslat = max(cos(deg2rad(latd)), eps(NF))   # prevents division by zero at the poles
+    dlon = u * Δt_deg / coslat
+
+    particle = mod(move(Particle{NF}(lond, latd), dlon, dlat))
```
milankl: does it make sense here to reuse the Particle machinery or is there overhead we'd want to avoid? Given this creates a particle, uses the `move` and `mod` functions and then unpacks the particle again ... I agree with you it's better to reuse those functions just checking whether this would actually compile away

SpeedyWeather/src/dynamics/semi_lagrangian.jl
```diff
@@ -0,0 +1,178 @@
...
+    sl = vars.scratch.semi_lagrangian
+    (; clock) = vars.prognostic
+
+    # grid u, v carry 2 steps for this time stepper: 1 = previous, 2 = current
+    u_new = field_view(vars.grid.u, :, 1, 2)
+    v_new = field_view(vars.grid.v, :, 1, 2)
+    u_old = field_view(vars.grid.u, :, 1, 1)
+    v_old = field_view(vars.grid.v, :, 1, 1)
+
+    # no previous step available on the very first time step
```
milankl: don't we copy over the state from step 1 to 2 in the initial `transform!` call? Because then old = new and the extrapolation, despite the weights, would be u_star = u_new anyway?

SpeedyWeather/src/dynamics/semi_lagrangian.jl
```diff
@@ -0,0 +1,178 @@
...
+    return nothing
+end
+
+@kernel inbounds = true function _add_coriolis_kernel!(
+        out, @Const(vor), @Const(f), scale, @Const(whichring), add::Bool
+    )
+    ij = @index(Global, Linear)
+    j = whichring[ij]
+    out[ij] = vor[ij] + ifelse(add, f[j] * scale, -f[j] * scale)
+end
```
milankl, line 300: Would it be easier to change planetary vorticity f to be a field instead of a latitude vector? I've been wondering about this change for a while, yes you then need to read in more data (as every ring has the same value of f, essentially duplicated across every longitude value) but I doubt it hurts performance much and then you could just write this as one-linear like vor .+ scale*f

SpeedyWeather/src/dynamics/semi_lagrangian.jl
```diff
@@ -0,0 +1,178 @@
...
+        vars::Variables,
+        time_stepping::AbstractSemiLagrangian,
+        model::AbstractModel,
+    )
+    (; locator, geometry) = time_stepping.interpolator
+    (; f) = model.coriolis
+    scale = vars.prognostic.scale[]
+
+    sl = vars.scratch.semi_lagrangian
+    absolute_vorticity = sl.absolute_vorticity
```
milankl: Could this not simply be done with one of the generic scratch arrays like `simulation.variables.scratch.a` because you only temporarily add/subtract f, correct? So no other place in the code needs to know the absolute vorticity? Then better to not bother allocating an additional array at initialize! ?

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
```diff
@@ -0,0 +1,218 @@
...
+prognostic_spectral_steps(::AbstractSemiLagrangian) = 1
+# ... but two grid time levels of u, v for the SETTLS wind extrapolation
+prognostic_grid_steps(::AbstractSemiLagrangian, ::Barotropic) = 2
+tendency_steps(::AbstractSemiLagrangian) = 1
+
+# WHICH STEP TO READ WHEN
+# spectral variables have a single step
+@inline which_prognostic_step(var, ::AbstractSemiLagrangian, ::STEP_COMPONENT) = 1
+# grid variables have two, the 2nd is the current one (the 1st is the previous time step)
+@inline which_prognostic_step(var::AbstractField, ::AbstractSemiLagrangian, ::STEP_COMPONENT) = 2
```
milankl: `STEP_COMPONENT` is convenience within steps.jl only, can you explicitly write AbstractModelComponent and AbstractSpectralTransform here if they both use indeed the same step?

milankl (commented):
Looks great already, curious of how the performance roughly plays out compared to Eulerian advection. It's OK to have the bilinear AnvilInterpolator for now, we'll improve on that in another PR. Some comments

milankl:
> ## Aside
> 
> `ParticleAdvection2D.Δt` looks like it may be missing a division by the planetary radius — `Δt[] = every_n_time_steps * Δt * (360/2π)` with `u` in m/s, while the docstring says `[s*˚/m]`. Not touched here; this PR derives its own conversion (`trajectory_time_step`). Worth a separate look.

This is already being fixed in #1215

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
milankl-claude: Done — dropped both fields. `model.geometry.londs/latds` are exactly what I needed (`Vector{Float32}`, length `npoints`, already on the right architecture), and `departure_points!` takes `model` anyway. One less thing duplicated and one less thing to keep in sync with the grid.

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
milankl-claude: Agreed, changed to `DynamicsVariable`s. They are named working arrays of the dynamical core, not write-before-read scratch — `departure_lond` in particular is meaningful state you would want to inspect. Now at `vars.dynamics.semi_lagrangian.*`.

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
milankl-claude: Good call, done — all four are `Grid2D` now. Your framing is the right one and I put it in the docstring: entry `ij` is the departure point of the trajectory *arriving* at grid cell `ij`, so the association with the arrival cell is in the type rather than implied by index order.

One wrinkle worth flagging: the interpolation calls need `.data` (the bare vector), because `interpolate!(::Field, ::Field2D, ...)` short-circuits to `copyto!` when the two fields share a grid — which they now do. Commented at the call sites so nobody "simplifies" it back.

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
milankl-claude: Done — `get_prognostic_step(vars.prognostic.vorticity, time_stepping, DynamicalCore(), model)`. Agreed on consistency; the hardcoded `1` was me special-casing something the dispatch already expresses.

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
milankl-claude: Done, `@kwdef` with the defaults on the type, matching `HyperDiffusion`. The generator now only builds the interpolator and fills the derived `Δt` fields.

SpeedyWeather/src/dynamics/semi_lagrangian.jl
milankl-claude: Checked, and it compiles away completely — 0 allocations over a 10⁴-point loop, inferred return `Tuple{Float32, Float32}`. `Particle{NF}` is `isbits` (a `Bool` plus 3 `NF`) and `move`/`mod` are both `@inline`, so the construct-move-mod-unpack round trip is just arithmetic on registers.

Worth recording that my first measurement said 192 bytes per call — that was my benchmark binding `displace` to a non-`const` global, forcing dynamic dispatch. Added a `Base.return_types` inference test so a future regression here gets caught.

SpeedyWeather/src/dynamics/semi_lagrangian.jl
milankl-claude: Agreed, removed. `scratch.a`/`.b` are spectral so not directly usable, but your underlying point holds: nothing outside this function needs absolute vorticity, so I now add `f` in place in `vars.grid.vorticity` and subtract it again from the interpolated result. Safe because `vars.grid.vorticity` is overwritten by the closing `transform!` of the same time step before anything (including output) reads it — noted in the docstring since it is the kind of thing that breaks silently if the ordering changes.

`vorticity_departure` stays, as interpolation needs distinct input and output.

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
milankl-claude: Done — spelled out as `AbstractModelComponent` and `AbstractSpectralTransform`. They do use the same step, so it is two methods each for the spectral/grid split rather than one.

SpeedyWeather/src/dynamics/semi_lagrangian.jl
milankl-claude: Good question, and the answer turned out to be no — I checked, and it is worth recording because the assumption is a reasonable one:

```
after initial transform: max|u step1| = 0.0   max|u step2| = 3.1805077e-6
```

`initialize!(vars, ::AbstractLeapfrog, model)` copies step 1 → 2 for the **prognostic** variables, and the PE `transform!` calls `move_prognostic_grid_variables_back!` for the grid ones — but the barotropic `transform!` never called that hook, so under my stepper grid step 1 stayed zero. The guard was load-bearing, and if the extrapolation had ever run on step 0 it would have computed `3/2 uⁿ - 1/2 · 0`, i.e. a 50% overshoot on the first trajectory.

But you are right that the guard is the wrong fix. The barotropic `transform!` now calls the hook like the PE one does, so `uⁿ⁻¹ = uⁿ` going into the first step and the extrapolation reduces to `uⁿ` on its own — guard deleted. `Leapfrog` needed an explicit no-op for `TwoDModels` since it keeps one grid step there and would otherwise index step 2 out of bounds. Test added asserting step 1 == step 2 after initialization, and I smoke-tested Leapfrog + NCycleLorenz on Barotropic/ShallowWater/PrimitiveDry since the hook now fires on a path it did not before.

SpeedyWeather/src/dynamics/semi_lagrangian.jl
milankl-claude: I agree with the change and it would make this a one-liner (`vor .+= scale*f`, no `whichring` lookup, no kernel), but I have left it out of this PR — `Coriolis` is used across all four models and a representation change there deserves its own PR rather than being smuggled in behind a semi-Lagrangian scheme. Added to Future work in the plan so it is not lost.

On the performance worry: for the barotropic grid that is 3168 floats instead of 48, so ~12KB at T31 — trivially cached, and it removes an indirection per grid point. I would expect it to be neutral-to-faster, but that is worth measuring rather than asserting.

SpeedyWeather/src/dynamics/horizontal_diffusion.jl
milankl-claude: I think it is always at least as good, yes: same storage, same kernel, same cost, `expl` unchanged, and both forms are unconditionally stable and monotone for a purely damping diagonal operator. So the option is deliberately independent of the time stepper already — `HyperDiffusion(spectral_grid, exponential=true)` works with `Leapfrog` and `NCycleLorenz` too, `SemiLagrangian` just opts in automatically because ETD1 needs it to be exact.

What I did not do is flip the **default**, and I would rather that were its own PR. The two differ by ~0.7% per step at the truncation limit at T32 (0.8825 vs 0.8889), which is small but not nothing — it shifts results for every existing simulation, and this PR has not run the wider suite. Trivial to change once someone has looked at what it does to the reference tests.

Worth noting the difference grows with `Δt/τ`: at `z = -1` it is 0.368 vs 0.5. So it matters most exactly where you would want it — large time steps, or the divergence filter with its 1-hour `time_scale`.

Also flagging for completeness: I only changed `HyperDiffusion`. `SpectralFilter` still uses `1/(1-z)` and has the same one-line fix available; left alone to keep the diff focused.

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
milankl-claude: Genuinely open, and I do not think I can answer it honestly yet — it needs the Courant sweep and enstrophy diagnostics that this PR does not have.

My expectation is that you will still want *some*, because the two damp different things. Bilinear interpolation damping is set by how far the departure point sits from a grid point, so it is flow-dependent and largest where the wind is strong, and it is not scale-selective in any controlled way. Hyperdiffusion is `∝ l^(2p)`, targeted at the truncation limit and independent of the flow. Removing it entirely leaves nothing specifically controlling the enstrophy cascade piling up at `lmax`.

The likely answer is a much longer `time_scale` rather than none, and `SemiLagrangian` keeping its own default rather than inheriting the Eulerian 4 hours. But that is a number to measure, not guess. Noted in Known limitations.

Mechanically it is already easy to test both ways — `HyperDiffusion(spectral_grid, time_scale=Hour(100))` or swapping the component out — so this can be settled as soon as the accuracy tests exist.

milankl-claude:
Thanks — pushed 850c9005 addressing these. Summary of what changed and what did not:

**Implemented (8):** departure arrays are `Grid2D` `DynamicsVariable`s; arrival coordinates read from `model.geometry` instead of being stored on the stepper; `@kwdef` on the type; `ζ+f` formed in place so the dedicated array is gone; `get_prognostic_step(..., DynamicalCore())` instead of `get_step(..., 1)`; `STEP_COMPONENT` spelled out; and the first-step guard replaced by wiring the barotropic `transform!` into the existing `move_prognostic_grid_variables_back!` hook.

**Checked, no change needed (1):** the `Particle` round trip in `displace` compiles away — 0 allocations, inferred `Tuple{Float32, Float32}`. My earlier concern was a broken benchmark on my side; added an inference test to pin it.

**Agreed but deferred (3):** `f` as a `Field` (touches `Coriolis` across all models — own PR, added to Future work); `exponential` as the *default* for `HyperDiffusion` (strictly better, but changes results for every existing simulation — own PR after validation); and whether SL needs hyperdiffusion at all, which I do not think can be settled without the accuracy tests this PR is still missing.

Two things your review turned up that were real:

- **Grid step 1 was zeros, not a copy of step 2.** The barotropic `transform!` never called the `move_prognostic_grid_variables_back!` hook, so my guard was load-bearing rather than redundant. Without it the first trajectory would have used `3/2 uⁿ`, a 50% overshoot. Now fixed at the source, with `Leapfrog` given an explicit no-op for 2D models where it keeps one grid step.
- That hook change means `transform!` now fires on a path it did not before, so I smoke-tested `Leapfrog` and `NCycleLorenz` on Barotropic, ShallowWater and PrimitiveDry — all still finite and unchanged in magnitude.

Tests now 98 passing in `test/dynamics/semi_lagrangian.jl`. **Still no accuracy verification** — no solid-body rotation, no Courant sweep, no conservation diagnostics, and the wider suite has not been run. Staying a draft until that exists.

And noted on #1215 for the `ParticleAdvection2D.Δt` radius factor — good, I will drop that from my notes.

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
milankl: @bgroenks96 ah yeah ☝🏼 that's why it's important to have `::Any` for the `interpolate!` calls!

bgroenks96: you mean the `interpolate!` calls need to handle plain `Array`s rather than `Field`s?

milankl:
T128 30-day default BarotropicModel, top Eulerian advection (30min-equivalent time step), middle/bottom semi-Lagrangian without/with diffusion (60min-equivalent time step). So clearly at the moment this is too diffusive. But they run at roughly the same speed. I guess for barotropic, there isn't many transforms we skip and the cost of the interpolation adds. Trying longer time steps now!

image

milankl:
Longer time steps don't help much, maybe that's different at higher flow speeds, not sure. However, highest velocity here is 2x higher than with NCycleLorenz maybe that's around the poles, doesn't sound good. But at least performance-wise this gets really speedy! About the same performance with 2x longer time step, but twice as fast with a 120min time step (divide by 4 to get the actual time step not the T32-equivalent time step)

```julia
run_0024 100% Time: 0:00:12 (2000-01-31, 553.92 years/day,  61 m/s)    # Eulerian advection 30min time step
run_0026 100% Time: 0:00:13 (2000-01-31, 528.80 years/day, 161 m/s)    # 60min time step
run_0027 100% Time: 0:00:09 (2000-01-31, 777.48 years/day, 140 m/s)    # 90min time step
run_0028 100% Time: 0:00:07 (2000-01-31, 1009.34 years/day, 124 m/s)   # 120min time step
run_0029 100% Time: 0:00:04 (2000-01-31, 1452.69 years/day, 264 m/s)   # 180min time step
```

milankl:
This is what the variables now look like

```julia
julia> simulation.variables
Variables{@NamedTuple{...}, ...} (3.29 MB)
├ prognostic (92 bytes)
│ ├ clock: Clock{DateTime, Second, Float32, Int64, Millisecond}
│ └ vorticity: 8384×1×1 (LMZT) LowerTriangularArray{ComplexF32, 3, Array{.....
├ grid (0 bytes)
│ ├ vorticity: 40320×1×2, 192-ring (XYZT) OctahedralGaussianField{Float32, ...
│ ├ u: 40320×1×2, 192-ring (XYZT) OctahedralGaussianField{Float32, 3} as Ar...
│ └ v: 40320×1×2, 192-ring (XYZT) OctahedralGaussianField{Float32, 3} as Ar...
├ tendencies (228.46 KB)
│ ├ vorticity: 8384×1×1 (LMZT) LowerTriangularArray{ComplexF32, 3, Array{.....
│ └ grid
│   ├ vorticity: 40320×1×1, 192-ring (XYZT) OctahedralGaussianField{Float32...
│   ├ u: 40320×1×1, 192-ring (XYZT) OctahedralGaussianField{Float32, 3} as ...
│   └ v: 40320×1×1, 192-ring (XYZT) OctahedralGaussianField{Float32, 3} as ...
├ dynamics (1.13 MB)
│ ├ u_tendency: 8384×1×1 (LMZT) LowerTriangularArray{ComplexF32, 3, Array{....
│ ├ v_tendency: 8384×1×1 (LMZT) LowerTriangularArray{ComplexF32, 3, Array{....
│ └ semi_lagrangian
│   ├ departure_lond: 40320-element, 192-ring (XY) OctahedralGaussianField{...
│   ├ departure_latd: 40320-element, 192-ring (XY) OctahedralGaussianField{...
│   ├ departure_u: 40320-element, 192-ring (XY) OctahedralGaussianField{Flo...
│   ├ departure_v: 40320-element, 192-ring (XY) OctahedralGaussianField{Flo...
│   ├ u_star: 40320-element, 192-ring (XY) OctahedralGaussianField{Float32,...
│   ├ v_star: 40320-element, 192-ring (XY) OctahedralGaussianField{Float32,...
│   └ vorticity_departure: 40320-element, 192-ring (XY) OctahedralGaussianF...
...
```

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
milankl: yes!

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
milankl: I've run the semi Lagrangian advection scheme without diffusion and that seems absolutely fine. This might be because the interpolation at the moment is really diffusive, but I generally propose the take out the control of the diffusion from the `SemiLagrangian` type and leave that to the diffusion itself. At the moment if one chooses `time_stepping = SemiLagrangian(...)` then one would get diffusion, and so would need to manually disable it but in the future we may make the SL + no diffusion the default

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
milankl: Apologies, I think your previous version was better, matching the other time steppers, please revert

SpeedyWeather/src/dynamics/horizontal_diffusion.jl
milankl: So I don't see a difference with/without diffusion (see other comment) hence I propose to keep the exponential keyword but leave it to the diffusion itself (default false for now) and separate this choice from the time stepper? Or do you think it's crucial to flag that SemiLagrangian should not be used with exponential=false? Then add a warning during initialization?

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
milankl-claude: Reverted in ef2a3e05 — back to the plain mutable struct plus generator, consistent with `Leapfrog` and `NCycleLorenz`. Agreed, uniformity beats my local preference here.

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
milankl-claude: Agreed and done — `SemiLagrangian` no longer has any diffusion option and no longer touches `model.horizontal_diffusion`. `time_stepping = SemiLagrangian(...)` now gives you whatever `horizontal_diffusion` you asked for, nothing more. If SL + no diffusion becomes the default later, that is a change to the *model* default and nothing in the stepper has to move.

Your empirical result matches mine — weakening the diffusion 25× (`time_scale=Hour(100)`) at T128/30d recovers only ~4% enstrophy (3.582e-5 → 3.742e-5 against 4.055e-5 Eulerian). The interpolation is doing essentially all of the damping.

SpeedyWeather/src/dynamics/horizontal_diffusion.jl
milankl-claude: No, it is not crucial, and no warning — I had this wrong and I am glad you pushed.

My claim was "ETD1 needs `exponential=true` to be exact". True for the diffusion sub-step in isolation, but it does not propagate: the scheme is a Lie–Trotter split of transport and diffusion, and since those two do not commute the splitting error is already `O(Δt²)`. The difference between `φ₁(z)` and `1/(1-z)` is *also* `O(z²)`. So exactness in the damping factor buys a better constant on a term that is already the same order as an error I am making anyway. Calling it required was overselling.

So: `exponential` stays a pure `HyperDiffusion` keyword, default `false`, independent of the time stepper — exactly as you propose. Done in ef2a3e05, with a test asserting the stepper leaves `horizontal_diffusion.exponential` alone so the coupling cannot creep back in.

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
milankl-claude: Yes — `interpolate!(::AbstractVector, ::Field2D, ...)` already exists and works, the problem is only that passing a `Field` as output hits `interpolate!(::Field, ::Field2D, ...)`, which starts with `fields_match(Aout, A) && return copyto!(...)`. For semi-Lagrangian both fields live on the same grid (the departure points simply are not the grid points), so that short-circuit silently turns the interpolation into a copy. Hence `.data` at every call site here.

Worth fixing properly in RingGrids rather than relying on everyone remembering `.data`: the `fields_match` short-circuit is only valid when the locator is known to be the identity, which the function cannot know. Either a keyword to opt out, or splitting "resample this field onto that grid" from "evaluate this field at these locator points" into two names. Happy to do it in a follow-up if you have a preference — it is a RingGrids API decision rather than something I should settle inside this PR.

milankl-claude:
Pushed ef2a3e05. Two requested changes plus one I think is a real defect.

## Found a pole singularity in the trajectories

Your "2x higher velocity, maybe around the poles, doesn't sound good" was the thread worth pulling. The lat/lon displacement needs

```
dlon = u*Δt / (radius * cos(lat))
```

which is singular at the poles. On the T128 octahedral Gaussian grid the outermost ring sits at **89.28˚N where cos(lat) = 0.0125**:

| | u = 20 m/s | u = 50 m/s |
|---|---|---|
| 60min-equiv step (Δt = 900s) | 13.0˚ | 32.4˚ |
| 180min-equiv step (Δt = 2700s) | 38.9˚ | 97.2˚ |

A 97˚ longitude jump in one step, treated as a straight line in (lon, lat). The departure points near the poles were meaningless, which is both a source of spurious velocity and a source of interpolation error exactly where you saw it.

Replaced with 3D Cartesian great-circle trajectories (the standard fix — Ritchie 1987, what the IFS does): convert to a unit vector, build the velocity as a 3D tangent vector, rotate by `α = |V|Δt/radius` along the great circle, convert back. No coordinate singularity, exact for solid-body rotation, and it follows the actual great circle rather than a straight line in lon/lat. About 25 lines, no `1/cos(lat)` anywhere.

**Please A/B this against your reference runs** — I have not reproduced your exact configuration, so I cannot claim it is the whole story. What I can say is that after the fix, max|U| at T128/30d is 27.1 m/s (60min) and 21.1 m/s (180min) against 24.3 m/s Eulerian, i.e. comparable rather than 2x.

## Diffusion decoupled

`SemiLagrangian` no longer has a diffusion option and no longer mutates `model.horizontal_diffusion`. And to answer the direct question: **no, ETD is not needed for SL, and no warning is warranted.** I had this wrong. The scheme is a Lie–Trotter split of transport and diffusion, which do not commute, so the splitting error is already `O(Δt²)` — the same order as the difference between `φ₁(z)` and `1/(1-z)`. Exactness in the damping factor improves a constant on a term of the order I am already erring at. `exponential` stays a pure `HyperDiffusion` keyword, default `false`.

`@kwdef` reverted as requested.

## Measurements

T128, 30 days, default forcing (statistical steady state, so these are the forcing/dissipation balance rather than pure decay — but the forcing is identical across rows):

| configuration | enstrophy | KE |
|---|---|---|
| Eulerian `NCycleLorenz` 30min | 4.055e-5 | 2.851e7 |
| SL 60min, default diffusion | 3.582e-5 | 2.354e7 |
| SL 60min, `time_scale=Hour(100)` | 3.742e-5 | 2.494e7 |
| SL 60min, `n_iterations=4` | 3.854e-5 | 2.493e7 |
| SL 180min, default diffusion | 2.764e-5 | 6.506e7 |

Weakening diffusion 25x recovers ~4% — confirms your finding that diffusion is not the culprit. **`n_iterations=4` recovers ~7%, more than I expected** — I had assumed iterations only fix trajectory phase error, not amplitude damping, and that was wrong: an unconverged departure point *is* a wrong shift, and a wrong shift of a smooth field smooths it further.

**The 180min row is a red flag**: KE is 2.3x the Eulerian value while enstrophy is the lowest. That is energy going somewhere it should not, and it smells like the SETTLS extrapolation rather than the interpolation. Cheap diagnostic: rerun with `SemiLagrangian(spectral_grid, extrapolate_winds=false)`. If the excess KE goes away, the extrapolation is the problem at long steps and needs a stability limiter, not a better interpolator.

## Other options for the diffusivity, roughly in order of value per effort

1. **Attribute before optimising.** Three separable damping mechanisms are currently conflated: the interpolation, the grid→spectral→grid round trip (a truncation filter applied *every step*, which the Eulerian core does not do to a freshly-interpolated field), and the trajectory error. A zero-wind run should be an exact identity — if it damps, the round trip is implicated. A displacement of exactly one grid cell in longitude on a full grid should also be exact for any interpolator.
2. **Cubic interpolation, via a cascade (1D) split.** The 2D stencil on reduced grids is awkward, but doing it as two 1D passes is not: within a ring the longitude spacing is *uniform*, so cubic Lagrange along a ring is a trivial 4-point stencil, and the latitude pass is 4 rings. That gets cubic accuracy at close to bilinear cost. Crucially, **only the transported field needs it** — the winds are interpolated `2*n_iterations` times per step but only set trajectory accuracy, where bilinear is fine. So the expensive interpolation is paid once per step, not 5 times.
3. **Skip interpolation entirely and use spectral derivatives.** SpeedyWeather-specific and possibly less work than a new interpolator: compute `∇ζ` spectrally (exact, `SpectralTransform` already carries the gradient machinery), find the grid point *nearest* the departure point, and Taylor-expand from there. The residual offset is then at most half a grid cell, so a second-order expansion with exact derivatives is very accurate. Costs two extra transforms per step and no new stencil code. The trick is expanding about the nearest grid point rather than the arrival point — expanding about the arrival point only converges for displacements under a grid cell, which defeats the purpose.
4. **More iterations** — the data above says this is worth more than I assumed, and it is free to try. Possibly bump the default to 3.
5. **Quasi-monotone limiting** — worth noting this is *not* a fix for this problem. It suppresses overshoots and in doing so adds damping. Only relevant once positivity matters (tracers).
6. **Spectral anti-diffusion.** The bilinear damping is a fairly predictable function of wavenumber and offset, so it can be partially inverted in spectral space. I would leave this until last; it amplifies noise and hides the real problem.

My ranking: (1) then (2), with (3) as the interesting alternative if writing a reduced-grid cubic interpolator looks unappealing. And the 180min SETTLS check is worth doing before any of them, since if the extrapolation is unstable it will contaminate every comparison.

Tests now 103 passing, including one asserting no pole blow-up and one asserting the stepper leaves `horizontal_diffusion.exponential` alone.

SpeedyWeather/src/dynamics/semi_lagrangian.jl
```diff
@@ -0,0 +1,205 @@
...
+
+Done in 3D Cartesian coordinates on the unit sphere rather than by incrementing longitude and
+latitude. The lat/lon form needs `dlon = u*Δt/(radius*cos(lat))`, which is singular at the poles:
+on an octahedral Gaussian T128 grid the outermost ring sits at 89.28˚N where `cos(lat) = 0.0125`,
+so a 50 m/s wind over a 45 min step gives a 97˚ longitude jump — a straight line in (lon, lat) that
+bears no relation to the actual trajectory. The Cartesian form has no coordinate singularity, is
+exact for solid-body rotation, and follows the great circle rather than a rhumb-like path.
+
+Backward trajectories are obtained by passing a negative `Δt_over_radius`."""
+@inline function displace(lond::NF, latd::NF, u, v, Δt_over_radius) where {NF}
```
milankl: I like the great circle-based displace function but is that not something that should also go in the particle `move` function so that particle advection also benefits from this?

SpeedyWeather/src/dynamics/semi_lagrangian.jl
```diff
@@ -0,0 +1,205 @@
...
+    u_new = field_view(vars.grid.u, :, 1, 2)
+    v_new = field_view(vars.grid.v, :, 1, 2)
+    u_old = field_view(vars.grid.u, :, 1, 1)
+    v_old = field_view(vars.grid.v, :, 1, 1)
+
+    NF = eltype(sl.u_star)
+    w, w_old = time_stepping.extrapolate_winds ? (NF(3 // 2), NF(-1 // 2)) : (one(NF), zero(NF))
+
+    @. sl.u_star = w * u_new + w_old * u_old
+    @. sl.v_star = w * v_new + w_old * v_old
```
milankl: provide comments here that these are weights (where are they from?) and maybe rename the extrapolated winds from u star, v star to something more telling (the "star" doesn't really communicate anything when you read it out of context)

milankl (commented):
More comments!

milankl-claude:
Done in f441474f — `CubicInterpolator` in RingGrids, and the flexibility fixes needed to make it drop in without special-casing.

## The interpolator

16-point stencil: 4 longitude points on each of the 4 rings `j-1, j, j+1, j+2`. Within a ring the longitude spacing is uniform (`360/nlon`), so those are cubic Lagrange weights on a uniform stencil; ring latitudes are *not* uniformly spaced, so those are the general Lagrange basis over the 4 ring latitudes. The two are combined into a single weight per stencil point, so the kernel is a plain 16-term weighted sum — no direction-specific logic at interpolation time.

**Poles.** Fewer than 4 rings are available near the poles. Rather than special-case, the poles are treated as two extra "rings" carrying the ring-average value — the same `0` / `-1` flag convention `AnvilLocator` already uses — and the latitude Lagrange basis is built over whichever slots are usable. So weights still sum to 1 and the scheme degrades cubic → quadratic → linear instead of extrapolating. Verified explicitly at ±90˚ and ±89.9˚.

## Measurements

Isolated from the model, on an octahedral Gaussian grid. The semi-Lagrangian-relevant number is the last one — repeatedly shifting a field by a third of a grid cell, which is what SL does every step, so the per-application loss compounds:

| | Anvil | Cubic |
|---|---|---|
| identity (onto grid points) | exact | exact |
| partition of unity | exact | 6.7e-16 |
| off-grid RMS error, smooth field | 7.89e-4 | **6.95e-6** (114x better) |
| amplitude after 100 repeated shifts | **88.7%** | **99.87%** |

11.3% loss versus 0.13%. That is the whole story of why SL looked so diffusive.

At T128/30d through the model (default forcing, so these are forcing/dissipation equilibria rather than decay):

| configuration | enstrophy | KE |
|---|---|---|
| Eulerian `NCycleLorenz` 30min | 4.094e-5 | 2.852e7 |
| SL 60min anvil | 4.133e-5 | 2.575e7 |
| SL 60min **cubic** | 5.584e-5 | 3.582e7 |
| SL 180min anvil | 3.177e-5 | 4.204e7 |
| SL 180min **cubic** | 4.952e-5 | 9.241e7 |

Cubic clearly stops the damping. **But read the 180min rows as a warning, not a win** — KE is now 3.2x Eulerian. That is the SETTLS long-step problem I flagged last round, previously masked by the interpolation soaking up the excess energy. Worth fixing before drawing conclusions about SL accuracy; `extrapolate_winds=false` is the one-line diagnostic.

## Interpolator-flexibility

Three things were in the way, all now fixed:

1. **`find_grid_indices!` claimed `::AbstractLocator` but destructured `ij_as, ij_bs, ij_cs, ij_ds, Δabs, Δcds`** — Anvil-specific fields. Any other locator hit a `MethodError` inside a function advertising itself as generic. It now dispatches on the concrete locator, which is right: the stencil shape *is* what distinguishes interpolators.
2. **The `Interpolator(grid, npoints; NF)` constructor was Anvil-only.** Now generic over `AbstractInterpolator`, resolving the locator through the existing `Locator(I)` hook and re-parameterising via a new `nonparametric_type(I)`.
3. **The `AbstractLocator` contract was implicit.** Now documented: a ring-based locator must provide `npoints_output`, `js` and `Δys` (filled generically by `find_rings!`); everything else is its own business, filled by its `find_grid_indices!` and consumed by its `_interpolate!`.

So adding a further interpolator is now three methods — `Locator(::Type{<:X})`, `find_grid_indices!(::XLocator, ...)`, `_interpolate!(..., ::XLocator, ...)` — plus the struct. There is a test asserting the generic path works for both interpolators, so this does not silently regress.

`SemiLagrangian(spectral_grid; Interpolator=...)` selects it, defaulting to cubic.

## Versions and tests

`RingGrids` 0.1.9+DEV → **0.2.0-DEV** (new public type). 25 new tests in `RingGrids/test/interpolation_cubic.jl`; **full RingGrids suite passes**, so the refactor is non-breaking; SpeedyWeather SL tests pass.

Two honest caveats: the cubic stencil stores 16 indices + 16 weights per output point (~7.7 MB at T128), where a leaner version could store 4 ring offsets and recompute weights in-kernel — I chose the explicit form for a first cut and noted it. And rings with `nlon < 4` will repeat points in the stencil; harmless (weights still sum to 1) but worth knowing for very coarse HEALPix.

milankl:
Okay cubic interpolation is already much better! This is with a 120min-equivalent time step (30min at T128, the resolution actually used). But it's also much more expensive to run, this is 260 SYPD vs 600 SYPD Eulerian despite the larger time step.

image

SpeedyWeather/src/dynamics/semi_lagrangian.jl
milankl: can you double check that the great circle calculation is correct? Because I still see velocities blow up towards the poles, negative -500m/s near the north pole, 500m/s near the south pole, quickly growing there from 0 over 30 days

image

RingGrids/src/interpolation_cubic.jl
```diff
@@ -0,0 +1,305 @@
...
+    npoints_output::IntType
+
+    "ring indices j such that [j, j+1) contains the point, as for every ring-based locator"
+    js::VectorIntType = zeros(Int, npoints_output)
+
+    "distance fractions between rings j and j+1, as for every ring-based locator"
+    Δys::VectorType = zero(VectorType(undef, npoints_output))
+
+    "flat indices ij of the 16 stencil points, 0 = north pole, -1 = south pole"
+    ijs::VectorIntType = zeros(Int, NSTENCIL_CUBIC * npoints_output)
```
milankl: Maybe clearer to make this a Matrix of nstencil x npoints size?

RingGrids/src/interpolation_cubic.jl
```diff
@@ -0,0 +1,305 @@
...
+    NF_str = "{$NF}"
+    println(io, styled"{warning:CubicInterpolator}{note:$NF_str} for $(L.geometry.grid)")
+    print(io, styled"└ {info:onto}: $(L.locator.npoints_output) points")
+    return nothing
+end
+
+# generator with NF default based on geometry and locator, mirroring AnvilInterpolator
+function CubicInterpolator(
+        geometry::AbstractGridGeometry,
+        locator::AbstractLocator;
```
milankl: should this not be constrained? Otherwise one would be able to create a CubicInterpolator with an AnvilLocator in it, no?

RingGrids/src/interpolation_cubic.jl
```diff
@@ -0,0 +1,305 @@
...
+@kernel inbounds = true function _interpolate_cubic_kernel!(
+        Aout, @Const(A), @Const(ijs), @Const(weights), A_northpole, A_southpole,
+    )
+    k = @index(Global, Linear)
+
+    base = (k - 1) * NSTENCIL_CUBIC
+    sum = zero(eltype(Aout))
+
+    for s in 1:NSTENCIL_CUBIC
+        ij = ijs[base + s]
```
milankl: allocated as matrix this would be simply `[s, k]` ? Maybe clearer?

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
```diff
@@ -0,0 +1,201 @@
...
+    "[DERIVED] Time step Δt [s] at specified resolution"
+    Δt::NF
+
+    "[DERIVED] Δt/radius [s/m], the great-circle angle per unit velocity, see `trajectory_time_step`"
+    Δt_trajectory::Base.RefValue{NF}
+
+    """[DERIVED] Interpolator (grid geometry + locator) used to evaluate fields at departure points.
+    Defaults to `CubicInterpolator`; `AnvilInterpolator` is the cheaper but much more damping
+    alternative (11% amplitude loss over 100 repeated half-cell shifts against 0.1% for cubic)."""
+    interpolator::IP
```
milankl: Should the semi Lagrangian interpolator maybe live in the variables instead? The particle advection for example splits the interpolator into geometry (lives in model as constant) and locator (lives in variables as varying throughout model integration)

milankl: Also please be brief in comments above fields as they are used for documentation or `?` help, one line only. This comment is 3 lines long.

milankl (commented):
@milankl-claude Can you address these points please?

milankl-claude:
Chased the performance in fad7377c. **Cubic is now roughly at parity with Eulerian rather than 2.3x slower** — and the interesting part is that the interpolation was never the problem.

## Where the time actually went

Profiled the components at T128 rather than guessing:

| | `find_rings!` | `find_grid_indices!` | `interpolate!` |
|---|---|---|---|
| anvil | 0.56 ms | 1.23 ms | 0.18 ms |
| cubic | 0.56 ms | **6.79 ms** | 1.03 ms |

The 16-point interpolation costs 1.03 ms. Building the stencil cost **6.79 ms** — 5.5x anvil for only 4x the stencil points, and ~7x the cost of actually using it. My locator kernel was just inefficient. Three fixes:

**1. Rewrote `find_cubic_indices_kernel!`.** The four latitude weights were computed by calling a per-node helper four times, each re-selecting its own node with nested `ifelse` chains and doing four guarded divisions — about 40 `ifelse` and 16 divisions per grid point. Now one pass: ~12 `ifelse`, 4 divisions. The ring loop is unrolled, and the longitude wrap uses `ifelse` instead of `mod`, removing another 16 integer divisions per point. **6.79 → 4.08 ms.**

**2. The first trajectory iteration was doing nothing.** The first guess for the departure point *is* the arrival point, where the wind is already `u_star` at the grid point — so iteration 1 was paying a full locator update and two interpolations to recompute a field it already had, via what is provably the identity. Seed the departure winds directly instead. Symmetrically, the winds at the *final* departure points are never read (the next step needs the points, not the winds), so that refresh goes too. `n_iterations = 2` now costs **one** locator update and **two** wind interpolations rather than two of each. This is not an approximation — it removes work whose result was already known.

**3. Split the interpolators.** `trajectory_interpolator` (default `AnvilInterpolator`) places the departure point; `interpolator` (default `CubicInterpolator`) shifts the transported field. This is the point I raised earlier and it matters here: the winds are rebuilt from the spectral state every step, so no damping accumulates on them — only the transported field needs the expensive stencil, and it is interpolated exactly once per step.

## Result

T128, same machine, relative to Eulerian since absolute SYPD will differ from yours:

| configuration | SYPD | vs Eulerian |
|---|---|---|
| Eulerian `NCycleLorenz` 30min | 213.3 | 1.00x |
| **SL 120min, cubic field / anvil trajectory** | **232.4** | **1.09x** |
| SL 120min, cubic everywhere | 188.0 | 0.88x |
| SL 120min, anvil everywhere | 277.1 | 1.30x |

Your 260 vs 600 SYPD was 0.43x. The component numbers say the per-step SL overhead went from ~19.8 ms to ~7.8 ms, a 2.5x reduction, which would land 0.43x at roughly parity — consistent with what I measure here, though your machine is clearly faster in absolute terms so please re-run it.

## Accuracy is not paying for this

The cheap trajectory interpolator is a real approximation change, so I checked it rather than assuming. T128, 30 days:

| | cubic field / anvil traj | cubic everywhere |
|---|---|---|
| 60min enstrophy | 5.288e-5 | 5.073e-5 |
| 120min enstrophy | 5.093e-5 | 5.159e-5 |

Differences are within the scatter of a forced statistical steady state and not systematically in either direction. (Eulerian reference: 3.955e-5.)

Both interpolator slots are independently selectable, so `TrajectoryInterpolator=CubicInterpolator` reproduces the old behaviour if you want to A/B.

Full RingGrids suite passes, SL tests pass — 24 new ones covering both slots and `n_iterations` of 1, 2 and 4.

Still outstanding and unchanged: the 120–180min KE excess, which I still think is the SETTLS extrapolation rather than the transport. `extrapolate_winds=false` remains the one-line diagnostic.

SpeedyWeather/src/dynamics/semi_lagrangian.jl
milankl-claude: The great-circle calculation is correct — but you were right that something was wrong, and it turned out to be three things.

**`displace` itself checks out.** Against the independent direct-geodesic formula (`φ₂ = asin(sinφ₁cosδ + cosφ₁sinδcosβ)` etc.), it agrees to 1e-12 at every latitude and time step I tried. The only mismatches are at *exactly* ±90˚, where the local east/north frame is undefined — latitude agrees there too, only longitude differs, i.e. the two formulas pick different conventions for "which meridian is east at the pole". Distance travelled is `|V|Δt/R` to 1e-13.

**Bug 1 — the trajectory winds were averaged across mismatched frames.** This is your symptom. `u, v` are components in the *local* east/north frame, and `_departure_point_kernel!` added the arrival and departure winds componentwise. That's only valid where the two frames coincide; the mismatch is the meridian convergence, O(Δλ sinφ) — negligible in midlatitudes, O(1) next to the poles where a trajectory spans a large Δλ. The giveaway is that it got *worse* with more iterations, i.e. the fixed point was diverging:

| T31, 30d, polar ring-mean u | north | south |
|---|---|---|
| `n_iterations = 1` (no wind interpolation at all) | −9.6 | −5.7 |
| `n_iterations = 2` | **+218** | **−222** |
| `n_iterations = 4` | **−8800** | **+8766** |
| Leapfrog | +2.5 | +0.3 |

`n_iterations = 1` is the one configuration that never interpolates the departure wind, and it was clean. Both winds are now mapped to a common 3D Cartesian frame (`cartesian_velocity`) before averaging, and `displace` projects the mean back onto the tangent plane — an average of two vectors tangent at *different* points isn't tangent at either.

**Bug 2 — the trajectory was half the right length.** The trapezoidal rule is `x_d = x_a − Δt(V(x_a) + V(x_d))/2`. The kernel formed the mean `(V_a+V_d)/2` and then *also* halved the time step. Against the exact trajectory for a uniform wind the departure points sat at a ratio of 0.5000 at every latitude. The docstring already had the right formula; only the code disagreed.

Fixing this made things *worse*, which is how I found the third one.

**Bug 3 — SETTLS was a silent no-op.** `move_prognostic_grid_variables_back!` was called at the *end* of the barotropic `transform!`, after the current grid step had already been overwritten — so step 1 was copied from the **new** state, `u_old == u_new`, and `3/2 uⁿ − 1/2 uⁿ⁻¹` collapsed to `uⁿ` exactly. It now runs before the transforms, and only afterwards on `initialize`. That is the same `initialize ||` / `initialize &&` pattern the primitive-equation `transform!` already used — I'd just got the barotropic one wrong.

**Why that was fatal.** The trajectory carries the Rossby term through `f(x_d) − f(x_a)`, so *the time level of the trajectory wind is the time discretisation of the Rossby wave*. With the wind at `tⁿ` that's forward Euler — unconditionally unstable on the imaginary axis. Small-amplitude m=2, n=3 mode, free decay, no forcing/drag, negligible diffusion:

| Δt | measured, wind at tⁿ | `1 + (ωΔt)²/2` | measured, SETTLS |
|---|---|---|---|
| 60 min | 1.0038185 | 1.003827 | 1.0000468 |
| 30 min | 1.0009562 | 1.000957 | 1.0000049 |
| 15 min | 1.0002394 | 1.000239 | 1.0000007 |

(ω = 2Ωm/(n(n+1)) = Ω/3 for this mode.) Forward Euler confirmed to 3e-6, and growth−1 ∝ Δt² exactly — a factor of 4 per halving. With SETTLS actually running, growth drops 82× and the scaling becomes ≈Δt⁴.

**End to end**, T31, RMS wind, `KolmogorovFlow` + `LinearVorticityDrag`, plus an unforced control:

| | d10 | d20 | d30 | d40 | d50 | d60 |
|---|---|---|---|---|---|---|
| Leapfrog, forced | 17.4 | 21.9 | 24.8 | 27.4 | 29.3 | 30.6 |
| SL forced, before | 677 | 685 | 642 | 587 | 569 | 547 |
| **SL forced, after** | 14.3 | 18.5 | 21.4 | 22.3 | 22.6 | 24.7 |
| Leapfrog, free decay | 13.1 | 12.9 | 12.6 | 12.7 | 12.6 | 12.5 |
| SL free decay, before | 747 | 719 | 724 | 709 | 701 | 696 |
| **SL free decay, after** | 11.8 | 11.5 | 11.5 | 12.3 | 13.9 | 16.7 |

Polar ring-mean u is now −15.2/+12.3 against Leapfrog's +2.1/−1.7, and global max|u| 44.9 against Leapfrog's 54.1 — SL is now *quieter* than the Eulerian run rather than 8× louder.

Two things I should flag rather than bury:

- **I was wrong earlier** when I said the 120–180min KE excess looked like SETTLS instability. SETTLS was never running. It was the forward-Euler Rossby term.
- **`extrapolate_winds = false` is not a safe diagnostic option** — it *is* the forward-Euler case, and it blows up within days. I've documented it as such rather than as a cheaper first-order alternative. Suggesting it as a diagnostic was bad advice on my part.

Still open: with SETTLS the Rossby term is AB2-like, which is *weakly* unstable rather than neutral (1+4.7e-5 per step at 60min). Diffusion absorbs it at these resolutions, but free decay does creep up over 60 days. A genuinely centred trajectory wind needs a predictor–corrector pass at roughly double the per-step cost; noted under known limitations, not attempted.

Regression tests added for all three (frame mismatch, trajectory length incl. an assertion that a half-length trajectory would fail the tolerance, previous-step wind, and Rossby amplification).

RingGrids/src/interpolation_cubic.jl
milankl-claude: Done — `ijs` and `weights` are now `NSTENCIL_CUBIC × npoints_output` matrices. Clearer, and each output point's stencil is a contiguous column rather than a strided slice.

`CubicLocator` needed its own `(::Type{CubicLocator})(NF, npoints; architecture)` for this: the generic `AbstractLocator` zero-generator hardcodes three type parameters (`VectorType, VectorIntType, IntType`) and there are now five. The generic one still serves `AnvilLocator` unchanged.

One knock-on: `RingGrids.extrema_in` was restricted to `AbstractVector`, so the `@boundscheck` on `ijs` failed. Widened to `AbstractArray` — the implementation was already shape-agnostic.

RingGrids/src/interpolation_cubic.jl
milankl-claude: Yes — it is `[s, k]` now, same change as above. The kernel is a plain `for s in 1:NSTENCIL_CUBIC; ij = ijs[s, k]; sum += weights[s, k] * value; end` with no offset arithmetic, and `write_ring_stencil!` writes rows `off+1:off+4` of column `k`.

RingGrids/src/interpolation_cubic.jl
milankl-claude: Correct, you could — the outer constructor took any `AbstractLocator`, so `CubicInterpolator(geometry, anvil_locator)` would have built happily and then produced nonsense in `_interpolate!`, which dispatches on the locator. Constrained to `CubicLocator`.

`AnvilInterpolator` had exactly the same hole, so I constrained it to `AnvilLocator` too. Nothing in the tree constructs either with a mismatched locator, and the generic `(::Type{I})(grid, npoints)` path goes through the default inner constructor with explicit type parameters, so neither is affected.

SpeedyWeather/src/time_stepping/steppers/semi_lagrangian.jl
milankl-claude: Agreed, and done — that is the right split. The locators are rewritten every single time step, so they are exactly the kind of state that should not be sitting in an otherwise-immutable model component.

Following particle advection: `geometry` (a property of the grid, constant) stays in the stepper; the two locators are now `DynamicsVariable`s under the `:semi_lagrangian` namespace, alongside the departure points they index.

That needed `LocatorDim` generalising — it hardcoded both `AnvilLocator` and `model.particle_advection.nparticles`:

```julia
@kwdef struct LocatorDim{L} <: AbstractVariableDim
    n::Int = 1
end
LocatorDim(n::Integer) = LocatorDim{RingGrids.AnvilLocator}(n)   # the existing default

allocate(v::AbstractVariable{LocatorDim{L}}, model::AbstractModel) where {L} =
    L(model.spectral_grid.NF, v.dims.n; architecture = model.spectral_grid.architecture)
```

so particle advection keeps its behaviour via the one-arg constructor, and it no longer reaches back into `model.particle_advection` to find its own size. The stepper stores the locator *types* (`Locator`, `TrajectoryLocator`, derived from the `Interpolator`/`TrajectoryInterpolator` kwargs, which are unchanged) so `variables()` knows what to allocate. Particle advection tests pass.

milankl-claude: Fixed — every field docstring in `SemiLagrangian` is one line now. The detail that was crammed into them (why cubic by default, why the trajectory uses the cheap one, the `n_iterations` cost model) moved into the struct docstring and the relevant function docstrings, where it is not being rendered as a field description.

Two of the fields are gone entirely anyway, since the locators moved into `Variables`.

milankl:
Getting there!!!

image

milankl-claude:
**Correction to the performance numbers I gave earlier.**

I told you the cubic SL ran at 1.09x Eulerian throughput. That was measured before the trajectory fixes, and it no longer holds. Same idle machine, pre- vs post-fix commit, T128:

| | pre-fix `fad7377c` | post-fix `0371e7c4` |
|---|---|---|
| Eulerian `NCycleLorenz` | 334.8 | 337.7 |
| SL 120min cubic field / anvil traj | 359.3 (1.07x) | 320.5 (**0.95x**) |
| SL 120min cubic everywhere | 297.6 (0.89x) | 270.0 (0.80x) |
| SL 120min anvil everywhere | 417.0 (1.25x) | 370.2 (1.10x) |

So the original 1.09x claim reproduces (1.07x here), but **the correctness fixes cost ~11%** and cubic SL is now slightly *below* Eulerian parity rather than slightly above. Absolute SYPD differs from my earlier post because that run was contended by a test suite on the same box — the Eulerian baseline was 213 then and 335 now. The ratios above are from an idle machine, three repeats, +/-2%.

The cost is entirely in `departure_points!` (5.46 -> 6.95 ms/step); `semi_lagrangian_transport!` is unchanged at 3.2 ms. It is the two extra `sincos` pairs per grid point per iteration needed to map both winds into a common Cartesian frame.

`c5b87aeb` recovers 0.28 ms of that by fusing `cartesian_velocity` and `displace` into `departure_point`, so the arrival point's `sincos` is evaluated once instead of twice. Modest, and throughput moves within noise, but it removes work that was provably redundant.

The remaining pair is the departure point's own frame, which genuinely changes every iterate. The way to remove it is to interpolate the wind's **Cartesian components** instead of its local `u, v`: three interpolations per iteration instead of two, against one `sincos` pair per point per iteration. At T128 an anvil interpolation is ~0.2 ms against ~0.75 ms for the sincos, so it should be a net win — and it is also the more correct treatment, since interpolating `u, v` averages components in a frame that rotates between the stencil points. That is how the IFS does it. I have not done it: it changes the numerics rather than just the arithmetic, so it wants to be its own change with its own verification.

Happy to take the parity hit as the price of a scheme that no longer blows up, unless you would rather I chase it now.

SpeedyWeather/src/dynamics/semi_lagrangian.jl
milankl-claude: Both done in `ce0b8069`.

**Weights.** Spelled out where 3/2 and -1/2 come from — linear extrapolation from `uⁿ⁻¹, uⁿ` to the middle of the step, which is where the trapezoidal trajectory wants its wind:

```julia
# Linear extrapolation in time from uⁿ⁻¹, uⁿ to the middle of the step, which is where the
# trapezoidal trajectory wants its wind:
#   u(tⁿ + Δt/2) ≈ uⁿ + (Δt/2) (uⁿ - uⁿ⁻¹)/Δt = 3/2 uⁿ - 1/2 uⁿ⁻¹
# hence the weights. Without it the wind stays at tⁿ, one step behind (see the docstring:
# that is forward Euler on the Rossby term and unstable, not merely first order).
```

**Rename.** Agreed on the problem — I went with `u_trajectory`/`v_trajectory` rather than `u_extrapolated`, for one reason: with `extrapolate_winds = false` the field is not extrapolated at all, it is just `uⁿ`, so `u_extrapolated` would be actively wrong in one of the two modes. `u_trajectory` says what it is *for*, which holds either way, and it matches the `Δt_trajectory` / `trajectory_locator` / `trajectory_interpolator` vocabulary already in the file. It also reads well against `departure_u`/`departure_v`, which are that same field evaluated at the departure points.

Say the word if you prefer `u_extrapolated` anyway and I will switch it — it is a one-line change either way.

The `desc` strings changed to match ("Wind defining the backward trajectories, zonal/meridional").

milankl:
100km HEALPix at 12-hour long time steps (yes 12 hours!, so equivalent to 2 days at T32) only works when switching off rotation, the Coriolis term seems to cause instabilities at much shorter time steps already. Maybe this can be fixed, not sure, also not sure whether relevant. At such long time steps, the semi-Lagrangian model is 16x faster than normal (8k SYPD vs 500 SYPD on a single macbook M4 cpu)

image

